PR #4, opened 2024-01-11
Support standalone operation without docker

bukodi:
During development, or in a small environment (Raspberry or OpenWrt) it is useful to start an application without docker.

This pull request contains the following modifications:
- MySQL specific sessionstore implementation replaced with a generic GORM based.
- SQLite database type added. (It works with GORM, only one index name needed to be changed.)
- Access static files and templates through the fs.FS interface, so the embed.FS can be used also.
- Smart defaults defined for environment variables, so starting without any configuration results a working demo instance.

Please, consider to accept these changes! Thank You!


leodip:
@bukodi Thanks very much for the contribution! 

Your idea is great, if we can support it running in SQLite, without a container, that should cover some more interesting use cases.

Please let me take a closer look at the PR over the weekend, and I'll give feedback soon.

Thanks again 🙂

leodip:
@bukodi, I think your PR is awesome!

My plan is to merge it as it is, then make a few modifications in a subsequent commit:

- For consistency sake, I'll move gormstore.HttpSession to entities.go, and move gormDB.AutoMigrate in gormstore to the existing database.go file

- I'll try to add a method to distinguish if the application is running inside a docker container, and if it is, the default value for "Host" environment variable should remain empty. Otherwise, if we make the application listen on 127.0.0.1 or localhost, when running in a container, the port forwarding to inside the container doesn't work.

- This is maybe the most important: I'll replace the SQLite driver gorm.io/driver/sqlite with the alternative [github.com/glebarez/sqlite](https://github.com/glebarez/sqlite). The first one depends on C libraries linking, and depends on having CGO_ENABLED=1, or it panics. I would like to keep the Alpine base image and it wouldn't be straightforward to have CGO_ENABLED=1, with the required dependencies, in that Alpine image. The second SQLite driver (glebarez) is written purely in Go, so we can have CGO_ENABLED=0.

I love your enhancements, overall 🙂 The GORM store you added will allow for an easy addition of other database types, such as Postgres.

Thanks again for contributing!

leodip:
@bukodi  About that `GORMStore.Close()`, I'm considering removing the method, as it's not currently being used. If you see any use case for that, please let me know.

bukodi:
@leodip Thank you!
- The glebarez's driver seems better, just I had no experience with it.
- Yes you have right, the Close() is unnecessary.

PR #10, opened 2024-02-18
Limiting the maximum length of deviceName and deviceOS when parsing t…

aboyadzhiev:
I'm proposing a logic to limit the outcome length for `deviceName` and `deviceOS` when parsing the `User-Agent` HTTP header.


leodip (approved):
I think it is very reasonable to limit those strings. Thanks for sending this!

PR #99, opened 2026-07-05
Pin JWT signing algorithm

### LINKED ISSUE #81: JWT parsers do not pin signing algorithm (jwt.WithValidMethods)
## Summary

Neither JWT parser pins the accepted signing algorithm (`jwt.WithValidMethods`). Not exploitable today, but worth adding as defense-in-depth against algorithm-confusion / `alg:none` attacks.

Severity: **Low (defense-in-depth)**

## Details

Both parsers build parser options without `jwt.WithValidMethods([]string{"RS256"})`:

- `src/core/oauthdb/token_parser.go:84-89`
- `src/core/oauth/token_parser_jwks.go:75-80`

This is currently **not** exploitable: both keyfuncs return `*rsa.PublicKey`, and `golang-jwt/v5` rejects HMAC verification when the key is not `[]byte` and rejects `alg:none` unless the unsafe sentinel key is returned. So HS256 algorithm-confusion and `none` attacks fail as written.

However, pinning the algorithm is cheap insurance: a future refactor to the keyfunc (e.g. returning a type that some `SigningMethod` accepts) could silently reintroduce the risk.

## Suggested fix

Add `jwt.WithValidMethods([]string{"RS256"})` to the parser options in both parsers.

## Location

- `src/core/oauthdb/token_parser.go:84-89`
- `src/core/oauth/token_parser_jwks.go:75-80`

_Found during a structured security review (Area A: OAuth2/OIDC protocol correctness)._

ahfoysal:
Fixes #81.

Pins both JWT parsers to RS256 and adds coverage for non-RS256 tokens.

Tests: `go test ./oauth ./oauthdb` from `src/core`.

### DIFF
--- a/src/core/oauth/token_parser_jwks.go
+++ b/src/core/oauth/token_parser_jwks.go
@@ -72,7 +72,7 @@ func (tp *JWKSTokenParser) DecodeAndValidateTokenString(token string, _ *rsa.Pub
 
 	claims := jwt.MapClaims{}
 
-	opts := []jwt.ParserOption{}
+	opts := []jwt.ParserOption{jwt.WithValidMethods([]string{jwt.SigningMethodRS256.Alg()})}
 	if withExpirationCheck {
 		opts = append(opts, jwt.WithExpirationRequired())
 	} else {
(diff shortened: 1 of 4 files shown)


leodip (approved):

leodip:
Thank you

PR #100, opened 2026-07-06
fix: pin JWT signing algorithm

ahfoysal:
Pins JWT verification to RS256 and surfaces token algorithm mismatch as explicit failures.


leodip:
This landed on main as 8f3a38f, applied directly with your authorship preserved, which is why GitHub didn't auto-close this PR. Closing as already merged. Thanks

Issue #134, opened 2026-08-02
Follow-up to #129: terminated-session registry markers are retained indefinitely


leodip:
## Summary

#129 introduces a **terminated-session registry**: one small row per explicitly terminated session,
holding a session identifier and a timestamp. Those rows are retained **indefinitely and by design**,
because no finite retention period can be shown safe. This issue tracks making them reclaimable.

Filed as an explicit follow-up rather than left as a documented residual of #129, so the growth keeps
an open tracking item after #129 closes. It does **not** block #129: indefinite retention of two
small columns is a deliberate, cheap trade, and the alternative would have been a horizon that is
provably wrong.

Depends on #129 having landed, since the table does not exist yet.

## Why no finite horizon is safe today

The obvious reasoning is that the marker only has to outlive any authorization code that could
reference it, and a code expires 60 seconds after creation (`authCodeExpirationInSeconds` in
`ValidateTokenRequest`). That reasoning is wrong, and #129's decision 7 records why.

The 60 seconds runs from `code.CreatedAt`, not from the termination. The dangerous interleaving is:

1. `/auth/issue` resolves the ceremony's bound session and finds it live.
2. Termination writes the marker, sweeps and commits.
3. The `/auth/issue` request **stalls**.
4. A cleanup horizon measured from `terminated_at` elapses and the marker is deleted.
5. The request finally inserts its code, with a **fresh** 60-second window.
6. Redemption finds no marker, classifies the missing session as ordinary housekeeping deletion, and
   succeeds.

Nothing bounds step 3. Verified: the authserver constructs no `http.Server` with `ReadTimeout` or
`WriteTimeout`, so handler execution time is unbounded, and database waiting and process scheduling
sit outside the code lifetime calculation entirely.

## What would make reclamation safe

A provable upper bound on how late a code can be created relative to the session observation that
preceded it. That means serializing issuance against explicit termination, or an equivalent atomic
protocol, after which the retention horizon is:

```
last possible code creation + authorization code lifetime + clock allowance
```

That is the same missing serialization boundary #131 and #132 describe for their own races, which is
why this issue should be scoped alongside them rather than solved in isolation. All three want a
portable way to order a write against a sweep across SQLite, MySQL, PostgreSQL and SQL Server.

## Scope

- A cutoff whose safety argument is written down, not assumed, and which depends on a serialization
  property that actually exists by then.
- A cleanup path for the registry, presumably in `Worker.performTask` alongside the existing
[...]

Issue #132, opened 2026-08-01
Make refresh-token family containment atomic with rotation


leodip:
## Summary

Refresh-token family containment, as specified for #128, is best-effort. Rotation commits its
parent claim separately from its child insert, so a child whose insert commits after the
containment statement survives containment. This issue tracks making the two atomic with
respect to one another.

Filed as an explicit follow-up rather than left as a documented residual of #128, so the gap
keeps an open tracking item after #128 closes. It does **not** block #128: best-effort
containment is a substantial improvement over today, where detected replay contains nothing at
all.

## The mechanism

#128 gives refresh rotation a compare-and-set on the presented row, and gives replay a
family-scoped containment statement:

```sql
UPDATE refresh_tokens
SET revoked = true, updated_at = ?
WHERE first_refresh_token_jti = ?
  AND revoked = false;
```

The containment statement is atomic over the rows visible to it. It cannot reach a row that
does not exist yet. Rotation inserts its child from inside the token issuer, after the parent
claim has already committed and after JWT signing, so there is a window in which the parent is
revoked and the child is not yet persisted.

The damaging interleaving, with a family of RT1 (already retired) and RT2 (live):

1. An attacker holding RT2 presents it. The compare-and-set succeeds and RT2 becomes revoked.
   The attacker's RT3 is being signed but is not yet inserted.
2. The legitimate client presents RT1. This is a correctly detected replay.
3. Containment runs. The family contains RT1 (revoked) and RT2 (revoked). **Zero rows are
   transitioned.**
4. The attacker's RT3 insert commits. RT3 is live and outside the family sweep that was
   supposed to include it.

The attacker keeps a working grant, which is the outcome #128 exists to prevent.

### The audit consequence, which is the sharper part

Because the cascade transitioned zero rows, #128's audit gate suppresses the
`refresh_token_replay_detected` event for that interleaving. So the one interleaving in which
containment fails is also the one in which it emits no record that it ran. A subsequent replay
of RT1 would contain RT3 and would emit the event, but the legitimate client has already
received `invalid_grant` and will typically stop rather than retry.

## Why #131 does not own this

#131 coordinates rotation with a **user-scoped** authentication-generation change. This race
requires rotation and **family-scoped replay containment** to serialize on the token family.
The two touch different row sets and different triggers. #131's eventual serialization boundary
may supply reusable machinery, but satisfying #131 does not by itself order rotation against a
family-replay cascade, and neither issue should be made to depend on the other.

[...]

Issue #131, opened 2026-07-31
Refresh rotation is not coordinated with user-scoped generation changes, so a preserved session can be handed an unusable token


leodip:
## Summary

A refresh token rotation running concurrently with a user-scoped authentication generation change
can succeed and return a child refresh token that is **already unusable**. The user must
re-authenticate. This is fail-closed and not a security defect, but it defeats the preservation
promise of the self-service password change path.

Accepted as a documented residual of #106, whose design work weighed this coordination and deferred it. This issue tracks the durable fix.

Related machinery: #128. **#128 is not a superset of this issue.** It proposes compare-and-set
rotation and family containment, and explicitly leaves open whether stronger persistent coordination
is required. A compare-and-set on the parent refresh-token row does not by itself coordinate
correctly with a user-scoped sweep that touches a different set of rows, so #128 landing does not
automatically satisfy the criteria below.

## The mechanism

`RevokeUserAuthState` (`src/authserver/internal/handlers/revocation.go`) runs inside a transaction.
It increments `users.auth_state_generation`, then discovers the user's refresh tokens and sessions,
revokes everything outside the preserved set, and promotes the preserved session and its tokens to
the new generation.

Refresh rotation does not participate in that transaction. In `HandleTokenPost`'s `refresh_token`
case, the parent refresh token is loaded during validation, marked revoked with
`UpdateRefreshToken(nil, ...)`, and the issuer then stamps the child from the **in-memory** parent.
So the child inherits the generation the parent held when validation read it, and the window spans
an entire token issuance including JWT signing.

If the child row commits outside what the sweep's transaction can see, it is neither revoked nor
promoted. It keeps the previous generation, the user is now on the next one, and the child is
rejected the next time it is presented. The parent is already revoked, so that grant is dead.

Observable outcome by grant type:

| Grant | Returned refresh token | Returned access token |
|---|---|---|
| Session-bound (auth code) | rejected on next use | still works: the middleware defers to the session's generation, and the session was promoted |
| Offline (`offline_access`) | rejected on next use | rejected immediately: the sid-less branch reads the token's own generation claim |
| ROPC | rejected on next use | rejected immediately, same reason |

## Why query ordering does not fix it

`RevokeUserAuthState` deliberately runs its user-scoped discovery query **before** its sid-scoped
one. On an engine with per-statement read views, that makes one sub-case harmless: a child
committed between the two queries is absent from the first, so it is never swept, and present in
[...]

Issue #109, opened 2026-07-25
RP-Initiated Logout: post_logout_redirect_uri wrongly required, unencoded state, and non-standard sid in the redirect


leodip:
## Summary

Review of `src/authserver/internal/handlers/handler_account_logout.go` against [OpenID Connect RP-Initiated Logout 1.0](https://openid.net/specs/openid-connect-rpinitiated-1_0.html) turned up one functional bug that breaks the most common conforming logout request, one URL-construction bug with four distinct failure modes, and a non-standard parameter that discloses the session identifier. Two further design divergences are noted at the end.

`/auth/logout` implements RP-Initiated Logout only. Front-Channel Logout 1.0, Back-Channel Logout 1.0 and Session Management 1.0 are not implemented, and the discovery document correctly advertises only `EndSessionEndpoint` (`handler_well_known_oidc_config.go:41`). This matters for item 3 below.

---

## 1. `post_logout_redirect_uri` is treated as required, so a valid logout request fails and the user stays logged in

**Impact: functional. This is the most important item here.**

`doLogoutWithIdToken:268-273`:

```go
postLogoutRedirectURI := httpHelper.GetFromUrlQueryOrFormPost(r, "post_logout_redirect_uri")
if len(postLogoutRedirectURI) == 0 {
    renderAuthError(w, r, httpHelper, i18n.NewLocalizedError(i18n.ErrCodeLogoutPostLogoutRedirectRequired, nil))
    return
}
```

The spec lists this parameter as **OPTIONAL**:

> "post_logout_redirect_uri - OPTIONAL. URI to which the RP is requesting that the End-User's User Agent be redirected after a logout"

So `GET /auth/logout?id_token_hint=<valid token>`, which is the single most common conforming RP-initiated logout request, renders an error page. Worse, the session teardown at `:328-334` and the cookie wipe at `:336-345` both sit **after** this early return, so **the user is not logged out at all**. An RP that simply wants to end the session without being redirected back cannot do so.

The spec's model is that the redirect is the optional extra: perform the logout first, then redirect only if asked.

### Fix

Make the *redirect* conditional, not the *logout*. Perform the teardown unconditionally, then either redirect (when `post_logout_redirect_uri` was supplied and validated) or render a "you have been signed out" page.

[...]

PR #138, opened 2026-08-04
fix(sessions): make ending a session durably cut off access

### LINKED ISSUE #129: Security: ending a user session does not durably cut off access (offline grants survive, and an in-flight ceremony recreates the session)
## Summary

Ending a user session does not durably cut off access. Three endpoints tear down session state, and
none of them prevents the terminated session's authentication state from being recreated or its
grants from continuing.

| Endpoint | What it does today | Revokes refresh tokens? |
|---|---|---|
| `DELETE /api/v1/admin/user-sessions/{id}` (`HandleAPIUserSessionDelete`) | deletes the `UserSession` | no |
| `DELETE /api/v1/account/sessions/{id}` (`HandleAPIAccountSessionDelete`) | deletes the `UserSession`, ownership-checked | no |
| `GET/POST /auth/logout` (`handleExistingSessionOnLogout`) | deletes one `UserSessionClient`, and the `UserSession` when it was the last client | no |

Split out of #106 after review. #106 fixes credential-invalidating actions (password reset, password
change, admin password set, account disable) with a per-user generation boundary, which is complete
for that scope. Session termination cannot reuse that mechanism, because incrementing a user-wide
counter would invalidate every one of that user's other devices, which is the opposite of what
"sign this device out" means.

Severity: **Medium.** An administrator or user ending a session reasonably expects that device to
lose access. Today it keeps renewable access through an `offline_access` grant for up to that
grant's max lifetime, seeded at one year.

## Gap 1: refresh tokens are not revoked

Session deletion contains a session-bound (`typ=Refresh`) refresh token, because its validation
requires a live, in-bounds session. It does **not** contain an offline one: `typ=Offline` validation
deliberately does not consult the session, and offline refresh tokens store no session identifier
of their own (`TokenIssuer.generateRefreshToken` stores either a max lifetime or a session
identifier, never both).

So after "end session", an `offline_access` client that authorized through that session keeps
refreshing indefinitely.

The obvious fix is sid-scoped revocation using the existing
`GetRefreshTokensBySessionIdentifier`, which matches `codes.session_identifier` and therefore does
reach offline tokens. That is necessary but not sufficient, per gaps 2 and 3.

## Gap 2: a concurrent refresh inserts a surviving replacement

Refresh rotation marks the presented token revoked with `UpdateRefreshToken(nil, ...)` and inserts
its replacement with `CreateRefreshToken(nil, ...)`, both outside any transaction. A refresh that
validated before the termination sweep therefore inserts its child after the sweep commits, and for
an offline grant nothing invalidates that child.

A `codes.revoked` marker was considered for this and does work here specifically, because a rotated
refresh token inherits `code_id` from its parent, so revoking the code catches every present and
future descendant. It requires more than a column:

1. `codes.revoked` plus a model field tagged `fieldtag:"dont-update"` so a stale `Code` model cannot
   restore `revoked = false`.
2. A narrow bulk method, `RevokeCodesBySessionIdentifier(tx, sid)`, plus per-engine wrappers.
3. Initial code redemption must reject `code.Revoked`, or a terminated session's unused code can
   still mint its first offline token family.
4. `MarkCodeAsUsed` must fold the marker into its compare-and-set:
   `WHERE id = ? AND used = false AND revoked = false`. Verified its current predicate is
   `id = ? AND used = false`, so validation and claiming are separate steps and a code revoked
   between them would still be claimed.
5. Auth-code refresh validation must reject a revoked grant.
6. Cleanup must reach unused revoked codes. Verified `DeleteUsedCodesWithoutRefreshTokens` filters
   on `used = true`, so an unused revoked code would otherwise never be reaped. **See the separate
   note below: that predicate is also currently broken for an unrelated reason.**

## Gap 3, the hard one: an in-flight ceremony recreates the session

This is why the marker alone is not a durable boundary, and it is what took this work out of #106.

Verified sequence:

1. An SSO ceremony validates session S and proceeds through the auth state machine.
2. "End session" revokes S's codes and deletes S.
3. The ceremony resumes at `/auth/completed`.
4. `HandleAuthCompletedGet` evaluates `hasValidUserSession`, finds none, and takes the else branch,
   which calls `userSessionManager.StartNewUserSession(...)` **unconditionally**, from
   `authContext.UserId`, with no requirement that anyone authenticated. It computes
   `userReallyAuthenticated` but consults it only inside the *valid session* branch, to refresh
   `AuthTime`.
5. `/auth/issue` inserts a brand-new, unrevoked code, outside the termination transaction.
6. Nothing rejects it: the code is new so no marker applies, and the user's generation never moved.

The consent screen can delay code insertion past the sweep in the same way.

Note gap 3 is arguably a defect in its own right, independent of revocation: a ceremony whose
session an administrator deleted mid-flight silently receives a **new** session without
re-authenticating. Deciding what should happen instead (almost certainly: fail and force
re-authentication) is an auth-flow behaviour change that has to be reasoned about alongside
`prompt=none`, step-up authentication and the `Level2AuthConfigHasChanged` path.

## Candidate designs

Neither is small; choosing between them is the point of this issue.

**A. Session tombstone.** Mark sessions revoked instead of deleting them, carry the originating
session identifier on `AuthContext`, and reject at code issuance when the ceremony's originating
session is revoked. Requires deciding gap 3's behaviour change, plus a reaping policy for
tombstones so `user_sessions` does not grow without bound. Composes with the gap 2 marker.

**B. Transactional serialization.** Serialize session termination against issuance so a ceremony
cannot complete across a termination. Needs consistent row-lock semantics across SQLite, MySQL,
PostgreSQL and SQL Server, and the issuance paths currently pass nil transactions throughout. It
also cannot by itself identify a ceremony that began before the termination and completes after it,
so it still needs a persisted marker.

A per-session generation, mirroring #106's per-user generation, is a third framing of A and may be
the cleanest: `user_sessions.generation` bumped rather than the row deleted, inherited by
`AuthContext` on SSO reuse and by codes, and checked at issuance and refresh.

## Scope of a full fix

- Sid-scoped refresh-token revocation at all three endpoints, or a documented decision to leave
  RP-initiated logout alone (see below).
- Whichever durable marker design is chosen, with its schema change across four engines.
- Gap 3's auth-flow decision.
- Cleanup and reaping policy for whatever persistent state is introduced.

## What RP-initiated logout should do is a separate policy question

Sid-scoped revocation is the **wrong** scope for `/auth/logout`. A user logging out of one client
leaves the session alive for the others, so revoking sid-wide would silently break them. Covering
logout properly needs client-scoped revocation within a session, which no query provides today.

And no specification compels it either way. OIDC RP-Initiated Logout is silent on token revocation.
RFC 9700 section 4.14.2 says "Authorization servers MAY revoke refresh tokens automatically in case
of a security event, such as: password change or logout at the authorization server", which permits
but does not require it. OIDC Core section 11 defines `offline_access` as a refresh token usable
"even when the End-User is not present (not logged in)", which is the one normative pull, and it
points toward leaving ordinary browser logout alone.

Recommendation: treat the two explicit "end this session" endpoints as security actions and
RP-initiated logout as routine, unless a product decision says otherwise.

## Adjacent bug found while investigating, worth its own issue

`DeleteUsedCodesWithoutRefreshTokens` builds
`NotIn("id", SELECT code_id FROM refresh_tokens)`. ROPC refresh tokens have `code_id = NULL`, and
in SQL `x NOT IN (…, NULL)` evaluates to UNKNOWN rather than TRUE, so **the predicate matches
nothing at all once any ROPC refresh token exists** and used codes accumulate forever. The fix is a
correlated `NOT EXISTS`, or at minimum filtering `code_id IS NOT NULL` inside the subquery. Live
today, unrelated to session termination, and only surfaced here because gap 2's design would have
changed that predicate.

## Test requirements

- **Integration, dev container.** Authorize with `offline_access`, end the session, then refresh:
  expect rejection. Plus the reverse assertion, that a session-bound grant on an unrelated session
  is unaffected.
- **Integration, concurrency.** A refresh in flight across a session termination must not yield a
  usable child. This is the gap 2 test and it mirrors
  `TestToken_AuthCode_ConcurrentDoubleSpend_IssuesOnlyOnce` in `token_authcode_concurrent_test.go`.
- **Integration, ceremony.** Begin an SSO ceremony, terminate the session mid-flight, complete the
  ceremony: assert no usable code is issued. This is the gap 3 test and the one that fails against
  any design that only marks existing rows.
- **Data tests**, all four engines, for whatever query and schema the chosen design adds.
- **Regression.** `session_deletion_test.go` and `session_bearer_revocation_test.go` must still
  pass, since access tokens on session-bound grants are already rejected once the session is gone.

## Relationship to other issues

- **#106** fixes the credential-change scope with a per-user generation boundary and deliberately
  excludes session termination, which is why this issue exists. #106's own design record carries the reversal and the reasoning.
- **#128** covers refresh-token replay containment and the non-atomic single-use claim. Adjacent:
  both concern durable state in the token path, and #128's compare-and-set is the same technique
  gap 2 needs.
- **#127** (closed, not planned) proposed an authorization-grant entity. A grant-level revoked state
  is the minimal piece of that proposal which gap 2 would need.
- **#77** fixed the analogous non-atomic claim for authorization codes; `MarkCodeAsUsed` is the
  precedent for gap 2's item 4.

leodip:
Closes #129. Built by `/leo-run` from the sealed agreement, eight stages, each reviewed by a second
agent before it was committed.

**Issue:** #129, ending a user session does not durably cut off access (`bug`, `security`)
**Agreement:** written and reviewed alongside this change and deliberately not tracked in the repository. The decisions that constrain future edits live in the code comments beside the code they constrain.
**Read first:** [the deferred decisions comment](https://github.com/leodip/goiabada/pull/138#issuecomment-5196348942)
(nothing was deferred, and it says what came to you instead), then
[the follow-ups comment](https://github.com/leodip/goiabada/pull/138#issuecomment-5196349557)
(four drafted issues, since reviewed and filed as #139, #140, #141 and a comment on #109, see
"Follow-ups filed" at the bottom).

## What the change does

Ending a session at either explicit "end session" endpoint becomes a security action with a durable
consequence. Three gaps closed, all of which failed open, meaning they preserved access somebody
believed they had removed:

1. **An offline grant outlived its session.** `codes.revoked` now carries the boundary: terminating a
   session marks every authorization code that session issued, in one transaction with the sid-scoped
   refresh-token sweep and the session delete. Redemption rejects a revoked code, and so does any
   refresh token descended from one.
2. **A refresh racing the termination inserted a surviving replacement.** Closed structurally rather
   than by serializing anything: a rotated refresh token inherits its parent's `code_id`, so marking the
   code marks every present and future descendant. The racing child is born already rejected, because
   the fact predates its existence.
3. **An in-flight ceremony recreated the session it rode.** `/auth/completed` now requires that this
   ceremony performed level 1 authentication before it mints a session, and `/auth/issue` refuses to
   bind a code to a session that no longer resolves, with a compensating `UPDATE` covering the insert
   that lands just after a termination.

A session **expiring** still leaves an offline grant working, which is what `offline_access` is for. A
session **explicitly ended** does not. That distinction is the one the whole design rests on, and the
documentation now draws it.

The 16 decisions behind this, including the alternatives that lost and why, are in section 3 of the
agreement. Section 4 states the property that makes it safe; section 1 has the three gaps in full.

| Stage | What landed | Commit |
|---|---|---|
| 1 | `codes.revoked` and its sweep: migration 000026 on four engines, `RevokeCodesBySessionIdentifier`, and `AND revoked = false` on the code claim. Inert by construction, nothing reads it yet | `066d289` |
| 2 | The two rejections in `ValidateTokenRequest`, at redemption and at refresh, placed behind client authentication and PKCE per decision 7 | `266615b` |
| 3 | `TerminateUserSessionTx`, the three writes in one transaction, plus the `terminated_user_session` audit event. Still uncalled | `d19d8b5` |
| 4 | **The behaviour changes here.** Both endpoints terminate, both audit events fire after the commit, and everything that describes the action lands in the same commit: three confirmation modals in both locales, and four documentation pages | `f964e54` |
| 5 | The `/auth/completed` gate, on a dedicated level 1 proof rather than on `AuthenticatedAt`, per decision 15 | `5a7c280` |
| 6 | The `/auth/issue` liveness check, the compensating revoke, and `login_required` rather than a login form for a `prompt=none` ceremony whose session ended, per decision 16 | `422dd8a` |
| 7 | Retention: unused revoked codes are reaped, so the new state cannot accumulate | `803f593` |
| 8 | Gap 2's evidence, two integration tests and no production line | `682d51e` |

26 commits over 60 files, one code commit per stage plus the agreement following it. The agreement's
section 7 is the run log: what landed, which tiers ran, every deviation, every review finding with its
evidence and disposition, and the mutations measured at each gate.

## Tests

**Full suite at the close, every tier, all four engines, green.** `where.sh test --type all`, exit 0,
zero `FAIL` and zero panics, "All tests completed successfully". Modules across all three Go modules;
the data tier on mysql, postgres, mssql and sqlite; the integration tier on the same four (64.8s, 74.3s,
99.2s, 43.2s), 9022 passing test lines in total.

Per tier, this change adds: migration and data-layer coverage on four engines for both new methods and
for the marker's `dont-update` protection; unit coverage for both validator rejections including the
ordering rows that pin them behind client authentication; unit coverage for the termination helper's
transaction threading and its zero result on every failure path; two new handler test files that did not
exist before, covering the audit payloads and that a 403 terminates nothing; and integration coverage of
both endpoints, the three ceremonies that must not recreate a session, the `prompt=none` refusal, and
gap 2 both reconstructed and raced.

**The race reaches the window rather than approaching it.** In the closing run every engine produced one
replacement token alongside the termination, and on mssql and sqlite that replacement was still live in
storage when it was presented, so the sweep never saw it and the code marker was the only thing between
it and a working grant. On mysql and postgres it landed before the sweep. Both are legitimate and
neither is asserted, so the test logs which happened.

**Mutation-tested at every gate**, 16 in all, each applied to the real code and reverted. The two for
stage 8 are the argument for that stage existing: with the marker's read removed, and separately with
its write removed, stage 4's headline endpoint case and stage 6's consent ceremony both stayed green,
and only stage 8's two tests failed, answering a terminated grant with a working access token.

## Deviations from the plan

Recorded per stage in section 7, and none changed what the change does. The load-bearing ones:

- **Stage 2** shipped ten test rows rather than nine. Review round 1 showed the planned missing-secret
  row could not pin what it claimed, because a missing secret is refused by an earlier guard, so a
  present-but-wrong secret row was added. Demonstrated by relocating the check and watching the new row
  fail while the old one passed.
- **Stage 3** corrected a claim the implementation session had made in the mailbox: four failure rows
  catch a partially populated result, not one. The prediction was left in place and the measurement
  recorded beside it, because editing the prediction would erase the evidence that it was wrong.
- **Stage 5** stopped the run and escalated decision 15: the gate the seal specified was satisfied by
  OTP alone. Rebuilt on a dedicated level 1 proof after you answered.
- **Stage 6** stopped the run and escalated decision 16 (`prompt=none` cannot be restarted into a login
  form), and its round 2 found the new branch clearing the auth context *after* committing the client
  response, so the clear reached nothing. Fixed inside the stage, with assertions that read the
  committed response rather than a mock expectation. That defect turned out to exist at seven
  pre-existing sites, now #141.
- **Stage 8** staggers its race by 5 milliseconds. Released at the same instant, sqlite consistently
  produced no replacement at all, so the interesting assertions never ran. Every assertion holds at any
  value including zero, which is what keeps it a probability knob rather than a synchronization point.

## Known limits, stated rather than implied

- **An offline grant's access token stays valid until it expires**, 5 minutes by default, because it
  carries no `sid` for the session middleware to check and termination must not move the user's
  generation. Decision 11, and the documentation says so rather than letting the credential-change
  sentence be read as covering this case too.
- **One interleaving is still open**, a code insert landing between a termination's `UPDATE` and its
  `COMMIT`. Decision 12 accepted it, and it is now #139: closing it needs the same portable
  order-a-write-against-a-sweep primitive that #131 and #132 want.
- **RP-initiated logout is unchanged**, deliberately, and now pinned by tests. Decision 3, with
  decision 13 leaving the interactive path's own defect to #109, where it is now recorded as a
  [comment](https://github.com/leodip/goiabada/issues/109#issuecomment-5197711587).

## Follow-ups filed

The four drafted follow-ups were re-verified against the code after the run closed, then filed. Two
bodies changed on the way in, and both changes narrow a claim rather than widen it.

| Drafted as | Filed as | Change on the way in |
|---|---|---|
| 1, the `UPDATE`/`COMMIT` window | **#139** (`bug`, `security`, `go`) | Body now says the window is **not uniform across engines**. It is open on PostgreSQL; on MySQL, SQL Server and SQLite it is closed incidentally, by InnoDB next-key locks over `idx_codes_session_identifier`, by a blocking READ COMMITTED read, and by writer serialization respectively. Reframed as "correctness rests on three engines' defaults by accident", which is the standard #132's criterion 4 already sets. Without this a reader reproducing on SQLite would conclude the issue is wrong |
| 2, the logout path that writes nothing | **[comment on #109](https://github.com/leodip/goiabada/issues/109#issuecomment-5197711587)** | Two reachability claims in the draft were **wrong and are removed**. The admin console's account menu does *not* take this path: it mints a signed `id_token_hint` through `/api/v1/account/logout-request` and lands on `doLogoutWithIdToken`, which does tear down. And no first-party UI reaches the no-hint POST at all, since the auth server has no logout link of its own. The reachable callers are an RP omitting `id_token_hint`, which is conforming, and a direct request |
| 3, the stale `amr` | **#140** (`bug`, `security`, `go`) | Unchanged in substance. Severity stated as low: `acr` stays correct, so the token is internally inconsistent rather than silently over-claiming, and the narrow reachability the draft gave was confirmed, since the `AuthMethods` copy sits inside the `hasValidUserSession` block and an idle-expired session therefore leaks nothing |
| 4, `ClearAuthContext` after the response | **#141** (`bug`, `security`, `go`) | All seven sites re-read and confirmed, as were the two success paths that get the order right. Body gains a second defect on the same lines: two of the seven also miss a `return` after `httpHelper.InternalServerError`, so a failed redirect writes a 500 and falls through |

Each was checked for duplicates against the open and closed tracker. #139 is distinct from #131 and
#132 (session-scoped code issuance, versus a user-scoped generation sweep and a family-scoped
containment cascade), and #134 held its previous shape but was closed as moot when decision 4 moved
the marker onto `codes.revoked`, so nothing tracked it. Nothing at all tracked #140 or #141.

Refs #129

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,6 +19,9 @@
 tmp/
 .devdata/
 
+# Agent review mailbox (throwaway ping-pong between Claude Code and Codex)
+.review/
+
 # Go workspace file
 go.work
 
--- a/docs/issue-129-durable-session-termination.md
+++ b/docs/issue-129-durable-session-termination.md
@@ -1,0 +1,1042 @@
+# Issue 129: ending a user session does not durably cut off access
+
+**Issue:** [#129](https://github.com/leodip/goiabada/issues/129)
+**Issue state:** open (labels: bug, security)
+**Written:** 2026-08-04
+**Last synced:** 2026-08-04 (issue has zero comments; body is the whole specification)
+**Agreement sealed:** 2026-08-04
+**Run state:** not started
+**PR:** none
+
+**Related:** #106 (closed) built the per-user generation boundary and the revocation helper this
+work sits next to. #128 (closed) built the atomic refresh claim and family containment. #77
+(closed) built the atomic code claim. #127 (closed, not planned) proposed a grant entity. #130
+(open) owns the used-code cleanup NULL bug this issue's body flagged. #131 and #132 (open) are
+residuals of #106 and #128 in the same code. #133, #134, #135 and #137 (open) were filed out of an
+earlier, discarded attempt at this issue and cite decision numbers from a document that no longer
+exists; see "Prior attempt" below. None of them blocks this work.
+
+**Prior attempt.** An earlier `/leo-spec` plus `/leo-run` cycle for #129 reached stage 4 and was
+discarded by `git reset --hard origin/main` on 2026-08-04. The user chose to start over from zero.
+That document, its 14 decisions and its 7 stages are deliberately **not** consulted here, and every
+claim below is re-derived from the code as it stands on `main` at `f8d093f`. The four issues filed
+out of that attempt survive it, but their `decision N` citations point at numbering that no longer
+exists and will need reconciling against this document's section 3.
+
+---
+
+## 0. Code anchors
+
+| Label | File | Function | Locate by | Note |
+|---|---|---|---|---|
+| `admin/session-delete` | `src/authserver/internal/handlers/apihandlers/handler_api_users_sessions.go` | `HandleAPIUserSessionDelete` | `err = database.DeleteUserSession(nil, sessionId)` | termination site 1, bare delete, no revocation |
+| `account/session-delete` | `src/authserver/internal/handlers/apihandlers/handler_api_account_sessions.go` | `HandleAPIAccountSessionDelete` | `if err := database.DeleteUserSession(nil, sessionId); err != nil {` | termination site 2, ownership checked, bare delete |
+| `logout/last-client-delete` | `src/authserver/internal/handlers/handler_account_logout.go` | `handleExistingSessionOnLogout` | `if len(userSession.Clients) == 1 {` | termination site 3, deletes the session only when the last client logs out |
+| `completed/start-new-session` | `src/authserver/internal/handlers/handler_auth_completed.go` | `HandleAuthCompletedGet` | `newSession, err := userSessionManager.StartNewUserSession(` | gap 3: reached unconditionally, no proof anyone authenticated |
+| `completed/really-authenticated` | `src/authserver/internal/handlers/handler_auth_completed.go` | `HandleAuthCompletedGet` | `userReallyAuthenticated := authContext.AuthenticatedAt != nil` | computed, then read only inside the valid-session branch |
+| `completed/set-acr-level` | `src/authserver/internal/handlers/handler_auth_completed.go` | `HandleAuthCompletedGet` | `err = authContext.SetAcrLevel(targetAcrLevel, userSession)` | runs after the branch, on the ambient session, whoever it belongs to |
+| `issue/sid-from-context` | `src/authserver/internal/handlers/handler_auth_issue.go` | `HandleIssueGet` | `sessionIdentifier = r.Context().Value(constants.ContextKeySessionIdentifier).(string)` | the sid stamped on the code comes from the cookie, not the AuthContext |
+| `issue/create-code` | `src/authserver/internal/handlers/handler_auth_issue.go` | `HandleIssueGet` | `code, err := codeIssuer.CreateAuthCode(createCodeInput)` | the insert gap 3 has to stop; no session lookup precedes it |
+| `pwd/authenticated-at` | `src/authserver/internal/handlers/handler_auth_pwd.go` | `HandleAuthPwdPost` | `authContext.AuthenticatedAt = &utcNow` | the only writer on the level 1 path, so the discriminator for gap 3 |
+| `authorize/prompt-login` | `src/authserver/internal/handlers/handler_authorize.go` | `HandleAuthorizeGet` | `if authContext.HasPromptValue("login") {` | returns before the session lookup, leaving the cookie session live (#133) |
+| `authorize/hint-mismatch` | `src/authserver/internal/handlers/handler_authorize.go` | `HandleAuthorizeGet` | `if authContext.IdTokenHintSub != "" && userSession.User.Subject.String() != authContext.IdTokenHintSub {` | second entry to the same defect, from inside the valid-session block |
+| `authorize/sso-reuse` | `src/authserver/internal/handlers/handler_authorize.go` | `HandleAuthorizeGet` | `authContext.AuthState = oauth.AuthStateLevel1ExistingSession` | the SSO path that never sets `AuthenticatedAt` |
+| `authcontext/generation` | `src/core/oauth/auth_context.go` | `n/a` | `AuthStateGeneration int64` | the AuthContext carries no session identifier of its own |
+| `sessionmgr/start-new` | `src/core/user/usersession_manager.go` | `StartNewUserSession` | `sess.Values[constants.SessionKeySessionIdentifier] = userSession.SessionIdentifier` | rewrites the cookie, so `/auth/issue` sees the replacement sid |
+| `sessionmgr/bump-no-userid` | `src/core/user/usersession_manager.go` | `BumpUserSession` | `if authMethods != "" && authMethods != userSession.AuthMethods {` | assigns methods and ACR, never `UserId` (#133 consequence 2) |
+| `issuer/grant-is-offline` | `src/core/oauth/token_issuer.go` | `grantIsOffline` | `func grantIsOffline(authorizedScope string, sessionIdentifier string) bool {` | an empty sid alone makes a grant offline |
+| `issuer/refresh-type-branch` | `src/core/oauth/token_issuer.go` | `generateRefreshToken` | `if grantIsOffline(scope, code.SessionIdentifier) {` | Offline stores a max lifetime, Refresh stores the sid, never both |
+| `issuer/store-sid` | `src/core/oauth/token_issuer.go` | `generateRefreshToken` | `refreshTokenEntity.SessionIdentifier = claims["sid"].(string)` | so an offline token's own `session_identifier` is empty |
+| `issuer/create-refresh-token` | `src/core/oauth/token_issuer.go` | `generateRefreshToken` | `err := t.database.CreateRefreshToken(nil, refreshTokenEntity)` | gap 2: the child insert, outside any transaction |
+| `issuer/suppress-sid-claim` | `src/core/oauth/token_issuer.go` | `generateAccessTokenCore` | `if len(input.SessionIdentifier) > 0 && !input.GrantIsOffline {` | an offline grant's ACCESS token carries no sid either |
+| `validator/authcode-preauth` | `src/core/validators/token_validator.go` | `ValidateTokenRequest` | `if !wasReused {` | user, generation and expiry checks, all ahead of client auth (#137) |
+| `validator/code-generation` | `src/core/validators/token_validator.go` | `ValidateTokenRequest` | `if codeEntity.AuthStateGeneration != codeEntity.User.AuthStateGeneration {` | #106's redemption gate, the precedent for a code-level marker |
+| `validator/session-branch` | `src/core/validators/token_validator.go` | `ValidateTokenRequest` | `case "Refresh":` | already fails once the session row is gone |
+| `validator/offline-branch` | `src/core/validators/token_validator.go` | `ValidateTokenRequest` | `case "Offline":` | checks only the max lifetime, never the session |
+| `validator/refresh-generation` | `src/core/validators/token_validator.go` | `ValidateTokenRequest` | `if refreshToken.AuthStateGeneration != refreshToken.Code.User.AuthStateGeneration {` | sits ahead of the client-ownership check below it |
+| `validator/client-ownership` | `src/core/validators/token_validator.go` | `ValidateTokenRequest` | `if tokenClientId != client.Id {` | the ownership gate a new refresh check should sit behind |
+| `db/refresh-by-sid` | `src/core/data/commondb/refresh_token.go` | `GetRefreshTokensBySessionIdentifier` | `selectBuilder.JoinWithOption(sqlbuilder.InnerJoin, "codes", "codes.id = refresh_tokens.code_id")` | joins through `codes`, so it does reach offline tokens |
+| `db/mark-code-used` | `src/core/data/commondb/code.go` | `MarkCodeAsUsed` | `ub.Equal("used", false),` | the compare-and-set predicate, no revoked term today |
+| `db/delete-used-codes` | `src/core/data/commondb/code.go` | `DeleteUsedCodesWithoutRefreshTokens` | `deleteBuilder.Equal("used", true),` | reaps used codes only; #130 owns its NULL bug |
+| `revoke/refresh-tokens` | `src/authserver/internal/handlers/revocation.go` | `revokeRefreshTokens` | `if err := db.UpdateRefreshToken(tx, rt); err != nil {` | #106's helper, read-modify-write rather than compare-and-set |
+| `middleware/session-check` | `src/authserver/internal/middleware/api_auth.go` | `RequireValidSession` | `session, err := database.GetUserSessionBySessionIdentifier(nil, sid)` | reached only when the token carries a sid |
+| `worker/perform-task` | `src/authserver/internal/workers/background_worker.go` | `performTask` | `func (w *Worker) performTask(ctx context.Context) {` | where any reaper for new durable state would go |
+| `schema/user-sessions` | `src/core/data/sqlitedb/schema.sql` | `n/a` | `CREATE TABLE user_sessions (` | documentation snapshot, not loaded by Go |
+| `test/fresh-ceremony-after-delete` | `src/authserver/tests/integration/session_deletion_test.go` | `TestSessionDeletedDuringAuthFlow_LoginSucceeds` | `assert.Equal(t, 1, len(userSessions2), "Should have a new session after second login")` | #46's regression guard; the case gap 3's fix must NOT break |
+| `test/completed-new-session` | `src/authserver/internal/handlers/handler_auth_completed_test.go` | `TestHandleAuthCompletedGet` | `t.Run("Successful flow, new session, consent not required", func(t *testing.T) {` | one of three subtests reaching `StartNewUserSession` with a nil `AuthenticatedAt` |
+
+Counts in this document carry their command:
+
+- 3 termination sites (`admin/session-delete`, `account/session-delete`, `logout/last-client-delete`).
+- 3 subtests reach `StartNewUserSession`
+  (`grep -c 'StartNewUserSession", rr, req' src/authserver/internal/handlers/handler_auth_completed_test.go`).
+- Next free migration number is **000026** on all four engines
+  (`for db in sqlitedb mysqldb postgresdb mssqldb; do ls src/core/data/$db/migrations | grep -o '^[0-9]*' | sort -u | tail -1; done`).
+
+---
+
+## 1. Context
+
+### The three termination sites
+
+All three delete the `user_sessions` row and revoke nothing. Each passes `nil` for the
+transaction, so none of them is a boundary against anything running concurrently.
+
+| Site | What it does | Revokes tokens? |
+|---|---|---|
+| `DELETE /api/v1/admin/user-sessions/{id}` (`admin/session-delete`) | deletes the `UserSession` | no |
+| `DELETE /api/v1/account/sessions/{id}` (`account/session-delete`) | deletes the `UserSession`, ownership checked | no |
+| `GET/POST /auth/logout` (`logout/last-client-delete`) | deletes one `UserSessionClient`, and the `UserSession` when it was the only one | no |
+
+The logout case is per-client and only incidentally ends a session. The predicate is
+`len(userSession.Clients) == 1`, evaluated after the client row is deleted, so the session dies
+exactly when the departing client was the sole one on it.
+
+### Gap 1: an offline grant outlives its session
+
+Session deletion already contains a **session-bound** refresh token. `validator/session-branch`
+loads the session by identifier and rejects with `invalid_grant` when the row is missing.
+
+It does not contain an **offline** one. `validator/offline-branch` checks only the
+`offline_access_max_lifetime` claim and never consults the session. Offline refresh tokens also
+carry no session identifier of their own: `issuer/refresh-type-branch` chooses one of two storage
+shapes, and `issuer/store-sid` sets `session_identifier` on the Refresh branch only, so the Offline
+branch leaves it empty and stores a `max_lifetime` instead.
+
+So after "end session", a client holding an `offline_access` grant that was authorized through that
+session keeps refreshing, bounded only by that grant's max lifetime.
+
+The join that reaches those rows does exist. `db/refresh-by-sid` inner-joins `codes` on
+`codes.id = refresh_tokens.code_id` and filters `codes.session_identifier`, and the **code** row
+does carry the sid even for an offline grant, because `issuer/grant-is-offline` returns true from
+the `offline_access` scope rather than from an empty sid. ROPC tokens have `code_id = NULL` and are
+correctly excluded, having no session to belong to.
+
+**Access tokens are affected too, and this is not in the issue body.** `issuer/suppress-sid-claim`
+omits the `sid` claim for an offline grant. `middleware/session-check` reaches its session lookup
+only for a token that has a sid, so an offline grant's access token is checked against the user's
+generation instead. Session termination does not move that generation, and must not, so an offline
+access token issued before termination stays acceptable at Goiabada's own endpoints until it
+expires. Revoking refresh tokens alone leaves that window open.
+
+### Gap 2: a concurrent refresh inserts a surviving replacement
+
+Rotation claims the presented token and inserts its replacement as two separate commits.
+`issuer/create-refresh-token` passes `nil` for the transaction. A refresh that validated before a
+termination sweep therefore inserts its child after the sweep commits, and for an offline grant
+nothing invalidates that child.
+
+> **Correction to the issue body.** It describes rotation as marking the parent revoked with
+> `UpdateRefreshToken(nil, ...)`. #128 replaced that with a compare-and-set,
+> `MarkRefreshTokenAsRevoked`. The race the body describes survives the change, because it is the
+> child *insert* that escapes, not the parent claim, but the citation is stale.
+
+`revoke/refresh-tokens` is #106's sweep helper and is a read-modify-write over rows it already
+loaded, not a compare-and-set. It skips rows already marked revoked and reports only the JTIs it
+transitioned, which is a contract #77's reuse cascade depends on.
+
+**Gap 2 may cost nothing beyond gap 1, depending on decision 4.** The issue's sketch closes it with
+a `codes.revoked` column plus an edit to `db/mark-code-used`. Verified that a cheaper route exists:
+
+- A rotated child **inherits its parent's `code_id`**. `issuer/create-refresh-token` sets `CodeId`
+  from the `code` it was handed, and on the refresh path that code is
+  `validateResult.CodeEntity`, which the validator derives as `codeEntity = &refreshToken.Code`.
+  So every descendant of a grant points at the same `codes` row.
+- The refresh validator therefore **already holds the originating session identifier** for every
+  auth-code-derived refresh token, offline ones included, as `refreshToken.Code.SessionIdentifier`.
+  No new join and no new column are required to read it.
+
+A refresh-path check keyed on that value rejects every present *and future* descendant of the
+grant, because the racing child inherits the same code and so the same sid. Under a design that
+marks the *session* rather than the *rows*, gap 2 closes as a consequence of gap 1's check rather
+than as separate work, and `db/mark-code-used` needs no edit. This is a claim about the refresh
+path only; an unredeemed code still needs its own gate, which is decision 7.
+
+### Gap 3: an in-flight ceremony recreates the session
+
+Verified sequence, and the reason a marker on existing rows is not by itself a boundary.
+
+1. An SSO ceremony validates session S at `authorize/sso-reuse`, copies its `UserId`, `AcrLevel`,
+   `AuthMethods` and `AuthStateGeneration` onto the AuthContext, and proceeds. It never reaches the
+   password handler, so `AuthenticatedAt` stays nil.
+2. "End session" deletes S.
+3. The ceremony resumes at `/auth/completed`.
+4. `HandleAuthCompletedGet` finds no valid session and takes the else branch, which calls
+   `completed/start-new-session` **unconditionally**, from `authContext.UserId`, with no requirement
+   that anyone authenticated. `completed/really-authenticated` computes the very fact that would
+   distinguish the two cases and is consulted only inside the *valid session* branch, to refresh
+   `AuthTime`.
+5. `sessionmgr/start-new` writes the replacement session's identifier into the cookie.
+6. `/auth/issue` reads the sid from the request context at `issue/sid-from-context`, which is now
+   the replacement, and `issue/create-code` inserts a fresh, unmarked code.
+7. Nothing rejects it. The code is new, so no marker on old rows applies, and the user's generation
+   never moved.
+
+The consent screen delays code insertion the same way, widening the window from milliseconds to
+however long a person takes to click.
+
+**The AuthContext carries no session identifier.** `authcontext/generation` is the neighbouring
+field; there is no sid field beside it. The ceremony's originating session is knowable only from
+the cookie, and the cookie is rewritten in step 5. Nothing durable ties the code issued in step 6
+back to the session that was terminated in step 2.
+
+> **Departure from the issue body.** It calls the fix "almost certainly: fail and force
+> re-authentication". Verification says the predicate cannot be "no valid session", because
+> `test/fresh-ceremony-after-delete` covers a legitimate case with exactly that shape: session
+> deleted, then a **new** ceremony in which the user really does enter a password, which must keep
+> succeeding. That is #46's regression guard. The discriminator that separates the two is whether
+> *this ceremony* performed level 1 authentication, which is what
+> `completed/really-authenticated` already computes and `pwd/authenticated-at` is the only level 1
+> writer of.
+
+### Costs already visible in the tests
+
+Three subtests in `test/completed-new-session`'s file drive the else branch with a nil
+`AuthenticatedAt` and assert `StartNewUserSession` is called. Any gate on
+`completed/really-authenticated` breaks all three, so the gap 3 work is **not** additive at the
+unit tier.
+
+### What #106 already built, and why it does not extend
+
+#106 added a per-user `auth_state_generation`, checked at `validator/code-generation`,
+`validator/refresh-generation` and `middleware/session-check`, plus `RevokeUserAuthState` and
+`revoke/refresh-tokens` in `revocation.go`. The generation is the part that survives a race,
+precisely because a rotated child inherits its parent's generation rather than the user's current
+value.
+
+It cannot be reused here. Incrementing a user-wide counter to sign out one device invalidates every
+other device that user has, which is the opposite of what the action means. The sweep helper is
+reusable; the boundary is not.
+
+### The pre-authentication oracle in the same block
+
+`validator/authcode-preauth` runs the user-enabled check, the generation check and the expiry check
+**before** client-secret validation and before PKCE. So a presenter holding a stolen code learns
+whether the account's generation moved without proving anything. This is live today and #137 owns
+it. It matters here because a new redemption check placed in that block inherits the same leak.
+
+On the refresh path, `validator/refresh-generation` likewise sits ahead of
+`validator/client-ownership`.
+
+### Test landscape
+
+- **Style.** `testify/assert` throughout, `t.Run` subtests rather than table structs in the handler
+  tests, `mockery` mocks under `src/core/data/mocks` and `mocks_*` packages.
+- **Unit.** `handler_auth_completed_test.go` is a single `TestHandleAuthCompletedGet` with 10
+  subtests. `token_validator_test.go` is 33 tests over 5168 lines. `revocation_test.go` has 8.
+- **No unit tests exist for either session-delete endpoint.** Neither
+  `handler_api_users_sessions_test.go` nor `handler_api_account_sessions_test.go` exists. Absent
+  infrastructure, so whichever stage touches those handlers creates the file.
+- **Data.** `src/authserver/tests/data/` runs all four engines and holds per-migration tests
+  (`migration_000024_auth_state_generation_test.go`, `migration_000025_family_index_test.go`) as the
+  precedent for a new migration.
+- **Integration.** `session_deletion_test.go` (the #46 guard), `session_bearer_revocation_test.go`
+  (3 tests pinning that a deleted session rejects bearer tokens),
+  `token_refresh_concurrent_test.go` and `token_authcode_concurrent_test.go` (the double-spend
+  harness a gap 2 test would mirror), `api_users_sessions_test.go`, `api_account_sessions_test.go`,
+  `authorize_existing_session_test.go`.
+
+### Documentation
+
+`site/src/content/docs/concepts/user-sessions.mdx` documents credential changes and live sessions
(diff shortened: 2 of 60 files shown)


leodip:
## Blocking decision: does the `/auth/completed` gate require a password, or is OTP enough?

Stage 5 is implemented, its tests are green on every tier it claims, and the Codex review re-ran all of
them. The review then found one thing, on the security axis, and I have confirmed it against the code.
It is a question I am not allowed to answer alone: the run never auto-resolves anything in
authentication or session territory, and never settles a security finding by itself. So stage 5 is
**uncommitted** and waiting on you.

This is agreement decision 15, and it is the run's first blocking escalation.

### What stage 5 built

Decision 6's gate. `HandleAuthCompletedGet`'s no-valid-session branch used to call
`StartNewUserSession` unconditionally, so a ceremony whose session was ended mid-flight silently
recreated it. Stage 5 gates that on `!userReallyAuthenticated`, exactly as decision 6 specifies, reading
the `AuthenticatedAt` field the handler already computed.

### The problem

`AuthenticatedAt` has **two** writers, not one. `handler_auth_pwd.go` sets it, and so does
`handler_auth_otp.go`. So a ceremony that reached OTP by SSO reuse, never touching the password form,
satisfies the gate on OTP alone and recreates the session that was just explicitly terminated.

I verified all five hops rather than taking the reviewer's word for it:

1. `HandleAuthorizeGet`'s SSO branch copies the session's user and auth state onto the `AuthContext` and
   leaves `AuthenticatedAt` nil.
2. `HandleAuthLevel1CompletedGet` sends that ceremony to `/auth/level2` when the target ACR is higher
   than the session's, which is the ordinary step-up.
3. `HandleAuthLevel2Get` sets `AuthStateLevel2OTP` for `level2_mandatory`, and for `level2_optional`
   when the user has OTP. No password is needed to arrive, and no session liveness is re-read.
4. **The session is ended here.** `HandleAuthOtpPost` checks only that the state is
   `AuthStateLevel2OTP`. It reads no session row, so the termination is invisible to it. On success it
   sets `AuthenticatedAt = now` and redirects to `/auth/completed`.
5. `HandleAuthCompletedGet` finds no session row, computes `userReallyAuthenticated` as **true**, passes
   the gate, and calls `StartNewUserSession`. `HasValidUserSession` and `UserSessionLoadUser` both
   handle the nil session cleanly, so nothing intervenes with a 500.

**The sub-case that sets the severity.** For a user **without** OTP on a `level2_mandatory` client,
`HandleAuthOtpGet` generates a fresh TOTP secret, renders it into the enrollment page, and stores it in
the http session. The browser is handed the secret it will be tested on. So a holder of a stolen session
cookie can produce a valid code with no password and no second factor of their own, and ending the
session does not stop that ceremony.

### Why I cannot just pick an answer

The sealed text disagrees with itself about this exact ceremony, which is why it is yours to settle.

- **Decision 6's supporting facts** name both writers and call them "both real authentication". Read
  that way, stage 5 is correct as built and there is nothing to fix.
- **Section 1's departure note** says the discriminator is whether this ceremony performed **level 1**
  authentication, and that `pwd/authenticated-at` "is the only level 1 writer" of the field. Read that
  way, stage 5 is incomplete, and goal 1's "cannot be undone by anything already in flight" stays false
  for this ceremony.

Note that "the run built what was sealed" is true and still does not dispose of it. A conformance
defence does not answer a security axis: the gate fails open on a reachable path either way.

### Options

**A. A dedicated level 1 proof on the `AuthContext`. Recommended.**
A new field written only by `handler_auth_pwd.go`. The gate reads that; `AuthenticatedAt` keeps its
existing job of refreshing `AuthTime` after a step-up. Makes the gate mean what section 1 says it means,
in one place, at the line a later reader is looking at.

- Cost: one field on a shape already persisted as JSON in the cookie, one write, the gate expression,
  two unit rows at seam 6 and one integration case in `session_deletion_test.go`.
- Covers any future route that reaches `/auth/completed` without a password, not only the OTP one.
- Upgrade wrinkle, small: a ceremony already past the password form when the binary is replaced has the
  field nil, so it gets sent back to login once.

**B. A liveness read in the OTP handler.**
At the top of `HandleAuthOtpPost`, `AuthenticatedAt == nil` means this ceremony did no level 1, so
require the bound session to still resolve before accepting the OTP.

- Cost: no new field, one session lookup in a handler that does none today.
- Weaker: the gate keeps reading a field that does not mean what it says, and the reasoning now lives in
  two handlers, so a third writer of `AuthenticatedAt` would need both updated.

**C. Accept it and document the limit.**
Defensible on decision 6's literal text, and cheapest. I do not recommend it: the enrollment sub-case
makes it a fail-open reachable with a stolen cookie alone, which is the shape #129 exists to close.

**D. Defer to a follow-up issue.** Ruled out rather than weighed. Gap 3's first half is what goal 2 and
section 4.4 ask for, and drafting a follow-up for work the agreement owes would narrow this run's scope
and then report success.

### What either mistake costs

Choosing A or B when C was right costs one field or one query, plus one extra password entry for an
honest user in the rare race where their session expires mid-step-up. That outcome is arguably correct
anyway: a session created from OTP alone, in a ceremony where no password was entered, is the thing in
question.

Choosing C when A was right leaves explicit termination defeatable by an in-flight ceremony, on a path a
stolen cookie reaches, which is the defect this issue exists to close.

Worth saying plainly: **the legitimate step-up is untouched by any option.** When the session is still
valid the handler takes the `hasValidUserSession` branch and never consults this gate. Only the
no-valid-session branch changes. `TestSessionDeletedDuringAuthFlow_LoginSucceeds`, #46's guard, enters a
password and stays green under A and B.

### State of the run

- Stages 1 to 4 are committed and pushed. The behaviour change has landed.
- **Stage 5 is implemented and uncommitted**, deliberately, because the gate expression is the subject
  of this decision and nothing resting on an open decision gets committed. The working tree holds
  `handler_auth_completed.go`, `handler_auth_completed_test.go`, `handler_auth_otp.go`,
  `handler_auth_pwd.go` and `concepts/user-sessions.mdx`.
- Its tiers were green and the reviewer re-ran them: modules on three modules, integration on all four
  engines, 59 anchors, 42 site pages.
- Stages 6, 7 and 8 have not started. Stage 6 owns the other half of gap 3 and is unaffected by this.

Answer here and the run picks it up: it flips decision 15 to `Decided` with your reasoning, applies it
inside stage 5, extends the case table, and opens round 2 at the same gate.

leodip:
Let's go with option A.

leodip:
Applied. Decision 15 is `Decided` with your words, and stage 5 now carries option A.

**What landed:** `AuthContext.Level1AuthCompleted`, a bool written only by `handler_auth_pwd.go` and
read by the gate in `handler_auth_completed.go`. `AuthenticatedAt` keeps its existing job of deciding
whether to refresh the session's `AuthTime` on a step-up, and `handler_auth_otp.go` gained a comment
saying it deliberately does not set the new field, with a test assertion pinning that so the asymmetry
does not get tidied away later.

Three comments changed rather than two, and one is a reversal of what the first stage 5 pass did: it had
widened the comments on both `AuthenticatedAt` writers to say that field also decides whether a session
may be created, which option A makes false. They are back to their original wording, which is accurate
again. The comment that really was falsified is the one at the computation itself, which claimed the
field is "set by the password handler"; it now names both writers.

**Tests:** the zero-time `AuthenticatedAt` subtest was replaced rather than kept beside the new one. It
pinned an `!IsZero()` spelling of a discriminator this gate no longer reads, and the new OTP subtest
catches the same mistake through a shape the code can actually reach. Plus one assertion each in the
existing password and OTP handler subtests to pin who writes the field, and
`TestSessionEndedDuringStepUp_OtpAloneDoesNotRecreateTheSession` at the integration tier, sitting beside
#46's guard in the same file, varying exactly one thing: whether a password was entered.

**Verified on the amended tree:** modules green on three modules, integration green on all four engines,
63 anchors resolving. Five mutations with disjoint failure sets, and the one that matters is the round 1
gate restored: at the integration tier it issues a code to the client and recreates the session, while
#46's guard stays green. So the defect reproduces and the fix closes it end to end.

**The cost you accepted shows up as expected:** a user whose session expires while they are on the OTP
form enters a password once. The legitimate step-up on a live session is untouched, which the
pre-existing re-auth subtest pins.

Stage 5 is still uncommitted, deliberately: the gate changed after round 1 read it, so review round 2 is
opening on the current tree. Stages 6, 7 and 8 follow after that.

leodip:
## Blocking decision: for `prompt=none`, does `/auth/issue` show a login form or return `login_required`?

Stage 6 is implemented, green on all three tiers, and the Codex review re-ran every one of them
independently on all four engines. It came back with one finding, on the conformance axis, and I have
confirmed it against the code. It contradicts a sealed decision, so I am not allowed to answer it
alone: decision 6 does not merely omit this case, it decides it, and the run never auto-resolves
anything in authentication, session or public-interface territory. So stage 6 is **uncommitted** and
waiting on you.

This is agreement decision 16, and it is the run's second blocking escalation. Both came out of decision
6, which is worth knowing for next time.

### The defect

Stage 6 built the liveness refusal exactly as decision 6 words it: when the ceremony's session
identifier no longer resolves to a session row, refuse to mint a code, set `AuthStateRequiresLevel1`,
redirect to `/auth/level1`. For an interactive ceremony sitting on the consent screen that is right, and
it is what the new integration test asserts.

For a `prompt=none` ceremony it renders a password form, and `prompt=none` must show no UI at all.

Four hops, each checked against the code rather than reasoned about:

1. `handlePromptNone` runs its silent checks, bumps the session, and 302s to `/auth/issue`. Every one of
   its own failures goes back to the client through `redirToClientWithError`, and `login_required` is
   what it sends when its session lookup finds no row.
2. The session is ended between that 302 and the browser following it.
3. `MiddlewareSessionIdentifier` puts the identifier in the request context only when the row exists, so
   `/auth/issue` reads the empty string. Same shape as the consent-screen case, which is the whole
   reason the check fires at all.
4. The new branch redirects to `/auth/level1`, which redirects to `/auth/pwd`, which renders the form.
   Nothing in between knows the request forbids UI: `HasPromptValue` has exactly three production
   callers, in `handler_authorize.go`, `handler_auth_completed.go` and `handler_consent.go`, and none is
   on this path.

So a silent-renewal iframe gets a login page and no error at all, and waits for its own timeout instead
of reading `login_required`. `concepts/prompt-parameter.mdx` says of `none`: "Silent authentication. Must
not display any UI. Returns an error if the user is not authenticated or consent is not available." OIDC
Core 3.1.2.1 says the same. The window is narrow, one redirect hop, but it is a window this issue
introduces.

`HandleIssueGet` already answers a neighbouring case the other way, a few lines above the new check: its
`id_token_hint` mismatch branch returns `login_required` through `redirToClientWithError` and clears the
auth context.

### Why I can't just fix it

The fix is about as forced as they come and I recommend it. But decision 6 names `prompt=none` as the
other entry to this exact sub-case, says restart level 1 there, and explicitly **rejects** "failing to
the client with `login_required` or `access_denied`". Reading that rejection as scoped to interactive
ceremonies would be paraphrasing your decision rather than applying it, and overriding a sealed
rejection in authentication territory is your call, not mine.

### Options

**A, recommended: return `login_required` for a `prompt=none` ceremony, keep the level 1 restart for
every other one.** `redirToClientWithError` with `ErrorLoginRequired` then `ClearAuthContext`, copying
the `id_token_hint` branch directly above it. No code is minted either way, so the fail-open decision 6
closes stays closed, and the client learns what happened. Decision 6's reasoning survives intact where
it applies: it rejected `login_required` because it costs an interactive user "a round trip through the
client to arrive at the same password form", and a `prompt=none` client has no password form to arrive
at and is forbidden from showing one. Cost: one branch, one row in `TestHandleIssueGet`, one integration
case, cheap because `codeFromSameSession` already drives `prompt=none` to the `/auth/issue` redirect and
follows it on the next line, so the case ends the session between those two lines.

**B: restart level 1 for everything, as decision 6 words it, and document the divergence.** Keeps stage
6 exactly as built, no code at all. It means editing `concepts/prompt-parameter.mdx` to carve out an
exception to "Must not display any UI", and it leaves a silent-renewal client with no error to act on.

**C: mint the code for `prompt=none` and let the compensating statement handle it.** Ruled out rather
than weighed, and the reason is measured: on this path the identifier is **empty**, and
`RevokeCodeIfSessionGone` refuses an empty identifier at entry, so the compensating call errors and the
handler answers 500. A 500 in the iframe is worse than A or B on every axis.

**D: return `login_required` for every ceremony, reversing decision 6's rejection outright.** Simpler
than A, one branch and no `Prompt` read, and `/auth/issue` then behaves uniformly. It discards a ceremony
that could have completed correctly, which is what decision 6 weighed and rejected, and it reddens the
new consent-screen test, whose assertion is that the ceremony reaches `/auth/pwd`.

### What each mistake costs

Answering A when B was right costs one branch and one test row, and leaves the interactive behaviour you
chose untouched. Answering B when A was right ships an authorization server that renders a login form in
response to a request that forbids one, in a window #129 itself introduces, and the client cannot tell
that from a hang. Both are recoverable in one stage. The asymmetry is that B's cost lands on integrators
and A's lands on this run.

### Where things stand

Stages 1 to 5 are committed and pushed. Stage 6's thirteen-file diff is complete and green on modules,
data on all four engines, and integration on all four engines, but uncommitted, exactly as stage 5
waited on decision 15. The agreement is committed and carries decision 16 `Open` plus the full round 1
record in section 7, including everything the review found sound, so nothing is lost if this session
does not resume.

Reply with the option and I'll apply it inside stage 6, then take it back through review.

leodip:
Let's go with option A again.

leodip:
Applied. Decision 16 is `Decided` with your words, and stage 6 now carries option A.

**What landed:** one branch at the head of the existing refusal in `HandleIssueGet`. When the
ceremony's session is gone and it carries `prompt=none`, it goes back to the client with
`login_required` through `redirToClientWithError` and the auth context is cleared, copying the
`id_token_hint` branch a few lines above it in the same function. Every other ceremony still restarts
level 1, so decision 6's rejection of `login_required` ends up scoped rather than reversed: its stated
reason was that failing to the client costs a user "a round trip through the client to arrive at the
same password form", and a `prompt=none` client has no password form to arrive at.

**One judgement inside your answer, recorded rather than made silently.** The error description is
`handlePromptNone`'s own wording for this condition, "User authentication is required", rather than
something naming the mid-flight termination. The client cannot act differently on the two cases, and a
distinct description would only tell a caller which internal hop refused it. It has a useful side
effect: `concepts/prompt-parameter.mdx` stays accurate unedited, down to the `error_description` in
its own worked example, so the docs carve-out option B would have cost is avoided rather than just
deferred.

**Tests.** One row at `TestHandleIssueGet`, bringing it to 8 where `main` has 3, and
`TestSessionEndedBeforeIssue_PromptNoneGetsLoginRequired` at the integration tier, the fourth ceremony
in `session_deletion_test.go`. It uses `createSessionWithAcrLevel1` for the live session and ends it
between `assertRedirect` and `loadPage`, which is the entire window a silent ceremony has, since it
never waits for a person.

**Docs.** A third sentence in `concepts/user-sessions.mdx`, next to the two that both say "sent back
to the login page": an application renewing quietly with `prompt=none` is told rather than shown, and
gets `login_required` back. Without it a reader generalizes the login-page outcome to a background
renewal, which is exactly the outcome your answer rejected for that shape.

**Verified on the amended tree, re-run in full rather than narrowed to the change:** modules green on
three modules; data green on all four engines run separately; integration green on all four engines,
889 passing test lines per engine and zero failures, with all four of the file's ceremonies passing
once per engine; 74 anchors resolving; `npm run build` at 42 pages.

**Two mutations, and they are the pair that matters here.** Removing the new branch, which is stage 6
exactly as the review read it and is your option B, fails precisely the new unit row and the new
integration case while the three sibling ceremonies stay green. Making the branch unconditional, which
is option D, fails the two interactive unit rows and `TestSessionEndedOnConsentScreen_NoCodeIsIssued`
while the new case passes. Disjoint failure sets, so the branch is pinned as keyed on the prompt
rather than as replacing the refusal, and D's cost is measured rather than predicted.

Stage 6 is still uncommitted, deliberately: the branch the review found missing is now present, so
round 1's verdict no longer describes this tree. Review round 2 is opening on it, and the code commit
follows that. Stages 7 and 8 follow after.

leodip:
## Deferred decisions: none

The run deferred nothing, so there is no list of assumptions here to check. That is the first thing to
read on this PR and it is worth one paragraph of explanation rather than a bare "n/a", because an empty
section reads the same as a forgotten one.

`/leo-run` may settle a question on its own only when it is derivable from a sealed decision, forced by
the code, or local and reversible, and it may never settle one in authentication, token, session,
cryptography, schema or public-interface territory whatever the demonstration looks like. Every question
this run met was in that forbidden territory, so nothing was available to defer on a stated assumption.
Two of them stopped the run and came to you instead:

**Decision 15, raised at stage 5's review gate.** Decision 6 fixed the gate as
`!hasValidUserSession && !userReallyAuthenticated`, and stage 5 shipped exactly that. The review found
`AuthenticatedAt` has two writers, not one: the OTP handler sets it too, so a ceremony that reached
level 2 by SSO reuse and never touched a password form satisfied the gate on OTP alone and recreated the
session that had just been terminated. Worse in the enrollment sub-case, where the server hands the
browser the secret it is about to be tested on, so a stolen session cookie alone was enough. You
answered option A on 2026-08-05: a dedicated level 1 proof on the `AuthContext`, written only by the
password handler, with `AuthenticatedAt` keeping its existing job. Built as answered, at
`authcontext/level1-proof` and `completed/level1-restart`.

**Decision 16, raised at stage 6's review gate.** Stage 6 refuses to mint a code when the ceremony's
session is gone, and restarts level 1. For a `prompt=none` ceremony that restart is a contradiction: the
specification forbids showing that request a login form. You answered option A on 2026-08-05: return
`login_required` to the client for `prompt=none`, and keep the level 1 restart for every other ceremony.
Built as answered, at `issue/prompt-none-refusal`.

Both answers are quoted into section 3 of the agreement with the options that lost, and both are covered
by tests named in section 6.

**Two escalations on a change this size is inside the normal range**, and both were the same decision
(6) being harder than the interview found it. Section 3 records that, because the judgement belongs to
the next planning pass rather than to this run.

Judgement calls the run did make on its own are not hidden in this comment: each is recorded in its
stage's run-log entry in section 7 with the demonstration that made it forced, mostly test-shape and
naming choices plus the two documentation sweeps that found pages needing edits the plan had missed.

leodip:
## Follow-ups the run drafted, four, none filed

Each is a real defect verified against the code, each is outside this change under section 2 of the
agreement, and none is a way to avoid work #129 owed. Drafting is the run's job and filing is yours, so
below is everything needed to file each one, or to drop it.

The duplicate searches were re-run just now against the 33 open issues, since the tracker may have moved
while the run was working: nothing new covers any of the four. `#131` and `#132` are still the rotation
residuals and still do not cover code issuance, and there is still nothing on `amr` accuracy or on
`ClearAuthContext` ordering.

Full reasoning for each, including what was verified and why it is out of scope, is in section 9 of the
agreement.

---

### 1. A code insert can still land between a termination's `UPDATE` and its `COMMIT`

This is decision 12's accepted residual, and it is the one to read if you read only one. Termination
marks the codes that exist, and `/auth/issue` marks its own code when the session is already gone, so
the two sweepers cover each other. One interleaving escapes both: an insert committing after the
termination's sweep has read `codes` but before that transaction commits, where the compensating
statement's `NOT EXISTS` still sees the session row. What survives is an offline refresh token, for up
to `RefreshTokenOfflineMaxLifetimeInSeconds`. It needs the same missing primitive as #131 and #132, a
portable way to order a write against a sweep across all four engines, which is why it was accepted
rather than closed here.

```bash
gh issue create \
  --title "Security: a code insert can land between a termination's UPDATE and its COMMIT, escaping both marks" \
  --label bug --label security --label go \
  --body "$(cat <<'EOF'
#129 marks a terminated session's authorization codes revoked, and `/auth/issue` runs a compensating `UPDATE ... WHERE NOT EXISTS (SELECT 1 FROM user_sessions ...)` immediately after inserting a code, so a code created after the termination is marked even when the termination could not see it.

The remaining window is an insert committing after the termination's `UPDATE` has read `codes` and before that transaction commits: the compensating statement still observes the session as present, so the code carries no marker, and redeeming it yields an offline refresh token good for up to the configured offline maximum lifetime.

Closing it needs issuance ordered against the revocation sweep, which is the same missing primitive as #131 and #132. Scope the three together.
EOF
)"
```

---

### 2. Add to #109: the interactive logout path writes nothing to the database

A comment rather than a new issue, because #109 already owns this surface and its divergence B is the
same asymmetry one path over. Found while tracing which endpoint each "End session" button reaches, and
it is why decision 13 leaves logout alone.

```bash
gh issue comment 109 --body "$(cat <<'EOF'
Divergence B notes that the cookie is wiped unconditionally while the database teardown is per-client. There is a path where the teardown does not happen at all. `POST /auth/logout` without an `id_token_hint` clears the cookie session, writes an `AuditLogout` entry, redirects, and never touches `user_sessions`. Since `logout_consent.html` posts back carrying only a csrf field, that is the path taken by every logout a person performs in a browser, and by the admin console's account menu.

The row therefore survives with every client still attached. Session-bound refresh tokens keep working, access tokens keep passing `RequireValidSession`, and the row is orphaned from the browser because the cookie no longer carries the session identifier. The user still sees it listed as an active session on that device and can only end it with the "End session" button.

Whatever B decides about per-client versus whole-session teardown, this path should reach the same code as the `id_token_hint` path rather than bypassing it.

Found while building #129, which deliberately does not touch logout (its decision 13).
EOF
)"
```

---

### 3. A session-less completion inherits the terminated session's `amr`

Pre-existing, and #129 neither introduces nor widens it: before stage 5 the same stale value reached a
session created with no authentication at all, so the gate strictly improved it. Re-checked after
decision 15's answer and it stands unchanged.

```bash
gh issue create \
  --title "Security: a session-less completion inherits the previous session's amr, claiming otp that did not happen" \
  --label bug --label security --label go \
  --body "$(cat <<'EOF'
An authorization ceremony that reuses an existing session copies that session's `AuthMethods` onto its `AuthContext` in `HandleAuthorizeGet`, and nothing clears the copy if the session stops backing the ceremony. `HandleAuthCompletedGet`'s no-valid-session branch then passes the stale value to `StartNewUserSession`, and `HandleIssueGet` stamps it onto the authorization code, so it reaches the token as `amr`.

The result is a session row, and a token, recording `otp` for an authentication in which only a password was verified. `acr` is unaffected: `SetAcrLevel` recomputes it from the target when there is no session, so the two claims disagree, and a relying party trusting `amr` over `acr` reads a second factor that did not happen in that ceremony.

Reproduce with a user who has OTP enabled: authenticate at level 2 for one client, then send an authorization request from a client whose default ACR is level 1, and end the session while that request is in flight. The ceremony restarts at level 1 (#129 decision 6), the user enters a password, and the new session records `pwd otp`.

Clearing the inherited methods when the ceremony no longer has the session it inherited them from looks like the fix, but it is a claim-semantics change and wants its own decision.

Pre-existing rather than introduced by #129: before that change the same stale value reached a session created with no authentication at all.
EOF
)"
```

---

### 4. Every error response to a client clears the auth context after the response is committed

Seven sites, each read rather than inferred. #129 wrote one of them the wrong way round by copying the
neighbour, and fixing it inside stage 6 is the evidence this behaves as described: with the calls
reversed, a replayed `/auth/issue` on the same cookie jar answered `login_required` a second time from a
context still reading `ready_to_issue_code`.

```bash
gh issue create \
  --title "ClearAuthContext is called after the client response is committed in seven handlers, so the clear reaches nothing" \
  --label bug --label security --label go \
  --body "$(cat <<'EOF'
`AuthHelper.ClearAuthContext` deletes the auth context from the cookie session and persists that deletion through `sessionStore.Save(r, w, sess)`, which writes a `Set-Cookie` header. Seven handlers call it immediately *after* `redirToClientWithError`, which has already committed the response: `http.Redirect` for the query and fragment response modes, and `t.Execute(w, m)` for `form_post`. Go snapshots the header map when the status line is written, so the `Set-Cookie` is dropped and the browser keeps the auth context.

The sites are `HandleAuthorizeGet`'s validation-error closure and `handlePromptNone`'s `redirectWithError` closure in `handler_authorize.go`; the disabled-user and no-authorized-scopes refusals in `handler_auth_completed.go`; both consent-denied refusals in `handler_consent.go`; and the `id_token_hint` subject mismatch in `HandleIssueGet`. The success paths in the same files do it in the right order, clearing before writing the response, so this reads as a copied idiom rather than an intended difference.

No bypass is claimed for any single site: replaying a retained ceremony re-runs the same checks. The defect is that a cleanup written seven times never happens, and that the `id_token_hint` site leaves the browser holding a context in `ready_to_issue_code`, the state that mints codes.

The fix is to clear before writing the response at each site, returning a 500 if the clear fails rather than committing a client response that cannot be recorded. A mock expectation cannot catch a regression here, since it proves only that the method was called; the assertion has to read the committed response, either `httptest.ResponseRecorder.Result()` at the unit tier or a second request on the same cookie jar at the integration tier.

Found while building #129, whose own `prompt=none` refusal was written the same way round and fixed inside that change, which is where the demonstration above comes from.
EOF
)"
```

leodip:
## The four drafted follow-ups are filed

Re-verified against the code first, since the drafts were written during the run. Three went in as
issues, one as a comment, and two bodies changed on the way. The PR description now carries this in a
**Follow-ups filed** section, and its three dangling "follow-up N" references point at real numbers.

| Drafted as | Filed as | Change on the way in |
|---|---|---|
| 1, code insert between a termination's `UPDATE` and its `COMMIT` | **#139** | **Corrected.** The draft asserted the window uniformly. It is open on **PostgreSQL** only; MySQL, SQL Server and SQLite close it incidentally, by InnoDB next-key locks over `idx_codes_session_identifier`, by a blocking locking-READ-COMMITTED read on the uncommitted `DELETE`, and by writer serialization. Reframed as "correctness rests on three engines' defaults by accident", the standard #132's criterion 4 already sets. Without this, a reader reproducing on sqlite concludes the issue is wrong |
| 2, the interactive logout path writes nothing | **[comment on #109](https://github.com/leodip/goiabada/issues/109#issuecomment-5197711587)** | **Two claims were wrong and are removed.** The admin console's account menu does *not* take the no-hint path: it mints a signed `id_token_hint` via `/api/v1/account/logout-request`, so it lands on `doLogoutWithIdToken` and does tear down. And no first-party UI reaches the no-hint POST at all, since the auth server has no logout link of its own. Reachable callers are an RP omitting `id_token_hint`, which is conforming, and a direct request. The defect itself stands |
| 3, a session-less completion inherits the terminated session's `amr` | **#140** | Substance unchanged, severity stated as low. The narrow reachability the draft claimed was confirmed rather than assumed: the `AuthMethods` copy sits inside `HandleAuthorizeGet`'s `hasValidUserSession` block, so an idle-expired session leaks nothing and it really does take a mid-flight termination |
| 4, `ClearAuthContext` after the response is committed | **#141** | All seven sites re-read, as were the two success paths that order it correctly. Gained a second defect on the same lines: two of the seven also fall through after `httpHelper.InternalServerError` instead of returning, so a failed redirect writes a 500 and keeps going |

Duplicates checked against the open and closed tracker. #139 is distinct from #131 and #132, which
are the rotation residuals: session-scoped code issuance against a revocation sweep, versus a
user-scoped generation sweep and a family-scoped containment cascade. Different triggers, different
row sets. #134 held #139's previous shape and was closed as moot when decision 4 moved the marker onto
`codes.revoked`, so nothing tracked it. Nothing tracked #140 or #141 at all.

**Still open, and it is yours to call.** Section 9 of the agreement still reads `Status: **Drafted**`
on all four, and still carries the two claims corrected above. That file is part of this PR, so those
go into the merged history as written. I left it alone because the run log's convention is to record
rather than revise, which is why stage 3's wrong prediction is still sitting there with the
measurement beside it. The consistent fix is to append the corrections and the issue numbers under
each entry rather than edit the drafts. Say the word and I will.

Issue #142, opened 2026-08-05
Narrow the TOTP skew window so future-period codes are not accepted


leodip:
## Summary

Goiabada accepts a TOTP code from the previous, current and next 30 second time step, so roughly a 90 second window is valid at any instant. The next-step arm exists only to tolerate device clocks running ahead of the server. Accepting only the previous and current step would halve the window, at a cost to users whose clock runs fast.

This is recommendation 3 of #111, deliberately left out of that change because it is independent of replay and changes behaviour for a population of users in a way a security fix should not do quietly.

## Current behaviour

A TOTP code is derived from a time step, `floor(unix_seconds / 30)`, not from the clock directly. The authenticator app and the server each compute that number from their own clock, so a verifier accepts neighbouring steps to absorb the difference.

Goiabada uses the library defaults throughout. `GenerateOTPSecret` in `src/core/otp/generator.go` passes no `Period`, `Digits` or `Algorithm` to `totp.Generate`, and validation goes through `totp.Validate`, which hardcodes `Skew: 1`. In `pquerna/otp v1.5.0` that builds the candidate set as `counter`, then `counter+i` and `counter-i` for each skew step:

```go
counters = append(counters, uint64(counter))
for i := 1; i <= int(opts.Skew); i++ {
    counters = append(counters, uint64(counter+int64(i)))
    counters = append(counters, uint64(counter-int64(i)))
}
```

So with the server at step N, codes for N-1, N and N+1 are all accepted.

The two arms do not serve the same population:

- **N-1** covers device clocks running slow, and the ordinary case of reading a code near the end of its period and submitting it just after the rollover. Network latency and typing time put every user in this bucket some of the time.
- **N+1** covers only device clocks running ahead of the server. Nothing else produces a future code.

## Why it matters

The window is how long a code obtained by an attacker stays usable. After #111 a code is dead once the legitimate user spends it, so the residual exposure is a code the attacker obtained that the user then did not submit: shoulder-surfing, a code read from a lock screen notification, a phished code from a login the user abandoned. Halving the window is defence in depth on top of the one-time-use guarantee, not a fix for a distinct break.

## The cost, which is the reason this needs deciding rather than just doing

A device whose clock runs fast by `d` seconds is one step ahead exactly when `(t mod 30) >= 30 - d`, which is `d/30` of the time. If future-period codes stop being accepted:

| Device clock ahead by | Resulting login failure rate |
|---|---|
| 2s | ~7% |
| 5s | ~17% |
[...]

Issue #140, opened 2026-08-05
Security: a session-less completion inherits the previous session's amr, claiming otp that did not happen


leodip:
An authorization ceremony that reuses an existing session copies that session's `AuthMethods` onto its `AuthContext` in `HandleAuthorizeGet`, and nothing clears the copy if the session stops backing the ceremony. `HandleAuthCompletedGet`'s no-valid-session branch then passes the stale value to `StartNewUserSession`, and `HandleIssueGet` stamps it onto the authorization code, so it reaches the token as `amr`.

The result is a session row, and a token, recording `otp` for an authentication in which only a password was verified.

Found while building #129 and recorded there as follow-up 3. **Pre-existing rather than introduced by #129**: before that change the same stale value reached a session created with no authentication at all, so #129's `/auth/completed` gate strictly improved it.

Citations are by symbol rather than line number.

## The chain

Each hop verified against the code:

1. `HandleAuthorizeGet`'s SSO reuse branch assigns `authContext.AuthMethods = userSession.AuthMethods`. This is inside the `hasValidUserSession` block, which matters for reachability below.
2. The ceremony reaches `HandleAuthCompletedGet` with the session no longer valid, so it takes the else branch.
3. Since #129, that branch requires `authContext.Level1AuthCompleted` before creating a session, so the ceremony restarts at level 1 and the user enters a password. `AuthContext.AddAuthMethod` appends rather than replaces and is a no-op when `pwd` is already present, so `pwd otp` survives the restart unchanged.
4. `StartNewUserSession` is called with `authContext.AuthMethods` verbatim.
5. `HandleIssueGet` sets `AuthMethods: authContext.AuthMethods` on the code, which becomes the token's `amr`.

`acr` is unaffected: `AuthContext.SetAcrLevel` recomputes it from the target when there is no session. So the two claims disagree, and the token comes out internally inconsistent, `acr` at `urn:goiabada:level1` with `amr` `["pwd", "otp"]`. A relying party trusting `amr` over `acr` reads a second factor that did not happen in the ceremony that produced the token.

## Reachability, stated because it is narrow

The target ACR must not require level 2, or `HandleAuthLevel1CompletedGet` sends the ceremony to `/auth/level2` and the `otp` is genuinely earned. And the copy in step 1 sits inside the `hasValidUserSession` block, so an idle-expired session does not leak its methods: the session has to be valid at `/auth/authorize` and invalid by `/auth/completed`.

[...]

Issue #113, opened 2026-07-25
No rate limit on password verification in /api/v1/account/password and /api/v1/account/otp


leodip:
## Summary

Two account API endpoints verify the user's current password against `user.PasswordHash` with no rate limiting. An attacker holding a stolen access token but not the password can brute-force it offline-speed against the live endpoint, and the payoff on one of them is disabling the victim's second factor.

Severity: **Low to Medium**. It requires an already-compromised access token, so it is an amplification path rather than an entry point. But the specific amplification is 2FA removal, which is what makes it worth fixing.

## Affected code

`src/authserver/internal/server/routes.go:317` and `:319`:

```go
r.Put("/password", apihandlers.HandleAPIAccountPasswordPut(s.database, passwordValidator, auditLogger))
...
r.Put("/otp", apihandlers.HandleAPIAccountOTPPut(s.database, auditLogger))
```

Neither carries a `rateLimiter.*` middleware. The `/api/v1/account` group applies only `APIDebugMiddleware`, `authHeaderToContext`, `RequireBearerTokenScope` and `RequireValidSession` (`routes.go:304-308`).

Both handlers perform an unbounded bcrypt comparison:

`handler_api_account_password.go:66-69`

```go
if !hashutil.VerifyPasswordHash(user.PasswordHash, req.CurrentPassword) {
    writeJSONError(w, "Authentication failed. Check your current password and try again.", "AUTHENTICATION_FAILED", http.StatusBadRequest)
    return
}
```

`handler_api_account_otp.go:110-113`

```go
if !hashutil.VerifyPasswordHash(user.PasswordHash, req.Password) {
    writeJSONError(w, "Authentication failed. Check your password and try again.", "AUTHENTICATION_FAILED", http.StatusBadRequest)
    return
}
```

By contrast the browser-facing password form is protected: `routes.go:104` wraps `POST /auth/pwd` with `rateLimiter.LimitPwd`, which enforces both a per-IP and a per-account ceiling.

## Impact

The password check on `PUT /api/v1/account/otp` exists precisely so that a stolen session cannot disable 2FA. Leaving it unthrottled undermines that:

1. Attacker obtains an access token with `authserver:manage-account` (XSS on an RP, a leaked token in logs, a compromised client).
2. Attacker cannot authenticate as the user interactively because OTP is enabled.
3. Attacker brute-forces `req.Password` against `PUT /api/v1/account/otp` at whatever rate the server will sustain.
[...]

PR #143, opened 2026-08-05
fix(otp): make TOTP codes single use

### LINKED ISSUE #111: TOTP codes are replayable: no used-code tracking, RFC 6238 5.2 requires one-time use
## Summary

TOTP codes can be submitted repeatedly. `totp.Validate` has no notion of a consumed code, and goiabada stores nothing to track one, so a single intercepted OTP stays valid for its entire acceptance window. The default skew of one period widens that window to roughly 90 seconds.

RFC 6238 section 5.2 makes this a MUST NOT:

> "The verifier MUST NOT accept the second attempt of the OTP after the successful validation has been issued for the first OTP, which ensures one-time only use of an OTP."

Severity: **Low**. Exploiting it requires already having the code, so this is a defence-in-depth gap against real-time phishing proxies and shoulder-surfing rather than a standalone break.

## Affected code

Three call sites, all using the default `totp.Validate`:

- `src/authserver/internal/handlers/handler_auth_otp.go:264` (login, user already enrolled)
- `src/authserver/internal/handlers/handler_auth_otp.go:274` (login, enrolling now)
- `src/authserver/internal/handlers/apihandlers/handler_api_account_otp.go:163` (self-service enable)

```go
otpValid := totp.Validate(otpCode, otpSecret)
```

A grep for any used-code or counter tracking (`lastotp`, `otp_counter`, `usedotp`, `last_used_otp`) across `*.go` and `*.sql` returns nothing, and there is no such column on the `users` table.

## Reproduction

```go
key, _ := totp.Generate(totp.GenerateOpts{Issuer: "goiabada", AccountName: "poc@example.com"})
secret := key.Secret()
code, _ := totp.GenerateCode(secret, time.Now())

accepted := 0
for i := 0; i < 5; i++ {
    if totp.Validate(code, secret) {   // the exact call from handler_auth_otp.go:264
        accepted++
    }
}

prev, _ := totp.GenerateCode(secret, time.Now().Add(-30*time.Second))
```

Output:

```
REPLAY CONFIRMED: one OTP accepted 5/5 times by totp.Validate
SKEW CONFIRMED: previous-period code "347601" still accepted (default skew=1)
```

## Impact

An OTP obtained once can be used more than once inside the window. Realistic paths to obtaining one:

- A real-time phishing proxy that relays the victim's code, then reuses it on a second session.
- Shoulder-surfing, or a code read from a notification on a lock screen.
- A code entered into the wrong site and replayed by the attacker before it expires.

`LimitOtp` caps attempts at 10 per minute per user, which bounds brute force but does nothing about replay: replay needs one attempt, not many.

The `handler_auth_otp.go:274` and `handler_api_account_otp.go:163` enrolment sites are lower risk, since the secret is not yet bound to the account, but they should share whatever mechanism is adopted for consistency.

## Recommendation

Track the last accepted time-step counter per user and reject any code whose counter is less than or equal to it.

1. **Schema**: add a nullable `last_otp_counter BIGINT` (or equivalent per engine, following the existing datetime/int conventions) to `users`. Next migration number should be confirmed with `ls src/core/data/sqlitedb/migrations | tail` rather than assumed.

2. **Validation**: switch to `totp.ValidateCustom`, which returns the validation result, and derive the counter as `time.Now().Unix() / period`. Because skew is in play, the accepted code may belong to the previous, current or next step, so the implementation needs to identify which step matched and store that. Simplest approach is to iterate the skew window explicitly:

```go
const period = 30
now := time.Now().UTC()
for _, delta := range []int64{-1, 0, 1} {
    step := now.Unix()/period + delta
    candidate, err := totp.GenerateCode(secret, time.Unix(step*period, 0))
    if err != nil {
        return err
    }
    if subtle.ConstantTimeCompare([]byte(candidate), []byte(code)) == 1 {
        if user.LastOTPCounter.Valid && step <= user.LastOTPCounter.Int64 {
            return errOTPAlreadyUsed   // surfaced to the user as the generic incorrect-OTP error
        }
        user.LastOTPCounter = sql.NullInt64{Int64: step, Valid: true}
        // persist
        return nil
    }
}
```

Note the constant-time comparison: the current `totp.Validate` already does this internally, so a hand-rolled loop should not regress it.

3. **Consider narrowing skew.** The default of 1 accepts the previous and next periods. Reducing to accept only the previous period (not the future one) halves the window at minor cost to users with slow clocks. Optional and separate from the replay fix.

4. **User-facing behaviour**: a replayed code should produce the same generic "incorrect OTP" error as a wrong code, and emit `AuditAuthFailedOtp` as the existing failure paths do (`handler_auth_otp.go:266`, `:276`). Do not tell the caller the code was already used.

5. **Concurrency**: two simultaneous submissions of the same code could both read the old counter and both pass. If that matters, the counter update wants to be a conditional write (`UPDATE ... WHERE last_otp_counter IS NULL OR last_otp_counter < ?`) with the same compare-and-set shape already used by `MarkCodeAsUsed` in the auth-code flow.

## References

- RFC 6238 section 5.2 (Validation and Time-Step Size)
- OWASP ASVS 5.0 V2.8 (One-Time Verifier requirements)

leodip:
Closes #111.

A TOTP code accepted once is never accepted again, at all three call sites, and the refusal is
indistinguishable from a wrong code to the caller. RFC 6238 5.2.

The agreement and the full run trace are deliberately not tracked in the repository. What survives is here: the decisions in this description and in the comments below, and the reasoning in the code comments beside the code it constrains.

## How it works

`users.last_otp_step` records the most recently consumed time step, one column, `NOT NULL DEFAULT 0`,
tagged `dont-update` so no ordinary whole-user write can regress it. The only write is a
compare-and-set:

```sql
UPDATE users SET last_otp_step = ?, updated_at = ? WHERE id = ? AND last_otp_step < ?
```

**That statement is both the claim and the replay check**, which is the property that makes this safe:
there is no window between deciding a code is good and recording that it was used, so two concurrent
submissions of one code contend on a single `UPDATE` and at most one row transitions.

`otp.MatchStep` replaces `totp.Validate` and reports *which* step produced the passcode, iterating the
skew window over `hotp.ValidateCustom` so the constant-time comparison stays the library's. It accepts
exactly what `totp.Validate` accepted, with one deliberate exception: an empty secret now matches
nothing (decision 9), which closes a gap where the library derives a real, publicly computable code
from the empty key.

| Stage | What | Commit |
|---|---|---|
| 1 | The step matcher, `otp.MatchStep`, uncalled. Reports the bottom of a colliding-step chain, per decisions 11 and 12 | `01a186c` |
| 2 | `users.last_otp_step`, migration 000027 on four engines, `TryConsumeUserOTPStep` and `ResetUserOTPStep`, uncalled | `a10c6bb` |
| 3 | Enforcement at the two browser sites, `/auth/otp`, plus the `otp_code_replay_detected` audit event | `cc208c5` |
| 4 | The account API, reset on disable at both disable sites, and the docs bullet | `f14696f` |

Each stage landed uncalled-then-called deliberately: the matcher and the column ship with their
exhaustive tables before anything depends on them.

## Tests

Full suite on the merge tree, `where.sh test --type all`, 7m14s: **9122 pass, 0 fail, 4 skip**.

| Tier | Engines | Result |
|---|---|---|
| Unit | all three modules | 2770 pass |
| Data | sqlite, mysql, postgres, mssql | 1564 pass, 4 skip |
| Integration | mysql, postgres, mssql, sqlite | 4788 pass |

The 4 skips are pre-existing engine-capability skips, unchanged by this branch. `check-anchors.sh` 25
of 25, `go vet` clean.

**Every new test was mutation-checked**, because most of them pin something whose absence previously
left the whole suite green. Twelve mutations across stages 3 and 4; each failed the case it was meant
to fail and was reverted with the source restored byte-identical. One of them found a real defect in a
new test rather than in the code: the `apihandlers` package's `TestMain` never initialised the data
cipher, so a case asserting a 500 on the claim-error branch passed with that branch deleted.

## Deviations worth knowing

- **Disabling OTP resets the marker** (decision 4). A code refused as a replay becomes acceptable again
  after the authenticator is removed and re-enrolled. That is deliberate: the marker belongs to the
  authenticator, and it is the only in-product remedy if a clock jump strands a marker in the future.
  The security page says so rather than promising permanence.
- **Both disable sites commit their two writes in one transaction** (decision 13, raised by stage 4's
  own review). Committed separately they leave a window in which an enrolment can claim a step that the
  reset then erases, leaving an already-consumed code claimable again.
- **The account API's replay branch emits `otp_code_replay_detected` and no `AuditAuthFailedOtp`**,
  unlike the two browser sites, because a wrong code at that endpoint has never audited anything and
  enabling OTP is not an authentication ceremony.
- One existing integration test needed repair (decision 8): it disables OTP by bypassing the handlers,
  so it now resets the step alongside, standing in for the disable handler.

## Read these two comments before merging

- **[Deferred decisions](https://github.com/leodip/goiabada/pull/143#issuecomment-5207560716)**: none
  were deferred, and the comment says what was accepted on purpose instead.
- **[Follow-ups](https://github.com/leodip/goiabada/pull/143#issuecomment-5207560828)**: two drafted
  and ready to file, with their `gh issue create` commands. #142 is already filed.


leodip:
## Decision 10: does a disable reset reopen a consumed code to a request already in flight?

The plan gate's review (round 1, six findings) turned up one question I should not answer alone. Five
findings are resolved and committed in `82ef76a`; this is the sixth. **No code has been written yet**,
and stage 1 is unaffected, so answering this is cheap right now and expensive after stage 2 lands.

### The hole

Decision 4 says reset-on-disable is "not a bypass" and that "nothing here lets a replayed code buy
anything". There is an interleaving it does not cover. I verified all four steps against the code:

1. A browser request at `otp/verify-enabled` loads the user, decrypts the stored secret, and holds
   both in memory for the rest of the request.
2. A disable commits `ClearOTPSecret` plus `UpdateUser`, and then resets the marker, as **two
   separate writes**.
3. The designed claim predicate is `id` plus `last_otp_step < step`. It binds nothing about
   enrolment, so the in-flight request claims the already-consumed step successfully.
4. `handler_auth_completed.go:HandleAuthCompletedGet` reloads the user but checks only `user.Enabled`,
   never `user.OTPEnabled`, so nothing downstream catches it.

The race is genuinely required. A fresh request after the disable takes the enrolling branch and
validates against the session secret, so only a request that loaded the old enabled state can do this.

### What it actually buys, which is less than the review said

The reviewer concluded the attacker "obtains level 2". I checked, and that overstates it. Once OTP is
disabled, someone holding the password already reaches level 2 more cheaply:
`handler_auth_level2.go:HandleAuthLevel2Get` sends `AcrLevel2Optional` with `!user.OTPEnabled`
**straight to `/auth/completed`, skipping OTP entirely**, and sends `AcrLevel2Mandatory` to the
enrolment form, where an attacker can enrol their own secret.

So the replay grants no access the disable had not already granted. What it grants is a token asserting
`amr: ["pwd","otp"]` for an authenticator that was just removed, obtained without the
`AuditEnabledOTP` event and the visible new authenticator that self-enrolling would leave. A false
factor assertion and a stealth gain, not an access gain. That reframing is why this is a judgement
call about what our tokens may claim, rather than an obvious emergency.

### Options

**A. Bind the claim to enrolment state.** Add `AND otp_enabled = 1` to the claim. It cannot be
unconditional, because the two enrolment sites claim while `otp_enabled` is still false (§4 puts the
claim before the enable write deliberately), so the method takes a flag:
`TryConsumeUserOTPStep(tx, userId, step, requireOTPEnabled bool)`. Fixes the cause. Costs a signature
change and one branch at three call sites. Does not close the narrower variant where a disable is
immediately followed by re-enrolment with a new secret while the stale request is in flight, since
`otp_enabled` is true again by then; closing that needs an enrolment-generation column, a second
column and a notably larger change.

**B. Reset to the current step instead of to 0.** `last_otp_step = now/30` on disable. A code consumed
in the current window stays refused, and a marker stranded in the future by a clock jump still comes
back to now, so decision 4's operator remedy survives. No signature change, no new column, one line
different. Costs: after a disable, re-enrolling with a code from the current window is refused and the
user waits up to 30 seconds, and seam 4's pinned "enable, disable, re-enable with the same code
succeeds" case has to become a later-step code, which weakens what that case proves.

**C. Accept and document it.** Given the corrected impact, record the false AMR assertion as a known
narrow window and build the design as sealed. Costs nothing now, and leaves an authentication surface
where a token can name a factor that no longer exists.

**D. Drop the reset entirely**, reversing decision 4. Closes the hole completely and reinstates the
permanent clock-jump lockout with no in-product remedy, which decision 4 rejected for good reason.
Listed for completeness; I do not recommend it.

### Recommendation: A

It is the only option that fixes the cause rather than the blast radius, and the signature change is
at its cheapest right now because stage 2 has not been written and no caller exists. The flag also
reads honestly at each site: verification requires enrolment, enrolment does not.

B is the fallback if the signature change is unwelcome, and the thing to weigh there is the 30 second
re-enrolment delay. C is defensible, but only on a judgement about what an `amr` claim is allowed to
assert, and that is yours rather than the run's.

### What getting it wrong costs

Choosing **C** and reversing later means revisiting stage 2's interface and stage 4's tests after both
have landed, which is the expensive direction. Choosing **A** where C would have done costs one
boolean parameter and a little noise at three call sites. That asymmetry is most of why I recommend A.

---

Full reasoning is in decision 10 of the design record, together with the plan-review round that raised it. **The run is stopped until this is answered.**
Reply with A, B, C or D (or something else) and the next session applies it and builds stage 1.

leodip:
A. Bind the claim to enrolment state.

leodip:
## Decision 10 applied: option A, the claim binds to enrolment state

Recorded in `dafdc88`. No code written yet; stage 1 is next.

`TryConsumeUserOTPStep` takes a fourth argument, `requireOTPEnabled`, and adds `otp_enabled = true` to
the claim predicate when it is set. Verification passes true; the two enrolment sites pass false, since
the design puts the claim before the enable write and `otp_enabled` is still false there.

**Two things I checked before writing it down**, because option A rested on them and the escalation had
not verified either:

- `api-otp/verify` is only ever reached with OTP off, since the enable branch refuses with
  `OTP_ALREADY_ENABLED` when `user.OTPEnabled`. So `false` is honest there, not just convenient.
- The bound bool is portable. `users.otp_enabled` carries the same column type as `users.enabled` on all
  four engines, and `TrySetUserEnabled` already binds a Go bool in exactly this position, so nothing is
  engine-specific. Checked against all four schema snapshots.

**Two things your answer settled that I had left implicit:**

1. **The disable writes need no transaction, but their order is load-bearing.** The reset must follow the
   `otp_enabled = false` write. Reversed, the marker reads 0 while the authenticator still reads enabled,
   and an in-flight verification claims a consumed step through it, reintroducing the hole by ordering.
   Section 4's order is already the right one, so stage 4 just adds the comment.
2. **A false at a verification site now has a second cause,** the authenticator being removed under the
   request, so `otp_code_replay_detected` is named for the likely cause rather than the certain one.
   `MarkCodeAsUsed` documents the same imprecision about its own three-way false, so this follows
   precedent rather than inventing one.

Stage 2 gains two data cases pinning both directions of the flag, including that a reset does not reopen
a consumed step to a verification claim. Neither is endpoint-observable, because the interleaving needs a
request holding state loaded before a disable, so both sit at the data tier that owns the column.

**One residual, drafted as a follow-up rather than built.** Option A does not close a re-enrolment
landing while a stale verification request is in flight, as its text said. Working through it, it is
narrower than that text: the re-enrolment claims a step of its own, so the replayed code must come from
the +1 window and the re-enrolment from the current one or earlier, all inside one 30 second period.
Closing it needs an enrolment-generation column, which is a second migration. Drafted in `closing.md`
with the `gh issue create` command, and it is adjacent to #140, the same false-`amr` class by a different
mechanism.

leodip:
## Decision 11: one passcode can be produced by two steps. Which one does `MatchStep` return?

Stage 1's code review found the matcher replayable, and it returned a `forced_answer`, which is
normally a repair I apply without asking. I checked it before applying it, and **the suggested fix
does not actually close the hole**, so what to do instead is a choice about the matcher's contract.
That is why this is here rather than in a commit.

Stage 1 is written and green but **not committed**: the step it returns is the question. Stages 2 to
4 have not started, so answering now is cheap.

### The defect, reproduced

A six-digit passcode does not name a time step. `MatchStep` returns the first step in delta order
`0, -1, 1` that produces the passcode, and decision 2's marker records that step. When two steps that
are both live for one passcode produce the same six digits, the lower is claimed first and the higher
is still claimable once the window slides, so the same passcode is accepted twice. For the secret the
test table pins on, steps 3710568 and 3710569 both yield `874294`, and steps 818665 and 818667 both
yield `475244`.

I kept the probe, which prints everything below and
says in its header how to run it. It swept 600,000 consecutive steps, 208 days of one secret, and
found 5 same-passcode pairs within a spread of 5, against 5 expected. So about 8 authentications in a
million meet one at all, and about 3 in a million meet one that matters. Meeting one is not yet an
exploit: the attacker still needs the code, which is what #111 already assumes.

### The boundary I think this should be decided against

Step A is acceptable while the current step is `A-1`, `A` or `A+1`. So two matching steps are
*continuously* acceptable only when they are at most 3 apart, which is `2 * skew + 1` at today's skew.
A wider pair has a gap in which the code is refused, and by the time it is accepted again the
authenticator is legitimately showing those same six digits for a new step. **No marker that records a
step can refuse that**; refusing it means storing the code itself, which is stronger than the
consumed-codes table decision 1 already rejected. Pairs 1 to 3 apart are the target. Wider ones are
inherent to the design you chose, and I think correctly.

### What each candidate does

A trace reads "first accepted at current `c1`, accepted again at `c2`". A rule is safe for a pair when
no presentation of that passcode leaves a later one acceptable. Every rule accepts exactly the codes
`totp.Validate` accepts today, the empty secret aside, so none of this changes which codes work. Only
the recorded step moves.

| Rule | spread 1 | spread 2 | spread 3 | spread 5 (inherent) |
|---|---|---|---|---|
| shipped, first match | replayable from 2 presentations | 3 | 3 | 3, again at `A+4` |
| **A**, greatest inside the window | 1, only from `A-1` | 2 | 3 | 3, again at `A+4` |
| **B**, lowest within 3 below the window | safe | safe | safe | 4, again at `A+5` |
| **B2**, as B but claiming the greatest in the window | safe | safe | safe | 3, again at `A+5` |
| **C**, greatest within 3 above the window | safe | safe | safe | 2, again at `A+1` |

**A, the reviewer's answer as given.** Return the greatest match inside the window. One line, reverse
the delta order. It closes the trace the reviewer demonstrated and nothing more. The spread 1 pair
survives when the code is presented a step early, which is exactly what the `+1` window exists to
allow, and spreads 2 and 3 survive from an ordinary presentation, because the colliding step is not in
the window yet when the claim is made.

**B, return the lowest step within three below the window that produces the passcode.** Acceptance
still requires a match in `c-1 .. c+1`; on a match, walk down to `c-4` and return the lowest step there
producing the same code. That value is the same at every presentation of a continuously live passcode,
which is what makes the first claim block the rest. **No signature moves**: `MatchStep` keeps
`(int64, bool)`, `TryConsumeUserOTPStep` keeps decision 10's shape, stages 2 to 4 are untouched. Costs
four extra HMAC computations on a successful match, microseconds. In the same 3-in-a-million case the
marker sits up to three steps below the code's own step, which shrinks the incidental refusal of lower
unused codes that decision 1 calls the counter's imprecision, and nothing depends on it. It can refuse
one legitimate code, when a collision coincides with a consumption in the previous four steps: the
next code works, so it fails toward a retry.

**B2, as B but returning two values**, the lowest to guard on and the greatest in the window to claim,
so the marker keeps sitting at the top of the window. Still a single statement:
`SET last_otp_step = <claim> WHERE id = ? AND last_otp_step < <guard>`. Marginally stronger on the
inherent wide pairs. Costs a wider `MatchStep` return and a wider `TryConsumeUserOTPStep`, both in
stage 2, which has no callers yet.

**C, claim upward instead**, the greatest match within three above the window. Equal on the cases that
matter and no signature change, but the guard rises with the claim, so on a wide pair it accepts the
second presentation 30 seconds later instead of after a refused gap, the worst timing of any
candidate. When it fires it also burns up to three of the user's upcoming codes, up to 90 seconds of
refusals, so it fails toward a lockout.

**D, take A and record the rest as a known limitation**, with a drafted follow-up. Cheapest, and it
ships #111 with a reachable instance of the defect #111 exists to close.

### Recommendation: B

It is the only candidate that closes every continuously live case, changes no signature and no later
stage, and fails toward a retry rather than a lockout. The `3` would be derived from the window rather
than written as a literal, so if #142 ever narrows the skew it narrows with it.

B2 is the answer if the marker sitting below the current step is worth two signature changes. C is
strictly worse than B on both failure modes. D is defensible only if 3 in a million is acceptable for
this property, which is a judgement about how literally RFC 6238 5.2 binds here, and that is yours
rather than mine.

Getting it wrong either way: choosing A or D and later reversing means revisiting the matcher and its
table after stages 2 to 4 have built on the returned step, and the release note for this fix would be
wrong in the meantime. Choosing B when A would have done costs four HMAC computations, one test row
and a longer doc comment.

Whichever wins, seam 1's existing 15 rows are unaffected. The probe confirms nothing collides within
eight steps of the instant they pin, for both the valid secret and the empty one, and the table gains
one collision row pinned at its own instant. The pairs are already located: spread 1 at Unix
1687378770 (`249386`), spread 2 at Unix 24559950 (`475244`), spread 3 at Unix 1682324250 (`194241`).

Reply with the letter, or with something else you would rather do. Full reasoning is decision 11 of the design record; the numbers are reproducible from the probe.

leodip:
B

leodip:
## Decision 11 applied: B, the lowest colliding step

Recorded in `7a82cc3`. Stage 1's two files are written and green, still uncommitted until the gate
closes.

`MatchStep` keeps `(int64, bool)` and accepts exactly what it accepted before. What moves is the step
it reports: the lowest one within `lookbackSteps` below the window that produces the same passcode,
with `skewSteps = 1` and `lookbackSteps = 2*skewSteps + 1` so #142 narrows the lookback with the
window, as your answer's reasoning asked.

**What I checked rather than argued.** The probe gained a sweep section, and the new test rows are
transcribed from its output rather than reasoned out: for each of the three located pairs it prints
every current step from which the passcode is accepted and what B reports there. All three report one
constant step across the whole run, and no third step within eight either side produces those
passcodes, which is what makes the answer constant rather than merely lower. `output.txt` is committed
beside the probe.

**Three mutations, run and reverted**, because a new table that passes proves nothing about what it
catches:

1. Drop the lookback, report the step matched inside the window: all three collision subtests fail,
   and the original 15-row table passes. That is the point of adding them. Nothing collides near the
   instant that table pins, so every rule decision 11 rejected passes all 15 of its rows.
2. `lookbackSteps = 2*skewSteps`, one too narrow: only the spread 3 subtest fails. The pair at the
   continuous-liveness boundary is what pins the derivation.
3. `skewSteps = 2`: the two window-boundary rows fail and all three collision subtests fail with them.
   The spreads are written as literals, not in terms of the constant, so if #142 ever narrows the skew
   it fails here loudly instead of quietly adjusting.

**Two small things your answer settled that I had left implicit.** The cost is at most three further
HMAC computations, not four, and only on a successful match: the failure path computes the window and
stops, exactly as before. And the lookback cannot error once the window has matched, since the
library's two errors are an unparseable secret and a wrong-length passcode and neither depends on the
step; it is collapsed to no-match anyway so the file has one error rule and the unreachable branch
fails closed.

Core tier green, `gofmt` and `go vet` clean, anchors 24 of 24. Round 3 is requested at this gate.

leodip:
## Decision 12: a fixed lookback closes colliding pairs but not chains of three

Stage 1's third code review found that the answer to decision 11, which I applied, does not give the
constant reported step its own reasoning claims. I confirmed it, then found it is wider than the
review said. It is a security property in authentication, and the reviewer returned it with an empty
`forced_answer`, so it is yours rather than mine.

Nothing is committed that rests on it. Stage 1's two files stay uncommitted and stage 1 stays
`In progress`.

### The defect

`MatchStep` reports the lowest step in `current-4 .. current+1` that produces the passcode. That is
constant across an isolated colliding pair, which is what decision 11 fixed. It is not constant across
a **chain**: three steps producing one passcode, each adjacent gap at most 3, chain their acceptance
ranges into one unbroken run, and the lowest of them falls out of the fixed interval while the run is
still live.

For steps 100, 103 and 106 the passcode is accepted continuously from current 99 through 107, and the
reported step is 100 through current 104, then **103 from current 105**. A marker holding 100 admits
103, so the same passcode is consumed a second time with no refused step in between.

Executed, not argued: the reviewer's model is `MatchStep`'s two scans with the HMAC predicate replaced
by the explicit set, and I reproduced and extended it in `probe/step-collisions`.

### Worse than the review reported

Replaying both rules against every continuously live shape:

- **6 of the 9 chains of three are replayable as built**, every one whose total span exceeds the
  lookback of 3. `{A, A+1, A+2}`, `{A, A+1, A+3}` and `{A, A+2, A+3}` are safe.
- The span 4 pair leaks from 5 presentations rather than 3.

### How often, measured

20,000,000 steps over 4 secrets, 19 years of one authenticator:

| | measured | expected |
|---|---|---|
| colliding pairs at most 3 apart | 55, 2.75e-06 per step | 3.0e-06 per step |
| chains of three | **0** | 1.8e-04 over the whole sweep |

The pair rate matching expectation is what validates the 1e-06 per-pair model everything else rests
on. From it: a presented code sits in a chain about **2.7e-11** of the time, roughly **3 in 100
billion authentications**. That is five orders of magnitude below the 8.3e-06 pair rate you answered
decision 11 against.

### The options

**A. Walk the lookback down transitively** instead of applying it once. From the matched step, look
for a match in the 3 steps below it, and repeat from whatever it finds.

- Closes all 9 chain shapes and all 3 continuously live pairs. Executed.
- Acceptance untouched, `MatchStep` keeps `(int64, bool)`, nothing in stages 2 to 4 moves.
- **Free in the case every real authentication takes**: one lookback scan of 3 steps, exactly what the
  code does now, because the walk stops the moment a scan finds nothing. 2 scans on a pair, 3 on a
  chain of three. Needs a cap to stay a total function, at a length nothing can reach.
- Two real costs. On the inherent span 4 pair it re-accepts at `A+3` rather than `A+5`, 60 seconds
  sooner, though from 3 presentations rather than 5; that whole class is what decision 11 already
  accepted as inherent. And it **cannot be tested at seam 1**, because no reachable secret exhibits a
  chain, so the scan has to move behind a step predicate and be pinned with a synthetic match set,
  which is a boundary below the seam you sealed in section 5.

**B. Accept the residual and narrow the claim.** One sentence in the doc comment and in decision 11
saying the constant answer holds for a pair and for a chain spanning at most 3, not beyond. Costs
nothing and ships #111 with a 3-in-100-billion instance of the defect #111 exists to close.

**C. Bind consumption to the passcode** rather than to the step, storing the consumed code or its
hash. Closes the chains and the inherent wide pairs with them. Reopens decision 1, which rejected a
consumed-codes table as too expensive for a 90-second window, and costs a second column in stage 2.

**D. Widen the lookback to a larger fixed number.** Recorded because it is the obvious first thought
and it does not work: any fixed `L` fails for a chain spanning more than `L`. Strictly dominated by A,
which is D taken to its limit at lower cost.

Either way the existing tests are unaffected: the three located pairs report identically under both
rules at every current step, executed in the probe, so `TestMatchStepWithCollidingSteps` and the
15-row table stand.

### Recommendation: A

It is the only candidate that makes the invariant decision 11 chose actually hold, it is free in the
case every real authentication takes, and no signature or later stage moves. Its honest price is the
test boundary, not the code.

B is defensible if 3 in 100 billion is acceptable for this property. That is the same question you
answered at 3 in a million, and it is why I am asking rather than applying: how literally RFC 6238 5.2
binds is not mine to decide. C is a larger change than #111 has called for anywhere else.

**Getting it wrong either way.** Choosing B and later reversing costs the matcher and one test
function, both still uncommitted, so it is cheap now and less so once stages 2 to 4 have built on the
returned step. Choosing A when B would have done costs a loop, a cap and a test at a lower boundary
than section 5 sealed.

---

Reply with a letter. Decision 12, the probe that measured it and the stage-1 account are all in the design record.

One note for later, not for you to answer now: this is the run's fourth escalation, which the workflow
treats as a signal the agreement was sealed too early. Decisions 10, 11 and 12 all came out of stage 1
and 11 and 12 are the same question twice, what a step means when a passcode does not name one. The
log entry records that so the next planning session sees it.

leodip:
A. Walk the lookback down transitively.

It is the only option under which the invariant decision 11 chose actually holds, rather than holding
for pairs and quietly failing for chains. It costs nothing whenever nothing collides, since the walk
stops the moment a scan finds nothing, and it moves no signature, so stages 2 to 4 are untouched. B
ships #111 with an instance of exactly the defect #111 exists to close. C reopens decision 1 for a
rate five orders of magnitude below what decision 11 already accepted as inherent.

Three things to fold in while applying it.

1. **Amend section 5 explicitly.** The walk cannot be exercised at seam 1, because no reachable
   secret exhibits a chain, so it needs a synthetic match set at a boundary below the seam the
   interview sealed. Record that as a named amendment to section 5 with the reason, not as an
   implementation detail inside a step.

2. **State the cap's reasoning where the cap is written.** Derive it from what the sweep can actually
   reach rather than from a round number, and say in the comment why nothing can exceed it.

3. **This answer settles the class, not just this shape.** The transitive walk canonicalises a chain
   of any length within the cap, so no further escalation about colliding steps is wanted. If a
   fourth shape turns up, apply the same rule and record it in the log entry.

For the record, on proportion. Pairs spread four or more apart stay replayable at roughly a millionth
per presentation, and decision 11 accepted that knowingly. That is about a hundred thousand times more
likely than the chain case being closed here. Option A is worth taking because it is nearly free, not
because chains are a threat. So do not spend further rounds narrowing this class: the residual that
dominates is the one already accepted, and closing that one needs the consumed-codes table decision 1
rejected.

_Drafted by Claude Code on my instruction, with its own recommendation._

leodip:
## Decision 12 applied: A, the transitive walk

Recorded in `e6c626a`. Stage 1's two files are rewritten and green, still uncommitted until the gate
closes. `MatchStep` keeps `(int64, bool)` and accepts exactly what it accepted before, so stages 2 to
4 are untouched. All twelve decisions are now `Decided`.

**The walk.** From the matched step, scan the `lookbackSteps` below it, and repeat from whatever that
finds until a scan finds nothing, so the answer is the bottom of the whole chain rather than the bottom
of one interval. The doc comment carries the sub-claim that makes one pass down sufficient: taking the
lowest match in each scan reaches at least as far as any higher match in the same scan could, since
anything that higher step could reach is itself within `lookbackSteps` of the lowest one.

**Your three amendments, all in.**

1. **§5 has a named amendment**, not a note inside a step: seam 1 also owns one boundary below itself,
   `matchStepWith` over a step predicate, pinned with a synthetic match set. It states why the walk
   cannot be reached through `MatchStep` at all, the three things the boundary owns, what it does not
   move, and why two of its rows assert a replay rather than safety.
2. **The cap is `maxLookbackScans = 5`, derived where it is written.** Not a round number: every scan
   that finds something needs one more colliding step, the sweep measured that link rate (2.75e-06 per
   step against 3.0e-06 modelled) rather than assuming it, and the stated criterion is that the cap
   must first refuse at a chain whose expected count stays under one in a billion across 1e18
   presentations. `capLadder` in the probe prints the ladder and names the first value clearing that
   criterion, independently of the constant in the code; it prints 5. The comment also says that
   reaching the cap refuses rather than answers, and that the cap bounds work rather than termination,
   since each scan already searches strictly below the last.
3. **Settled as a class.** No further escalation about colliding steps; a fourth shape would get the
   same rule and a line in the log entry.

**What I checked rather than argued.** Three mutations, run and reverted with `diff` confirming the
bytes back:

1. Apply the lookback once, which is decision 11's rule exactly: **6 of the 9 chain subtests fail and
   the other 3 pass**, the same 6 and the same 3 the probe reported from a separate model. The
   production code under test and the probe's model agreeing on which shapes leak is the strongest
   evidence in this round. The 15-row table and the three located pairs pass throughout: neither can
   see this defect.
2. Answer with the lowest step found so far when the cap fires: only the cap's refusal subtest fails.
3. Answer with the match when the walk errors: only that subtest fails, and it is a branch nothing
   could reach before this round.

The accepted residual stands as you restated it: the span 4 pair now re-accepts at `A+3` rather than
`A+5`, 60 seconds sooner and from 3 presentations rather than 5, and that class stays open at roughly a
millionth per presentation. Two test rows assert that, with a note saying asserting safety there would
assert something the design does not provide.

Core tier green on tree `ba3b1f75278d`, 54 `PASS` lines over 7 test functions in the `otp` package,
`go vet` and `gofmt` clean, anchors 24 of 24. Round 4 is requested at this gate.

leodip:
## Decision 13: the marker and the enrolment flag are written unbound, so an operation that loaded one authenticator generation can act on a later one. Bind the disable, or accept it?

Stage 4's code review found a new interleaving. It is real, I reproduced the mechanism in the code, and
it is **wider than the reviewer stated**. Decision 10 already answered one member of this family and
knowingly left another open, so I am asking the general question once rather than bringing you the next
shape separately.

Nothing is committed. Stage 4's diff is uncommitted in the working tree and stage 4 stays `In progress`,
exactly as stage 1 stood when decisions 11 and 12 were asked.

### The class

`otp_enabled` and `last_otp_step` are written by **separate, unbound statements**. Neither the claim nor
the reset carries any identity of the authenticator generation the request loaded, so any request that
loaded generation N can write over generation N+1. Three members, and the third is new:

- **(a)** A stale *verification* claim landing after a disable plus a re-enrolment, when `otp_enabled` is
  true again by the time the claim arrives. Decision 10 recorded this as "not closed, and knowingly",
  drafted as follow-up 2.
- **(b)** The mirror-image unbound enrolment shape stage 3's review found. Also follow-up 2.
- **(c) New.** A stale *disable reset* erasing a newer enrolment's successful claim. This one is not in
  follow-up 2: there the stale operation is a claim that succeeds when it should fail, here it is a reset
  that erases a claim that already succeeded, leaving the database persistently at `otp_enabled = 1,
  last_otp_step = 0` with a code already consumed.

### (c), the mechanism, verified in the code

Both disable sites do two unconditional writes with no transaction between them, in the order decision 10
requires:

```go
database.UpdateUser(nil, user)        // clears otp_enabled
database.ResetUserOTPStep(nil, user.Id)   // WHERE id = ?   and nothing else
```

`ResetUserOTPStep`'s predicate is `id` alone (`commondb/user.go`), so it is bound to nothing the disable
request observed. The schedule:

1. Disable D commits `otp_enabled = false`.
2. Enrolment E loads the user, sees OTP off, matches code C at step N, and **claims N successfully**.
3. D resumes and resets the marker to 0.
4. E's enable write lands (or landed at 2). End state: `otp_enabled = 1`, `last_otp_step = 0`, C consumed.
5. C is now claimable again at the browser prompt, where `requireOTPEnabled` is true and `otp_enabled`
   is true, for the remainder of C's ~90 second acceptance window.

The reviewer's `.review/scratch/issue111_stage4_disable_race.sql` executes those transitions on SQLite and
reports `enrollment claim rows|1`, `state after stale reset|1|0`, `replay claim rows|1`. That demonstrates
the database permits the state; the code permits the schedule, which is what I checked by reading it.

**Wider than the review states, and this is the part that decides feasibility.** The reviewer reasoned
about the account API as the racing enrolment. That path verifies the password between loading the user
and claiming, so the disable would have to stall for the whole bcrypt to lose the race, which is not
plausible. But the **browser** enrolment site is also an enrolment claim with `requireOTPEnabled` false,
and it has **no password verify in that gap**: `handler_auth_otp.go` loads the user at the top of
`HandleAuthOtpPost` and claims a few lines later, with a client fetch and a few HMAC computations in
between. So E needs microseconds inside D's inter-statement gap, not tens of milliseconds. The attacker
holds a prepared enrolment POST and fires it against their own disable.

**Impact, stated honestly, because decision 10's escalation overstated its own and you should not have to
discount mine.** Both operations need the account password, and anyone holding the password can already
disable OTP and walk in at `level2_optional` or self-enrol at `level2_mandatory`, which decision 10
established. So this buys **no access that the disable did not already grant**. What it buys is one further
acceptance of an already-consumed code, and a token asserting `amr: ["pwd","otp"]` for it. That is the same
false-factor-assertion harm decision 10 was asked about, and you chose to close it at its cause then.

**Why I am not treating this as rare enough to absorb under the proportion test.** The attacker chooses
the timing, supplies both requests, and can repeat without limit, because rate limiting this endpoint is
out of scope (#113). I cannot state plainly why an attacker cannot steer it, so the rarity carve-out does
not apply and it blocks. I also cannot produce a meaningful measured rate: it depends on the deployment's
load and connection-pool behaviour, not on anything I can pin in a probe.

### Options

**A. Wrap each disable's two writes in one transaction.** `BeginTransaction`, `UpdateUser(tx, user)`,
`ResetUserOTPStep(tx, user.Id)`, `CommitTransaction`, at both disable sites. A concurrent enrolment then
either sees `otp_enabled` still true, and is refused by `OTP_ALREADY_ENABLED` at the account API or takes
the verification branch in the browser where decision 10's flag refuses it, or sees the fully committed
post-disable state including the reset marker, and its claim stands. Closes (c) completely.

Costs: **no interface change, no column, no migration, no user-visible behaviour change.** Both methods
already take `tx`, `TestTryConsumeUserOTPStep_EnlistsInTransactionAndFailsClosed` already proves
enlistment on all four engines, and `BeginTransaction` is already the idiom in
`handler_api_users_crud.go`, one of the two files this touches. The real price is rollback error handling
on two paths and one new case proving the two writes are atomic. Leaves (a) and (b) exactly where
decision 10 knowingly left them.

**B. Make the reset a compare-and-set on the marker the disable loaded.**
`ResetUserOTPStep(tx, userId, expectedStep)` with `AND last_otp_step = ?`. Also closes (c). Costs a
`data.Database` signature change in the last stage, regenerated mocks, and changes to stage 2's
four-engine data table. It also weakens decision 4: a reset can now silently no-op, so if the disable's
loaded marker is stale for any unrelated reason the clock-jump remedy fails for that request. An
`otp_enabled = false` predicate on its own is **not** sufficient, because enrolment claims while that flag
is false.

**C. Accept it, fold it into follow-up 2, ship stage 4 as it stands.** Costs nothing now. Ships #111 with a
reachable state in which a consumed code is accepted again, which is the defect #111 exists to close.

**D. Add the enrolment-generation column now.** Closes (a), (b) and (c) together, which is what follow-up 2
drafts. A new column, a migration on four engines, and a fifth stage. Larger than anything else in #111.

### Recommendation: A

It closes the new shape at its cause, changes no interface and no schema, depends on enlistment that is
already proven on all four engines, and leaves the two residuals you already accepted knowingly where they
are. B closes the same shape while paying an interface change and weakening decision 4's remedy. C is the
only option that ships a reachable instance of the defect. D is the complete fix and it is follow-up 2's
job, not this change's.

Getting it wrong either way: choosing C and later reversing means reopening both disable sites and their
tests after stage 4 has landed and merged. Choosing A when C would have done costs a transaction on two
disable paths and one test.

### What else this round found, and what I did with it

Two further findings, both with the reviewer's own forced answer, both applied in the stage rather than
brought to you, and both independent of whichever option you pick here:

- **No test reached the account API's replay or claim-error branches.** Real, and §5's reason for
  rejecting a unit file there ("the claim logic is identical to seam 5's") stopped being true when this
  stage deliberately made the audit behaviour diverge. Added.
- **The `security.mdx` bullet overpromised.** It said an accepted code is refused everywhere afterwards,
  which decision 4's reset on disable deliberately contradicts, and two tests pin that contradiction.
  Qualified to the currently enrolled authenticator.

The run stops here until you answer. Everything else in stage 4 is done and its tiers were green.

leodip:
A

leodip:
## Decision 13 applied: A, one transaction at each disable site

Applied in the working tree, still uncommitted until the gate closes. All thirteen decisions are now
`Decided`, and stage 4 is the last stage.

**What changed.** Both disable sites, `api-otp/disable` and `admin-otp/disable`, now commit
`ClearOTPSecret` plus `otp_enabled = false` plus the marker reset as one transaction. A concurrent
enrolment therefore sees either the pre-disable state, where `OTP_ALREADY_ENABLED` refuses it at the
account API and decision 10's `requireOTPEnabled` refuses it at the browser verification branch, or the
fully committed post-disable state including the reset, where its claim stands. Member (c) is closed at
its cause. Members (a) and (b) stay exactly where decision 10 knowingly left them, as follow-up 2; your
answer carried no amendment and I have not widened it.

**Exactly what option A promised, and nothing more:** no interface change, no column, no migration, no
user-visible behaviour change. The two enrolment claim sites are untouched.

**One thing option A did not prescribe.** The five statements are a shared unexported function rather
than written out at both sites. The rollback belongs in a `defer`, and deferring at handler scope would
hold the transaction handle across the session write, the reload and the response encoding, which is
wider than the operation; a function of its own is what makes the `defer` correct. The two sites also
run the identical sequence, and a repair to one that misses the other is the failure mode this stage has
already met twice. Each site keeps its own 500 response.

**The ordering comment changed, and that is deliberate.** Decision 10's order is kept, `otp_enabled`
cleared before the marker, but inside a transaction no outside reader sees either write until both have
landed, so the commit boundary is now what closes the window rather than the ordering. Leaving the old
comment in place would assert something false about a reset-first arrangement.

**Three new cases, at the unit tier, because no sequential caller can see this repair.** The four-engine
reset cases already run this code on every engine and still pass, but they observe the identical end
state whether or not the two writes share a transaction. Only the call shape distinguishes them, so:
both writes and the commit in order at each site, registered against one opaque handle so a write that
slipped back to the pool fails as an unexpected call; and a failing reset that must roll back, not
commit, and not audit a disable that did not happen.

**What I checked rather than argued.** Three mutations, run and reverted, the handler restored
byte-identical by `sha256sum -c`:

1. Both writes unbound again, which is exactly the pre-decision-13 shape: all three cases fail, and the
   admin case's stack names `HandleAPIUserOTPPut`'s own call, so the second site's failure is
   attributable to the second site.
2. Commit moved before the reset, so the reset lands outside the transaction: all three fail.
3. Reset error swallowed and the transaction committed anyway: only the rollback case fails, which is
   what makes that case attributable to the rollback path alone.

**Tiers, all three on four engines**, because the repair touched production code. One
`where.sh test --type all`, 7m06s, `All tests completed successfully`, 0 FAIL, against tree
`6de62e71d30b`. Unit 2770 PASS. Data 1564 PASS, 4 pre-existing engine-capability skips. Integration 0
FAIL 0 SKIP on mysql, postgres, mssql and sqlite, with stage 4's three OTP cases each running four
times. Anchors 25 of 25, `gofmt` and `go vet` clean.

Code-review round 2 is requested at this gate, since the repair is production code the reviewer has not
seen.

leodip:
## Deferred decisions: none

Nothing was deferred. The deferred-decisions section of the design record is empty, and that is a
result rather than an omission: every question the run met was either resolved with a demonstration
recorded in that stage's log entry, or blocked and escalated to you. Nothing was carried forward on an
assumption, so there is no list here for you to settle before merging.

What that traded off is visible in the other direction. **Five questions blocked and came to you on
this PR**, and all five are applied:

| # | Raised | Question | Your answer |
|---|---|---|---|
| 10 | stage 1, plan review | Should the claim bind to enrolment state? | A, bind it |
| 11 | stage 1, code review | One passcode can come from two time steps. Which one is claimed? | B, the lowest in the chain |
| 12 | stage 1, round 3 | A fixed lookback closes pairs but not chains of three | A, walk it transitively |
| 13 | stage 4, code review | A stale disable reset can erase a newer enrolment's claim | A, one transaction |
| (6) | grounding | Is narrowing the skew window part of this change? | No, filed as #142 |

`decisions.md` §5 treats four escalations as evidence the agreement was sealed too early, so the note
for the next `/leo-spec` is written into decision 13 rather than left here: those five were **two
families asked twice each**. Decisions 11 and 12 are both "the time step is not a unique identity for a
passcode, how completely do you want that closed". Decisions 10 and 13 are both "the marker and the
enrolment flag are not bound to an authenticator generation, how completely do you want that closed".
Each family was answerable at its first escalation, and asking the general question then would have
cost you two replies instead of four.

**Two residuals are accepted on purpose**, both already decided rather than deferred, both restated
here because they are the honest limits of what shipped:

1. **Colliding step pairs spread 4 or more apart stay replayable**, at roughly one in a million per
   presentation. No marker that records a *step* can refuse them: the gap in acceptance means the
   authenticator is legitimately redisplaying those digits for a new step. Refusing it needs storing
   the code itself, which is the consumed-codes table decision 1 rejected. Decision 11 accepted this
   knowingly and decision 12 restated it; `probe/step-collisions` measured it.
2. **A consumed code can still assert `otp` if a disable is immediately followed by a re-enrolment
   while a stale request is in flight**, and the mirror-image case where two overlapping enrolment
   ceremonies both succeed. Decisions 10 and 13 both left these where they were. Closing them needs an
   enrolment-generation column, a migration and a fifth stage, which is follow-up 2 in the next
   comment.

leodip:
## Follow-ups: 2 to file, 1 already filed

Drafted, not filed: creating issues is yours. Duplicate searches were re-run at the closing pass on
2026-08-06 and returned nothing for either. Full text is in the follow-ups section of the design record.

**Already filed:** follow-up 1, narrowing the TOTP skew window, is
[#142](https://github.com/leodip/goiabada/issues/142) (decision 6).

---

### 1. A consumed code can still assert `otp` across a re-enrolment, and a stale enrolment can replace a fresh authenticator

`security`, `go`. Found while applying decision 10, widened by stage 3's review. This is the residual
decisions 10 and 13 both knowingly left open, and it is the one worth filing.

Two shapes, same class: **the claim binds to the enrolment flag, not to the authenticator**.

- After #111, `TryConsumeUserOTPStep` refuses a verification claim unless `otp_enabled` is still true,
  so removing an authenticator invalidates a claim from a request that loaded it. A re-enrolment
  restores `otp_enabled`, so a verification request holding the *old* secret can still claim a code it
  already spent, provided the new enrolment claimed a strictly lower time step. That needs the replayed
  code to come from the `+1` window and the re-enrolment from the current window or earlier, all three
  operations overlapping inside one 30 second period, and the attacker already holding both the level 1
  credential and a code from the removed authenticator.
- The enrolment claim has the mirror image, which predates #111: `otp/verify-enrolling` passes
  `requireOTPEnabled` false, so it binds nothing about *which* authenticator it establishes, while the
  enable is a whole-user `UpdateUser`. Two password-authenticated enrolment ceremonies that both loaded
  `otp_enabled` false both succeed, and the second one's write replaces the first one's secret. One-time
  use is not violated, since each ceremony spends a different code, which is why it was not a stage 3
  finding.

No access is gained that the disable did not already grant (see decision 10): a holder of the level 1
credential reaches level 2 more cheaply by disabling OTP or self-enrolling. The harm is a token
asserting `amr: ["pwd","otp"]` for an authenticator that no longer exists, without the
`AuditEnabledOTP` trail enrolling would leave. Adjacent to #140, the same false-factor-assertion class
by a different mechanism, so whoever picks one up should read the other.

Closing both needs the claim bound to the authenticator rather than to the flag: an enrolment
generation column on `users`, incremented on every enable and disable, carried on the request and
compared in the claim predicate. That is a migration plus a column, which #111 deliberately did not
take on.

```bash
gh issue create \
  --title "Security: the OTP claim binds to otp_enabled, not to the authenticator" \
  --label security --label go \
  --body 'After #111, `TryConsumeUserOTPStep` refuses a verification claim unless `otp_enabled` is still true, so removing an authenticator invalidates a claim from a request that loaded it. A re-enrolment restores `otp_enabled`, so a verification request holding the *old* secret can still claim a code it already spent, provided the new enrolment claimed a strictly lower time step. The resulting token asserts `otp` for an authenticator that no longer exists, without the `AuditEnabledOTP` trail that enrolling would leave. No access is gained that the disable did not already grant (see #111 decision 10), so this is a false factor assertion rather than an authentication bypass.

The enrolment claim has the mirror image of the same gap, which predates #111: it passes `requireOTPEnabled` false and so binds nothing about which authenticator it establishes, while the enable is a whole-user write. Two overlapping enrolment ceremonies both succeed and the later one'"'"'s write replaces the earlier one'"'"'s secret.

Closing both properly needs the claim bound to the authenticator rather than to the enrolment flag: an enrolment generation column on `users`, incremented on every enable and disable, carried on the request and compared in the claim predicate. That is a migration plus a second column, which #111 deliberately did not take on. Decide whether the window is worth that before building it.

Adjacent to #140, which is the same class of defect by a different mechanism.'
```

---

### 2. Redact credentials and OTP material from API debug request logging

`security`, `enhancement`. Raised by stage 4's code review as a follow-up rather than a finding, and
verified against the code before drafting. Out of scope for #111: §2 asks for nothing about debug
logging.

`src/authserver/internal/middleware/api_debug_middleware.go`, `debugLog`, at
`slog.Info(fmt.Sprintf("[DEBUG API]   Request Body:\n%s", prettyReq.String()))`. The function already
redacts the `Authorization` header to `Bearer ***`, so the intent to keep credentials out of logs is
there, but the bodies go to the log verbatim. `PUT /api/v1/account/otp` alone carries `password`,
`otpCode` and `secretKey`: the account password, a live TOTP code and the authenticator seed.

Debug-only and off by default, which is why it is a follow-up and not a blocker.

```bash
gh issue create \
  --title "Security: redact passwords, client secrets and OTP material from API debug logging" \
  --label security --label enhancement \
  --body 'When `GOIABADA_AUTHSERVER_DEBUG_API_REQUESTS` is enabled, `debugLog` in `src/authserver/internal/middleware/api_debug_middleware.go` pretty-prints the entire JSON request and response body to the log. The function already redacts the `Authorization` header to `Bearer ***`, so the intent to keep credentials out of logs is there, but the bodies are written verbatim. `PUT /api/v1/account/otp` alone carries `password`, `otpCode` and `secretKey`, which is the account password, a live TOTP code and the authenticator seed; other account and admin endpoints carry passwords and client secrets on the same path.

Redact by key before formatting, so the shape of the body is still readable for debugging while the values are not. Worth covering with tests over at least `password`, `currentPassword`, `newPassword`, `clientSecret`, `otpCode` and `secretKey`, and worth applying to the response body too, since enrollment responses return `secretKey`.

This is a debug-only path and off by default. The cost of leaving it is that anyone who turns the flag on to diagnose an unrelated problem writes reusable credentials and authenticator seeds into their console and log aggregator.'
```

Issue #149, opened 2026-08-07
Decide whether to support the logout_hint parameter


leodip:
OpenID Connect RP-Initiated Logout 1.0 defines `logout_hint`:

> "OPTIONAL. Hint to the Authorization Server about the End-User that is logging out. The value and meaning of this parameter is left up to the OP's discretion."

Goiabada does not read it. Nothing is broken by that, and the parameter is OPTIONAL with OP-defined semantics, so this is a product question rather than a defect: is there a use for it here?

The plausible use is the one the spec has in mind, telling the OP which End-User is logging out when no `id_token_hint` is available, so a multi-account session could preselect the right account. Goiabada has no multi-account browser session today, which is most of the argument for closing this.

Raised while working on #109, which fixes RP-initiated logout conformance defects and deliberately does not add optional parameters.

**Decide:** either define what `logout_hint` means for Goiabada and implement it, or close this as "won't do" so the question is settled and does not come back in the next conformance review.

Issue #148, opened 2026-08-07
Refactor: /api/v1/account/logout-request resolves the client by scanning every client


leodip:
`HandleAPIAccountLogoutRequestPost` in `src/authserver/internal/handlers/apihandlers/handler_api_account_logout.go` resolves the client by brute force when the request omits `clientIdentifier`:

```go
clients, err := database.GetAllClients(nil)
...
for i := range clients {
    c := &clients[i]
    if derr := database.ClientLoadRedirectURIs(nil, c); derr != nil { ... }
    for _, uri := range c.RedirectURIs {
        if uri.URI == postLogout { matches = append(matches, c); break }
    }
}
```

That is one `GetAllClients` plus one `ClientLoadRedirectURIs` **per client**, on every call, and it runs before the session and ownership checks below it. On a deployment with a few hundred clients this is a few hundred queries to answer a question a single indexed lookup against `redirect_uris` answers directly.

Found while working on #109 and left alone there, since that issue rewrites `/auth/logout` rather than this endpoint's client resolution.

**Fix:** resolve the client with one query joining `redirect_uris` on the supplied URI. Keep the existing "exactly one match, otherwise ask the caller for `clientIdentifier`" semantics: ambiguity is a real case, because two clients may register the same redirect URI, and the current error message for that case is good. Moving the resolution after the cheap validation above it would also avoid the work entirely on rejected requests.

Issue #147, opened 2026-08-07
Register post-logout redirect URIs separately from login redirect URIs


leodip:
`post_logout_redirect_uri` is validated against `client.RedirectURIs`, the OAuth **login** redirect URIs, because no separate list exists. `validateClientAndRedirectURI` in `src/authserver/internal/handlers/handler_account_logout.go` loops over exactly that set.

This conforms. OpenID Connect RP-Initiated Logout 1.0 says the value "MUST have been previously registered with the OP, either using the `post_logout_redirect_uris` Registration parameter **or via other mechanisms**", and the comparison is already exact, which also satisfies RFC 9700 section 2.1. So this is not a compliance defect.

The cost runs in one direction only, and that direction is the problem.

**Harmless direction.** Every login redirect URI automatically becomes a valid logout target. That is somewhere the server already delivers an authorization code, and the logout redirect carries only `state`, so nothing is widened.

**The direction that costs something.** To register a logout landing page, an operator must add it to `redirect_uris`. That makes the landing page a valid **authorization-code delivery endpoint**. A plain "you have been signed out" page, which may log its query string, be served statically, or simply lack the real callback's protections, becomes somewhere a code can be sent. The operator has no way to register the one capability without the other.

**Related gap.** `/connect/register` silently drops `post_logout_redirect_uris`; `grep -n "post_logout\|PostLogout" src/authserver/internal/handlers/handler_dynamic_client_registration.go` returns nothing. A client self-registering per OpenID Connect Dynamic Client Registration has the metadata ignored, and then finds its logout target rejected unless the same URI also happens to be a login redirect URI.

**Scope.** Found while working on #109, and deliberately excluded there: this needs a migration across all four engines, a data layer with its engine wrappers and regenerated mocks, an admin console UI, DCR handling, API request and response shapes, and data tests on four engines. That is larger than everything #109 does put together, and #109 otherwise needs no migration.

**Fix:** add a `post_logout_redirect_uris` list to clients, prefer it when validating `post_logout_redirect_uri`, and fall back to `redirect_uris` when it is empty so existing deployments keep working. Accept the metadata at `/connect/register`.

Issue #146, opened 2026-08-07
Bug: redirToClientWithError emits two state parameters when the redirect URI already carries one


leodip:
`redirToClientWithError` in `src/authserver/internal/handlers/handler_authorize.go` seeds its query from the registered redirect URI and then appends `state` with `values.Add`:

```go
values := redirUrl.Query()
values.Add("error", code)
values.Add("error_description", description)
if len(strings.TrimSpace(state)) > 0 {
    values.Add("state", state)
}
redirUrl.RawQuery = values.Encode()
```

When the registered `redirect_uri` already contains a `state` parameter, the RP receives **two**. Which one it reads depends on its own query parser: Go's `Query().Get` and most frameworks take the first, which is the registered value rather than the one the client sent, so the client's CSRF check compares against something it never generated.

Verified by running both constructions over the same input. With a registered URI of `https://app.example.com/cb?state=fixed` and a supplied state of `abc`:

- `Add` produces `?state=fixed&state=abc`, parsed as `["fixed" "abc"]`
- `Set` produces `?state=abc`, parsed as `["abc"]`

Two smaller issues in the same block: `TrimSpace` is used to decide whether to emit `state`, so a whitespace-only state is silently dropped, and `state` is an opaque value the authorization server has no licence to alter. The fragment branch above has the same shape.

The identical defect was found in the RP-initiated logout redirect while working on #109 and is fixed there by switching to `Set` and by extracting `state` presence-aware rather than as a bare string. This call site was deliberately left alone, because it is a different flow with a different error surface and #109's scope is the logout endpoint.

**Fix:** use `values.Set("state", state)` in both the query and fragment branches, stop trimming the value, and decide emission on whether the parameter was present rather than whether it is non-empty. Add a test case with a registered redirect URI that already carries `state=`.

PR #150, opened 2026-08-07
fix(authserver): make RP-initiated logout conform to OpenID Connect RP-Initiated Logout 1.0

### LINKED ISSUE #109: RP-Initiated Logout: post_logout_redirect_uri wrongly required, unencoded state, and non-standard sid in the redirect

leodip:
Closes #109.

`/auth/logout` implements OpenID Connect RP-Initiated Logout 1.0 and nothing else. Three defects the
issue reports, plus four more found while verifying it, are fixed here. The headline one is that the
most common conforming request failed closed in the worst possible way: a `GET /auth/logout` with a
valid `id_token_hint` and no `post_logout_redirect_uri` rendered an error page **and left the user
signed in**, because both the database teardown and the cookie wipe sat after that early return.
`post_logout_redirect_uri` is OPTIONAL in the spec.

## What changed

**One rule replaced seven special cases.** `doLogoutWithIdToken` and the no-hint half of
`HandleAccountLogoutPost` collapsed into a single pipeline that both HTTP methods enter, with four
independent stages: classify the hint, decide whether to ask the End-User, resolve the redirect
target, tear down and respond.

The property that fixes the reported defect and its six siblings at once, and the one to check first
when reviewing: **the teardown reads nothing the redirect resolution produced.** Hint validity decides
whether the End-User is asked. Redirect-target validity decides only whether a redirect happens.
Neither decides whether the logout happens, which always does.

| Request | Response |
|---|---|
| Valid hint matching the current session | tear down, then redirect if the target validates, else the signed-out page |
| No hint, or a hint that cannot be confirmed | consent page; on the confirming POST tear down, then redirect if `client_id` validates the target, else the signed-out page |
| Any redirect target that cannot be validated | tear down anyway, then the signed-out page, never a redirect |

**The hint gets three states, not two: confirmed, absent, and supplied-but-rejected.** Collapsing
"rejected" into "absent" is what let a detected `client_id`/`aud` mismatch go on to authorize a
redirect, which section 4 of the spec forbids twice over: information that failed to validate MUST NOT
be used, and on detecting an error the OP MUST NOT perform post-logout redirection. A rejected hint now
permits consent and teardown and forbids any redirect, whatever `client_id` says.

**`state` is built with `url.Values` instead of string concatenation.** All four failure modes were
reproduced before being fixed: a base64 `state` had its `+` decoded to a space, breaking the RP's CSRF
check; a `#` truncated the value into a fragment; an `&` injected arbitrary parameters; and a
registered URI with its own query was corrupted to `lang=en?sid=...`. `state` now round-trips byte for
byte, is not trimmed (the spec calls it opaque), and arrives exactly once even when the registered URI
already carried one. An RP that sent `state=` gets `state=` back and one that sent nothing gets none,
which needed a presence-aware parameter lookup, since the existing helper returns `""` for both.

**The non-standard `sid` is gone from the redirect.** RP-Initiated Logout defines exactly one parameter
on the way back, `state`. `sid` belongs to Front-Channel Logout: a different endpoint, the opposite
direction, and a spec that requires `iss` alongside it. Nothing consumed it, and it was an identifier
disclosure into the RP's URL bar, history and access logs. Nothing replaces it.

**A hintless logout now actually ends the session.** It used to clear the cookie and redirect without
touching the database, so session-bound refresh tokens kept working and the row was orphaned from a
browser that could no longer reach it. Every confirmation of the consent page takes that branch. It now
loads the session by the cookie's identifier and deletes it, emitting `AuditDeletedUserSession` beside
the existing `AuditLogout`. Offline grants survive, which is OIDC Core section 11's reading and what
#129 adopted.

**A hint naming a different session no longer returns HTTP 500.** The spec says to ask the End-User
when the supplied ID Token does not belong to the current session, so it renders the consent page.
Reachable with no attacker: sign in, the RP holds an ID token, the session is replaced, the RP
initiates logout with the older one.

**An expired hint is accepted while its session lives.** The spec: the OP SHOULD accept ID Tokens whose
`aud` or `sid` has a current or recent session "even when the `exp` time has passed". The admin console
mints a 60-second hint, so an admin who paused on the way through failed logout. Parsing now runs with
the library's claims validation off and every claim is checked by hand, because the library's `false`
branch disables **all** claims checking rather than just `exp`. A past `exp` is tolerated only when the
`sid` resolves to a live session row.

**`POST /auth/logout` works cross-site, when and only when it carries an `id_token_hint`.** The spec is
a MUST: "OpenID Providers MUST support the use of the HTTP GET and POST methods defined in RFC 7231".
`/auth/logout` was absent from the CSRF exempt paths, so gorilla/csrf refused every cross-origin POST
and the binding was unreachable for the RPs it exists for. The exemption is a predicate over the
request rather than a path entry: POST only, and only on hint **presence**, since middleware cannot
judge validity.

A hintless POST keeps full CSRF protection, and that is what makes the exemption safe: a hintless POST
is exactly what the consent screen submits, so a cross-site one with a garbage hint gets the consent
page rather than a teardown. The consent form drops the hint, so confirming it is always hintless and
always CSRF-checked. A skipped POST with a rejected hint is answered `303 See Other` to
`GET /auth/logout`, carrying `post_logout_redirect_uri`, `state` and `ui_locales` and dropping
`id_token_hint` and `client_id`, because a CSRF-skipped response carries no token and its own consent
page would otherwise be unsubmittable.

**Ten logout error codes and `renderAuthError` are deleted**, 40 deletions across four files each. The
diagnostics rendered to the End-User, for whom "the `id_token_hint` parameter is invalid" is noise; the
reason belongs in the server log, and the four sites that lacked one gained it. A new `logged_out.html`
replaces them as the terminal page, with one optional note when a redirect target was supplied and not
honoured. It also replaces the old redirect to the auth server's base URL, which bounced a logged-out
RP user onto an admin sign-in page they had no reason to see.

## What did not change, deliberately

**Per-client teardown on the hinted path.** #129 decided grants survive logout and gave the reason: a
user logging out of one client would silently break the other clients still on that session. Its two
integration tests pass unedited, which was the check that the change had not exceeded its scope.
Client-scoped revocation is #135.

**The cookie wipe stays whole-session** while the database teardown stays per-client. The asymmetry is
real and recorded in a code comment: the RP asked the OP to log the End-User out, and that is the OP
session.

**`client_id` never scopes a teardown.** It is unauthenticated, so a caller could name any client. It
is trusted for exactly one thing, deciding whether a supplied URI is registered to the client it names,
which is self-limiting because the URI must already be in that client's set. Only a confirmed hint,
whose `aud` is signed, scopes the per-client teardown.

## Stages

| Stage | Commit | What landed |
|---|---|---|
| 1 | `a4a586d` | the post-logout redirect builder and the presence-aware parameter lookup |
| 2 | `7a6c039` | the hintless path end to end: real teardown, the signed-out page, 10 error codes deleted |
| 3 | `ad43d76` | hint classification into confirmed, absent and rejected |
| 4 | `6a0f47d` | the hinted pipeline, collapsing both bindings into one |
| 5 | `3972dc4` | the conditional CSRF exemption |

## Tests

Full suite green on all four engines: **9308 passing, zero failures**, 10m11s.

| Tier | Engines | Result |
|---|---|---|
| unit | n/a | 2800 pass |
| data | mysql, postgres, mssql, sqlite | 1564 pass |
| integration | mysql, postgres, mssql, sqlite | 4852 pass |

No migration and no new `Database` method: the teardown uses `GetUserSessionBySessionIdentifier` and
`DeleteUserSession`, and `user_session_clients` already cascades in all four engines.

Coverage sits at four seams. The redirect builder owns the encoding table exhaustively, including
`state` containing `+`, `/`, `=`, `#` and `&`, absent versus empty versus whitespace-only, and a
registered URI that already carries its own `state`. Hint classification owns one case per rejection
reason, each differing from the confirmed case in exactly one field, 34 rows. The two handlers own the
method-by-hint-state-by-target table. `shouldSkipCsrf` owns the exemption, including that a handler
downstream still reads the body after the middleware parsed it looking for the hint. Integration owns
only what mocks cannot show: a real `Location` header, a real cross-origin POST, the row actually gone,
and an offline grant still refreshing afterwards.

Mutation testing found a defect in this change's own tests, which is the most useful thing the run
produced. The cross-origin confirmed-hint case asserted a teardown happened, and passed under a
mutation that broke the feature, because a hintless POST tears the whole session down too. It now
requires the `302` that only a confirmed hint can authorize. A later round found the same shape again:
every cross-origin case ran with a session cookie a real browser would withhold under `SameSite=Lax`,
so a case was added with no cookie at all, where the signed hint is the request's only authority.

## Notes for review

Two **deferred decisions** are in [a comment below](https://github.com/leodip/goiabada/pull/150#issuecomment-5223581795):
whether the registered URI's query is preserved byte for byte or canonicalised, and whether the
consent page's integration coverage should drive a real browser. Both are cheap to reverse and both
name what reversing costs.

Three **drafted follow-ups** are in [a second comment](https://github.com/leodip/goiabada/pull/150#issuecomment-5223585941),
each with the `gh issue create` command that files it: a client/resource identifier namespace collision
that makes an access token indistinguishable from an ID token to audience validation (hardening, since
the one live instance is closed here), the consent page's No button submitting without JavaScript, and
the CSS build harvesting class names from the whole Go module. #146, #147, #148 and #149 were filed
during the run.

One decision was escalated and answered here: how a CSRF-skipped POST reaches a submittable consent
page, answered `A`, the `303`.

### DIFF
--- a/site/src/content/docs/concepts/user-sessions.mdx
+++ b/site/src/content/docs/concepts/user-sessions.mdx
@@ -73,7 +73,7 @@ Nor can a sign-in that had already reached the consent screen. Goiabada checks t
 
 An application renewing quietly in the background with [`prompt=none`](/concepts/prompt-parameter/) hits the same check and is told rather than shown: it gets a `login_required` error back, never a login page, so it can decide when to ask the person to sign in again.
 
-Logging out is not the same thing. It is navigation rather than a security decision, so it leaves refresh tokens alone.
+Logging out is not the same thing. It revokes nothing: no refresh token is marked revoked and no authorization code is touched. It can still end the session itself, and a normal refresh token stops working once the session behind it is gone. Offline refresh tokens are not tied to a session and keep working either way. See [/auth/logout](/integration/endpoints/) for which logout requests end the whole session.
 
 Each termination writes two entries to the [audit log](/concepts/audit-log/): `terminated_user_session`, recording what was revoked, and `deleted_user_session`, the lifecycle record beside it. Count `terminated_user_session` if you want to know how many sessions were ended.
 
--- a/site/src/content/docs/integration/endpoints.mdx
+++ b/site/src/content/docs/integration/endpoints.mdx
@@ -366,6 +366,24 @@ You can still be redirected back to your application on this path, by sending `p
 Sending `post_logout_redirect_uri` with neither `id_token_hint` nor `client_id` never redirects. The spec requires the OP to confirm the legitimacy of the redirection target first, and with neither parameter there is nothing to confirm it against.
 </Aside>
 
+### Choosing GET or POST
+
+Both bindings accept the same parameter names. Send them where the method puts them: in the query string on a GET, and in an `application/x-www-form-urlencoded` body on a POST. A GET's body is never read, so a parameter sent only there is ignored.
+
+A POST reads the query string as well as its body, so send each parameter once. Which copy wins when the same one arrives in both places is not worth relying on.
+
+What differs between the two methods is not the parameters but whether the request is accepted at all, which is what the rest of this section is about.
+
+POST is worth using when you have an `id_token_hint`, because it keeps the ID token out of the browser's address bar, its history and the referrer headers it sends on to other sites. A self-submitting form from your own page is the usual way to do it, and that form is a cross-site request: it comes from your origin and posts to Goiabada's.
+
+**A cross-site POST needs an `id_token_hint`.** Without one Goiabada answers `403 Forbidden`, because a POST with no hint is exactly what the logout consent screen submits, and treating an arbitrary site's request as a user's confirmation would let any page sign your users out. With a hint the request is accepted and evaluated on the hint's merits: a hint Goiabada can confirm logs the user straight out, and one it cannot is answered with a `303 See Other` back to `GET /auth/logout`, which is where the consent screen comes from. Follow the redirect and the user is asked to confirm, as they would have been on a GET.
+
+That `303` carries your `post_logout_redirect_uri`, `state` and `ui_locales` onward, and deliberately drops `id_token_hint` and `client_id`. Dropping `client_id` is why a hint that fails validation earns no redirect even when a valid `client_id` was sent beside it.
+
+<Aside>
+This is the only reason to think about the method at all. A GET is never refused, so if you are sending users to `/auth/logout` with a link or a browser redirect, nothing here applies to you.
+</Aside>
+
 ### RP-initiated logout (with parameters)
 
 For proper logout with redirection back to your application, use either GET or POST with the following parameters:
@@ -402,7 +420,7 @@ For proper logout with redirection back to your application, use either GET or P
     <tr>
       <td style="white-space: nowrap;"><code>ui_locales</code></td>
       <td>No</td>
-      <td>Space-separated list of preferred locales for the pages this endpoint renders, such as <code>pt-BR en</code>. Read from the query string or the form body.</td>
+      <td>Space-separated list of preferred locales for the pages this endpoint renders, such as <code>pt-BR en</code>. Read from the query string, and on a POST from the form body as well.</td>
     </tr>
   </tbody>
 </table>
--- a/site/src/content/docs/integration/endpoints.mdx
+++ b/site/src/content/docs/integration/endpoints.mdx
@@ -349,10 +349,22 @@ The response body contains error details:
 
 This endpoint enables the client application to initiate a logout. This implementation aligns with the [OpenID Connect RP-Initiated Logout 1.0 protocol](https://openid.net/specs/openid-connect-rpinitiated-1_0.html).
 
-### Basic logout (no parameters)
+### Logout without an id_token_hint
 
-- **GET /auth/logout** - Displays a logout consent screen, prompting the user to confirm their intention to log out. No redirection occurs.
-- **POST /auth/logout** - Immediately logs out the user and redirects to the auth server base URL.
+`id_token_hint` is recommended but not required. Without one, Goiabada cannot tell which client is asking, so it asks the user instead:
+
+- **GET /auth/logout** - Displays a logout consent screen, prompting the user to confirm their intention to log out.
+- **POST /auth/logout** - What that screen submits. The user is logged out: the session record is deleted and the session cookie is cleared.
+
+Once the user confirms, they land on a page telling them they have been logged out.
+
+Ending the session reaches the tokens that depend on it, and no further. A normal refresh token from that session stops working, because there is no session behind it any more. Offline refresh tokens are not tied to a browser session and keep working. Access tokens that were already issued stay cryptographically valid: Goiabada's own endpoints, `/userinfo` and the account and admin APIs, reject one whose session is gone on the very next request, but a third-party resource server validating a token by signature alone cannot see that the session ended and keeps accepting it until it expires.
+
+You can still be redirected back to your application on this path, by sending `post_logout_redirect_uri` together with `client_id`. The URI must exactly match one registered for that client, and only `state` is added to it. Your `state` comes back byte for byte, including any `+`, `/`, `=`, `#` or `&`; sending it empty gives you back an empty `state`, and not sending it gives you no `state` at all. If the URI cannot be validated, because `client_id` is missing or unknown, or because the URI is not registered for that client, the user is **still logged out** and lands on the logged-out page with a short note saying we could not return them to the application.
+
+<Aside>
+Sending `post_logout_redirect_uri` with neither `id_token_hint` nor `client_id` never redirects. The spec requires the OP to confirm the legitimacy of the redirection target first, and with neither parameter there is nothing to confirm it against.
+</Aside>
 
 ### RP-initiated logout (with parameters)
 
@@ -369,28 +381,33 @@ For proper logout with redirection back to your application, use either GET or P
   <tbody>
     <tr>
       <td style="white-space: nowrap;"><code>id_token_hint</code></td>
-      <td>Yes</td>
-      <td>The previously issued ID token (can be encrypted or unencrypted).</td>
+      <td>Recommended</td>
+      <td>The previously issued ID token (can be encrypted or unencrypted). Send it when you have it: it identifies the client and session, so the user is not asked to confirm. Without it, the logout still happens, after the consent screen.</td>
     </tr>
     <tr>
       <td style="white-space: nowrap;"><code>post_logout_redirect_uri</code></td>
-      <td>Yes</td>
-      <td>A redirect URI that must be pre-registered with the client.</td>
+      <td>Conditional</td>
+      <td>A redirect URI that must be pre-registered with the client. Required when you send an <code>id_token_hint</code>. Without a hint it is optional: leave it out and the user is logged out and lands on the logged-out page.</td>
     </tr>
     <tr>
       <td style="white-space: nowrap;"><code>client_id</code></td>
       <td>Conditional</td>
-      <td>Required only if <code>id_token_hint</code> is encrypted with the client secret.</td>
+      <td>Required if <code>id_token_hint</code> is encrypted with the client secret, to select whose secret derives the key. Also required to authorize <code>post_logout_redirect_uri</code> when you send no <code>id_token_hint</code>.</td>
     </tr>
     <tr>
       <td style="white-space: nowrap;"><code>state</code></td>
       <td>No</td>
       <td>Any arbitrary string that will be echoed back in the redirect.</td>
     </tr>
+    <tr>
+      <td style="white-space: nowrap;"><code>ui_locales</code></td>
+      <td>No</td>
+      <td>Space-separated list of preferred locales for the pages this endpoint renders, such as <code>pt-BR en</code>. Read from the query string or the form body.</td>
+    </tr>
   </tbody>
 </table>
 
-### Response
+### Response (with an id_token_hint)
 
 After successful logout, the user is redirected to the `post_logout_redirect_uri` with the following query parameters:
 
--- a/site/src/content/docs/integration/endpoints.mdx
+++ b/site/src/content/docs/integration/endpoints.mdx
@@ -386,8 +386,8 @@ For proper logout with redirection back to your application, use either GET or P
     </tr>
     <tr>
       <td style="white-space: nowrap;"><code>post_logout_redirect_uri</code></td>
-      <td>Conditional</td>
-      <td>A redirect URI that must be pre-registered with the client. Required when you send an <code>id_token_hint</code>. Without a hint it is optional: leave it out and the user is logged out and lands on the logged-out page.</td>
+      <td>No</td>
+      <td>A redirect URI that must be pre-registered with the client. Optional in every case: leave it out and the user is logged out and lands on the logged-out page.</td>
     </tr>
     <tr>
       <td style="white-space: nowrap;"><code>client_id</code></td>
@@ -407,14 +407,27 @@ For proper logout with redirection back to your application, use either GET or P
   </tbody>
 </table>
 
-### Response (with an id_token_hint)
+### Response
 
-After successful logout, the user is redirected to the `post_logout_redirect_uri` with the following query parameters:
+When a `post_logout_redirect_uri` is supplied and can be validated, the user is redirected to it after the logout. Exactly one parameter is added:
 
-- `sid` - The session identifier from the ID token
-- `state` - The state parameter if provided in the request
+- `state` - only if you sent one, and byte for byte what you sent
 
-Example redirect: `https://your-app.com/logged-out?sid=abc123&state=xyz`
+Anything the registered URI already carried in its own query is preserved.
+
+Example redirect: `https://your-app.com/logged-out?state=xyz`
+
+With no `post_logout_redirect_uri`, or when the URI cannot be validated, the user lands on Goiabada's logged-out page instead. Either way the logout has already happened.
+
+### When the id_token_hint cannot be used
+
+Goiabada checks the hint's signature, issuer, audience, session and expiry, and that `client_id` matches the audience when you send both. If any of those fail, the request is not an error: the user sees the logout consent screen, and once they confirm, they are logged out.
+
+A hint that fails these checks earns no redirect, even if you also sent `client_id` and a registered `post_logout_redirect_uri`. The user is logged out and lands on the logged-out page with the note. Send a hint you can vouch for, or send none at all and let `client_id` authorize the redirect.
+
+<Aside>
+An **expired** `id_token_hint` is still accepted, as long as the session it names is still alive. This is deliberate, and the spec recommends it: hints are short lived, and a user who pauses between receiving one and following it should not have their logout fail. A hint whose session has already ended is treated as one that cannot be confirmed.
+</Aside>
 
 ### Encrypting the ID token hint
 
--- a/src/authserver/internal/handlers/handler_account_logout.go
+++ b/src/authserver/internal/handlers/handler_account_logout.go
@@ -3,6 +3,7 @@ package handlers
 import (
 	"log/slog"
 	"net/http"
+	"net/url"
 	"strings"
 
 	"github.com/gorilla/csrf"
@@ -351,4 +352,84 @@ func doLogoutWithIdToken(
 		logoutUri += "&state=" + state
 	}
 	http.Redirect(w, r, logoutUri, http.StatusFound)
+}
+
+// buildPostLogoutRedirect returns the Location value for a post-logout redirect: the
+// registered URI with the RP's state written into its query, properly escaped. Building
+// it by string concatenation instead is what let a state containing "+", "/", "=", "#"
+// or "&" reach the RP altered or truncated, and let a registered URI that already
+// carried a query gain a second "?" (#109).
+//
+// The registered query is copied field by field, verbatim and in order, empty fields
+// included, and only the RP's state is encoded. Decoding it into url.Values and
+// re-encoding it, which is what
+// redirToClientWithError does in handler_authorize.go, does not round-trip: url.Query
+// discards the error from url.ParseQuery, so a field separated by a literal semicolon
+// ("?lang=en;mode=dark") is deleted outright and the caller never learns; Encode sorts
+// by key, so a query whose order the RP signs over comes back reordered; a valueless
+// field ("?flag") gains an "="; and percent-escapes are normalised ("%7E" to "~"). The
+// registered URI is an operator-registered target that was just matched exactly, so
+// altering its query sends the RP somewhere it did not register. None of those shapes is
+// rejected at registration: validateRedirectURI's excluded-character set is
+// "<>\"{}|\\^` " and admits a semicolon.
+//
+// Copying registered bytes into a Location header is safe here because url.Parse above
+// rejects CR, LF and NUL outright ("net/url: invalid control character in URL"), so a
+// registered URI that reaches this point cannot carry a header-splitting byte. The RP's
+// state is attacker-influenced and is escaped, never copied.
+//
+// Three further properties, each load-bearing, so changing one back is a behaviour
+// change rather than a tidy-up:
+//
+//   - Exactly one state reaches the RP. Every registered field whose decoded name is
+//     "state" is dropped and one is appended, so the RP reads the value it sent rather
+//     than the registered one, and never both with the choice left to its parser.
+//   - No TrimSpace. OpenID Connect RP-Initiated Logout 1.0 section 2 calls state an
+//     "Opaque value used by the RP to maintain state between the logout request and the
+//     callback", so the OP has no licence to alter it. Trimming would turn a
+//     whitespace-only state into no state at all.
+//   - url.Parse rather than url.ParseRequestURI. A fragment is not part of an HTTP
+//     request URI, so ParseRequestURI keeps a literal "#" in the path and the result
+//     comes back as ".../out%23frag?state=abc". url.Parse still rejects the genuinely
+//     malformed, so the looser parse gives up nothing.
+//
+// statePresent, rather than a len(state) > 0 guard, is what keeps three cases distinct:
+// a state that was absent writes nothing, a state supplied empty comes back as "state=",
+// and a whitespace-only state survives byte-identical.
+func buildPostLogoutRedirect(registeredURI string, state string, statePresent bool) (string, error) {
+	redirUrl, err := url.Parse(registeredURI)
+	if err != nil {
+		return "", errors.Wrap(err, "unable to parse post-logout redirect URI")
+	}
+
+	if !statePresent {
+		return redirUrl.String(), nil
+	}
+
+	fields := make([]string, 0, 4)
+	// An empty RawQuery is the only field the loop must not see. strings.Split("", "&")
+	// yields one empty string rather than nothing, so iterating unconditionally would
+	// append a leading "&" to a URI that had no query at all. Guarding here rather than
+	// skipping empty fields inside the loop is what keeps the empty fields a registered
+	// query really did carry: "?a=1&&b=2" stays that way instead of collapsing to
+	// "?a=1&b=2", which is a different target from the one the operator registered and
+	// the OP just matched exactly (#109).
+	if redirUrl.RawQuery != "" {
+		for _, field := range strings.Split(redirUrl.RawQuery, "&") {
+			name := field
+			if i := strings.IndexByte(field, '='); i >= 0 {
+				name = field[:i]
+			}
+			// An unescapable name cannot be "state", and is kept as it stands rather than
+			// dropped, since the point is to preserve what was registered.
+			if decoded, err := url.QueryUnescape(name); err == nil && decoded == "state" {
+				continue
+			}
+			fields = append(fields, field)
+		}
+	}
+	fields = append(fields, "state="+url.QueryEscape(state))
+
+	redirUrl.RawQuery = strings.Join(fields, "&")
+	return redirUrl.String(), nil
 }
--- a/src/authserver/internal/handlers/handler_account_logout.go
+++ b/src/authserver/internal/handlers/handler_account_logout.go
@@ -646,7 +646,12 @@ func doLogout(
 // class of defect this endpoint was rewritten to remove. The GET binding is never exempt, so the page
 // it renders has a real token and a real cookie (#109).
 //
-// id_token_hint is dropped because sending it back would land on this branch again, forever.
+// id_token_hint is dropped, and not to prevent a loop: the follow-up is a GET, and the switch above
+// sends a rejected hint on any method but POST to the consent page, so sending the hint back would
+// terminate rather than return here. It is dropped because carrying it puts an ID token in the
+// address bar of a top-level navigation, its history and its referrers, which is the exposure the
+// POST binding exists to avoid, and because feeding input already judged unusable into the next
+// request gives that request nothing it can act on anyway.
 // client_id is dropped because the follow-up GET is then the hint-absent shape, where client_id is
 // what authorizes a post-logout redirect: keeping it would hand the request the redirect authority a
 // rejected hint is specifically denied. With no client_id nothing validates the target, so the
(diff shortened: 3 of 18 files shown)


leodip:
## Decision 18: a CSRF-skipped `POST /auth/logout` renders a consent page that cannot be submitted. How does that shape reach a submittable one?

This is a blocking question for #109. The run has stopped at stage 4 of 5 and needs an answer
before it can implement "decide whether to ask".

### Where this comes from

#109 makes `/auth/logout` conform to OpenID Connect RP-Initiated Logout 1.0. Two of its sealed
decisions collide here.

- The spec requires the POST binding: "OpenID Providers MUST support the use of the HTTP GET and
  POST methods defined in RFC 7231." Today `/auth/logout` is absent from the CSRF exemption list, so
  gorilla/csrf rejects every cross-origin POST and the binding is unreachable for the RPs it exists
  for. This change exempts it, but only for a request that carries an `id_token_hint`, because
  middleware cannot judge whether the hint is genuine. A hintless POST keeps full CSRF protection.
- The spec also requires the consent question: "the OP MUST ask the End-User this question if an
  `id_token_hint` was not provided or if the supplied ID Token does not belong to the current OP
  session with the RP and/or currently logged in End-User." So a POST carrying a hint that fails
  validation renders the consent page rather than tearing anything down.

Put together: a cross-site POST with a hint that does not validate skips CSRF and renders the
consent page.

### The defect

gorilla/csrf v1.7.3 issues its token *inside* the call that the skip flag makes it return from.
`ServeHTTP` opens with

```go
if val, err := contextGet(r, skipCheckKey); err == nil {
    if skip, ok := val.(bool); ok {
        if skip {
            cs.h.ServeHTTP(w, r)
            return
        }
    }
}
```

and returns there, **before** `cs.st.Save` writes the `_gorilla_csrf` cookie and before
`contextSave(r, tokenKey, ...)` and `contextSave(r, formKey, ...)`. `csrf.TemplateField` returns
`template.HTML("")` when `formKey` is absent from the context.

`MiddlewareSkipCsrf` is registered ahead of `MiddlewareCsrf` in both apps
(`src/authserver/internal/server/server.go:269-270`), so the flag is set before gorilla runs.

The result: the consent page renders with an empty CSRF field. Its own confirming POST is hintless
by design (the form deliberately drops `id_token_hint`, which is what stops a bogus hint from
converting the exemption into a forced logout), so that POST is not exempt and gorilla refuses it
with 403. **The user cannot complete the logout.**

Reachable with no attacker at all: the RP holds an ID token, the user's session is replaced (reaped,
ended elsewhere, signed in again), the RP posts cross-site with the older token, the `sid` claim
mismatches, the hint is rejected. #109's own goal statement says no request shape may reach a
terminal page while leaving the user signed in, and it relies on the consent page not being
terminal. A consent page that cannot be submitted is terminal.

### Scope of the problem: exactly one shape

The templates that render a CSRF field are `auth_pwd`, `auth_otp`, `auth_otp_enrollment`, `consent`,
`logout_consent`, `account_register`, `forgot_password` and `reset_password`. No currently exempt
path renders any of them: `/auth/authorize` is the only exempt path that renders at all, and it
renders `auth_error.html`, which has no form. So this is not a pre-existing class, and closing it
here closes all of it.

Reordering the two middlewares does not help, which is why the option space is only the three below.
The skip flag has to reach the context before gorilla reads it, and gorilla issues the token inside
the call it then returns from.

### The options

**A. Answer the skipped POST with `303 See Other` to `GET /auth/logout`**, carrying
`post_logout_redirect_uri`, `state` and `ui_locales`, and dropping `id_token_hint` and `client_id`.
The browser re-requests by GET, which is not exempt, so the consent page gets a real token and
cookie and confirming it tears down. The handler needs no gorilla introspection to know it was
skipped, since the skip condition is exactly "POST with the hint present", which it already knows.

Dropping `client_id` is load-bearing: without it, the follow-up GET is the hint-absent case, where
`client_id` is what authorizes a post-logout redirect, and a rejected hint's `client_id` would
regain the redirect authority this change deliberately took away from it. Carrying the redirect URI
without a `client_id` means no client validates the target, so no redirect happens and the user
lands on the signed-out page with the "we could not return you to the application" note, which is
the intended outcome for a rejected hint.

Costs an extra round trip, and the endpoint answers 303 where it answered 200.

**B. Render an interstitial page whose only control is a GET link to `/auth/logout`.** Same
destination, no redirect. Needs a new template and new catalog keys in both languages that this
change did not budget for, and costs the user a second action.

**C. Accept it and document it.** A cross-site POST with an unconfirmable hint returns a consent
page that cannot be submitted, and the End-User logs out by visiting the auth server directly.
Cheapest, no new code, and it amends #109's goal statement, which is yours to amend.

### Recommendation: A

It is the only option that leaves every sealed decision of this change intact. The confirming POST
stays hintless and stays CSRF-checked, so teardown without consent still requires a hint that
actually validates. The failed request's `client_id` is dropped, so a detected error still never
becomes a redirect. The "could not return you" note still appears. The requested locale still
survives in `ui_locales`.

It opens nothing new: an attacker's cross-site POST with a garbage hint gets a 303 to a consent page
they cannot confirm, which is exactly where a plain link to `GET /auth/logout` already lands today.

**Cost of getting it wrong.** A when you wanted C spends one branch in the pipeline and one
integration case. C when you wanted A ships a no-attacker path that ends with the user still signed
in, which is the defect this whole issue exists to remove. B when you wanted A spends a template and
two catalog keys in both languages.

**All three conform.** The spec says nothing about how an OP is to make its own consent page
submittable. What separates the options is #109's own commitment that no request shape ends on a
terminal page with the user still signed in.

### To answer

Reply with `A`, `B` or `C`, with or without reasoning. The run resumes at stage 4 from your answer.

Stages 1 to 3 are landed and pushed (`a4a586d`, `7a6c039`, `ad43d76`); nothing is blocked behind
this except stages 4 and 5.

leodip:
A

leodip:
## Decision 18 applied: a CSRF-skipped POST is answered with a 303 to the GET

Answered `A`. Recorded as decided, and stage 4 builds it.

### What that means in the code

A `POST /auth/logout` whose `id_token_hint` is present but cannot be confirmed no longer renders the
consent page directly. It answers `303 See Other` with a `Location` of `/auth/logout` carrying:

- `post_logout_redirect_uri`, when one was supplied
- `state`, when it was present, empty value included
- `ui_locales`, when one was supplied

and deliberately **not** `id_token_hint` and **not** `client_id`.

The browser re-requests by GET. The GET binding is never CSRF-exempt, so gorilla/csrf issues a token
and a cookie on that request, and the consent page it renders is submittable. Confirming it is a
hintless POST, which is CSRF-checked, so teardown without consent still requires a hint that actually
validates.

Dropping `client_id` is what preserves the rule that a detected error never becomes a redirect: the
follow-up GET is then the hint-absent shape, where `client_id` is the only thing that can authorize a
post-logout redirect, so with none supplied no client validates the target, no redirect happens, and
the user lands on the signed-out page with the "we could not return you to the application" note. That
is the intended outcome for a rejected hint.

The handler needs no knowledge of what the CSRF middleware did: the exemption keys on "POST with the
hint present", which is exactly the condition the handler already evaluates.

### Why this shape and not the others

B, an interstitial page with a GET link, reaches the same destination but needs a new template and new
strings in both languages, and costs the user a second click. C, accepting the dead end and documenting
it, ships a path reachable with no attacker at all (an RP holding an ID token for a session that has
since been replaced) that ends with the user still signed in, which is the defect this change exists to
remove.

### Cost

One branch in the pipeline, one unit case per parameter carried and per parameter dropped, and one
integration case. The endpoint answers 303 on that shape where it answered 200.

leodip:
## Deferred decisions, for you to settle before merge

Two questions came up during the run that were safe to proceed past on a stated assumption. Neither
blocks the change, both are cheap to reverse, and this is the list to read first.

---

### 1. When the post-logout redirect is built, is the registered URI's own query preserved byte for byte, or canonicalised?

Raised during stage 1, by the round-1 review and refined by the round-2 review.

**Assumed: preserved byte for byte, in order, empty fields included.** The design states the property,
"the 302's query preserves whatever the registered URI already carried", and the no-`state` path
already delivered it, so canonicalising only when the RP sends a `state` was an accident of the
mechanism rather than a choice. Round 2 found the same accident one level down: the field-by-field
copy skipped empty fields, so `?a=1&&b=2` became `?a=1&b=2` when a `state` was sent and stayed intact
when one was not. Empty fields now survive, which is what makes the two paths agree.

**The alternative** is to decode the query into `url.Values` and re-encode it, which is what
`redirToClientWithError` does today and what stage 1 first shipped. It is simpler to read and it
normalises the query into a canonical form. It also, executed side by side rather than reasoned
about:

- deletes any field separated by a literal semicolon, because `url.URL.Query` discards the error
  `url.ParseQuery` returns for one,
- reorders the fields, which breaks a signature computed over the raw query,
- adds an `=` to a valueless field,
- decodes registered percent-escapes.

Nothing rejects those URI shapes at registration: checked against `validateRedirectURI` and
`urlutil.IsAbsoluteRedirectURI`.

**Cost to reverse:** the body of `buildPostLogoutRedirect`, one unexported function, and thirteen rows
of `TestBuildPostLogoutRedirect`. The signature is the same either way and nothing downstream reads
anything but its return value.

---

### 2. Should the integration tier drive a real browser for the logout consent page, or is a form validator in the test helper enough?

Raised during stage 2, by the round-6 review, which named both mechanisms and said the repository
forces neither.

**Assumed: the validator is enough, for a claim narrowed to what it can actually establish.** That
claim is: the payload a browser would submit from this form is what the test submits, and nothing the
document *declares* sends that payload elsewhere or forbids it. Not that a browser can use the page.

The helper refuses to submit a page whose rules it has not been taught, along four dimensions: the
element types allowed inside the form, the attributes each may carry, the values of `class` against
the tokens the template actually uses, and the document around the form, which is where the
submission's destination and its permission live. Thirty-three mutations have been executed against
it and every one is killed.

**The residual, stated plainly:** three inputs to whether a browser can use this page are outside a
parsed document altogether, and no refusal reaches them.

- **Script from another file.** A handler cancelling the submit event would leave the page unusable
  with the tier green. Inline handlers and `<script>` inside the form are refused; one registered from
  `/static/utils.js` needs a JavaScript engine to see. Nothing registers one today.
- **Style from an external stylesheet.** The layout links `/static/main.css` and two CDN stylesheets,
  any of which can hide a control the helper just approved. Declared style is refused; linked style is
  not reachable.
- **The ancestor attribute denylist's open tail.** Every other set in the helper is an allowlist and
  therefore closed. That one is a denylist on purpose, because the auth layout is shared by every page
  in the app.

All three are properties of every one of the 61 templates in the two apps that carry a form, none of
which any test observes through a browser, so closing them here would close them in one place out of
61. If you answer "browser", the honest reason is that the validator took four review rounds to close,
not that the closed version is unsound.

**Cost to reverse:** a browser dependency in the devcontainer plus the body of one helper. Every case
reaches the consent page through `logoutThroughConsentPage`, so swapping the mechanism is one
function; the cases and their assertions do not move.

leodip:
## Follow-ups, ready to file

Seven real problems turned up during this work that belong to a different change. Four you already
filed during the run as #146, #147, #148 and #149. Three are still drafted below, each with the
`gh issue create` command that files it.

Duplicate searches re-run just now against the 30 open issues: nothing matches any of the three.

---

### 1. Security: a resource identifier and a client identifier may be the same string, and nothing checks

Found during the plan review.

Client identifiers are checked for uniqueness only against other clients (`HandleAPIClientsCreate`
calls `GetClientByClientIdentifier` before inserting) and resource identifiers only against other
resources (`handler_api_resources.go` calls `GetResourceByResourceIdentifier`). No code compares the
two namespaces, so a deployment can hold a client `myapp` and a resource `myapp` at once. That is not
a perverse configuration: naming a client and the API it protects after the same product is the
obvious thing to do.

It matters because an access token's `aud` is built from resource identifiers
(`generateAccessTokenCore`) while an ID token's `aud` is the client identifier (`generateIdTokenCore`),
so under a collision the two token types become indistinguishable to any code that resolves `aud`
through `GetClientByClientIdentifier`. This PR found one such place, the `id_token_hint` validation,
where a session-bound access token satisfied every check the logout endpoint made and could force a
logout. It is closed here by requiring ID-Token shape in the hint classifier: reject a `typ` of
`Bearer`, `Offline` or `Refresh`, and require `sub` and `iat`.

So the issue below is hardening rather than a second live hole. `aud` is resolved against client
identifiers in exactly one place repo-wide, and that place is fixed here. The reason to close it
anyway is that the next piece of code to resolve an audience will not know to distrust it, and a
per-call-site defence has to be remembered every time.

```bash
gh issue create --label bug --label security --label go \
  --title "Security: a resource identifier and a client identifier may be the same string, and nothing checks" \
  --body 'Client identifiers and resource identifiers live in separate tables with separate uniqueness checks, so the same string can name both. Under such a collision an access token'"'"'s `aud` resolves to a client through `GetClientByClientIdentifier`, which makes an access token indistinguishable from an ID token to anything validating an audience.

#109 hit this in the `id_token_hint` path and closed it there by requiring ID-Token shape (rejecting a `typ` of `Bearer`, `Offline` or `Refresh`, and requiring `sub` and `iat`), but the ambiguity itself remains.

Reject a client identifier that already names a resource and vice versa, on create and on update, in the admin API, the admin console and Dynamic Client Registration.

Decide separately what to do about deployments that already hold a collision: a startup warning is probably enough, since renaming either one is a breaking change for the integrations using it.'
```

---

### 2. Bug: the logout consent page's No button submits the form when JavaScript is unavailable

Found during stage 2.

`logout_consent.html` renders the declining control as
`<button class="..." onclick="NoBtnClick(this, event);">`, with no `type` attribute. HTML's
missing-value default for `<button type>` is `submit`, so the only thing keeping No from submitting is
`NoBtnClick` calling `event.preventDefault()`. With JavaScript unavailable, clicking No submits the
consent form, which is a confirming hintless POST carrying a valid CSRF token, and the handler tears
the session down. The End-User declined and was logged out.

Pre-existing rather than introduced here: before this change the hintless POST already cleared the
session cookie, so No already ended the OP session. This PR only makes the teardown durable by
deleting the `user_sessions` row as well. Low severity, since it fails in the safe direction and needs
JavaScript to be off, but it is the one End-User-facing page the spec makes mandatory.

One attribute is not the fix, which is why it is a separate change rather than a line here.
`NoBtnClick` also calls `history.back()`, so a `type="button"` No would be entirely inert without
JavaScript instead of wrong.

```bash
gh issue create --label bug --label enhancement \
  --title "Bug: the logout consent page's No button submits the form when JavaScript is unavailable" \
  --body 'The logout consent page'"'"'s No button has no `type` attribute, so it is a submit control by HTML'"'"'s default, and only its inline `onclick` handler'"'"'s `preventDefault()` stops it submitting. With JavaScript unavailable, clicking No submits the consent form, which the server reads as the End-User confirming, and the session is torn down. Declining should never log the user out.

A `type="button"` alone is not enough, because the same handler'"'"'s `history.back()` also needs JavaScript, so the control would become inert rather than correct: give the declining control a script-free destination and decide what it does with no history to go back to.

The same `<button onclick="...">`-inside-a-form shape appears throughout the admin console (roughly 20 buttons), which is JavaScript-dependent by design, so this is about the auth server'"'"'s End-User pages.

Note for whoever fixes it: `logoutThroughConsentPage` in `src/authserver/tests/integration/api_account_logout_request_test.go` reads the two controls off the rendered page and refuses anything it has not been taught. Giving No an explicit `type="button"` needs no change there. Turning No into a link does: `a` is not in the helper'"'"'s element allowlist for the form'"'"'s subtree and the helper expects exactly two buttons. The failure says so, and teaching it the new shape is the intended way to change the page.'
```

---

### 3. Bug: the CSS build harvests class names from the whole Go module, not from the template globs its config declares

Found during stage 2.

`src/authserver/build.sh` runs `tailwindcss -c ./web/tailwindcss/tailwind.config.js`, whose `content`
key declares the three template globs and nothing else. That does not hold: `input.css` starts with
`@import "tailwindcss"`, which is Tailwind v4, and v4 performs automatic source detection over the
project tree in addition to whatever a v3-style config lists. So the shipped stylesheet is a function
of every file in the module rather than of the templates.

Executed rather than reasoned. Writing the string for Tailwind's screen-reader-only utility into a
`_test.go` file under `src/authserver/tests/integration/`, with every template byte-identical, added
an eleven-line rule for it to `web/static/main.css`. Restoring the generated file and rebuilding
brought the rule straight back, and it disappeared only when the string was removed from the Go
source. No template in the module contains that class.

The consequence that matters: a class deleted from a template keeps shipping for as long as any Go
file, comment included, happens to contain the word, so the generated file stops tracking the
templates. It also means an unrelated Go change surfaces as a diff in a generated CSS artifact, which
is how this was found. `src/adminconsole` has the identical build line, config and import.

```bash
gh issue create --label bug --label go \
  --title "Bug: the CSS build harvests class names from the whole Go module, not from the template globs its config declares" \
  --body 'Both apps build their stylesheet with `tailwindcss -c ./web/tailwindcss/tailwind.config.js`, whose `content` key lists only `./web/template/*.{html,js}` and the partials and layouts beside it. Because `input.css` uses `@import "tailwindcss"` (Tailwind v4), automatic source detection also runs and scans the whole module, so the generated `web/static/main.css` reflects every file in the tree rather than the templates.

Demonstrated by adding a Tailwind class name to a `_test.go` file with the templates unchanged: the corresponding rule appeared in `main.css`, and removing the string removed it again.

The practical cost is that a class removed from a template keeps shipping while any Go source mentions the word, and unrelated Go changes surface as diffs in a generated artifact.

Pin the sources so the stylesheet is a function of the templates, for example `@import "tailwindcss" source(none);` plus explicit `@source` lines in each app'"'"'s `input.css`, then diff the regenerated `main.css` in both apps to confirm nothing rendering today loses a rule.'
```

---

### Not a new issue: a second instance for #136

The full-suite run at stage 5's gate failed once on MySQL in `TestGetUserByUsername`, then passed on a
re-run of the whole MySQL data tier and in isolation. #109 touches no query, no model and no
migration, so it cannot reach the data layer. #136 already documents the mechanism in full, so this is
a comment rather than an issue:

> A second instance of this, seen on MySQL during a full `./run-tests.sh`: `TestGetUserByUsername`
> failed with every field mismatching and `ForgotPasswordCodeIssuedAt` reading `2026-08-05`, two days
> before the run. It passed on a re-run of the whole MySQL data tier and in isolation.
>
> Worth recording because it is a stronger form of the same collision. `TestSearchUsersPaginated`
> needs only a shared substring; this one needs the whole username to repeat, which the duplicates
> already in the table (`Parker8278`, `Weber9487`) show does happen. It also narrows the remedy:
> relaxing the search assertion would not help here, because `GetUserByUsername` returns one row and
> the test compares every field of it. Unique usernames per run, or resetting the shared databases, is
> what closes both.

```bash
gh issue comment 136 --body 'A second instance of this, seen on MySQL during a full `./run-tests.sh`: `TestGetUserByUsername` failed with every field mismatching and `ForgotPasswordCodeIssuedAt` reading `2026-08-05`, two days before the run. It passed on a re-run of the whole MySQL data tier and in isolation.

Worth recording because it is a stronger form of the same collision. `TestSearchUsersPaginated` needs only a shared substring; this one needs the whole username to repeat, which the duplicates already in the table (`Parker8278`, `Weber9487`) show does happen. It also narrows the remedy: relaxing the search assertion would not help here, because `GetUserByUsername` returns one row and the test compares every field of it. Unique usernames per run, or resetting the shared databases, is what closes both.'
```

Issue #145, opened 2026-08-07
Security: redact passwords, client secrets and OTP material from API debug logging


leodip:
When `GOIABADA_AUTHSERVER_DEBUG_API_REQUESTS` is enabled, `debugLog` in `src/authserver/internal/middleware/api_debug_middleware.go` pretty-prints the entire JSON request and response body to the log. The function already redacts the `Authorization` header to `Bearer ***`, so the intent to keep credentials out of logs is there, but the bodies are written verbatim. `PUT /api/v1/account/otp` alone carries `password`, `otpCode` and `secretKey`, which is the account password, a live TOTP code and the authenticator seed; other account and admin endpoints carry passwords and client secrets on the same path.

Redact by key before formatting, so the shape of the body is still readable for debugging while the values are not. Worth covering with tests over at least `password`, `currentPassword`, `newPassword`, `clientSecret`, `otpCode` and `secretKey`, and worth applying to the response body too, since enrollment responses return `secretKey`.

This is a debug-only path and off by default. The cost of leaving it is that anyone who turns the flag on to diagnose an unrelated problem writes reusable credentials and authenticator seeds into their console and log aggregator.

Issue #121, opened 2026-07-26
Custom-scheme redirect URIs are broken with response_mode=form_post


leodip:
## Summary

A client whose redirect URI uses a custom scheme (`myapp://callback`, `com.example.app:/oauth`) and which requests `response_mode=form_post` receives a form that posts back to the authorization server instead of to the app. The authorization code reaches nobody and the flow dead-ends with no error.

The cause is that `form_post.html` renders the redirect URI into a `<form action>` attribute, and Go's `html/template` URL filter rejects every scheme outside http, https and mailto, replacing the value with `#ZgotmplZ`. The filter is doing exactly what it is designed to do; the template is relying on it in a context where legitimate non-http schemes also arrive.

Severity: **Low**. It fails closed: nothing leaks, no code is delivered to a wrong party, the flow simply does not complete. Bounded further by `form_post` being uncommon for native apps, which normally use `query`.

## Affected code

`src/authserver/web/template/form_post.html:13`:

```html
<form method="post" action="{{.redirectURI}}">
```

Reached from `issueAuthCode` (`src/authserver/internal/handlers/handler_auth_issue.go:251`), which selects the template when `responseMode == "form_post"`.

Nothing cross-validates the response mode against the redirect URI scheme. `authorize_validator.go:321` checks only that the value is one of `query`, `fragment` or `form_post`:

```go
if !slices.Contains([]string{"query", "fragment", "form_post"}, input.ResponseMode) {
```

So any client can request `form_post` regardless of what its registered redirect URI looks like.

## Verification

Rendering the template's action attribute through `html/template` with a range of schemes:

| Redirect URI | Rendered `action` |
|---|---|
| `http://127.0.0.1:9000/cb` | `http://127.0.0.1:9000/cb` |
| `myapp://callback` | `#ZgotmplZ` |
| `com.example.app:/oauth` | `#ZgotmplZ` |
| `javascript:alert(1)` | `#ZgotmplZ` |
| `vbscript:msgbox(1)` | `#ZgotmplZ` |

The `javascript:` row is the filter working as intended and is why the template is not an XSS. The `myapp://` row is the bug: the same rule that neutralises the dangerous scheme also neutralises the legitimate one, because the filter allowlists schemes rather than denylisting them.

Query mode is unaffected. `issueAuthCode` builds that redirect with `url.ParseRequestURI` plus `http.Redirect` (`handler_auth_issue.go:271-279`), which does not go through template escaping, so a custom scheme survives it.

[...]

Issue #114, opened 2026-07-25
LimitOtp returns a blank 200 when the auth context is unreadable


leodip:
## Summary

When `LimitOtp` cannot read the auth context it logs and returns without calling `next` and without writing a response. Go then emits a bare `200 OK` with an empty body, so the user gets a blank page instead of an error or a redirect.

This is a **correctness bug, not a vulnerability**. The behaviour is fail-closed in the sense that the OTP handler never runs, so nothing is bypassed. Filing it because it sits in security middleware and the failure mode is confusing to diagnose.

## Affected code

`src/core/middleware/middleware_ratelimiter.go:83-87`:

```go
authContext, err := m.authHelper.GetAuthContext(r)
if err != nil {
    slog.Error("Rate limiter - unable to get auth context", "error", err)
    return          // <-- no next.ServeHTTP, no w.Write, no status code
}
```

Contrast the sibling limiters, which either call `next` or let `RespondOnLimit` write a `429`. This is the only branch in the file that terminates a request without producing a response.

## When it triggers

`GetAuthContext` fails when the session cookie is missing, expired or undecryptable. Reaching `POST /auth/otp` in that state is normal enough:

- The user leaves the OTP form open long enough for the cookie to expire, then submits.
- The cookie is cleared in another tab (a logout elsewhere) before the form is posted.
- A stale bookmark or a replayed form post.

Only applies when the rate limiter is enabled, since the `!m.enabled` guard at `:78-81` returns early otherwise.

## Impact

The user sees a completely blank page mid-authentication with no error, no redirect and no way forward other than starting over. From an operator's perspective the request logs as a `200`, so it does not surface in error-rate monitoring despite being a failed request. There is no security impact: the handler is not reached, so no OTP is validated and no session is issued.

## Recommendation

Make the branch behave like the other error paths in the auth flow. Two reasonable options:

**Option A, treat it as a lost auth context** and surface an error, in the spirit of how the handlers themselves handle `ErrNoAuthContext` (see `handler_auth_issue.go:33-36` and `handler_auth_completed.go:32-35`):

```go
authContext, err := m.authHelper.GetAuthContext(r)
if err != nil {
    slog.Error("Rate limiter - unable to get auth context", "error", err)
    http.Error(w, "Your session has expired. Please start again.", http.StatusBadRequest)
    return
}
```

[...]

PR #154, opened 2026-08-08
fix(authserver): clear the auth context before the client response is committed

### LINKED ISSUE #141: ClearAuthContext is called after the client response is committed in seven handlers, so the clear reaches nothing
`AuthHelper.ClearAuthContext` deletes the auth context from the cookie session and persists that deletion through `sessionStore.Save(r, w, sess)`, which writes a `Set-Cookie` header. Seven handlers call it immediately *after* `redirToClientWithError`, which has already committed the response: `http.Redirect` for the query and fragment response modes, and `t.Execute(w, m)` for `form_post`. Go snapshots the header map when the status line is written, so the `Set-Cookie` is dropped and the browser keeps the auth context.

The store is `sessionstore.ChunkedCookieStore`, so the auth context lives in the cookie itself and a `Set-Cookie` is the only way the deletion reaches the browser. There is no server-side row to fall back on.

Found while building #129 and recorded there as follow-up 4. Citations are by symbol rather than line number.

## The sites

Seven, each read rather than inferred, all of the shape `redirToClientWithError(...)` then `authHelper.ClearAuthContext(w, r)`:

- `handler_authorize.go`, `HandleAuthorizeGet`'s local `redirToClientWithError` closure, which every request-validation failure goes through.
- `handler_authorize.go`, `handlePromptNone`'s `redirectWithError` closure, which every silent refusal goes through, `login_required` included.
- `handler_auth_completed.go`, the disabled-user refusal, and the no-authorized-scopes refusal.
- `handler_consent.go`, both consent-denied refusals.
- `handler_auth_issue.go`, the `id_token_hint` subject mismatch in `HandleIssueGet`.

The success paths in the same files do it in the right order, clearing before writing the response: `HandleIssueGet` clears before `issueAuthCode`, and `handleImplicitFlow` clears before `issueImplicitTokens`. So this reads as a copied idiom rather than an intended difference.

## What is and is not claimed

No bypass is claimed for any single site: replaying a retained ceremony re-runs the same checks, including at the sharpest site, where the `id_token_hint` comparison fires again. The defect is that a cleanup written seven times never happens, and that one of the seven leaves the browser holding a context in `ready_to_issue_code`, the state that mints codes, with nothing but the re-run of the same checks between it and a replay.

**The behaviour is measured, not inferred.** #129's own `prompt=none` refusal in `HandleIssueGet` was written the wrong way round by copying the `id_token_hint` branch beside it, and was fixed inside that change. With the calls reversed, a replayed `/auth/issue` on the same cookie jar answered the client `login_required` a second time from a context still reading `ready_to_issue_code`; reversing them back sent it to the account profile instead. That was mutation 9 of #129's stage review: the reversal was executed and both outcomes observed, not argued.

## Two of the seven also miss a return

Same lines, same fix pass, so recording it here rather than as a separate issue. In `handler_auth_completed.go`'s no-authorized-scopes refusal and in `handler_consent.go`'s first consent-denied refusal, the `if err != nil` branch after `redirToClientWithError` calls `httpHelper.InternalServerError(w, r, err)` and then falls through instead of returning, so a failed redirect writes a 500 and the handler keeps going. The other five sites return.

## The fix

Clear before writing the response at each site, returning a 500 if the clear fails rather than committing a client response that cannot be recorded, which is the order #129 established at its own site.

A mock expectation cannot catch a regression here, since it proves only that the method was called. The assertion has to read the committed response: `httptest.ResponseRecorder.Result()` at the unit tier, whose `WriteHeader` snapshots headers the same way `net/http` does, or a second request on the same cookie jar at the integration tier.

leodip:
Fixes #141.

## What this fixes

`ClearAuthContext` ran **after** the client's response had already been committed at seven call sites
in the authserver handlers. The session store is `sessionstore.ChunkedCookieStore`, cookie only, so
the deletion is persisted through a `Set-Cookie` and nothing else. A header written after the status
line never reaches the wire, so the clear was dropped and the browser kept a usable auth context for a
ceremony the server had just refused.

Seven sites, but **22 refusal paths**: two of the seven are closures with 5 and 12 callers each.

| Site | Retained state | Paths | What a replay could do |
|---|---|---|---|
| `HandleAuthorizeGet`'s refusal closure | `initial` | 5 | inert, no handler accepts it |
| `handlePromptNone`'s redirect closure | `initial` | 12 | inert |
| `/auth/completed`, disabled user | `authentication_completed` | 1 | re-refuses, but re-bumps the session and re-emits the audit event |
| `/auth/completed`, no authorized scopes | `authentication_completed` | 1 | re-refuses |
| `/auth/consent`, nothing consented | `requires_consent` | 1 | renders consent again; the user may approve, so the client gets `access_denied` **and then a code** |
| `/auth/consent`, cancelled | `requires_consent` | 1 | same |
| `/auth/issue`, `id_token_hint` mismatch | `ready_to_issue_code` | 1 | the state that mints codes; only the sub comparison stands between a replay and a code |

No bypass is claimed, and the issue does not claim one either. Every replay re-runs the check that
refused in the first place, and those checks are deterministic on values fixed in the cookie. What
this closes is the retained state itself: the sharpest path is the `requires_consent` pair, where a
relying party could receive two responses for one authorization request.

Fixed, a replay lands on the account profile instead, which is what the shared no-context branch
already does.

## How it works

One statement moves at each site, and the shape is identical at all of them:

```go
err := authHelper.ClearAuthContext(w, r)   // was below the redirect
if err != nil { /* answer the client with server_error, see below */ }
err = redirToClientWithError(w, r, templateFS, ...)
```

This is safe because **`redirToClientWithError` reads nothing from the session**. It takes the error
code, description, response mode, redirect URI, state and response type as explicit parameters, and
every caller already holds those in memory. Clearing the cookie first cannot change what the client is
told, in any response mode.

Three other things changed with it:

- **A failed clear now answers the client** with `error=server_error` rather than a bare HTTP 500 that
  the client never sees, at all seven sites plus the failure branch of the one site whose ordering was
  already correct. OIDC Core 1.0 3.1.2.2 with 3.1.2.6 requires an error response for a redirect URI
  validated upstream, which all eight are, and RFC 6749 4.1.2.1 mints `server_error` for exactly this
  condition. This was escalated and decided by the repository owner; the reasoning and what reversing
  it costs are in the first closing comment below.
- **Four missing `return`s** were added. After the reorder the trailing branch is last in its block at
  all four, so they are defensive rather than load-bearing, but they leave all seven sites in one
  shape.
- **`redirToClientWithError` buffers its `form_post` render.** It wrote straight to the
  `ResponseWriter`, so a template that parsed and then failed mid-render committed a partial 200 and
  the caller's `InternalServerError` could no longer change the status: the last-resort 500 was not a
  last resort. `form_post.html` is operator supplied whenever `GOIABADA_AUTHSERVER_TEMPLATEDIR` is
  set, so a custom template makes this reachable. Found by the reviewer at stage 5.

The `id_token_hint` site also gets its OIDC citation corrected. The "MUST NOT reply with an ID Token
or Access Token for a different user" sentence is in Core 1.0 **3.1.2.2**, not 3.1.2.1.

**No interface, schema, migration, wire format or persisted shape changes**, and no user-facing
documentation moves, because no observable protocol behaviour does except the failure branch above.

| Stage | Commit | What |
|---|---|---|
| 1 | `ac4ffac` | Pin the premise: three subtests in `TestClearAuthContext` driving a real `ChunkedCookieStore` through a real `net/http/cookiejar`. The existing test was entirely mock-backed and never observed a `Set-Cookie`, which is how the defect survived. |
| 2 | `b52850f` | `handler_authorize.go`, both closures, 17 of the 22 paths |
| 3 | `81924bf` | `handler_auth_completed.go`, the disabled-user and no-scopes refusals |
| 4 | `e80da8e` | `handler_consent.go`, both deny branches |
| 5 | `da1d6c8` | `handler_auth_issue.go`, the `id_token_hint` mismatch and the `prompt=none` failure branch, plus the buffered `form_post` repair |

## Tests

`./run-tests.sh --type all` on the final tree: **9347 passing, 0 fail, 4 skip**, 13m04s.

| Tier | Passing | Engines |
|---|---|---|
| Unit | 2923 | three modules |
| Data | 1564, 4 skip | sqlite, mysql, postgres, mssql |
| Integration | 4860 | sqlite, mysql, postgres, mssql |

The 4 skips are pre-existing engine-capability ones: sqlite holds a single connection, so two
contention tests cannot create overlap on it.

What the new coverage actually observes, at three seams:

1. **The committed response.** Each site gets a case that stubs `ClearAuthContext` to write a sentinel
   `Set-Cookie`, then asserts on `rr.Result().Header` rather than `rr.Header()`. That distinction is
   the whole point: handler and stub share one live header map, so `rr.Header()` shows the sentinel
   under either ordering, while `rr.Result()` reads the snapshot taken when the status line was
   written. Plus one case per site asserting the client still gets `server_error` when the clear fails.
2. **The real store.** `TestClearAuthContext` now proves a clear written before the commit reaches a
   browser, one written after does not, and that when a response carries two session writes the later
   wins. This is the only permanent test where a real `Set-Cookie` is written and read back.
3. **The ceremony end to end.** Four integration replays on the same cookie jar, asserting a refused
   ceremony now lands on the profile redirect instead of re-refusing or re-rendering consent.

Every reordering was mutation-tested: reverting any one site fails its own cases and leaves the others
green, and reverting the `id_token_hint` site also fails its integration replay with the browser
reaching the retained `ready_to_issue_code` context.

## Read first

Two comments below:

- **Deferred decisions** — there are none, and the one escalated decision is restated with what
  reversing it costs. It also flags that the escalation is still sitting unanswered-looking on issue
  #141 itself, since it was raised before this pull request existed.
- **Follow-ups** — two, both verified and ready to file, neither of them work this change owed.

### DIFF
--- a/src/authserver/internal/handlers/handler_auth_completed.go
+++ b/src/authserver/internal/handlers/handler_auth_completed.go
@@ -197,14 +197,29 @@ func HandleAuthCompletedGet(
 			auditLogger.Log(constants.AuditUserDisabled, map[string]interface{}{
 				"userId": user.Id,
 			})
-			err := redirToClientWithError(w, r, templateFS, "access_denied", "The user account is disabled.",
-				authContext.ResponseMode, authContext.RedirectURI, authContext.State, authContext.ResponseType)
+			// The clear goes FIRST. ClearAuthContext persists the deletion through a Set-Cookie
+			// on w, and redirToClientWithError commits the response in every response mode, so
+			// clearing afterwards leaves the header on a response already written and the
+			// browser keeps an auth context it can replay (#141).
+			err := authHelper.ClearAuthContext(w, r)
 			if err != nil {
-				httpHelper.InternalServerError(w, r, err)
+				// The clear failed, so Save wrote no cookie and the browser still holds the
+				// auth context. The client is owed an error response regardless: its redirect
+				// URI was validated upstream, so OIDC Core 1.0 3.1.2.2 with 3.1.2.6 applies,
+				// and RFC 6749 4.1.2.1 mints server_error for exactly this condition (#141).
+				slog.Error("failed to clear the auth context, answering the client with server_error",
+					"error", err)
+				err = redirToClientWithError(w, r, templateFS, "server_error", "Internal server error",
+					authContext.ResponseMode, authContext.RedirectURI, authContext.State, authContext.ResponseType)
+				if err != nil {
+					// Nowhere left to send the client, so the 500 is the last resort here.
+					httpHelper.InternalServerError(w, r, err)
+				}
 				return
 			}
 
-			err = authHelper.ClearAuthContext(w, r)
+			err = redirToClientWithError(w, r, templateFS, "access_denied", "The user account is disabled.",
+				authContext.ResponseMode, authContext.RedirectURI, authContext.State, authContext.ResponseType)
 			if err != nil {
 				httpHelper.InternalServerError(w, r, err)
 				return
@@ -225,13 +240,28 @@ func HandleAuthCompletedGet(
 		// in the auth context replace the scope with the effective scope
 		authContext.SetScope(effectiveScope)
 		if len(authContext.Scope) == 0 {
-			err = redirToClientWithError(w, r, templateFS, "access_denied", "The user is not authorized to access any of the requested scopes", authContext.ResponseMode,
-				authContext.RedirectURI, authContext.State, authContext.ResponseType)
+			// The clear goes FIRST, for the same reason as the disabled-user refusal above: a
+			// Set-Cookie written after redirToClientWithError has committed never reaches the
+			// wire, so the browser keeps a replayable auth context (#141).
+			err = authHelper.ClearAuthContext(w, r)
 			if err != nil {
-				httpHelper.InternalServerError(w, r, err)
+				// The clear failed, so Save wrote no cookie and the browser still holds the
+				// auth context. The client is owed an error response regardless: its redirect
+				// URI was validated upstream, so OIDC Core 1.0 3.1.2.2 with 3.1.2.6 applies,
+				// and RFC 6749 4.1.2.1 mints server_error for exactly this condition (#141).
+				slog.Error("failed to clear the auth context, answering the client with server_error",
+					"error", err)
+				err = redirToClientWithError(w, r, templateFS, "server_error", "Internal server error",
+					authContext.ResponseMode, authContext.RedirectURI, authContext.State, authContext.ResponseType)
+				if err != nil {
+					// Nowhere left to send the client, so the 500 is the last resort here.
+					httpHelper.InternalServerError(w, r, err)
+				}
+				return
 			}
 
-			err = authHelper.ClearAuthContext(w, r)
+			err = redirToClientWithError(w, r, templateFS, "access_denied", "The user is not authorized to access any of the requested scopes", authContext.ResponseMode,
+				authContext.RedirectURI, authContext.State, authContext.ResponseType)
 			if err != nil {
 				httpHelper.InternalServerError(w, r, err)
 				return
--- a/src/authserver/internal/handlers/handler_auth_completed_test.go
+++ b/src/authserver/internal/handlers/handler_auth_completed_test.go
@@ -550,7 +550,14 @@ func TestHandleAuthCompletedGet(t *testing.T) {
 
 		auditLogger.On("Log", constants.AuditUserDisabled, mock.Anything).Return()
 
-		authHelper.On("ClearAuthContext", rr, req).Return(nil)
+		// The clear has to reach the browser, so it must happen before the response is
+		// committed. rr.Header() is the live map the handler and this stub share, so it shows
+		// the sentinel under either ordering; rr.Result() reads the snapshot taken when the
+		// status line was written, which is the only unit-tier view that tells them apart.
+		const clearedContextCookie = "cleared-auth-context"
+		authHelper.On("ClearAuthContext", rr, req).Run(func(args mock.Arguments) {
+			args.Get(0).(http.ResponseWriter).Header().Set("Set-Cookie", clearedContextCookie)
+		}).Return(nil)
 
 		handler.ServeHTTP(rr, req)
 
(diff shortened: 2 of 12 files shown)


leodip:
## Deferred decisions: none

Nothing in this change rests on an assumption the run took on your behalf. Every question it met was
either forced by the code and demonstrated in the record, or blocked and asked. This is the list you
would normally read first, and it is empty on purpose.

One decision was escalated and you answered it, and it is worth restating here because it is the only
thing in this change that alters what a relying party can observe.

### The one escalation: what a client gets when clearing the auth context fails

You chose **option C**: the client is sent `error=server_error` through the normal redirect, at all
seven refusal sites plus the eighth site's failure branch. Before this change that branch answered
HTTP 500 and sent the client nothing at all.

The reason it was worth asking, and the reason C rather than the smaller B:

- OIDC Core 1.0 section 3.1.2.2 says "If the Authorization Server encounters any error, it MUST return
  an error response, per Section 3.1.2.6", and 3.1.2.6 returns the client to the redirection URI
  unless that URI is invalid. All eight sites sit after `ValidateClientAndRedirectURI`, so the
  exception does not apply and the old bare 500 was non-conforming.
- RFC 6749 section 4.1.2.1 defines `server_error` for exactly this: "(This error code is needed
  because a 500 Internal Server Error HTTP status code cannot be returned to the client via an HTTP
  redirect.)"
- The rejected option B, log the fault and send the original refusal, conforms too and is the smaller
  diff. It was rejected because it tells the relying party the user denied consent when the server
  also had an internal fault the client will never learn about. B remains the right fallback if
  `server_error` ever proves disruptive to a real relying party.

**Reversing it costs one line at each of eight sites**, changing the error code and description
passed to `redirToClientWithError`, plus the eight unit cases that assert on it. No interface, schema
or persisted shape depends on it.

### One thing to know about where that question was asked

The escalation was posted on **issue #141 itself**, as comment
[5223898983](https://github.com/leodip/goiabada/issues/141#issuecomment-5223898983), because it was
raised by the plan review before any code had landed and this pull request did not exist yet. You
answered it directly rather than in the thread, so that comment is still sitting on the public issue
looking unanswered. Worth a short reply or a resolution note when you close #141, so the next reader
does not think a security question went unaddressed.

### The one residual this change knowingly does not close

A failed `ClearAuthContext` leaves the browser holding the auth context, and no option on the table
changed that. Every error return in `ChunkedCookieStore.Save` sits above its first `http.SetCookie`,
so a failed clear writes zero cookies whatever the handler does next. The choice above only governs
what the client is told; the retained context on that branch is equally live under all three options,
and closing it would need a different mechanism than ordering.

leodip:
## Follow-ups: two, both ready to file as they stand

Neither is work this change owed. Both were found during stage 5's review, verified against the code
rather than taken on the reviewer's word, and left alone because acting on them would have widened a
sealed scope. Duplicate searches were re-run at the closing pass on 2026-08-09 and are recorded with
each.

---

### 1. Enabled HTTP request logging writes complete `id_token_hint` JWTs to the log

**Not a duplicate.** `gh issue list --state open --search "request log redact token"` returns nothing.
#145 is the nearest neighbour and is a different path: request and response *bodies* in
`api_debug_middleware.go` under `GOIABADA_AUTHSERVER_DEBUG_API_REQUESTS`. This is the query string,
under a different flag, in a different middleware. Worth cross-referencing when filing, not merging.

**Body:**

> When `GOIABADA_AUTHSERVER_LOG_HTTP_REQUESTS` is enabled,
> `src/authserver/internal/server/server.go` skips `/health`, `/static/` and `/favicon.ico` and hands
> everything else to Chi's `middleware.Logger`. Chi's `DefaultLogFormatter` formats `r.RequestURI`
> (chi v5 `middleware/logger.go`), which is the path *and the raw query string*. `GET /auth/authorize`
> carries `id_token_hint` in that query string, so every such request writes the complete signed ID
> token into the log.
>
> This is not a test-only configuration. `src/cmd/goiabada-setup/main.go` writes
> `GOIABADA_AUTHSERVER_LOG_HTTP_REQUESTS=true` into both the compose file and the environment script
> it generates, and every shipped compose file except `docker-compose-test.yml` sets it to `true`. A
> default install logs these.
>
> An operator or a log collector then holds a signed ID token and all of its claims: subject, session
> identifier, email, issuer, audience and authentication metadata. The token also stays usable as an
> `id_token_hint` for authorization and as a logout hint for as long as the server accepts it, so this
> is a reusable credential sitting in a log file rather than only a privacy leak.
>
> Replace the standard formatter with one that logs the path and redacts security-bearing query
> parameters, `id_token_hint` at minimum. Worth a test that feeds a recognisable JWT-shaped value
> through the logger and asserts it is absent from the captured output.
>
> Unrelated to the auth-context cleanup ordering of #141; found while running that issue's
> integration tier.

```bash
gh issue create --title "Security: enabled HTTP request logging writes complete id_token_hint JWTs to the log" --label security --label go --body "$(cat <<'EOF'

EOF
)"
```

---

### 2. `issueAuthCode` commits a partial 200 when the `form_post` template fails part way through

This is the success-path twin of the defect #141 fixed in `redirToClientWithError`. It was found while
closing that family, and it is out of scope because the agreement listed `issueAuthCode` as read-only
and the defect predates this change without being made worse by it.

**Not a duplicate.** `gh issue list --state open --search "form_post template"` returns only #121
(custom-scheme redirect URIs with `response_mode=form_post`). #146 is in the same helper family but is
about duplicate `state` parameters. #114 (`LimitOtp` returns a blank 200 when the auth context is
unreadable) is the nearest in symptom and is a different handler and mechanism. None covers this.

**Body:**

> `issueAuthCode` in `src/authserver/internal/handlers/handler_auth_issue.go` renders `form_post.html`
> with `t.Execute(w, m)`, writing straight to the `ResponseWriter`. `Execute` writes as it walks the
> template, so a template that parses cleanly and then fails part way through has already emitted its
> prefix and triggered an implicit `WriteHeader(200)` by the time the error is returned. Its only
> caller answers that error with `httpHelper.InternalServerError(w, r, err)`, whose `WriteHeader(500)`
> is then a no-op.
>
> The client is left with HTTP 200 and a truncated page for an authorization code that was minted and
> persisted but never delivered, and nothing in the response says the request failed.
>
> `form_post.html` is operator supplied whenever `GOIABADA_AUTHSERVER_TEMPLATEDIR` is set, since
> `server.New` serves templates from `os.DirFS` in that case, so this is reachable in a real
> deployment. The template shipped in `web/template/form_post.html` cannot trigger it: it only
> dereferences map keys, which yield a zero value rather than an error.
>
> The same defect was fixed in the sibling helper `redirToClientWithError` in #141, by rendering into
> a `bytes.Buffer` and copying to the `ResponseWriter` only after `Execute` returns successfully. That
> repair transfers directly, and a regression test can follow
> `TestRedirToClientWithError_FormPostExecutionFailureLeavesResponseUncommitted` in
> `handler_authorize_test.go`: a parse-valid, execution-failing template, then assert the body is
> empty and a later `WriteHeader(500)` still lands.

```bash
gh issue create --title "issueAuthCode commits a partial 200 when the form_post template fails part way through rendering" --label bug --label go --body "$(cat <<'EOF'
`issueAuthCode` in `src/authserver/internal/handlers/handler_auth_issue.go` renders `form_post.html` with `t.Execute(w, m)`, writing straight to the `ResponseWriter`. `Execute` writes as it walks the template, so a template that parses cleanly and then fails part way through has already emitted its prefix and triggered an implicit `WriteHeader(200)` by the time the error is returned. Its only caller answers that error with `httpHelper.InternalServerError(w, r, err)`, whose `WriteHeader(500)` is then a no-op.

The client is left with HTTP 200 and a truncated page for an authorization code that was minted and persisted but never delivered, and nothing in the response says the request failed.

`form_post.html` is operator supplied whenever `GOIABADA_AUTHSERVER_TEMPLATEDIR` is set, since `server.New` serves templates from `os.DirFS` in that case, so this is reachable in a real deployment. The template shipped in `web/template/form_post.html` cannot trigger it: it only dereferences map keys, which yield a zero value rather than an error.

The same defect was fixed in the sibling helper `redirToClientWithError` in #141, by rendering into a `bytes.Buffer` and copying to the `ResponseWriter` only after `Execute` returns successfully. That repair transfers directly, and a regression test can follow `TestRedirToClientWithError_FormPostExecutionFailureLeavesResponseUncommitted` in `handler_authorize_test.go`: a parse-valid, execution-failing template, then assert the body is empty and a later `WriteHeader(500)` still lands.
EOF
)"
```

---

Filing is yours, not the run's. Both commands are ready to paste.

Issue #159, opened 2026-08-09
Security: enabled HTTP request logging writes complete id_token_hint JWTs to the log


leodip:
When `GOIABADA_AUTHSERVER_LOG_HTTP_REQUESTS` is enabled, `src/authserver/internal/server/server.go` skips `/health`, `/static/` and `/favicon.ico` and hands everything else to Chi's `middleware.Logger`. Chi's `DefaultLogFormatter` formats `r.RequestURI` (chi v5 `middleware/logger.go`), which is the path *and the raw query string*. `GET /auth/authorize` carries `id_token_hint` in that query string, so every such request writes the complete signed ID token into the log.

This is not a test-only configuration. `src/cmd/goiabada-setup/main.go` writes `GOIABADA_AUTHSERVER_LOG_HTTP_REQUESTS=true` into both the compose file and the environment script it generates, and every shipped compose file except `docker-compose-test.yml` sets it to `true`. A default install logs these.

An operator or a log collector then holds a signed ID token and all of its claims: subject, session identifier, email, issuer, audience and authentication metadata. The token also stays usable as an `id_token_hint` for authorization and as a logout hint for as long as the server accepts it, so this is a reusable credential sitting in a log file rather than only a privacy leak.

Replace the standard formatter with one that logs the path and redacts security-bearing query parameters, `id_token_hint` at minimum. Worth a test that feeds a recognisable JWT-shaped value through the logger and asserts it is absent from the captured output.

Related to #145, which covers request and response bodies in `api_debug_middleware.go` under a different flag. Unrelated to the auth-context cleanup ordering of #141; found while running that issue's integration tier.

Issue #82, opened 2026-07-04
TOTP secrets stored in plaintext at rest


leodip:
## Summary

TOTP secrets (`user.OTPSecret`) are persisted in plaintext, unlike client secrets and email/phone verification codes, which are AES-GCM encrypted with the same available key. A read-only database compromise exposes every user's TOTP seed and defeats 2FA for all users.

Severity: **Medium**

## Details

`src/core/models/user.go:45`:

```go
OTPSecret string `db:"otp_secret"`
```

The secret is written in the clear and read back directly for validation:

- `src/authserver/internal/handlers/apihandlers/handler_api_account_otp.go:168` — `user.OTPSecret = normalizedSecret` then `UpdateUser`
- `src/authserver/internal/handlers/handler_auth_otp.go:279` — `user.OTPSecret = secretKey`
- `src/authserver/internal/handlers/handler_auth_otp.go:259` — `totp.Validate(otpCode, user.OTPSecret)`

This is inconsistent with the rest of the model: `ClientSecretEncrypted` and `EmailVerificationCodeEncrypted` are AES-GCM encrypted via `encryption.EncryptText`/`DecryptText` using `settings.AESEncryptionKey` — the same key is available in these code paths.

## Impact

A read-only DB compromise (backup/snapshot leak, SQL injection elsewhere, stolen replica) yields every user's TOTP seed. An attacker can then generate valid second factors and defeat 2FA for all users — precisely the failure mode 2FA is meant to survive when the primary credential store is compromised.

## Suggested fix

Encrypt the OTP secret at rest like the other secret columns:

- Add an `otp_secret_encrypted []byte` column (+ migration; migrate/re-enroll existing secrets).
- Encrypt on enable, decrypt just before `totp.Validate`.
- Remove the plaintext `otp_secret` column once migrated.

## Location

- `src/core/models/user.go:45`
- `src/authserver/internal/handlers/apihandlers/handler_api_account_otp.go:168`
- `src/authserver/internal/handlers/handler_auth_otp.go:259,279`

_Found during a structured security review (Area B: secrets & cryptography at rest)._

leodip:
Fixed in 1ecd116.

- New \`users.otp_secret_encrypted\` column (migration 000020, all 4 DBs); TOTP secrets are now AES-GCM encrypted at rest with \`settings.AESEncryptionKey\`, like client secrets and verification codes.
[...]

Issue #187, opened 2026-08-09
Repository hygiene: tracked files that mislead or churn — citations to a gitignored doc, two committed binaries, and an orphan mocks directory


leodip:
Eleven citations of `docs/ci-v2-design.md` sit in eight tracked, public files. `.gitignore` excludes `docs/`, so that document exists on one machine and in no clone. Anyone reading a workflow or a build script and following the pointer finds nothing, and the reasoning those comments defer to is unreachable from a fresh checkout.

## Where

```
.github/dependabot.yml:9                     (5.9)
.github/workflows/check.yml:186              (5.8)
.github/workflows/check.yml:244              (7)
.github/workflows/docs.yml:4                 (4.9)
.github/workflows/publish.yml:4              (4.6)
.github/workflows/release.yml:3              (4.6)
src/build/build-binaries.sh:6                (4.1)
src/build/build-docker-images.sh:7           (4.1)
src/build/build-docker-images.sh:137         (4.6)
src/build/build-docker-images.sh:178         (4.6)
src/cmd/goiabada-setup/build-binaries.sh:5   (4.1)
```

Reproduce:

```
ls docs/ci-v2-design.md
grep -rn "docs/ci-v2-design.md" --include="*.yml" --include="*.sh" . | grep -v "^./docs/"
```

The first reports no such file on a clean tree; the second returns the eleven lines above.

## Why it matters

These are not decorative. Several of them carry the *reason* a job is shaped the way it is, and the comment defers the detail to a section number in a file the reader cannot open. The `check.yml:244` case is the sharpest: it explains why a check must not be made required, and a maintainer deciding whether that still holds has no way to read the rationale.

## Suggested fix

Per site, either restate the constraint inline, which is what the surrounding comments already do well, or repoint at something durable such as an issue number. Listing the sites is mechanical; the judgement is deciding how much of the design note each one has to inline.

Note that `check.yml:244` is being removed independently, for an unrelated reason, so it may be gone by the time this is picked up.

leodip:
## Two more tracked-tree defects, from the #219 run

Folding these in rather than filing them separately: all three are the same shape — something in the tracked tree that misleads a reader or churns a diff — and one cleanup pass closes them together. Retitled and relabelled (`bug`, `go` alongside `documentation`) to match. The original citation problem is unchanged and is still item 1.

### 2. `go build ./...` in `src/cmd/goiabada-setup` silently rewrites a tracked 11.7 MB binary

[...]

Issue #93, opened 2026-07-04
DCR endpoint has no request body size limit


leodip:
## Summary

The Dynamic Client Registration endpoint decodes the request body with no size limit. It is the public, unauthenticated endpoint, yet the only body-reader in the codebase missing the `http.MaxBytesReader` guard that all upload handlers use.

Severity: **Low**

## Details

`src/authserver/internal/handlers/handler_dynamic_client_registration.go:40`:

```go
var req api.DynamicClientRegistrationRequest
if err := json.NewDecoder(r.Body).Decode(&req); err != nil { ... }
```

No `http.MaxBytesReader` wraps `r.Body`. By contrast, all three upload handlers do (`handler_api_account_profile_picture.go:56`, `handler_api_users_profile_picture.go:57`, `handler_api_client_logo.go:57`).

## Impact

Memory/CPU pressure from an oversized JSON body. Mitigated by:

- DCR is feature-flagged off by default (`DynamicClientRegistrationEnabled`).
- The endpoint is rate-limited (though the limiter is IP-spoofable — see the companion `X-Forwarded-For` rate-limiter issue).

## Suggested fix

Wrap the body before decoding, e.g.:

```go
r.Body = http.MaxBytesReader(w, r.Body, 16*1024) // DCR metadata is small
```

## Location

- `src/authserver/internal/handlers/handler_dynamic_client_registration.go:40`

_Found during a structured security review (Area E: input handling & injection)._

leodip:
Merged into #205, which now covers all three places the server reads unbounded input before anything bounds it: the missing `MaxHeaderBytes`, this body, and the debug middleware's pre-authentication buffering. The detail and the suggested 16 KB guard are carried over in full.

One thing was corrected on the way: the note here that the DCR limiter is IP-spoofable is stale. #219 established that no limiter key is spoofable, and gave `POST /connect/register` a 10/min per-/64 bound.

Closed during a backlog consolidation pass.

PR #189, opened 2026-08-09
fix(security): redact credentials from API debug logging

### LINKED ISSUE #145: Security: redact passwords, client secrets and OTP material from API debug logging

leodip:
Closes #145.

## What this is for

When `GOIABADA_AUTHSERVER_DEBUG_API_REQUESTS=true`, the API debug middleware writes the request and
response bodies of every route under `/api/v1/admin` and `/api/v1/account` to the log. It redacts one
thing today, the `Authorization` header. Everything else goes in verbatim.

Executed against the code before this change, a single `PUT /api/v1/account/otp` writes 2865 bytes of
log containing the account password, the live TOTP code, the TOTP seed, and the enrollment QR PNG,
which is a second copy of the seed because `pquerna/otp` QR-encodes the `otpauth://...?secret=...`
URL. The bearer token is the only credential the old code keeps out.

That sits against a decision this repository already made elsewhere: TOTP secrets are encrypted at
rest (#82), and so are client secrets and signing keys. The debug log was the one place a plaintext
copy was written.

No OAuth or OIDC document governs what an authorization server writes to its own logs, and the
silence is itself the finding: this is a product decision, constrained by documents about credential
custody rather than about logging. RFC 6819 section 4.6.7 is titled "Token Leakage via Log Files and
HTTP Referrers" and RFC 9700 section 1 keeps 6819 in force. RFC 6238 section 5.1 says TOTP keys
"SHOULD be protected against unauthorized access and usage" and recommends "exposing them only when
required". A log file is neither a secure area nor a program required by the validation system.

## What this changes

Every body reaching the log goes through one formatter, and there is no path from a body to the log
that does not pass through it. A named set of credential-bearing JSON keys, plus a substring net over
`password`, `secret`, `otp` and `token`, decides what is replaced. A matched value becomes
`"[redacted]"` whatever its type was; the key stays, because "this request carried a `secretKey`" is
most of the debugging value and none of the secret.

**Keys are matched by folding runes, not by lowercasing them.** That is what `encoding/json` does to
struct-tag names, so the decoder accepts `{"paſſword": ...}` into the `Password` field. A handler
honours that request, and a `strings.ToLower` matcher returns false for it, which would log that
password in the clear. Case-sensitivity is the obvious bypass and lowercasing is the same bypass one
rune further out.

**The set is eleven keys, not the six the issue names.** `base64Image` is the TOTP seed in QR form,
`verificationCode` is a live credential on two endpoints, `smtpPassword` is an operator credential,
and `logoutUrl` is the account logout-request response's URL carrying a signed `id_token_hint` that
`/auth/logout` accepts. `client_secret` is kept although the DCR endpoint is not under this
middleware today, so the entry is not the one missing on the day someone mounts it there. The
substring net over-redacts about fourteen harmless keys such as `otpEnabled`, which is accepted: none
is a value anyone debugs with and every one is visible in the admin UI.

**The response body's raw fallback is deleted.** It printed an unparseable response verbatim, where
the request body was silently dropped, and a redactor walking only parsed JSON would have inherited
that asymmetry untouched. Both sides now log a one-line placeholder carrying the byte count and the
reason, which also stops a multipart upload being indistinguishable from a bodyless GET.

**Two bounds, both landing at that same placeholder:** 256 KB on the raw body and 32 levels of
nesting outside a redacted value. The map round-trip costs about 12x what `json.Indent` did, and
indentation cost grows as the square of the depth, so a 60 KB body nested 10000 deep expands into a
200 MB log line. This middleware is the first `r.Use` on both API mounts, ahead of
`RequireBearerTokenScope`, `RequireUserBoundToken` and `RequireValidSession`, so it formats bodies
before any authentication middleware runs and the amplification is unauthenticated.

The `Authorization` line becomes `Bearer [redacted]`, so one word means one thing across a log entry.
`None` is unchanged, because an absent credential must stay distinguishable from a redacted one.

The redactor is unexported and lives beside `debugLog` in `src/authserver/internal/middleware`.
Nothing new is exported and no package is created. #159 is in the same module, so if it wants this key
vocabulary it can call `isSensitiveKey` directly; the two share the key list and nothing else, since
this one walks decoded JSON and #159 walks `url.Values`.

## The goal, executed

Re-running the same probe that produced the 2865-byte leak above, against the merged code:

```
== Q1: what reaches the log today? ==
  account password:                    present in log: false
  live OTP code:                       present in log: false
  TOTP seed (request):                 present in log: false
  QR PNG (response, first 40 chars):   present in log: false
  bearer token:                        present in log: false

  total log bytes for this one request: 707
```

All four credentials that were present are absent, the bearer token is still absent, and the same
request now writes 707 bytes instead of 2865.

## What landed

| Stage | Commit | What |
|---|---|---|
| 1 | [`6794f4c7`](https://github.com/leodip/goiabada/commit/6794f4c7fc8c947ad5863957d4d0399158f7ce95) | the redactor, `bodyForLog`, `debugLog`, the exhaustive key table, and the `### Logging` docs subsection |
| 2 | [`a2e48f9c`](https://github.com/leodip/goiabada/commit/a2e48f9c6152f5a9b574cdb6df32a24b6f188eee) | the middleware wiring, end to end |

### Stage 1

`api_debug_middleware.go` gains the key vocabulary (`foldKey`, `foldRune`, `sensitiveKeys`,
`sensitiveKeySubstrings`, `isSensitiveKey`), the traversal (`redact`), the single body formatter
(`bodyForLog`) and the three constants. `debugLog` routes both bodies through `bodyForLog` and the
raw-response fallback is deleted. `security.mdx` gains a `### Logging` subsection saying what verbose
API logging writes and what it redacts.

**Tests: unit 3028 pass, 0 fail**, the middleware package going from 187 to 193 cases. **20 mutations
run, 20 caught, 0 survived.**

**Two review rounds, both productive, nothing contested.** Round 1 found that the trailing-content
check used `json.Decoder.More`, which is container-scoped and therefore reports no-more-input for a
top-level stray `]` or `}`, so `{"a":1}]` was logged as though it were valid JSON. It is now a second
`Decode` that must return `io.EOF`. No credential leaked by that path, since the accepted prefix still
went through the full redactor; what it cost was the guarantee that a malformed body is identifiable
as malformed. Round 1 also found the depth sentence in `security.mdx` overstated what is refused,
since a sensitive value is replaced whole and never traversed. Round 2 differentially tested the new
rule against `json.Indent` over 1,526,408 bodies and found no acceptance mismatch, and came back clean
on all three axes.

### Stage 2

Test-only, by design: no production code moved, which the tree hash confirms rather than asserts.
Stage 1 shipped the whole redactor, and this stage proves `APIDebugMiddleware` actually routes the
bytes it buffered and teed through it. Two cases drive a real `chi.NewRouter()` with the middleware
mounted exactly as `routes.go` composes it, one per direction: a real
`api.AccountOTPEnrollmentResponse` with a generated seed and QR on the response side, and a real
`api.UpdateAccountOTPRequest` with a live `totp.GenerateCode` on the request side. Both assert the
credentials are absent from the log while the bytes the client sends and receives are unchanged.

Two cases rather than one because the two bodies come from two different sources, the request buffer
and the response tee, and one case cannot fail if the other wire is cut. **Four mutations run, four
caught**: neutralising each formatter call, and separately cutting each wire at the call site. The
reviewer ran six of its own, including restoring an empty `r.Body` downstream and stopping
`responseWriter.Write` forwarding to the client, all caught.

**Tests: unit 3030 pass, 0 fail.** Review round 1 clean, zero findings, `unchecked` empty.

## Tiers

- **Local, per stage:** the unit tier, `3028` then `3030` passing with zero failures, plus the
  middleware package verbose and `go vet`. Every new test seam was mutation-tested: **24 mutations
  across the two stages, 24 caught, 0 survived.**
- **The check suite, on each pushed commit:** the data and integration tiers on sqlite, mysql,
  postgres and mssql as four parallel jobs, plus `Lint`, `Vulnerabilities`, `Build validation` and the
  version check. **Green on `a2e48f9c`:**
  [run 31338403002](https://github.com/leodip/goiabada/actions/runs/31338403002), all eleven jobs
  passing.
- **Data and integration were not applicable to either stage** and nothing local ran them. No query,
  interface method, migration or schema change exists here. The integration harness starts the
  authserver as a separate process, so a test cannot set the debug flag in its memory nor read its
  `slog` output; stage 2 is the in-process substitute for exactly that coverage.

## Deviations worth knowing

- **The trailing-content check.** The design said `dec.More()`; it shipped as a second `Decode`
  requiring `io.EOF`, because `More` is container-scoped and accepts a top-level stray `]` or `}`.
  Found by review, verified independently over 14 shapes and then over 1.5 million bodies.
- **`redact` counts depth on containers only**, at the top of the map and slice cases rather than on
  entry, so a body whose deepest container sits at exactly the limit is accepted. Both sides of that
  boundary are pinned by cases.
- **A sensitive value is replaced whole and never traversed**, so nesting below a redacted key is
  never counted against the depth bound. The documentation said otherwise for one round and now says
  this.
- **The seam 2 handler is the test's own**, not the production OTP handler:
  `handlers/apihandlers` imports `internal/middleware`, so calling it from a test in package
  `middleware` is an import cycle. What carries the credentials is real either way, the response and
  request types, the seed, the QR, a live TOTP code and the encoder call.

## Before merging

- [One deferred decision](https://github.com/leodip/goiabada/pull/189#issuecomment-5234391714), on
  `logoutUrl` being redacted whole and taking its other query parameters with it.
- [Two drafted follow-ups](https://github.com/leodip/goiabada/pull/189#issuecomment-5234391768), each
  ready to file: the middleware's unbounded pre-auth body read, and `.gitignore`'s unanchored `docs/`
  rule reaching into the documentation site.

### DIFF
--- a/site/src/content/docs/reference/security.mdx
+++ b/site/src/content/docs/reference/security.mdx
@@ -61,3 +61,7 @@ Goiabada includes several built-in security features:
 - **CORS configuration** - Explicit web origin configuration required for JavaScript clients
 - **Proxy header trust** - Forwarded IP headers are ignored by default; enable `TRUST_PROXY_HEADERS` only behind a proxy. Behind multiple proxies or a CDN, set `TRUSTED_PROXIES` to your proxy IPs/CIDRs so the client IP is resolved from the trusted end of `X-Forwarded-For` and cannot be spoofed
 
+### Logging
+
+- **Credentials are redacted from verbose API logging** - `GOIABADA_AUTHSERVER_DEBUG_API_REQUESTS` writes the request and response bodies of every `/api/v1/admin` and `/api/v1/account` call to the log, so credential values are replaced by `[redacted]` before anything is written: passwords, TOTP codes, TOTP secrets and their enrollment QR image, client secrets, the SMTP password, email verification codes, and the logout URL that carries a signed `id_token_hint`. Field names are kept, so the shape of the body is still readable. The `Authorization` header has always been reduced to `Bearer [redacted]`. A body that is not valid JSON, is larger than 256 KB, or nests more than 32 levels deep outside a redacted value is replaced by a one-line note giving its size and why it was not logged, rather than being written out. Depth is only counted through the parts of the body that get logged, since a redacted value is replaced whole and whatever it contains is never looked at. The flag is off by default and is meant for development
+
--- a/src/authserver/internal/middleware/api_debug_middleware.go
+++ b/src/authserver/internal/middleware/api_debug_middleware.go
@@ -9,9 +9,193 @@ import (
 	"net/http"
 	"strings"
 	"time"
+	"unicode"
 
 	"github.com/leodip/goiabada/core/config"
 )
+
+const (
+	// A body larger than this is never parsed or logged: the map round-trip below
+	// costs roughly 12x what json.Indent did, so an unauthenticated caller could
+	// otherwise make the server spend hundreds of milliseconds and hundreds of
+	// megabytes formatting one request. 256 KB sits well above anything this API
+	// produces: the list endpoints cap at size=200, an audit page of about 100 KB.
+	maxLoggedBody = 256 * 1024
+
+	// Indentation cost grows as the square of the nesting depth, so a small but
+	// deeply nested body expands into a log line hundreds of megabytes long.
+	// 32 is four times the deepest body any endpoint here returns (UpdateUserResponse,
+	// at 8 levels), so nothing legitimate is refused.
+	maxLoggedDepth = 32
+
+	redactedValue = "[redacted]"
+)
+
+// Keys whose values are credentials, written in their struct-tag spelling and
+// folded at startup. logoutUrl is here because POST /api/v1/account/logout-request
+// returns a URL carrying a signed id_token_hint that /auth/logout accepts: the name
+// says nothing about that, so dropping the entry would log a usable token in the
+// clear (#145).
+var sensitiveKeys = foldKeySet(
+	"password", "currentPassword", "newPassword",
+	"otpCode", "secretKey", "base64Image",
+	"smtpPassword", "clientSecret", "client_secret",
+	"verificationCode", "logoutUrl",
+)
+
+// The backstop for credential fields nobody remembers to add above. It over-redacts
+// about fourteen harmless keys on this surface (otpEnabled, tokenExpirationInSeconds
+// and the like), which is accepted: none of them is a value anyone debugs with, and
+// every one is visible in the admin UI (#145).
+var sensitiveKeySubstrings = foldKeyList("password", "secret", "otp", "token")
+
+// foldRune maps a rune to the smallest rune it is case-equivalent to, which is what
+// encoding/json's own foldName does to struct-tag names. Lowercasing is not the same
+// thing and is a bypass rather than a rough edge: the decoder accepts
+// {"paſſword": ...} into the Password field, and a strings.ToLower matcher
+// returns false for it, so that request's password would be logged in the clear (#145).
+func foldRune(r rune) rune {
+	for {
+		folded := unicode.SimpleFold(r)
+		if folded <= r {
+			return folded
+		}
+		r = folded
+	}
+}
+
+// foldKey folds every rune of s, so foldKey(a) == foldKey(b) exactly when
+// strings.EqualFold(a, b).
+func foldKey(s string) string {
+	var b strings.Builder
+	b.Grow(len(s))
+	for _, r := range s {
+		b.WriteRune(foldRune(r))
+	}
+	return b.String()
+}
+
+func foldKeySet(keys ...string) map[string]struct{} {
+	set := make(map[string]struct{}, len(keys))
+	for _, key := range keys {
+		set[foldKey(key)] = struct{}{}
+	}
+	return set
+}
+
+func foldKeyList(keys ...string) []string {
+	folded := make([]string, 0, len(keys))
+	for _, key := range keys {
+		folded = append(folded, foldKey(key))
+	}
+	return folded
+}
+
+// isSensitiveKey reports whether a JSON key's value must be redacted, matching the
+// named set and the substring net against the folded form of the key.
+func isSensitiveKey(key string) bool {
+	folded := foldKey(key)
+	if _, ok := sensitiveKeys[folded]; ok {
+		return true
+	}
+	for _, substring := range sensitiveKeySubstrings {
+		if strings.Contains(folded, substring) {
+			return true
+		}
+	}
+	return false
+}
+
+// redact returns v with the value of every sensitive key replaced by redactedValue,
+// recursing through objects and arrays. A sensitive value is replaced whole, without
+// recursing into it, so an object-valued secret leaves nothing behind. The top-level
+// value is depth 1; a container nested deeper than maxLoggedDepth returns an error and
+// the caller logs a placeholder instead of the body.
+func redact(v any, depth int) (any, error) {
+	switch value := v.(type) {
+	case map[string]any:
+		if depth > maxLoggedDepth {
+			return nil, fmt.Errorf("nested deeper than %d levels", maxLoggedDepth)
+		}
+		for key, element := range value {
+			if isSensitiveKey(key) {
+				value[key] = redactedValue
+				continue
+			}
+			redacted, err := redact(element, depth+1)
+			if err != nil {
+				return nil, err
+			}
+			value[key] = redacted
+		}
+		return value, nil
+	case []any:
+		if depth > maxLoggedDepth {
+			return nil, fmt.Errorf("nested deeper than %d levels", maxLoggedDepth)
+		}
+		for i, element := range value {
+			redacted, err := redact(element, depth+1)
+			if err != nil {
+				return nil, err
+			}
+			value[i] = redacted
+		}
+		return value, nil
+	}
+	return v, nil
+}
+
+// bodyForLog returns the text to log for one request or response body: the body
+// pretty-printed with every credential replaced, or a one-line placeholder giving the
+// byte count and the reason. No part of a body that could not be parsed, that is too
+// large, or that is nested too deeply is ever returned.
+func bodyForLog(body []byte) string {
+	notLogged := func(reason any) string {
+		return fmt.Sprintf("%d bytes, not logged (%v)", len(body), reason)
+	}
+
+	if len(body) > maxLoggedBody {
+		return notLogged(fmt.Sprintf("larger than %d bytes", maxLoggedBody))
+	}
+
+	decoder := json.NewDecoder(bytes.NewReader(body))
+	// UseNumber keeps integers exact. Without it an int64 such as 9223372036854775807
+	// round-trips through float64 and is logged as 9223372036854776000, which reads
+	// like a real value and is not one (#145).
+	decoder.UseNumber()
+
+	var parsed any
+	if err := decoder.Decode(&parsed); err != nil {
+		return notLogged(err)
+	}
+	// A second Decode that ends in io.EOF is what proves the body was one complete
+	// value and nothing else. Decoder.More cannot do this job: it answers "is there
+	// another element in the container I am inside", so at the top level it reports
+	// no-more-input for a stray ] or }, and {"a":1}] would be logged as though the
+	// body were valid JSON. Requiring io.EOF accepts exactly what json.Indent accepts,
+	// trailing whitespace included (#145).
+	var trailing any
+	if err := decoder.Decode(&trailing); err != io.EOF {
+		return notLogged("unexpected trailing content after the top-level value")
+	}
+
+	redacted, err := redact(parsed, 1)
+	if err != nil {
+		return notLogged(err)
+	}
+
+	var out bytes.Buffer
+	encoder := json.NewEncoder(&out)
+	// SetEscapeHTML(false) keeps the body readable. With the default on, every "&" in
+	// a URL is rewritten as its backslash-u escape, u0026, and every "<" as u003c, so
+	// a websiteUrl in the log stops matching the one the client sent (#145).
+	encoder.SetEscapeHTML(false)
+	encoder.SetIndent("", "  ")
+	if err := encoder.Encode(redacted); err != nil {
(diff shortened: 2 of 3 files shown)


leodip:
## Deferred decisions

One question came up during the run that was judged safe to proceed past on a stated assumption
rather than to stop for. It is recorded here so it is settled before this merges, and reversing it is
contained.

### 1. `logoutUrl` is redacted whole, so the debug log loses the logout URL's other parameters too

Raised by the reviewer at the plan gate.

`POST /api/v1/account/logout-request` returns

```json
{"logoutUrl":"<baseurl>/auth/logout?id_token_hint=<signed JWT>&post_logout_redirect_uri=...&state=..."}
```

That endpoint is mounted under the debug middleware, so with `GOIABADA_AUTHSERVER_DEBUG_API_REQUESTS`
enabled the whole signed ID token is written to the log today. The token carries `sub`, `sid` and
`aud`, and `/auth/logout` accepts it, so anyone reading the log can tear down that client session.
`AccountLogoutRedirectResponse`'s own type comment in `src/core/api/responses.go` already says this
form "exposes the token in the URL".

**Assumed, and shipped in this PR:** `logoutUrl` joins the exact sensitive-key set, so the whole value
becomes `"[redacted]"`. The log no longer shows the `post_logout_redirect_uri` or the `state` an
operator might have been debugging with.

**Why this and not something narrower.** Whole-value replacement is the only mechanism this change
has. Values are replaced whole rather than rewritten, and detecting sensitive material by the shape of
a value rather than by its key was explicitly kept out of scope, so redacting only the `id_token_hint`
query parameter would mean parsing values inside the redactor. That is a different design, and it is
the same design #159 needs for query strings on the other middleware, so if it gets built it should be
built once.

**Why it was not escalated mid-run.** It applies the same principle as `base64Image`: a credential
under a key whose name does not say so belongs in the exact set. The remedy narrows what is logged,
changes no interface, and costs nothing at runtime.

**Cost to reverse:** delete one entry from `sensitiveKeys`, one table row and one mutation row.
Nothing else depends on it. If you would rather keep the non-credential parameters visible, say so and
it comes out in one line, at the price of the signed token going back into the log.

leodip:
## Follow-ups, drafted rather than filed

Two real problems found while doing this work that belong to a different change. Each is verified
against the code, checked against the open tracker, and written so it can be filed as it stands. The
duplicate searches below were re-run at the closing pass.

---

### 1. Security: the API debug middleware buffers the whole request body before authentication and before the handler's size limit

Labels: `security`, `go`

`APIDebugMiddleware` is the first `r.Use` on both `/api/v1/admin` and `/api/v1/account`
(`src/authserver/internal/server/routes.go`), so it runs ahead of `RequireBearerTokenScope`,
`RequireUserBoundToken` and `RequireValidSession`. Its first act is `reqBody, _ = io.ReadAll(r.Body)`
with no bound. The three handlers that do limit a body install `http.MaxBytesReader` inside themselves
(`grep -rn "MaxBytesReader" --include=*.go src/ | grep -v _test.go` returns 3 hits, all in
profile-picture and client-logo handlers), which is strictly after the middleware has already read it.
The response side has the same shape: `responseWriter.Write` tees into an unbounded `bytes.Buffer`.

Out of scope for #145, which is about what the log contains rather than what the middleware allocates.
#145 caps what is *formatted* at 256 KB, but the read itself still happens first, so that cap does not
close this.

Searched `gh issue list --state open --search "MaxBytesReader"` and `--search "debug middleware body
size"` at the closing pass. #93 is the same class on `/connect/register`, a different endpoint and a
different code path, and is not this.

**Body to file:**

> When `GOIABADA_AUTHSERVER_DEBUG_API_REQUESTS` is enabled, `APIDebugMiddleware` reads the entire
> request body into memory with `io.ReadAll` before any authentication middleware runs and before any
> handler installs `http.MaxBytesReader`. An unauthenticated request to any route under
> `/api/v1/admin` or `/api/v1/account` is therefore buffered in full, whatever its size, and only then
> rejected with a 401. The response side buffers without a bound too, through `responseWriter.Write`
> teeing into a `bytes.Buffer`.
>
> This is behind a flag that is off by default, which is what keeps it out of #145. #145 bounds what
> is formatted into the log at 256 KB, but the read has already happened by then.
>
> Fix by bounding the read in the middleware itself, with `http.MaxBytesReader` or an
> `io.LimitReader`, and logging a placeholder when the limit is hit. A limit in the middleware also
> has to leave the handler's own `MaxBytesReader` working, since the handler replaces `r.Body`
> afterwards.

```bash
gh issue create \
  --title "Security: the API debug middleware buffers the whole request body before authentication and before the handler's size limit" \
  --label security --label go \
  --body 'When `GOIABADA_AUTHSERVER_DEBUG_API_REQUESTS` is enabled, `APIDebugMiddleware` reads the entire request body into memory with `io.ReadAll` before any authentication middleware runs and before any handler installs `http.MaxBytesReader`. An unauthenticated request to any route under `/api/v1/admin` or `/api/v1/account` is therefore buffered in full, whatever its size, and only then rejected with a 401. The response side buffers without a bound too, through `responseWriter.Write` teeing into a `bytes.Buffer`.

`APIDebugMiddleware` is the first `r.Use` on both mounts in `src/authserver/internal/server/routes.go`, ahead of `RequireBearerTokenScope`, `RequireUserBoundToken` and `RequireValidSession`. The three handlers that do limit a body install `http.MaxBytesReader` inside themselves (`grep -rn "MaxBytesReader" --include=*.go src/ | grep -v _test.go` returns 3 hits, all in profile-picture and client-logo handlers), which is strictly after the middleware has already read it.

This is behind a flag that is off by default, which is what kept it out of #145. #145 bounds what is formatted into the log at 256 KB, but the read has already happened by then.

Fix by bounding the read in the middleware itself, with `http.MaxBytesReader` or an `io.LimitReader`, and logging a placeholder when the limit is hit. A limit in the middleware also has to leave the handler'"'"'s own `MaxBytesReader` working, since the handler replaces `r.Body` afterwards.

Related: #93 is the same class on `/connect/register`, a different endpoint and a different code path.'
```

---

### 2. Bug: `.gitignore`'s `docs/` rule also excludes the documentation site's content tree, so a new docs page is silently untracked

Labels: `bug`, `documentation`

`.gitignore` line 443 is `docs/`. A pattern with no leading slash matches at **every** directory
level, not just the repository root, so it matches `site/src/content/docs/` as well as the intended
`docs/`. All 41 pages under `site/src/content/docs/` are already tracked, so they still commit and
nothing is broken today, which is exactly why this is invisible.

Reproduce:

```bash
git check-ignore -v site/src/content/docs/reference/a-brand-new-page.mdx
git check-ignore -v --no-index site/src/content/docs/reference/security.mdx
```

Both print `.gitignore:443:docs/`. This is how it was found: staging this PR's own edit to
`security.mdx` failed with "The following paths are ignored by one of your .gitignore files:
site/src/content/docs" and needed `git add -f`, which adds nothing new to the repository because that
file is already tracked.

Searched `gh issue list --state open --search "gitignore"` and `--search "documentation site
untracked"` at the closing pass. #187 is the only nearby one and it is a different defect: tracked
files citing `docs/ci-v2-design.md`, a path the same rule excludes on purpose. This one is about the
rule reaching a directory it was never meant to.

**Body to file:**

> `.gitignore` ends with `docs/`, which excludes the working-document directory at the repository
> root. Because the pattern has no leading slash, git applies it at every directory level, so it also
> excludes `site/src/content/docs/`, the entire content tree of the documentation site.
>
> The 41 pages there are already tracked and keep working, so the problem is invisible until someone
> adds a page: the new file is ignored, `git add -A` stages nothing, and `git add <path>` fails with
> "The following paths are ignored by one of your .gitignore files". The docs site builds from that
> directory, so a page that exists locally and not in the repository publishes nothing.
>
> Confirm with `git check-ignore -v site/src/content/docs/reference/a-brand-new-page.mdx`, which
> prints `.gitignore:443:docs/`.
>
> Fix by anchoring the pattern to the repository root as `/docs/`, which keeps every working-document
> path the rule was written for and stops it reaching into `site/`. Worth checking the rest of
> `.gitignore` for the same unanchored shape in the same pass.

```bash
gh issue create \
  --title "Bug: .gitignore's docs/ rule also excludes the documentation site's content tree" \
  --label bug --label documentation \
  --body '`.gitignore` ends with `docs/`, which excludes the working-document directory at the repository root. Because the pattern has no leading slash, git applies it at every directory level, so it also excludes `site/src/content/docs/`, the entire content tree of the documentation site.

The 41 pages there are already tracked and keep working, so the problem is invisible until someone adds a page: the new file is ignored, `git add -A` stages nothing, and `git add <path>` fails with "The following paths are ignored by one of your .gitignore files". The docs site builds from that directory, so a page that exists locally and not in the repository publishes nothing.

Confirm with `git check-ignore -v site/src/content/docs/reference/a-brand-new-page.mdx`, or with `git check-ignore -v --no-index site/src/content/docs/reference/security.mdx` for a page that already exists. Both print `.gitignore:443:docs/`.

Fix by anchoring the pattern to the repository root as `/docs/`, which keeps every working-document path the rule was written for and stops it reaching into `site/`. Worth checking the rest of `.gitignore` for the same unanchored shape in the same pass.

Found while committing #145, whose docs edit had to be staged with `git add -f`. Not the same as #187, which is about tracked files citing a path the rule excludes deliberately.'
```

Issue #160, opened 2026-08-09
issueAuthCode commits a partial 200 when the form_post template fails part way through rendering


leodip:
`issueAuthCode` in `src/authserver/internal/handlers/handler_auth_issue.go` renders `form_post.html` with `t.Execute(w, m)`, writing straight to the `ResponseWriter`. `Execute` writes as it walks the template, so a template that parses cleanly and then fails part way through has already emitted its prefix and triggered an implicit `WriteHeader(200)` by the time the error is returned. Its only caller answers that error with `httpHelper.InternalServerError(w, r, err)`, whose `WriteHeader(500)` is then a no-op.

The client is left with HTTP 200 and a truncated page for an authorization code that was minted and persisted but never delivered, and nothing in the response says the request failed.

`form_post.html` is operator supplied whenever `GOIABADA_AUTHSERVER_TEMPLATEDIR` is set, since `server.New` serves templates from `os.DirFS` in that case, so this is reachable in a real deployment. The template shipped in `web/template/form_post.html` cannot trigger it: it only dereferences map keys, which yield a zero value rather than an error.

The same defect was fixed in the sibling helper `redirToClientWithError` in #141, by rendering into a `bytes.Buffer` and copying to the `ResponseWriter` only after `Execute` returns successfully. That repair transfers directly, and a regression test can follow `TestRedirToClientWithError_FormPostExecutionFailureLeavesResponseUncommitted` in `handler_authorize_test.go`: a parse-valid, execution-failing template, then assert the body is empty and a later `WriteHeader(500)` still lands.

Issue #139, opened 2026-08-05
Security: a code insert can land between a termination's UPDATE and its COMMIT, escaping both marks


leodip:
#129 marks a terminated session's authorization codes revoked, and `/auth/issue` runs a compensating `UPDATE ... WHERE NOT EXISTS (SELECT 1 FROM user_sessions ...)` immediately after inserting a code, so a code created after the termination is marked even when the termination could not see it.

The remaining window is an insert committing after the termination's `UPDATE` has read `codes` and before that transaction commits: the compensating statement still observes the session as present, so the code carries no marker, and redeeming it yields an offline refresh token good for up to the configured offline maximum lifetime.

Accepted as a documented residual of #129, which weighed this window and chose to record it rather than close it. It does **not** block #129: the two sweepers cover every other interleaving, and the window is between two bounded statements rather than the unbounded handler stall a read-only check would leave. This issue tracks the durable fix. #134 held the previous shape of this residual and was closed as moot when #129's decision 4 moved the marker off a session-keyed registry onto `codes.revoked`, so nothing tracks it today.

## The mechanism

`TerminateUserSessionTx` (`src/authserver/internal/handlers/revocation.go`) opens one transaction and runs three writes in it: `RevokeCodesBySessionIdentifier`, the sid-scoped refresh-token sweep, then `DeleteUserSession`, then commits. `HandleIssueGet` inserts its code with `CreateCode` and then calls `RevokeCodeIfSessionGone` on a nil transaction, so that statement takes its own snapshot.

The damaging interleaving:

1. `/auth/issue` resolves the ceremony's bound session and finds it live.
2. The termination's `RevokeCodesBySessionIdentifier` reads and updates `codes`. The ceremony's code does not exist yet, so nothing matches it.
3. `/auth/issue` inserts its code and commits.
4. `RevokeCodeIfSessionGone` runs. The termination has not committed, so the `user_sessions` row is still visible to it and the `NOT EXISTS` term is false. No marker is written.
5. The termination commits. The session row is gone and the code carries no marker.

Redemption then classifies the missing session as ordinary housekeeping deletion, which it must, because the background worker reaps idle and expired sessions routinely while an offline grant is designed to outlive that. The code redeems, and the offline refresh token descended from it is not covered by the code marker.

## Correctness here currently rests on three engines' default isolation levels, by accident

The window is not uniform across the four supported engines, and reproducing it is engine-specific:

[...]

Issue #137, opened 2026-08-04
Security: the code exchange reveals generation and disabled-user state before the requester authenticates


leodip:
## Symptom

The authorization-code exchange reveals whether a user's authentication generation has moved,
before the requester has proved it is the legitimate client.

Present a stolen authorization code with a **wrong** PKCE verifier:

| User's generation | Response |
|---|---|
| unchanged | `invalid_grant`, "Invalid code_verifier (PKCE)." |
| moved (password reset, password change, admin password set, account disabled) | `invalid_grant`, "Code is invalid." |

The attacker never proves possession of the code verifier or the client secret, and still learns
that the account's credentials changed.

## Where

`ValidateTokenRequest`, the `authorization_code` case in
`src/core/validators/token_validator.go`. The `!wasReused` block runs the user-enabled check, the
generation check and the expiry check. Client-secret validation and PKCE validation both run
**after** it.

```go
if !wasReused {
    if !codeEntity.User.Enabled { ... }
    if codeEntity.AuthStateGeneration != codeEntity.User.AuthStateGeneration { ... }  // <- leaks
    if time.Now().UTC().After(codeEntity.CreatedAt.Time.Add(...)) { ... }
}
// client secret validated here
// PKCE validated here
```

The disabled-user check has the same shape: `customerrors.ErrUserDisabled` is distinguishable from a
PKCE failure in the same way, so an unauthenticated presenter also learns whether the account is
disabled.

## Why it is filed separately

Found while implementing #129 stage 3, which added a terminated-session registry check to the same
block. That check had exactly this problem and was moved after client authentication, PKCE, reuse
handling and expiry (#129 decision 14), with a negative control asserting the registry is not
consulted when PKCE fails.

The generation and disabled-user checks were deliberately left alone. They predate #129, they belong
to #106, and moving them is a behaviour change that issue should own rather than something to slip
into an unrelated stage. #129's spec records the inconsistency rather than hiding it.

So this issue exists to close the other half of the class.

## Impact

Low but real. The information disclosed is "this account's credentials recently changed" or "this
account is disabled", to someone holding a code they cannot redeem. It does not grant access. RFC
6749 Section 5.2 does not require distinguishable descriptions, and the codebase elsewhere goes out
of its way to avoid exactly this: `invalidGenerationMessage`'s comment says spelling out that a
credential change superseded the grant "would tell an attacker holding a stolen token exactly what
happened", and #129 added `terminatedSessionMessage` with the same reasoning.

[...]

PR #196, opened 2026-08-10
fix(auth): bind an authorization ceremony to the authenticating user's session

### LINKED ISSUE #133: Security: an authorization ceremony can bind to a session belonging to a different user
## Summary

An authorization ceremony can bind to a user session belonging to a **different user**. A person who
enters their own credentials on a shared browser ends up with tokens carrying someone else's session
identifier, and the browser stays signed in as the *other* user with the new user's authentication
strength grafted onto that session.

Found while specifying #129 and recorded there as decision 10. **It is independent of #129**: it
needs no session termination, no offline grant and no race. #129 implements the one-line guard
because the session binding it adds to `AuthContext` would otherwise persist the wrong association
deliberately rather than incidentally, but the defect, its consequences and its test coverage belong
here.

Severity: **Medium.** Requires a shared or reused browser and a client that sends `prompt=login` (or
an `id_token_hint` for a different user), which is a normal configuration rather than an attack
precondition.

## The mechanism

Two entry paths leave a live session cookie in place while sending the user to re-authenticate, and
neither is followed by any check that the person who authenticates owns that session.

In `HandleAuthorizeGet`:

- The **`prompt=login`** branch sets `AuthStateRequiresLevel1` and redirects to `/auth/level1`,
  returning **before** the session lookup and before the block that copies `UserId`, `AcrLevel`,
  `AuthMethods` and `AuthStateGeneration` off the reused session. It does not clear
  `SessionKeySessionIdentifier` from the cookie.
- The **`id_token_hint` subject mismatch** check sits inside the `hasValidUserSession` block and
  returns at the same point, also leaving the cookie session intact.

Verified sequence. User A is signed in, the cookie holds `sid_A`, and session `S_A` has `UserId = A`:

1. An authorize request arrives with `prompt=login`, or with an `id_token_hint` naming someone other
   than A. Either way the flow redirects to level 1 with `S_A` still live and still cookied.
2. User B enters B's credentials. `handler_auth_pwd`'s POST sets `authContext.UserId = B`. Its GET
   reads the ambient session identifier **only to prefill the email field** and never compares users.
3. `HandleAuthLevel1CompletedGet` loads `S_A` from the ambient identifier and reads **A's**
   `AcrLevel` to decide whether **B** needs step-up authentication.
4. `HandleAuthCompletedGet` sees `hasValidUserSession == true` and calls `BumpUserSession(sid_A, ...)`
   with **B's** auth methods and target ACR.
5. `authContext.SetAcrLevel(targetAcrLevel, userSession)` then runs with **A's** session, and
   `SetAcrLevel` returns the maximum of the target and the supplied session's ACR.
6. `/auth/issue` reads the session identifier from the request context, which is still `sid_A`, and
   `CreateAuthCode` stamps it onto a code whose `UserId` is B.

No step in that chain compares `userSession.UserId` with `authContext.UserId`. `BumpUserSession`
loads purely by session identifier and has no user parameter to compare against.

## Consequences

**1. The grant is cross-bound.** B's authorization code, and therefore B's access and refresh tokens,
carry A's session identifier. `RequireValidSession` then validates B's bearer token against A's
session: A's idle timeout and maximum lifetime govern B's access, terminating A's session kills B's
tokens, and the generation comparison runs `S_A.AuthStateGeneration` (a snapshot of **A's**
generation) against **B's** current `users.auth_state_generation`. Those are independent counters, so
they agree by coincidence when both are 0 and diverge the moment either user has a credential change.

**2. The browser stays signed in as A, with B's authentication strength.** This is the worse one.
`BumpUserSession` assigns `userSession.AuthMethods` and `userSession.AcrLevel` but **never touches
`UserId`**. So after step 4, `S_A` still has `UserId = A` while recording the methods and ACR that B
proved, and the cookie still points at `S_A`. The next authorization request performs SSO **as A**.
A user who typed their own password has silently resumed someone else's session, and that session's
recorded authentication level now reflects a factor A never presented.

## Proposed fix

Treat the valid-session branch in `HandleAuthCompletedGet` as reusable only when both hold:

```
hasValidUserSession && userSession.UserId == authContext.UserId
```

When the session is valid but belongs to another user, fresh level 1 authentication has already
happened, so `StartNewUserSession` is the legitimate path: create and bind a new session for
`authContext.UserId`. Note `StartNewUserSession` already deletes other sessions matching the same
device name, type, OS and IP, which is the correct outcome for a browser account switch.

On the ordinary SSO path the guard is a no-op, because `HandleAuthorizeGet` sets
`authContext.UserId` from the very session it validated.

The guard alone is not the whole fix. The ambient session must not be used **anywhere** after it
fails, and `SetAcrLevel` is the specific leak: it runs after the branch and would otherwise graft A's
ACR onto B's code. Feed it the newly created session or nil. Auditing the handler for other
post-branch uses of `userSession` is part of the work.

Also worth deciding here rather than leaving implicit: whether `HandleAuthLevel1CompletedGet` should
consult an ambient session that does not belong to `authContext.UserId` at all, since today it reads
A's ACR to make a step-up decision about B. Rejecting the mismatch earlier would make the
`/auth/completed` guard a backstop rather than the only defence.

## Acceptance criteria

1. A ceremony that authenticates user B while user A's session is cookied never issues a code
   carrying A's session identifier.
2. `S_A` is not mutated by B's ceremony: its `AuthMethods` and `AcrLevel` are unchanged.
3. After B authenticates, the browser's subsequent SSO request authenticates as **B**, not A.
4. B's code and tokens carry B's target ACR and B's authentication methods only, never the maximum
   taken with A's session.
5. The ordinary SSO path, where the session belongs to the authenticating user, is unaffected.

## Test requirements

- **Unit**, host tier. Extend `handler_auth_completed_test.go`. The branch table needs a row where a
  valid session belongs to a different user and the ceremony completed level 1, asserting
  `StartNewUserSession` rather than `BumpUserSession`. Give that row an ambient session with a
  **higher** ACR than the ceremony's target and assert the resulting session, auth context and code
  carry the authenticating user's ACR, since asserting only the branch choice leaves the
  `SetAcrLevel` leak invisible.
- **Integration**, dev container. Two flows on one HTTP client so the cookie is shared: sign in as A,
  then run a `prompt=login` ceremony authenticating as B, and assert the code's `user_id` is B and
  its `session_identifier` is not `sid_A`. Then issue a third, ordinary authorize request and assert
  it authenticates as B.
- **Integration**, the `id_token_hint` variant, since it reaches the same branch by a different route.
- **Regression.** `authorize_existing_session_test.go` and `authorize_id_token_hint_test.go` must
  stay green: the guard is a no-op whenever the session belongs to the authenticating user.

## Relationship to other issues

- **#129** implements the same-user guard as part of its stage 5, because the bound session
  identifier it adds to `AuthContext` must not record the wrong association. This issue owns the full
  write-up, the browser-stays-signed-in-as-A consequence, the `SetAcrLevel` leak, the
  `HandleAuthLevel1CompletedGet` question and the test coverage above. #129's design record carries the reasoning and the verification.
- **#106** introduced the per-user generation whose comparison this defect makes incoherent, since
  the session's snapshot and the token subject's counter belong to different users.

leodip:
Closes #133.

## What this is for

An authorization ceremony can bind to a session belonging to a **different user**.

Two paths in `HandleAuthorizeGet` send the browser off to re-authenticate while leaving the existing
session cookie in place: `prompt=login` (and `max_age=0`, which is equivalent) returns before the
session is ever looked up, and an `id_token_hint` naming someone other than the session's owner
returns from inside the valid-session block. Neither reaches the point where the session's user is
copied onto the auth context, so the cookie survives and nothing downstream compares the person who
just authenticated with the owner of that session.

User A is signed in. User B then authenticates in the same browser. Today:

- **B's second factor is skipped.** `/auth/level1completed` decides step-up by comparing the target
  ACR against **A's** session ACR, so when A's session already sits at or above the target, B reaches
  a `level2_mandatory` client having presented a password only, with `amr` `["pwd"]`, without needing
  OTP enabled at all.
- **The authorization code is stamped with A's session identifier.** B's tokens then live inside A's
  session: A's idle timeout and max lifetime govern B's access, ending A's session kills B's tokens,
  and the authentication-generation check compares A's snapshot against B's counter.
- **A's session row is damaged.** `BumpUserSession` overwrites its auth methods with B's, ratchets its
  ACR, and clears the flag that was about to force A to re-verify a changed OTP setting. The cookie
  still points at it, so the next request performs SSO **as A**, at an ACR A last proved and an AMR
  describing B's ceremony.
- **A's grants stay alive in a browser that changed hands.** `StartNewUserSession` only deletes
  sibling sessions belonging to the *new* user, so A's row is never a candidate.

No race and no attacker-controlled input are involved. A shared or reused browser plus a client
sending `prompt=login` or an `id_token_hint` is enough, and all of that is ordinary configuration.

## The rule this change establishes

A ceremony may read, mutate or bind to the browser's ambient session **only when that session belongs
to the user the ceremony authenticated.** One predicate on the auth context, consulted at every place
the ambient session is reachable: the step-up decision, the completion branch, and code issuance. Plus
two backstops on the paths that validate already-issued tokens, since grants minted before the upgrade
stay cross-bound.

When the browser genuinely changes hands, the session that was there is terminated rather than left
orphaned: its row deleted, its refresh tokens and unredeemed codes revoked, and the handover recorded
as its own audit event.

### Stage 1: the `OwnsSession` predicate and the step-up decision

Landed in `8317e51c`. `AuthContext.OwnsSession`
states the rule once, with both zero cases false so the fail-closed direction is the default. It is a
predicate rather than a filter that hands back the session or nil, because a later stage needs to see
the foreign session in order to terminate it. `/auth/level1completed` now conjoins it with the
lifetime check, so a foreign session takes the existing target-only arm and B is sent to
`/auth/level2` exactly as a user with no session would be; the write that cleared A's
`Level2AuthConfigHasChanged` sits inside the block that is now skipped. This closes the second-factor
bypass, which is the sharpest edge of the defect and the one decided a hop earlier than the issue
predicted.

Same-user SSO is untouched: the ordinary path copies the session's owner onto the auth context before
this handler runs, so ownership holds there.

Tiers: unit on all three modules, 3049 pass. Integration on sqlite, 1218 pass, carrying the 15
same-user SSO regressions that prove the new conjunct does not disturb them. The data tier is not
applicable, since the stage adds no query, interface method, migration or schema change. Seven
mutations against the new cases, three planned and four raised by review, all caught. The four-engine
data and integration tiers are the check suite's to report.

### Stage 2: `/auth/completed` binds to the right session, and terminates the wrong one

Landed in `260d95c3`. The reuse arm now requires both questions to be yes, validity **and**
ownership, so B's ceremony can no longer bump A's row with B's methods. `SetAcrLevel` is given the
session the ceremony actually bound to rather than the one the browser arrived with, which also
closes a same-user defect with no cross-user element in it: an expired `level2_optional` session
plus a fresh password-only login used to produce a `level1` session and a token claiming
`acr: level2_optional`.

A foreign session is ended rather than left behind, and the ordering is the substance: the level 1
gate comes first, so destroying a session must follow a real authentication in this ceremony rather
than a ceremony that merely arrived. `StartNewUserSession` does not do this on its own, contrary to
what the issue assumes: it sweeps siblings of the **new** user, so the previous user's row would
survive orphaned with its refresh tokens still working and still bumping it. Termination revokes the
codes and refresh tokens that session authorized, because those are grants this browser holds and
the browser has changed hands; only grants keyed to that session identifier are touched, so A's
sessions on other devices are unaffected. A termination failure is a 500, with nothing audited and
no replacement created.

The handover writes a new `cross_user_session_replaced` event naming both parties, beside the
`terminated_user_session` pair every caller of `TerminateUserSessionTx` owes, so an operator can
tell a browser handover from an administrator ending a session. All three follow the commit, so no
event can attest to a termination that rolled back.

Tiers: unit 3054 pass, 0 fail across all three modules. Integration on sqlite, 1218 pass, 0 fail.
Nine mutations, all caught, including the ACR source, each conjunct of the branch condition, the
termination error check, the audit ordering and the replacement-failure path. The four-engine data
and integration tiers are the check suite's to report.

### Stage 3: the `/auth/issue` backstop

Landed in `3d522e2b`. The last gate before a session identifier is stamped onto a grant asked only
whether the session was live, so a resolving row belonging to somebody else passed it. It now asks
whether the ceremony owns the session, and it sits **above** the response-type dispatch.

That placement is the substance rather than a tidy-up. `handleImplicitFlow` copies the ambient
identifier straight into `ImplicitGrantInput.SessionIdentifier` and never loads the row, so with the
gate underneath it a `response_type` of `token`, `id_token` or `id_token token` would mint a signed
token for B carrying A's `sid`. No later backstop can reach that: a resource server validating an
already-signed token cannot look the session up, and the code flow's own compensation works only
because a code is a row that can be marked after the fact.

A foreign session folds into the not-live branch: `prompt=none` gets `login_required` in the
response mode its family uses, everything else restarts level 1, and no code is minted either way.
It self-heals in one pass, since the second visit to `/auth/completed` terminates the foreign
session and starts one the ceremony owns. Each refusal arm carries its own warning naming the two
user ids that failed to match, because the client-visible answer is identical to the
session-is-gone case by design and the log is the only place an operator can tell a takeover attempt
from an expiry.

One carve-out, drawn narrowly: an implicit request with an **empty** identifier keeps today's
behaviour, since there is no ambient session to cross-bind to. A non-empty identifier whose row has
vanished is refused, which is a behaviour change for implicit and the direction the previous
session-liveness work settled.

Tiers: unit 942 pass, 0 fail. Integration on sqlite 1218 pass, 0 fail, including all 15 tests in
`authorize_existing_session_test.go` and all 3 in `authorize_id_token_hint_test.go`. Data on four
engines 1144 pass, 0 fail. Two review rounds and nineteen mutations, all caught: one per issuance
family through the shared gate, the carve-out's exact scope in both directions, and, after round 2,
that each warning is a single `WARN` record carrying the right two ids rather than the right strings
loose somewhere in the log. The four-engine data and integration tiers are the check suite's to
report.

### Stage 4: the two post-issuance backstops

Landed in `8d378a3d`. Stages 1 to 3 close issuance. This stage covers the grants that were already
issued, at the four places a grant is resolved against a session after the fact.

`RequireValidSession` refuses when the session a bearer token names belongs to another user. It sits
**before the settings lookup**, not merely before the lifetime check it guards: the comparison reads
two integers already in hand, so a cross-bound token is refused even on the path where absent
settings would otherwise produce a 500. `ValidateTokenRequest` makes the same comparison on three
arms, all in one function: the `Refresh` arm against the code's user, the `authorization_code` branch
at redemption, and the `Offline` arm. The offline one reads the session identifier **from the code**,
because only a `Refresh` token stores one of its own; for an `Offline` token the issuer puts the max
lifetime in that column, so reading the token row would have made the check a silent no-op.

All four reuse the refusal their site already had, so no new oracle appears, and each sits after the
caller has proven it holds the grant, so none becomes a probe for whether a session exists.

**A missing session row is accepted, deliberately, and that is the whole shape of the remedy.** The
last two arms were added by decision 8, which was escalated during this stage and answered on this
PR: option B, plus option A's documentation of what B cannot reach. Sessions are swept once they idle
out or reach their maximum lifetime, two hours and twenty-four by default, while an offline grant
runs for a seeded year. Refusing on absence would refuse every legitimate grant whose session simply
timed out, which for an offline grant is its steady state. So this reaches a pre-fix grant only while
its session survives, and both test triples carry a swept-session row asserting the accept.

The residual is documented in `concepts/user-sessions.mdx` so an operator can decide for themselves
at upgrade. Round 2 corrected that text on two counts worth repeating here, because both would have
misled a reader: **only `acr` was ever inherited** from the other user's session, since the password
handler set `AuthenticatedAt` and `BumpUserSession` wrote the authenticating user's own methods onto
the row, so `amr` and `auth_time` always described the sign-in that really happened; and **lowering
the offline maximum lifetime does not close an existing grant**, because rotation copies the deadline
the grant was issued with rather than recomputing it. Revoking the client's consent is what still
works once the session row has gone, since an offline grant is consent-checked on every refresh. That
one matters operationally: the earlier text would have let an administrator lower a setting and
conclude the legacy population was closed.

Tiers: unit 3081 pass, 0 fail. Data on four engines 1564 pass, 0 fail. Integration on sqlite 1218
pass, 0 fail, which this stage owes in a way the earlier ones did not, since the code branch now
performs a session lookup on **every** authorization code redemption, so a mistake there would break
interactive sign-in for everyone. Two review rounds and nine mutations, all caught, including
fail-open mutants at both new lookups and a reorder hoisting the offline lookup above the
max-lifetime gate. The four-engine data and integration tiers are the check suite's to report.

One thing the local tiers could not see: `gofmt` flagged a test file whose alignment round 1 had
disturbed. No tier runs `gofmt`, and CI lints in a separate job, so the stage would have gone red on
a check nothing local covers. Fixed before committing.

### Stage 5: the logout hint's session ownership

`4affa0f5`. `classifyIdTokenHint` read the hint's `sub` for presence and its `sid` for whether it
matched the browser's cookie, and never asked whether the two agreed. An ID token signed with
`sub = B` over `sid = S_A` was therefore *confirmed*, which is the one branch that skips the consent
page, and the teardown then deleted A's binding to that client, or A's whole session when that client
was the last on it. The gate now resolves the hint's subject and compares it with the owner of the
row its `sid` names, rejecting when they differ or when the subject names no user.

**Rejecting is the required outcome, not a cautious substitute for one.** RP-Initiated Logout 1.0
section 2 makes asking the End-User a MUST when "the supplied ID Token does not belong to the current
OP session with the RP and/or currently logged in End-User", and on this endpoint rejecting a hint is
precisely how the End-User gets asked.

Three choices beyond the comparison itself. The session lookup moved out of the expiry-tolerance
branch, since it used to run only for a hint whose `exp` had passed and an unexpired hint therefore
loaded no session to compare against; one lookup now serves both gates, which makes a database
failure on an unexpired hint a 500 where it was previously unreachable. A missing session row skips
the gate, the same deliberate limit stage 4 set: ownership cannot be decided without the row,
sessions are swept after hours while the artifacts naming them last far longer, so a swept session is
the ordinary state of a healthy older hint. And both lookups propagate a failure rather than
rejecting, matching the session lookup and deliberately not the client lookup above it, because a
lookup that decides whether the hint's session may be trusted must not be answered on the database's
health.

Nothing an RP legitimately sends starts being refused: after stages 1 to 3 a code's user and the
owner of the session it names always agree, and session ownership never changes afterwards, so only
an artifact issued before that rule existed can disagree. The evidence is
`TestAPIAccountLogoutRequest_Success_And_LogoutFlow_WithAndWithoutCookie`, the repository's only
legitimate producer of a hint, which still passes through the new gate.

Two review rounds. Round 2 found no production defect and two coverage gaps, both fixed here: the
classifier takes its session identifier from the middleware's context when a cookie exists and from
the hint's own `sid` when it does not, and only the first source had strict ownership coverage, so a
mutation running the gate only for a cookie-supplied identifier survived the entire suite. That is
the higher-risk source, because a caller with no browser session is exactly who would present a
pre-fix cross-bound hint directly. It now has both outcomes pinned strictly and the mutation is
caught. Six mutations at this stage, all caught. Unit 3088, data 1564 on four engines, integration
1218 on sqlite; the four-engine data and integration tiers are the check suite's to report.

### Stage 6: the end-to-end proof

`34d17392`. The capstone, and the only seam that observes the defect the way a user meets it: one new
integration file driving the whole ceremony from `GET /auth/authorize` to `POST /auth/token` and one
hop further, on a single `http.Client` whose cookie jar is real. **No production code changed**,
which is the point. Everything asserted here was already built by stages 1 to 5; this stage says so
without mocks.

Three routes leave `HandleAuthorizeGet` while another user's live session cookie stays in the
browser, and each gets a case with an assertion only that shape can make.

| Case | Its own assertion | What it pins |
|---|---|---|
| `prompt=login`, A's session and the client both at `level2_optional` | the chain passes through `/auth/level2` | the second-factor bypass, end to end |
| `id_token_hint` naming B, A at `level2_optional`, request asking `level1` | the code carries `level1` | the ACR taken against A's session |
| `max_age=0`, A and the client at `level1` | `/auth/authorize` answers `/auth/level1` | a third entry path |

The middle case is the shape that actually mints a cross-bound code today: the hint's subject is B, so
the issuance check is satisfied and nothing else refuses.

**`max_age=0` turned out to be a third entry path rather than a variant of the first.** `prompt` is
tested before the session is loaded while `max_age` is applied inside the validity check, so the
request leaves at the function's final `return`. The write-up's assumption that the two named entry
paths were exhaustive is therefore understated, and this case pins the third. It also demonstrates
that an *expired* foreign session is terminated rather than left to rot, since its row and its cookie
both outlive its validity.

All three then share one assertion block: the code is B's and names neither an empty session nor A's;
A's session row is gone; A holds no sessions and B holds exactly one, the one on the code; the code
redeems and the ID token, the access token and the refresh token all name B's session; both tokens
carry `amr` exactly `["pwd"]`; and a fourth ordinary request on the same jar authenticates as B. That
last hop is the one a reader most wants, because it says the browser now belongs to B rather than to a
ghost of A.

**The AMR assertions are provenance, not shape, and that took two rounds.** Round 1 found the fixture
could not tell whose methods reached B's grant: both users were password-only, so the assertion passed
whether the value came from B's ceremony or was copied off A's ambient session. The reviewer proved it
by mutation. Production was correct throughout, but the capstone could not have detected a regression
that broke it. A now enrols in OTP and really presents a live passcode through `/auth/otp` whenever
the client's level asks, so A's session records `pwd otp` while B's records `pwd`, and the forbidden
value exists in the database on the row the ceremony was cookied to. A `level1` client never asks A
for a second factor, so the `max_age` case cannot carry provenance and says so in the fixture; the
other two do.

Round 2 came back clean, having re-run the cases, re-derived every tree hash, and written its own
mutation against the fixture's precondition guard: changing only what A's session records, while A
still walks the OTP route, fails the case with `expected pwd otp, actual pwd`. That guard exists so a
later edit cannot silently return the suite to the round 1 state, and it is now demonstrated by
somebody other than its author.

The fixture builds its own confidential client with a real S256 pair rather than extending the shared
helpers, whose challenge is random letters no verifier matches, so their codes can never be redeemed.

Four mutations, all caught: copying the ambient session's methods into the ceremony, deleting the
level 1 ownership guard, taking the ACR from the ambient session instead of the bound one, and
disabling the termination gate. The 15 existing-session and 3 `id_token_hint` regression tests were
re-run by name, unedited and green. Unit 945, data 1144 on four engines, integration 914 on sqlite;
the four-engine data and integration tiers are the check suite's to report on this commit.

## The result

All six stages are on the branch. Issuance can no longer bind a grant to a session the ceremony does
not own; the four paths that resolve an already-issued grant against a session now refuse a foreign
one; RP-initiated logout no longer acts on a session its hint does not own; and the whole rule is
observed end to end through a real cookie jar rather than through mocks.

**The check suite is green on `34d17392`**, the tip of this branch with all six stages on it:
[run 31420740452](https://github.com/leodip/goiabada/actions/runs/31420740452), four minutes ten. Ten
jobs concluded `success`: the data and integration tiers on **sqlite, MySQL, PostgreSQL and SQL
Server**, unit on all three modules, and `Lint`, `Vulnerabilities` and `Versions`, none of which any
local tier covers. `Build validation` was skipped, its ordinary state on a branch whose diff touches
no build input.

## Testing

Each stage's section above gives its own counts. The shape across the run:

- **Locally, at every gate:** the unit tier, and the integration tier on sqlite. Data ran locally
  wherever a stage's evidence leaned on it. Nothing here adds a query, an interface method, a
  migration or a schema change, so no mock was regenerated and the data tier is owed by none of the
  stages; it was run anyway at four of them.
- **On four engines: the check suite, not this session.** Every stage's log entry names its commit and
  defers the MySQL, PostgreSQL and SQL Server tiers to CI. The run linked above is the one that
  answered for the whole branch.
- **Mutation testing at every gate: 54 mutations, all caught by the end.** Seven at stage 1, nine at
  stage 2, nineteen at stage 3, nine at stage 4, six at stage 5, four at stage 6. A good number were
  written by the reviewer rather than by the author, and the ones that survived when first run are
  what produced this run's substantive test findings: at stage 2 a termination whose error return went
  unchecked by any case, and at stage 6 `amr` assertions that could not tell whose methods had reached
  the grant. Each survivor was fixed and its mutation re-run until caught. That is the evidence these
  tests can fail, and it is why the stage 6 fixture makes one user really walk the OTP route.
- **Regression, by name:** the 15 tests in `authorize_existing_session_test.go` and the 3 in
  `authorize_id_token_hint_test.go`, unedited, at each stage that could disturb same-user SSO. Plus
  `TestAPIAccountLogoutRequest_Success_And_LogoutFlow_WithAndWithoutCookie` at stage 5, the
  repository's only legitimate producer of an `id_token_hint`, which still passes through the new
  gate.

Twelve review rounds, all ingested: one on the plan and eleven across the six stage gates, with the
two escalations they raised answered on this thread.

## Deviations worth knowing

**The run gained a stage.** Reviewing stage 4 as committed found one more site where an artifact
carrying both a user identity and a session identifier acts on that session without proving the two
belong together: RP-initiated logout confirms an `id_token_hint` without ever comparing the subject
it is signed over against the owner of the session its `sid` names. That is a departure from a MUST
in RP-Initiated Logout 1.0 section 2, which requires the End-User to be asked whenever a hint does
not belong to the currently logged in End-User. It was escalated as decision 9 and answered on this
thread: close the logout site, and leave the authorization-code reuse cascade alone. The reasoning
for leaving the cascade is in that comment, and it is that the cascade's writes are a strict subset
of the session termination this change already prescribes for a browser that has changed hands.

So stage 5 became that comparison and the end-to-end proof became stage 6, which put the last
production change before the capstone that proves the whole rule rather than after it. Both have now
landed.

**Stage 4 grew two comparisons and a documentation section**, for the same reason. Reviewing it found
two shapes descended from a cross-bound authorization code that the two backstops named up front do
not reach: an unredeemed code, and an auth-code-derived offline refresh token, which never consults
the session precisely because an offline grant must outlive it. Escalated as decision 8 and answered
on this thread: refuse while the evidence exists, and document the residual that no comparison can
reach. Both comparisons went into `token_validator.go`, the file the stage already changed, so it
stayed one stage and one commit.

**This change is wider than the issue proposed, in three places, all deliberate.**

- The issue puts the guard at `/auth/completed` alone. That stops the cross-bound code and leaves the
  second-factor bypass untouched, because step-up is decided a hop earlier at `/auth/level1completed`.
  Hence one predicate consulted at every site that reads the ambient session.
- The issue states that `StartNewUserSession` "already deletes other sessions matching the same device
  name, type, OS and IP, which is the correct outcome for a browser account switch". **It does not.**
  It queries sessions by the *new* user's id, so the previous user's row is never a candidate and
  would survive orphaned, valid, with its refresh tokens still working and still bumping it. That is
  why the change terminates it explicitly.
- Feeding `SetAcrLevel` the session the ceremony actually bound to also fixes a same-user defect the
  issue does not mention: an expired `level2_optional` session plus a fresh password-only login at a
  `level1` client used to produce a `level1` session and a token claiming `acr: level2_optional`.

**`max_age=0` turned out to be a third entry path**, found while writing the end-to-end proof at
stage 6. The write-up names two paths leaving `HandleAuthorizeGet` with a live foreign cookie, and
reasons they are exhaustive; `prompt` is tested before the session is loaded while `max_age` is
applied inside the validity check, so a `max_age=0` request leaves at the function's final `return`
instead. It is covered by its own end-to-end case. The guard's placement is what made this a
discovery rather than a defect: the predicate sits where the session is *used*, not where the ceremony
starts, so a path nobody had enumerated was already closed.

One process note, since it would otherwise look like luck: **no local tier runs `gofmt`**, and CI
lints in a separate job. Stage 4's round 1 fixture correction disturbed a struct literal's alignment,
every tier passed, and the stage would have landed and gone red on a check nothing local covers. It
was caught by running `gofmt -l` on the changed files before committing, which every later gate then
did too.

## Also on this thread

- **[Deferred decisions](https://github.com/leodip/goiabada/pull/196#issuecomment-5244572586)**, one
  reversible judgement call taken during the run rather than escalated, plus the two deliberate limits
  the escalated decisions settled. The shorter of the two comments, and the one to read first.
- **[Follow-ups](https://github.com/leodip/goiabada/pull/196#issuecomment-5244584588)**, three real
  defects found while doing this work and out of scope for it, each drafted as a ready-to-file issue
  with the `gh issue create` command that files it. Duplicate searches re-run against the tracker
  today. Filing them is yours, not the run's.

## Status

**Complete.** All six stages landed, reviewed and committed; the check suite is green on the final
commit; the deferred decisions and the drafted follow-ups are posted above. Ready for review.


leodip:
## Decision 8: what does this change owe a grant that was already cross-bound when it is deployed, once the evidence has been swept away?

Stage 4 added the two post-issuance ownership comparisons this change planned for. Reviewing them,
Codex found two more shapes descended from a cross-bound authorization code that neither comparison
reaches, and executed both against the current tree to prove they are accepted. I verified both by
reading the code. They are real.

This needs your call because the only complete remedy has a cost that falls on deployments that were
never exposed.

### What stage 4 does close

- A bearer access token carrying a `sid` that names a session belonging to somebody else is refused
  by `RequireValidSession`, before the settings lookup, so it is refused even on the path that would
  otherwise return 500.
- A session-bound refresh token (`typ: Refresh`) whose session belongs to somebody else is refused by
  `ValidateTokenRequest`. If that session has since been deleted, the pre-existing nil check refuses
  it anyway. This half is completely closed.

### What it does not close

Both were verified by execution, and I confirmed each by reading the code.

1. **The unredeemed authorization code.** The `authorization_code` branch of `ValidateTokenRequest`
   never loads the session at all, so a code minted for B carrying A's session identifier redeems
   normally. The window is small by construction: `authCodeExpirationInSeconds = 60`, so this only
   reaches codes minted in the last minute before the upgrade restarts the server.

2. **The auth-code-derived offline refresh token.** `case "Offline"` checks only
   `offline_access_max_lifetime` and deliberately never consults the session, because an offline
   grant is supposed to survive its session ending. That deliberate silence is what the cross-bound
   grant slips through. `RefreshTokenOfflineMaxLifetimeInSeconds` is seeded at 31536000, one year.

**What actually persists is a false authentication strength, not access to A's account.** The tokens
say `sub: B` throughout. But `createTokenInputFromCode` re-copies `Code.AcrLevel`,
`Code.AuthMethods` and `Code.AuthenticatedAt` on every rotation, and in a cross-user ceremony those
were stamped as `acr` taken from A's session with `amr` holding only what B proved. So B keeps
presenting, for example, `acr: urn:goiabada:level2_mandatory` with `amr: ["pwd"]`, to a relying party
that may well gate sensitive operations on exactly that claim. The offline access token is sid-less
by design, so stage 4's middleware check takes the generation-only arm and passes it; the ID token
keeps `sid` unconditionally, so it still carries A's session identifier.

### Why this is a question and not just a missing comparison

The comparison is only possible while A's session row still exists. It does not last:
`DeleteIdleSessions` and `DeleteExpiredSessions` in the background worker physically delete rows
after the idle timeout and the max lifetime, seeded at 2 hours and 24 hours.

So the grant outlives its own evidence by a factor of roughly four hundred. Within a day of the
pre-fix account switch, a cross-bound offline grant is byte-for-byte indistinguishable from a
legitimate offline grant whose session simply expired, which is the ordinary state of every healthy
offline grant. There is no signal in the schema that separates them.

Nothing in OpenID Connect Core decides this. The MUST NOT in 3.1.2.2, "The Authorization Server MUST
NOT reply with an ID Token or Access Token for a different user, even if they have an active session
with the Authorization Server", is about the token's subject, and the subject here is correctly B.
The spec defines what `acr` and `amr` mean but says nothing about remediating grants that were
stamped wrongly in the past. Every option below is permitted; the choice is genuinely yours.

### Options

**A. Ship stage 4 as it stands, and document the residual.**
No further code. A pre-fix cross-bound offline grant keeps asserting A-derived `acr`, `amr` and
`auth_time` until its offline maximum lifetime runs out. Documented so an operator can decide for
themselves whether to revoke offline grants at upgrade.

**B. Refuse while the evidence still exists.**
Add the ownership comparison to the `authorization_code` branch and to the `Offline` arm, refusing
only when the code's session identifier is non-empty and the row it names exists and has a different
owner. Never refuses a legitimate grant, because after this change no legitimate code can name a
foreign-owned session. Costs one extra session lookup per code redemption, which is every login, and
per offline refresh. Protects only operators who upgrade within hours of an exploit, since after that
the row is gone.

**C. Revoke the pre-fix auth-code-derived offline population at upgrade.**
Complete for the offline half, because it does not depend on the session row surviving. It also
breaks every legitimate offline integration in every deployment, each needing re-consent, in order to
remediate a population most deployments do not have.

**D. Targeted revocation on an incoherence signal.**
Revoke only pre-fix grants whose code claims `acr: urn:goiabada:level2_mandatory` while its
`auth_methods` lacks `otp`. That combination should be unreachable legitimately, because a
`level2_mandatory` target always ends in OTP verification or enrollment, so it reads as a fingerprint
of exactly this defect and it survives the sweep. Two caveats: it is a detector invented during this
review and would need verifying in its own right before anyone ran a revocation off it, and it does
not catch the `level2_optional` case, which is indistinguishable from a legitimate user who has no
OTP enrolled.

### Recommendation

**B, plus A's documentation of what B cannot reach.** Not C.

B is the only option with no false refusals, and the lookup it adds is the same one the `Refresh` arm
already performs. C's blast radius is every deployment, to remediate a population that needs a shared
browser, plus a client sending `prompt=login`, `max_age=0` or an `id_token_hint`, plus
`offline_access` on that client, plus someone completing the switch, all before the upgrade. D is the
most interesting option and the only one that reaches past the sweep, but running a revocation off a
fingerprint nobody has verified is the wrong way round in an authorization server; I would rather
file it as its own issue with the verification as its first step.

### What it costs to be wrong either way

**Wrong toward A or B:** a cross-bound offline grant keeps asserting a second factor its holder never
presented, for up to the offline maximum lifetime. The exposure is relying parties that authorize
sensitive operations from `acr` or `amr`. Any operator can close it without a code change by revoking
offline grants or shortening the offline maximum lifetime at upgrade, which is what A's documentation
would tell them.

**Wrong toward C:** every offline integration in every deployment stops working at upgrade and needs
re-consent, and the operators affected mostly had nothing to remediate.

### What happens next

- **A**: one paragraph of documentation, stage 4 closes, the run continues to stage 5.
- **B**: a further stage for the two comparisons and their tests, then stage 5.
- **C or D**: a further stage carrying a migration, which is the only part of this change that would
  touch the schema.

Stage 4's code is written, reviewed and passing, and it is not committed, because the answer here may
land in the same two functions. Reply with the letter and any reasoning you want recorded.

leodip:
B, plus A's documentation of what B cannot reach.

leodip:
## Decision 8 applied: B, plus A's documentation of what B cannot reach

Built as answered. Both comparisons landed, plus the documentation of the residual neither reaches.

### The two comparisons

Both in `ValidateTokenRequest` (`src/core/validators/token_validator.go`), each refusing only when
the code's session identifier is non-empty, the row it names still exists, and that row belongs to
somebody else.

- **Code redemption**, in the `authorization_code` branch, after the revoked-code check and before
  the success return. Refuses with `"Code is invalid."`, the wording a revoked code and a superseded
  generation already share. It sits below the `wasReused` return and below client authentication and
  PKCE deliberately: the `!wasReused` block runs before the requester has proven anything, so
  nothing added there may speak about a grant's state (#137).
- **Offline refresh**, in `case "Offline"`, after the max-lifetime check so an already-expired token
  costs no lookup. Refuses with the shared session message. It reads the session identifier from the
  **code**, not the token row: only a `Refresh` token stores one of its own, since the issuer puts
  the max lifetime in that column for an `Offline` token, so reading the token row there would have
  been a comparison that always passed.

A missing session row accepts on both. That is what bounds the remedy and it is also what keeps it
from refusing anything legitimate: a swept session is the ordinary steady state of an offline grant.

### The documentation

A new section, `Grants issued before this behaviour existed`, under the account-switch part of
`concepts/user-sessions.mdx`. It says plainly that the comparison needs the previous session to
still exist, that sessions go after two hours idle or twenty-four of lifetime while an offline grant
runs for a year, and that what such a grant carries is a false `acr`, `amr` and `auth_time` rather
than access to the other account. It tells an operator the way to close it at upgrade is to revoke
offline grants or lower the offline maximum lifetime, and that deployments without shared browsers
have nothing to do.

### Option D became a follow-up

The `level2_mandatory` with no `otp` incoherence detector is the only idea that reaches past the
session sweep, and it was invented during a review. It is drafted as an issue whose first step is
verifying the fingerprint against every path that writes a code, before anybody runs a revocation
off it. It will be posted with the rest of the follow-ups when the run closes.

### Tests and tiers

Six new subtests, three per comparison: accept, swept and refuse. The swept row is the one that
matters, since it pins the deliberate hole rather than leaving it to be closed by a well-meaning
later edit.

Unit 3078 pass 0 fail, data 1564 pass 0 fail on sqlite, mysql, postgres and mssql, integration 1218
pass 0 fail on sqlite. Six mutations, all caught, including one per comparison that makes it refuse
a *missing* row rather than neutralising it, which is what proves the swept-session cases are
load-bearing. The four-engine integration result is the check suite's to report once this commits.

### One deviation from what the escalation predicted

The escalation said option B meant a further stage. It landed inside stage 4 instead: that stage was
uncommitted precisely because the answer could land in `token_validator.go`, and both new
comparisons sit in the same function as the one it already added. One stage, one review gate, one
commit. Nothing about what was built changed.

Stage 4 now goes to its round 2 review before it commits.

leodip:
## Decision 9: does the session-ownership comparison reach post-issued artifacts outside the token endpoint?

Stages 1 to 4 are committed and CI is green on `8d378a3d`. Stage 5, the end-to-end integration
proof, has not begun. The review of stage 4 as committed found a family this change has only
partly closed, and closing the rest means opening a file no part of this work has touched. That is
a scope call, so it is yours.

### What this change established

An authorization ceremony may only bind to a session belonging to the user it authenticated, and
an artifact presented later that carries both a user identity and a session identifier is refused
when the two disagree. Stage 4 put that second half on four paths: bearer tokens at the API, the
`Refresh` arm and the `Offline` arm of the token endpoint, and authorization-code redemption. In
every case a missing session row is accepted deliberately, because a swept session is the ordinary
state of a healthy older grant, and a database failure is propagated rather than read as absence.

None of it can be reached by anything minted after the fix. It exists for artifacts issued
*before* the upgrade, on a shared browser, in the window where one person's session was still
cookied while another signed in.

### What the review found still open

Two more sites where such an artifact acts on the session it names without proving it belongs to
its user. I verified both against the code rather than taking them on trust.

**1. A replayed authorization code.** `ValidateTokenRequest` returns its code-reuse error above the
ownership lookup, so a pre-fix code whose user is B and whose session identifier is A's still
drives the reuse containment cascade: every refresh token keyed to A's session is revoked and A's
session row is deleted.

**2. RP-initiated logout.** `classifyIdTokenHint` in `handler_account_logout.go` checks that the
hint's `sub` is present and never compares it with the owner of the row its `sid` names. A pre-fix
ID token signed with `sub = B` and `sid = A's session` is therefore confirmed, skips the consent
page, and the teardown deletes A's binding to that client, or A's whole session when that client
was the last one on it.

### Why I think the first one is not a defect

The reviewer called both a cross-user denial of service. For the replayed code I think that is
wrong, and the argument comes from this change's own behaviour.

When a browser changes hands, this change terminates the previous session through the same helper
an explicit "end session" uses: it revokes that session's unredeemed codes, revokes its refresh
tokens, and deletes the row. The reuse cascade does the last two of those three and nothing else,
keyed on the same identifier. **It is a strict subset of the remedy this change already prescribes
for that exact session**, and a cross-bound code can only exist because the handover happened. So
replaying one delivers the fix late rather than harming a bystander, and it costs the replayer
their own tokens too, since those are keyed to the same session.

The proposed remedy would also make things worse. RFC 6749 section 4.1.2: "If an authorization code
is used more than once, the authorization server MUST deny the request and SHOULD revoke (when
possible) all tokens previously issued based on that authorization code." Suppressing the cascade
for these codes leaves alive the tokens issued from the very code being replayed, in order to
preserve a session this change wants terminated. I would not do it.

### Why the second one is different

Its blast radius is the same session and by the same argument is not a bystander harm either. What
makes it different is that it departs from a MUST, and a MUST is not settled by impact.

OpenID Connect RP-Initiated Logout 1.0, section 2, quoted rather than summarised:

> At the Logout Endpoint, the OP SHOULD ask the End-User whether to log out of the OP as well.
> Furthermore, the OP MUST ask the End-User this question if an `id_token_hint` was not provided
> or if the supplied ID Token does not belong to the current OP session with the RP and/or
> currently logged in End-User. If the End-User says "yes", then the OP MUST log out the End-User.

And, in the same section:

> If the ID Token's `sid` claim does not correspond to the RP's current session or a recent session
> at the OP, the OP SHOULD treat the logout request as suspect, and MAY decline to act upon it.

The first sentence is already quoted in the logout handler, at the step that decides whether to
ask. A browser cookied to A, plus a hint signed with B as its subject, is exactly "does not belong
to the ... currently logged in End-User", and the confirmed branch is the one that does not ask.

To be clear about the three different questions: the spec **requires** asking in this case, it
**permits** treating the request as suspect and declining outright, and what it **exists to
enable** is an RP ending its own user's session without a round trip through a consent page. Only
the first of those is at stake, and none of the options below touches the normal case: after this
change a code's user and its session's owner always agree, so a mismatch is unreachable except
from an artifact issued before the upgrade.

### The options

**A. Close both sites, as the reviewer proposed.** Move the code-ownership lookup above the reuse
return and suppress the cascade on a live mismatch, and add the subject comparison to the logout
classifier.
*Cost:* the reuse half is the change I argue against above, and it trades a SHOULD in RFC 6749 for
a session this change wants gone. Roughly a stage of work across three files.

**B. Close the logout site only.** One comparison in `classifyIdTokenHint`: when the row named by
`sid` exists and does not belong to the signed `sub`, reject the hint. Rejecting is what makes the
endpoint ask the End-User, which is what the MUST requires, and the endpoint already does exactly
that for every other unconfirmable hint. A missing row keeps being accepted, matching the rule the
other four comparisons use, and a lookup failure keeps propagating as a 500, which is this
endpoint's existing rule for that same lookup. Leave the reuse cascade alone.
*Cost:* one extra session lookup per confirmed hint, which the teardown then repeats immediately.
About a stage: one production comparison, table rows in `handler_account_logout_test.go` covering
matching, foreign, absent and lookup-error, and a sentence in the logout documentation. Nothing
legitimate is refused.

**C. Neither. Document the residual and file a follow-up issue.**
*Cost:* ships with a known departure from that MUST for the pre-fix population, on a path a
reviewer has already written down. Cheapest now, and the one that is least comfortable to explain
later.

### Recommendation: B

The conformance gap is real and its remedy is small, bounded and refuses nothing that legitimately
happens. The reuse cascade is not a defect and I would rather not change it on this evidence.

**What it costs to be wrong.** If you pick B and I have misjudged the reuse cascade, the replayed
code keeps behaving as it does today and the follow-up is a separate, contained change. If you pick
C and the logout gap matters, it is a conformance departure discovered by someone else later, in an
endpoint that was rewritten for exactly this class of problem once already. If you pick A, the
reuse half is reversible but it removes a revocation the RFC asks for, which is the kind of thing
that is only noticed when it is needed.

Reply with **A**, **B** or **C**. Stage 5 is otherwise ready to run and does not depend on this,
but the answer may add a stage, so the run is paused here.

leodip:
B

leodip:
## Deferred decisions

One judgement call was taken during the run rather than escalated. It is reversible, and this is what
it assumed and what undoing it costs.

### 1. `cross_user_session_replaced` is written before the replacement session exists

Raised by the stage 2 round 1 review. The event's name says "replaced", but it is emitted immediately
after `TerminateUserSessionTx` commits and **before** `StartNewUserSession` runs, so what it actually
attests is the handover and the ending, not a completed replacement.

**The assumption taken.** Decision 6 already splits the two halves: this event answers *why* a session
ended, and `started_new_user_session` is what says a replacement exists. So the placement is the one
§4's design block specifies. If session creation then fails, the ceremony returns 500 and the audit
trail holds `cross_user_session_replaced`, `deleted_user_session` and `terminated_user_session` with
no `started_new_user_session` after them. An operator reads that absence as a handover that did not
complete.

**Why not the exact alternative.** Emitting it only after `StartNewUserSession` succeeds would make
the name precise on every path, at the price of recording that same failure as a termination with no
actor and no reason. `loggedInUser` is deliberately empty on this path, because the cookie at that
instant still names the party being terminated rather than the actor, so the reason event is the only
record naming who ended the session and why. Losing it on the failure path is the worse forensic
outcome, which is why the reversible option was taken instead.

**What reversing costs.** Moving one `auditLogger.Log` call below the `StartNewUserSession` error
check in `HandleAuthCompletedGet`, plus the ordering row in the valid-foreign subtest and two
sentences in `constants.go`'s doc comment. No interface, no schema, no payload and no event name
changes either way.

**Fixed regardless of the deferral:** the overclaim the review actually found. The constant's doc
comment used to say the session "was ended and replaced"; it now states what the event does and does
not attest.

---

Two other limits are deliberate rather than deferred, and are recorded in the PR body and in the docs
rather than here, because both are the answer to an escalated decision the user settled on this
thread:

- **A missing session row is accepted** at all four post-issuance comparisons (decision 8, answered
  "B, plus A's documentation of what B cannot reach") and at the logout gate (decision 9, answered
  "B"). Ownership cannot be decided without the row, and a swept session is the ordinary state of a
  healthy offline grant, so refusing on absence would refuse legitimate traffic. The residual is
  documented in `concepts/user-sessions.mdx` under **Grants issued before this behaviour existed**,
  and the third follow-up in the next comment is the investigation that would close it.
- **The authorization-code reuse cascade is left as it is** (decision 9). Replaying a cross-bound code
  revokes the previous user's refresh tokens and deletes their session row, but those writes are a
  strict subset of the `TerminateUserSessionTx` this change already prescribes for a browser that has
  changed hands, so the cascade delivers this change's own remedy late rather than harming a
  bystander. Suppressing it would also leave alive the tokens issued from the very code being
  replayed, which RFC 6749 section 4.1.2 addresses directly.

leodip:
## Follow-ups, drafted and ready to file

Three real defects found while doing this work, none of them this change's to fix. Each is written to
stand on its own, so filing it needs no context from this PR. The duplicate searches were re-run today
against the current tracker (43 open issues, up from 40 when they were drafted); nothing new overlaps.

Filing is yours, not the run's.

---

### 1. Security: an implicit ceremony receives tokens for a session that was already ended

**Labels:** `bug`, `go`, `security`

Found during the plan review, narrowed to what survives stage 3, then widened to a second shape by the
stage 3 round 1 review. Stage 3 moved the ambient-session gate above the `response_type` dispatch in
`HandleIssueGet`, so all four issuance families now refuse a session belonging to another user and a
non-empty identifier whose row has vanished. One shape is deliberately carved out, and that carve-out
is what this is about: `isImplicitFlow && sessionIdentifier == ""` still issues.

That is exactly the shape a terminated session arrives in, since `MiddlewareSessionIdentifier`
publishes an identifier only when the row exists. Closing it inside #133 would have changed implicit
behaviour in a case with no cross-user element in it, which is why the carve-out was drawn rather than
absorbed: both shapes ask what a ceremony owes when the session backing it vanished mid-flight, which
is #129's question rather than this one.

Not a duplicate: #139 is the code-insert race against a termination's COMMIT, #140 is `amr` inherited
across a session boundary for the same user.

<details><summary>Body</summary>

#129 added a gate at `/auth/issue` refusing to mint a grant when the session backing the ceremony has
been ended, and #133 moved it above the `response_type` dispatch so it covers the implicit branches
too. Two cases still get through it, both ending with an implicit ceremony holding tokens for a
session that is gone.

**1. The ceremony arrives with no session identifier.** The gate exempts
`isImplicitFlow && sessionIdentifier == ""` deliberately, and that exemption covers the
terminated-session case: `MiddlewareSessionIdentifier` only publishes an identifier when the row
exists, so an ended session leaves an empty one. The issuer omits the `sid` claim for an empty
identifier (`if len(input.SessionIdentifier) > 0` at both the access-token and ID-token claim sites),
and `RequireValidSession` treats a user token with no `sid` as an offline or ROPC grant, validating it
against the token's generation claim instead of a session. A termination does not change that claim,
so the access token and ID token stay valid until they expire. Implicit issues no refresh token, so
the exposure ends there.

**2. The session is ended after the gate's lookup.** `HandleIssueGet` performs exactly one
`GetUserSessionBySessionIdentifier`; `handleImplicitFlow` receives only the identifier string and
never loads the row again, so nothing between the gate and `GenerateTokenResponseForImplicit`
re-asks. A termination committing in that window leaves the response carrying freshly signed tokens
with the ended `sid`. The code flow closes the identical window with `RevokeCodeIfSessionGone` right
after `CreateAuthCode` (#129 decision 12), which works because a code is a row that can be marked
after issuance. A signed token cannot be. `RequireValidSession` does reject these on Goiabada's own
endpoints, since the `sid` no longer resolves, so what is left is a third-party resource server
validating the signature against `/certs`: it cannot consult the session table and accepts the token
until expiry.

Fix case 1 by applying the same liveness rule to the implicit branch when the identifier is empty,
with #129 decision 16's answer for `prompt=none` (`login_required`, which `redirToClientWithError`
already returns in a fragment for implicit response types) and a level 1 restart otherwise. Decide
first whether an implicit ceremony must have a session at all: the gate as written would refuse every
implicit request reaching `/auth/issue` without one, which is a wider question than the terminated
case and the reason #133 left it alone. The unit cases live in
`TestHandleIssueGet_ForeignAmbientSession`, and every existing implicit subtest in
`handler_auth_issue_test.go` builds its request without an identifier, so those are the ones that
would move.

Case 2 has no revoke-after-the-fact remedy, so the options are narrowing the window (re-check liveness
inside `handleImplicitFlow` immediately before signing, which shrinks it rather than closing it) or
shortening what an escaped token buys (the implicit access-token lifetime). Worth sizing against #139,
which records the same read-then-write shape for codes as an accepted residual; if that residual is
acceptable there, the question here is whether a third party's inability to see the termination
changes the answer.

</details>

---

### 2. `StartNewUserSession` commits the new session before the work that can still fail

**Labels:** `bug`, `go`

Raised by the stage 2 round 2 review. Out of scope for #133, which is about a ceremony binding to a
session belonging to a different user: this is the same on every path that creates a session, and it
predates the change.

Not a duplicate: #160 is a partial commit in `issueAuthCode` and #139 is a code insert racing a
termination's COMMIT. Nothing open names `StartNewUserSession`.

<details><summary>Body</summary>

`UserSessionManager.StartNewUserSession` (`src/core/user/usersession_manager.go`) opens a transaction,
inserts the `user_sessions` row and its `user_session_clients` rows, and commits. Only after that
commit does it read the user's other sessions, delete the ones matching this device and IP, load the
cookie session, and write the new identifier into it. Each of those four steps returns `nil, err` on
failure and none of them undoes the commit, so a failure after that point leaves a session row that no
browser holds a cookie for, and it may already have deleted the session the browser was using.

Verified by reading the function: `CommitTransaction` is followed by `GetUserSessionsByUserId`, a
`DeleteUserSession` loop, `sessionStore.Get` and `sessionStore.Save`, with an early `return nil, err`
at each. The caller sees only the error and cannot tell that a session was created, so it returns 500
and the row stays.

The visible effect is an orphaned session row that no cookie points at, counted by the admin console's
session list and holding its own idle timeout and maximum lifetime, plus the possible loss of the
session the browser did have. Nothing is issued against the orphan, because the ceremony stops, so
this is a consistency and hygiene defect rather than an access-control one.

Decide between compensating cleanup (delete the row before returning the error) and moving the commit
boundary below the sibling sweep, then pin it with tests that assert database state through the real
manager rather than a nil return: today's coverage checks the returned error only, so the leftover row
is invisible to it.

</details>

---

### 3. Security: no way to find a grant that was cross-bound before #133, once its session row is gone

**Labels:** `security`, `enhancement`, `go`

This is option D of decision 8, which the user rejected for #133 on the grounds that running a
revocation off a detector invented during a review is the wrong way round in an authorization server.
Drafted here with its verification as step one. #133 chose option B plus documentation, so the
residual is written up in `concepts/user-sessions.mdx` for an operator to close by hand.

Not a duplicate: #140 is `amr` inherited across a session boundary for the same user, #135 is a
client-scoped disconnect capability.

<details><summary>Body</summary>

After #133 an authorization ceremony can no longer bind to a session belonging to a different user,
and four validation paths refuse a grant whose session belongs to somebody else. Every one of those
comparisons needs the session row to still exist. Sessions are deleted after their idle timeout and
maximum lifetime (defaults 2h and 24h) while an auth-code-derived offline grant runs to
`RefreshTokenOfflineMaxLifetimeInSeconds` (default 31536000, one year), so a grant cross-bound shortly
before an upgrade can outlive the only evidence of it by roughly four hundred to one and keep rotating
for up to a year. After that it is byte-for-byte indistinguishable from a legitimate grant whose
session merely expired.

Its tokens name the correct user. What is wrong is the authentication strength they assert:
`createTokenInputFromCode` re-copies `Code.AcrLevel`, `Code.AuthMethods` and `Code.AuthenticatedAt` on
every rotation, so the grant keeps presenting an `acr` taken from the other user's session while its
`amr` holds only what its own holder proved. It is not access to the other account.

Investigate whether a durable incoherence signal exists in the `codes` table. The candidate, proposed
during the review that raised this, is `acr_level = 'urn:goiabada:level2_mandatory'` with
`auth_methods` not containing `otp`, which should be unreachable through a legitimate ceremony because
a `level2_mandatory` target always ends in OTP verification or enrollment. Nothing has verified that
claim, so step one is establishing whether any legitimate path can produce that pair, over the real
`codes` table across the supported engines. Only then is a revocation off it defensible. Note it does
not catch the `level2_optional` case at all, which is indistinguishable from a user with no OTP
enrolled.

Then decide whether Goiabada should offer an operator-run revocation over the matching population, or
simply a report.

</details>

---

### Filing them

```bash
  --title "Security: an implicit ceremony receives tokens for a session that was already ended" \
  --label bug --label go --label security \
  --body "$(cat <<'ISSUE_BODY'
#129 added a gate at `/auth/issue` refusing to mint a grant when the session backing the ceremony has been ended, and #133 moved it above the `response_type` dispatch so it covers the implicit branches too. Two cases still get through it, both ending with an implicit ceremony holding tokens for a session that is gone.

**1. The ceremony arrives with no session identifier.** The gate exempts `isImplicitFlow && sessionIdentifier == ""` deliberately, and that exemption covers the terminated-session case: `MiddlewareSessionIdentifier` only publishes an identifier when the row exists, so an ended session leaves an empty one. The issuer omits the `sid` claim for an empty identifier (`if len(input.SessionIdentifier) > 0` at both the access-token and ID-token claim sites), and `RequireValidSession` treats a user token with no `sid` as an offline or ROPC grant, validating it against the token's generation claim instead of a session. A termination does not change that claim, so the access token and ID token stay valid until they expire. Implicit issues no refresh token, so the exposure ends there.

**2. The session is ended after the gate's lookup.** `HandleIssueGet` performs exactly one `GetUserSessionBySessionIdentifier`; `handleImplicitFlow` receives only the identifier string and never loads the row again, so nothing between the gate and `GenerateTokenResponseForImplicit` re-asks. A termination committing in that window leaves the response carrying freshly signed tokens with the ended `sid`. The code flow closes the identical window with `RevokeCodeIfSessionGone` right after `CreateAuthCode` (#129 decision 12), which works because a code is a row that can be marked after issuance. A signed token cannot be. `RequireValidSession` does reject these on Goiabada's own endpoints, since the `sid` no longer resolves, so what is left is a third-party resource server validating the signature against `/certs`: it cannot consult the session table and accepts the token until expiry.

Fix case 1 by applying the same liveness rule to the implicit branch when the identifier is empty, with #129 decision 16's answer for `prompt=none` (`login_required`, which `redirToClientWithError` already returns in a fragment for implicit response types) and a level 1 restart otherwise. Decide first whether an implicit ceremony must have a session at all: the gate as written would refuse every implicit request reaching `/auth/issue` without one, which is a wider question than the terminated case and the reason #133 left it alone. The unit cases live in `TestHandleIssueGet_ForeignAmbientSession`, and every existing implicit subtest in `handler_auth_issue_test.go` builds its request without an identifier, so those are the ones that would move.

Case 2 has no revoke-after-the-fact remedy, so the options are narrowing the window (re-check liveness inside `handleImplicitFlow` immediately before signing, which shrinks it rather than closing it) or shortening what an escaped token buys (the implicit access-token lifetime). Worth sizing against #139, which records the same read-then-write shape for codes as an accepted residual; if that residual is acceptable there, the question here is whether a third party's inability to see the termination changes the answer.

Found while working #133 (PR #196).
ISSUE_BODY
)"
```

```bash
  --title "StartNewUserSession commits the new session before the work that can still fail, so a late error leaves a row nobody can reach" \
  --label bug --label go \
  --body "$(cat <<'ISSUE_BODY'
`UserSessionManager.StartNewUserSession` (`src/core/user/usersession_manager.go`) opens a transaction, inserts the `user_sessions` row and its `user_session_clients` rows, and commits. Only after that commit does it read the user's other sessions, delete the ones matching this device and IP, load the cookie session, and write the new identifier into it. Each of those four steps returns `nil, err` on failure and none of them undoes the commit, so a failure after that point leaves a session row that no browser holds a cookie for, and it may already have deleted the session the browser was using.

Verified by reading the function: `CommitTransaction` is followed by `GetUserSessionsByUserId`, a `DeleteUserSession` loop, `sessionStore.Get` and `sessionStore.Save`, with an early `return nil, err` at each. The caller sees only the error and cannot tell that a session was created, so it returns 500 and the row stays.

The visible effect is an orphaned session row that no cookie points at, counted by the admin console's session list and holding its own idle timeout and maximum lifetime, plus the possible loss of the session the browser did have. Nothing is issued against the orphan, because the ceremony stops, so this is a consistency and hygiene defect rather than an access-control one.

Decide between compensating cleanup (delete the row before returning the error) and moving the commit boundary below the sibling sweep, then pin it with tests that assert database state through the real manager rather than a nil return: today's coverage checks the returned error only, so the leftover row is invisible to it.

Found while working #133 (PR #196). Related in shape but not the same site: #160 (partial commit in `issueAuthCode`).
ISSUE_BODY
)"
```

```bash
  --title "Security: no way to find a grant that was cross-bound before #133, once its session row is gone" \
  --label security --label enhancement --label go \
  --body "$(cat <<'ISSUE_BODY'
After #133 an authorization ceremony can no longer bind to a session belonging to a different user, and four validation paths refuse a grant whose session belongs to somebody else. Every one of those comparisons needs the session row to still exist. Sessions are deleted after their idle timeout and maximum lifetime (defaults 2h and 24h) while an auth-code-derived offline grant runs to `RefreshTokenOfflineMaxLifetimeInSeconds` (default 31536000, one year), so a grant cross-bound shortly before an upgrade can outlive the only evidence of it by roughly four hundred to one and keep rotating for up to a year. After that it is byte-for-byte indistinguishable from a legitimate grant whose session merely expired.

Its tokens name the correct user. What is wrong is the authentication strength they assert: `createTokenInputFromCode` re-copies `Code.AcrLevel`, `Code.AuthMethods` and `Code.AuthenticatedAt` on every rotation, so the grant keeps presenting an `acr` taken from the other user's session while its `amr` holds only what its own holder proved. It is not access to the other account.

Investigate whether a durable incoherence signal exists in the `codes` table. The candidate, proposed during the review that raised this, is `acr_level = 'urn:goiabada:level2_mandatory'` with `auth_methods` not containing `otp`, which should be unreachable through a legitimate ceremony because a `level2_mandatory` target always ends in OTP verification or enrollment. Nothing has verified that claim, so step one is establishing whether any legitimate path can produce that pair, over the real `codes` table across the supported engines. Only then is a revocation off it defensible. Note it does not catch the `level2_optional` case at all, which is indistinguishable from a user with no OTP enrolled.

Then decide whether Goiabada should offer an operator-run revocation over the matching population, or simply a report.

Raised during #133 (PR #196) as option D of its session-ownership decision, and deliberately not taken there: running a revocation off a detector invented during a review is the wrong way round in an authorization server. #133 documents the residual in `concepts/user-sessions.mdx` so an operator can close it by hand.
ISSUE_BODY
)"
```

PR #200, opened 2026-08-10
fix(security): stop writing credentials to the request log, and get the codes out of the URL

### LINKED ISSUE #159: Security: enabled HTTP request logging writes complete id_token_hint JWTs to the log

leodip:
## Issue 159: credentials in URLs, from the request log to the emailed link

Closes #159.

When HTTP request logging is on, both servers hand every request to chi's `middleware.Logger`, which
formats `r.RequestURI` — the path **and the raw query string**. So a default install writes complete
signed ID tokens, account activation codes and password reset codes into its own log, in the clear.

That is not only a privacy leak. A leaked `id_token_hint` replayed at `GET /auth/logout` from
anywhere, with no cookie at all, reaches the branch that tears down the session without asking the
End-User: with no session identifier in context the hint's own `sid` names the session to act on, the
ownership gate passes by construction for a token this server signed, and an **expired** hint is
still accepted while its session row lives. The log entry is a working teardown credential for as
long as the session survives.

This is the default configuration, not a test-only setting: 6 of the 7 shipped compose files set
`LOG_HTTP_REQUESTS=true`, and `goiabada-setup` writes it into both the generated compose file and
the generated environment script.

### The rule this establishes

**A query parameter's value reaches the log only if its name is on a fixed allowlist.** Everything
else becomes `[redacted]`, names always survive so a request stays diagnosable, and there is no path
from a malformed query string to raw output.

An allowlist rather than a denylist because this issue *is* the recurrence a denylist produces:
`id_token_hint` was not omitted from a list by accident, it was never considered. A future endpoint
adding a sensitive query parameter is safe by default under an allowlist and leaks by default under
a denylist.

**`state` and `nonce` are redacted too**, which the first draft of the allowlist got wrong. The
argument for keeping them was that neither is redeemable at this server, which is true, and that a
log entry is post-hoc, which is not: the line is written as the authorize handler returns, and both
values stay live for the rest of the flow. `state` is the client's CSRF binding, and PKCE is not a
guaranteed substitute because `Client.PKCERequired` is a per-client override, so a client can run
with PKCE off and have `state` as its only protection. `nonce` is the OIDC replay binding on the ID
token. `code_challenge` stays on the list: only `S256` is accepted, so it is a digest and the
verifier never leaves the token POST body.

Three properties, each verified against a real `net/http` server rather than reasoned about:

- **Redact, then cap.** A 1 MB `id_token_hint` collapses to a short line. Capping first would write
  the first 4096 bytes of the token.
- **Emit names exactly as they arrived, decode only to compare.** `r.URL.Path` is percent-decoded
  and carries whatever the caller encoded: `GET /auth%0d%0a2026/01/01%2000:00:00%20FAKE` yields a
  real CRLF, and `%1b` yields an ESC that an operator's terminal acts on. Logging the decoded form
  would trade a token leak for log forging. The escaped forms are safe by construction, because
  `net/http` answers 400 to a literal ESC, TAB, CR or LF in the request line.
- **Walk the query textually, never parse it whole.** `url.ParseQuery` returns an error *and* a map
  holding the token, so any "on error, log the raw query" fallback re-leaks it; and
  `id_token_hint=a;b` defeats it entirely while a per-field walk still redacts.

### Two more things fixed in the same pass

**A panicking request is logged with the status the client received.** Both servers register
`middleware.Recoverer` before the request logger, the opposite of what chi documents, so a panic
unwinds through the logger's deferred write before Recoverer turns it into a 500. The client always
got a correct 500, but the log line read `- 000 0B`, so an operator grepping the request log for 5xx
found nothing for the requests that actually crashed.

**Activation and reset codes leave the URL on the first hop.** Both flows answer the landing GET with
a 302 to a clean URL, carrying the code in the encrypted, `HttpOnly`, `SameSite=Lax` session instead
of the query string. A 302 does not create its own history entry, so once the redirect completes no
URL the browser retains holds a code and no page ever has one as its document URL. The reset code was
in the query string on the **POST** as well, not only the landing GET, because the form's empty
`action` posts back to the current URL. Old links keep working: the link shape does not change and
the landing handler still accepts `?email=&code=`.

What this does not close, and cannot: the code is in the email regardless, and a reverse proxy or CDN
in front of Goiabada still logs the landing request. Both codes remain 5-minute and one-time, which
is what keeps that residual small.

**Status:** in progress. No stage has landed yet. The plan has been written and reviewed: nine
findings, seven resolved into the plan (including the `state`/`nonce` narrowing above, the 4096-byte
cap actually bounding at 4096, and a skip-list predicate that matched a different path
representation from the one chi routes on). One design question is open and will be posted here as a
comment when stage 4 starts: moving the reset code out of the URL also removes the value the reset
rate limiter keys on, and leaves the reset form without a binding to the ceremony that produced it.
Stages 1 to 3 do not depend on the answer.

**Waiting on decision 16, posted as a comment below.** Stage 1 is written and has been through four
review rounds, three of which found and fixed a real defect. The fourth found one I cannot settle on
my own: the logged path is unbounded, so an unauthenticated caller can still put about a megabyte,
or 1.84 MB using raw UTF-8, into the log per request, and the obvious remedy contradicts an agreed
decision that the path is never truncated. Stage 1's code stays uncommitted until that is answered.

### Stage 1: The redactor and the request logger, in core
_Not started._

### Stage 2: Mount it in both servers, and move the logger ahead of Recoverer
_Not started._

### Stage 3: The API debug path uses the same redactor
_Not started._

### Stage 4: The reset flow carries its code in the session, not the URL
_Not started._

### Stage 5: Activation carries its outcome in the session, not the URL
_Not started._


leodip:
## Decision 16: does the request log line get one end-to-end size bound, including the path?

Stage 1 of this change replaces chi's `middleware.Logger` with our own, which redacts credential
values out of the query string. It is written and reviewed but **not committed**, because the fourth
review round found something that needs your call rather than mine.

### The problem

We agreed earlier that the redacted query is capped at 4096 bytes, and the reason recorded for that
cap was to stop an unauthenticated caller writing a megabyte per request into the operator's log. The
same paragraph also says, in those words, **"the path is never truncated"**.

Those two cannot both hold. The path is the other half of the same logged target, and nothing bounds
it, so the exact ability the cap exists to remove is still reachable, just through the path instead of
the query.

Measured on a real `net/http` listener and then end to end through the middleware itself:

| Sent | Server accepted | What chi used to log | What this logs | Resulting log line |
|---|---|---|---|---|
| a 921,601-byte ASCII path | **204** | 921,601 B | 921,601 B | **921,675 B** |
| a 614,401-byte raw UTF-8 path | **204** | 614,401 B | 1,843,201 B | **1,843,275 B** |
| the same UTF-8 path under `/static/` | **204** | — | 1,843,208 B built | none, skipped |

For comparison, every other caller-controlled field on that line is now bounded, and the test that
enforces it uses a 1,280-byte ceiling for the whole line.

No client, user, token or valid route is needed. HTTP request logging is on in 6 of the 6 shipped
compose files, so this is the default install.

Two details make it more than "one more field to cap":

- **The UTF-8 row is the one place the new logger is worse than chi's.** chi logged `r.RequestURI`,
  the bytes as they arrived, so its line could never exceed the request it was handed. We log
  `r.URL.EscapedPath()`, which is *computed*: each non-ASCII byte becomes a three-character percent
  escape, so 600 KiB on the wire becomes 1.8 MB in the log. The output is larger than the request the
  server accepted.
- **Truncating the logged text does not close all of it.** The skip predicate for `/health`,
  `/static/` and `/favicon.ico` matches on the escaped path, so the full escaped string is built
  *before* anything decides not to log the request. The third row above is that case: 1.84 MB of work
  to produce no log line at all.

### Options

**A. Bound the path like every other field on the line.** Truncate the logged target at 320 bytes,
with the same `[truncated, N bytes]` note the other fields use, which supersedes the "never
truncated" sentence. The longest route pattern either server registers is 84 bytes
(`/admin/resources/{resourceId}/groups-with-permission/remove/{groupId}/{permissionId}`), so 320
refuses nothing real; a longer 404 probe is logged as its first 320 bytes plus its true length.
Cheapest option, and it leaves the skipped-path work above open.

**B. A, plus build the escaped path under that bound.** One bounded escape, used both by the skip
predicate and by the logged target, so the full escaped string is never built. Closes the wasted work
too, including for skipped requests. Costs a little more code on the path every request takes, and it
has to keep matching the skip list on the escaped form, which is what stops a caller silencing their
own log line by sending `/%68ealth`.

**C. Leave it as agreed: the path is logged whole.** An unauthenticated caller can put about 1 MB, or
1.84 MB using UTF-8, into the log per request, and we document that in `security.mdx` rather than
close it.

### Recommendation: B

A is the fix; B is the fix that closes the whole family. This particular family has now been declared
closed and reopened three times, once per review round, each time because the boundary claimed after
the previous fix turned out to be narrower than the defect: first the query's output, then the query's
work, then four other fields on the same line, now the path. The extra cost of B over A is one
function that escapes into a fixed budget, which is the shape the query redactor already has.

**What being wrong costs.** Pick A or B wrongly and the price is the tail of a path longer than 320
bytes missing from the log; reversing it is one constant. Pick C wrongly and a default install ships
an unauthenticated log-volume amplifier, and reversing it later is this same piece of work again.

Reply with **A**, **B** or **C** (a different limit than 320 is fine, just say the number). The run is
stopped until then, with stage 1's code uncommitted.

leodip:
B

leodip:
Closing this without merging. The run behind it is being discarded and #159 will be started again from scratch under a reworked process.

Nothing here is a judgement on the findings, which were real: the stage 1 reviews turned up raw Unicode line and paragraph separators surviving into the log line, a literal quote closing the line's own request field, and `RedactQuery` allocating against the whole raw query before its 4096-byte cap applied. Those are all still true of `main` and #159 stays open to cover them.

What is being thrown away is the run, not the knowledge.

Issue #203, opened 2026-08-11
A request that panics is logged with status 0, not 500


leodip:
Both servers register `middleware.Recoverer` before the HTTP request logging block, so the logger is nested **inside** the recoverer. On a panic the logger's deferred write runs while the stack is still unwinding, before `Recoverer` has written anything, so the log line reports status 0 while the client receives 500.

## Where

`initMiddleware` in `src/authserver/internal/server/server.go` and in `src/adminconsole/internal/server/server.go`, which are identical but for the config getter:

```go
// Recoverer
s.router.Use(middleware.Recoverer)

// HTTP request logging
if config.GetAuthServer().LogHttpRequests {
    ...middleware.Logger(next).ServeHTTP(w, r)
}
```

`r.Use(A); r.Use(B)` makes A the outer wrapper, so `Recoverer` wraps the logger. chi's own documentation says the opposite: "IMPORTANT NOTE: Logger should go before any other middleware that may change the response, such as middleware.Recoverer."

## Verified

Both orderings, same panicking handler:

```
logger inside Recoverer (today's order in both servers)
  status the client got  = 500
  status the logger saw  = 0

logger outside Recoverer (chi's documented advice)
  status the client got  = 500
  status the logger saw  = 500
```

## Impact

Anyone counting 5xx responses in the request log undercounts panics to zero. Only visible when `LOG_HTTP_REQUESTS` is on, which the setup wizard and five of the six shipped compose files turn on.

## The trade

Moving the logger ahead of `Recoverer` fixes the status, and puts the logger outside the only middleware that catches its own panics: a bug in the logger would then break the connection instead of producing a 500. Worth deciding deliberately rather than inheriting.

Found while working on #159, which is about what the line contains rather than what status it reports, and which deliberately left this ordering alone.

Issue #202, opened 2026-08-11
Security: password and OTP form handlers accept their credential from the query string


leodip:
Every form handler reads its inputs with `r.FormValue`, which merges the URL query into `r.Form` **behind** the request body. When the body does not carry a name, the query string supplies it. So a password or a TOTP code sent as a query parameter is not merely logged somewhere, it is accepted.

## Verified

Running it, rather than reading the standard library's documentation:

```
POST /auth/pwd?password=QUERY-SECRET&otp=123456    (empty body)
  r.FormValue("password") = "QUERY-SECRET"
  r.FormValue("otp")      = "123456"
  r.Form                  = map[otp:[123456] password:[QUERY-SECRET]]
  r.PostForm              = map[]
```

With the name present in both, the body wins (`r.Form` becomes `[BODY-SECRET QUERY-SECRET]` and `FormValue` returns the first). It is only the fallback that is the problem, and the fallback is always there.

## Where

Nine call sites outside tests read a credential this way:

```
$ grep -rn 'FormValue("password")\|FormValue("otp")\|FormValue("currentPassword")' --include=*.go src/
src/authserver/internal/handlers/handler_auth_otp.go:245
src/authserver/internal/handlers/handler_reset_password.go:173
src/authserver/internal/handlers/handler_auth_pwd.go:132
src/authserver/internal/handlers/accounthandlers/handler_account_register.go:66
src/adminconsole/internal/handlers/accounthandlers/handler_account_change_password.go:62
src/adminconsole/internal/handlers/accounthandlers/handler_account_otp.go:80
src/adminconsole/internal/handlers/accounthandlers/handler_account_otp.go:126
src/adminconsole/internal/handlers/adminsettingshandlers/handler_admin_settings_email.go:104
src/adminconsole/internal/handlers/adminuserhandlers/handler_admin_user_new.go:90
```

## Why it matters

A credential in a request target reaches browser history, the `Referer` header, and the access log of every intermediary between the client and the server. Accepting one is what makes a mistaken integration, a copied `curl` command or a crafted link work at all, and a request that works is a request that gets repeated.

## Suggested fix

Read the credential-bearing names with `r.PostFormValue`, which is the body only, and leave `r.FormValue` for the parameters that legitimately arrive either way (`/auth/authorize` genuinely accepts both GET and POST per OpenID Connect, and its own reads should stay as they are). At minimum: `password`, `currentPassword`, `newPassword`, `otp`.

## Related

[...]

Issue #201, opened 2026-08-11
Security: activation and password reset codes travel in the URL query string


leodip:
`/account/activate` and `/reset-password` take a single-use credential as a query parameter, so it ends up everywhere a URL ends up: browser history, the `Referer` header of anything the resulting page loads, and any proxy, gateway or CDN in front of the deployment.

## Where

`HandleAccountActivateGet` (`src/authserver/internal/handlers/accounthandlers/handler_account_activate.go`):

```go
email := r.URL.Query().Get("email")
code := r.URL.Query().Get("code")
```

`HandleResetPasswordGet` (`src/authserver/internal/handlers/handler_reset_password.go`) does the same, and `HandleResetPasswordPost` reads them again from the query.

The links are built by string concatenation at three sites, all listed in #112: `handler_forgot_password.go`, `handler_api_users_crud.go` and `handler_account_register.go`.

## Why it matters

The code is not a hint or an identifier. It is the whole authorisation for the action: `forgotPasswordCodeMatches` plus `isForgotPasswordCodeExpired` are the only gates on setting a new password, and the activation code is the only gate on enabling the account. Anything that reads the URL holds the credential for as long as it is unexpired.

`Referrer-Policy: same-origin` is already sent, which limits cross-origin `Referer` leakage but does nothing about browser history, the address bar, a URL pasted into a support ticket, or an intermediary's access log.

## Suggested directions

Any one of these closes most of it, and they compose:

- Redirect once the code has been consumed, to the same page without the parameter, so history and the address bar keep only the clean URL.
- Exchange the code for a short-lived server-side handle on first GET, and carry that instead.
- Make the confirmation step a POST, so the credential never appears in a request target.

## Related

- #159 stops Goiabada's own request log from recording these codes. It does not move them out of the URL, which is why this is separate: that touches the email templates, the handlers and links already sitting in people's inboxes.
- #112 is the same three link-building sites, but about `+` in email addresses breaking the link. Fixing both together may be sensible.

PR #204, opened 2026-08-11
fix(security): keep credentials out of the HTTP request log

### LINKED ISSUE #159: Security: enabled HTTP request logging writes complete id_token_hint JWTs to the log

leodip:
## Issue 159: keep credentials out of the HTTP request log

Closes #159.

When `GOIABADA_AUTHSERVER_LOG_HTTP_REQUESTS` or its admin console twin is on, every request outside
the skip list goes to chi's `middleware.Logger`, which writes `scheme://host` plus `r.RequestURI`.
`r.RequestURI` is the raw request target, path **and query string**, so the log records:

- the complete `id_token_hint` JWT from `/auth/authorize` and `/auth/logout`, which carries `sub`,
  `sid` and, when the scope was granted, the user's email address,
- the activation code and email address from `/account/activate`,
- the password reset code and email address from `/reset-password`,
- and any credential a caller puts in a query string, which the handlers do read: `r.FormValue`
  merges the query into `r.Form` behind the body, so `POST /auth/pwd?password=...&otp=...` with an
  empty body is both accepted and logged.

That token stays usable. `/auth/authorize` parses the hint with the expiration check off, and
`/auth/logout` honours an expired hint while its session row is live, which is the branch that does
not ask the End-User anything.

This is not a test-only configuration: the setup wizard writes `LOG_HTTP_REQUESTS=true` into both the
compose file and the env script it generates, and five of the six shipped compose files set it to
`true`.

### The rule this establishes

**An allowlist, not a denylist.** Every parameter name is logged. The value is logged only for names
on a short assessed-safe list, and every other value becomes `[redacted]`. A parameter nobody has
thought about is redacted by construction, which matters because #149 would add `logout_hint`, a
parameter the spec says may carry an email address, a phone number or a username. A denylist would
have logged it in full on the day it landed.

The twelve names that keep their values are the protocol and pagination core, each drawn from a fixed
vocabulary or a number: `client_id`, `response_type`, `response_mode`, `scope`, `prompt`, `max_age`,
`acr_values`, `code_challenge_method`, `ui_locales`, `error`, `page`, `size`. Matching is
case-sensitive, because the server's own reads are: `Client_ID=x` is not a parameter this server
reads, so treating it as one would print an attacker-chosen value under a familiar-looking name.

`state`, `nonce`, `code_challenge`, `redirect_uri` and `post_logout_redirect_uri` are redacted along
with everything else.

**And a bound on every component, because an allowlist bounds names and not bytes.** The twelve are
each drawn from a fixed vocabulary, but that is true of what the server *accepts*, not of what a
client can *send*: nothing validates a value before it is logged. A relying party that misbuilds its
authorize URL and puts a real token where `client_id` goes would otherwise write a real user's
credential into the log in full, and neither party would find out. So every rendered parameter name
and every retained value is clipped at 512 bytes with a marker giving the true length, which cuts a
JWT well before its signature. One bound for every component rather than a per-name validation table
that would drift out of sync with the server's own validators.

This is a bound and not an absence guarantee, and the distinction is worth stating rather than
glossing: a secret shorter than 512 bytes sent under one of the twelve names is still written whole.
Anyone adding a thirteenth name is choosing to trust that parameter's contents to this degree.

### What else changes

- **One log stream instead of two.** chi's logger wrote to stdout through the standard library `log`,
  while every other line in the process goes to stderr through `slog`. The replacement is a shared
  middleware in `src/core/middleware` writing one `slog` record, so the process has one format and
  one destination. The line an operator sees changes shape, and that is a deliberate cost.
- **The scheme and `Host` are dropped**, because `Host` is whatever the client sent and a deployment
  serves one base URL that is already in the configuration.
- **The line cannot be forged or flooded.** The path is taken from `EscapedPath` rather than the
  decoded `Path`; every string field is escaped down to printable ASCII and clipped; each query
  component is capped at 512 bytes; and the rendered target is capped at 4096, which matters because
  redaction barely shortens a target whose parameter names survive. All three caps measure the bytes
  that reach the log and report the true length in a marker.
- **The API debug middleware's URL line gets the same treatment.** It passed `r.URL.String()`, query
  string included, to a log line with no redaction. Same defect, second flag, one line.

### Deliberately out of scope

Moving the activation and password reset codes out of the URL entirely, which touches the email
templates, the handlers and the links already in people's inboxes: [#201](../../issues/201). The
handlers accepting a credential from the query string at all: [#202](../../issues/202). Request and
response bodies, which #145 already redacted. Nothing about how the ID Token itself is validated.

## Status

**Complete and ready for review.** Both stages have landed, the final review is done, three rounds all
applied, and the check suite is green on the final commit `717113bf`
([run 31492804002](https://github.com/leodip/goiabada/actions/runs/31492804002), eleven jobs). The
redactor is mounted in both servers and in the API debug middleware, the two documentation pages are
updated, and the defect this PR is about is closed on the branch: an `id_token_hint` in a query
string no longer reaches either server's log. Round 1 of the whole-change review found three issues,
fixed in `c668df00`; round 2 found that one sibling round 1 had deliberately left was a real
regression against the code this PR replaces, fixed in `66603294`; round 3 read the `net/url`
reproduction that fix introduced, found the code right and its tests unable to notice if it stopped
being, fixed in `717113bf`. All three are described below.

Decision 9 was answered on this thread and is applied. The plan review had found that the allowlist
bounds *which* parameters keep a value and bounds nothing about the bytes inside a value or inside a
parameter name, so a JWT placed in `client_id`, in `scope`, or in a parameter name still reached the
log; four such shapes were executed against the reference renderer. The answer was a 512-byte cap on
every rendered query component, plus narrowing the stated guarantee to what the code actually
delivers. Both are folded into stage 1 above, which is why the section above says "a bound, not an
absence guarantee" rather than promising more.

### Review

Reviewed by `gpt-5.6-sol` at effort `xhigh`, three rounds over the whole branch against `main`. No
stage was reviewed on its own, so these rounds are the only reading this code got.

| Round | Findings | Blocking | Answered by |
|---|---|---|---|
| 1 | 3 | 1 | `c668df00` |
| 2 | 1 | 1 | `66603294` |
| 3 | 1 | 0 | `717113bf` |

- **blocking, security:** all three output caps clipped a string that had already been escaped and
  assembled in full, so an unauthenticated request carrying 900000 high bytes cost 31.2 ms and
  10,301,512 B to render 128 bytes of log. Bounded during encoding instead, counting the discarded
  suffix rather than building it: **0.35 ms and 778 B**, output byte-identical.
- **blocking, security:** round 1 deliberately left the path unbounded, arguing it was inherent to
  `url.EscapedPath`. Round 2 measured that against the code this PR replaces and found it a real
  regression, 5,417,170 B against chi's 2,708,478 B. Bounded too.
- **significant, quality:** nothing pinned that the target is rendered *before* chi's `StripSlashes`
  rewrites `r.URL.Path`. The code was already right; moving the render into the deferred closure left
  the whole module suite green. One test now fails under that mutation.
- **significant, quality:** three of the `net/url` escaping decisions round 2 copied in were pinned
  by no test, so the copy could drift from the original unnoticed. Pinned over their whole domain.
- **minor, conformance:** three documentation claims stated absolutes the code does not hold. All
  corrected, and the lossiness is now stated positively.

Nothing was refuted, and every round returned an empty `unchecked` list, so no part of the change
went unread. Full accounting, per finding and per round, in [this comment](https://github.com/leodip/goiabada/pull/204#issuecomment-5253473361).

### Stage 1: the redactor and the request logger, in `src/core/middleware`

`5681cdb3`. One new file, `src/core/middleware/middleware_request_logger.go`, and its tests. Nothing
is mounted yet, so no observable behaviour changes: stage 2 wires it into both servers and the debug
middleware.

`RequestTargetForLog(u *url.URL) string` is the redactor. Every value that reaches its output has
passed one of exactly two gates, an exact case-sensitive match against the twelve names or being
empty, and there is no third path: a query `url.ParseQuery` refuses renders as
`?[unparsable query, N bytes]` rather than falling back to what arrived. That last part is not
defensive tidiness. `ParseQuery` hands back a *partial* map beside its error, and a renderer that
split the raw query instead would log `?client_id=c;id_token_hint=<JWT>` as one value under a safe
name, complete token and all, because a semicolon is not a separator.

`MiddlewareRequestLogger(enabled bool)` is the replacement for chi's logger: the flag as a parameter
so both servers can mount it unconditionally, the existing skip list carried over verbatim, and
exactly one `slog` record per request emitted from a `defer` so a panicking request still produces a
line. `status` is written raw, which means 0 for a panic, as chi logged it: normalising it to 200
would make the line say 200 for a request that failed, and that quirk is [#203](../../issues/203).

The three caps and the escaping helper landed with it, each constant's comment carrying its
measurements and what breaks if it is deleted.

**Tiers:** unit on all three modules, 3058 pass, 0 fail, 46 packages. Data and integration are not
applicable at this stage, nothing here touches the database and no endpoint can reach this code yet.
The four-engine data and integration tiers are the check suite's to report for `5681cdb3`.

**Mutations:** nine, one per behaviour the stage claims to have made observable, all nine caught.
Forcing the allowlist true, disabling the `ParseQuery` error path, swapping `EscapedPath` for `Path`,
raising each of the three caps beyond any input, making the escaping branch unreachable, disabling
the skip list, and ignoring the `enabled` flag each broke a test.

**One deviation from the plan**, about *when* rather than *what*: the target is rendered before the
handler runs rather than in the `defer`, because chi's `StripSlashes` sits immediately below this
middleware and edits `r.URL.Path` in place. Reading it afterwards would log the rewritten target
rather than the one that arrived.

### Stage 2: mount it in both servers, redact the debug middleware's URL, and document it

`d42c210a`. This is the stage that changes what anyone can observe.

Each server's `initMiddleware` drops its `if LogHttpRequests { ...chi logger... } else { ... }` block
for one unconditional `Use` of `MiddlewareRequestLogger`, which returns the next handler untouched
when the flag is off. The middleware does not move in the chain: still after `Recoverer`, so a
brand-new component stays inside the only thing that catches its panics, and still before
`StripSlashes`, which is why stage 1 renders the target before calling the next handler.

**chi's `middleware.Logger` is now referenced nowhere in the repository.** That is decision 3 rather
than a side effect. The process used to run two log streams in two formats, chi's raw request target
on stdout through the standard library `log` package beside everything else on stderr through `slog`,
and it was the stdout one that wrote the query string verbatim.

The API debug middleware's URL line changes from `r.URL.String()` to the same redactor (decision 5).
It is one line: the same defect under a second flag, on routes that carry a free-text user search
string in `query`.

**Tests:** one wiring test per server, at the seam `server_csrf_test.go` established, asserting that
the middleware is mounted *and* that it is wired to the configuration flag rather than to a constant.
Neither claim is reachable from stage 1's unit table, which passes perfectly against a middleware
nobody wired up, and a deleted `Use` line still compiles in both packages. Two thin cases at the
debug middleware, one accept and one reject, both driving `APIDebugMiddleware` rather than
`debugLog`: `debugLog` takes the URL as a `string` parameter, so a test calling it directly would
pass with the call site unchanged.

**Docs:** `reference/security.mdx` gains two bullets in its Logging section, one for what the
redaction keeps and withholds, one for the destination change, and the existing `DEBUG_API_REQUESTS`
bullet gains a clause. Both `LOG_HTTP_REQUESTS` rows in `reference/environment-variables.mdx` are
rewritten. The redaction is stated at the width the code guarantees and no wider, per decision 9.

**Tiers:** unit on all three modules, 3064 pass, 0 fail, 47 packages. Integration on sqlite, 1221
pass, 0 fail: owed here because this stage changes the middleware chain every request in both servers
passes through, which a hand-built `Server` in a unit test cannot cover. Data not applicable, nothing
here touches the data layer. The four-engine data and integration tiers are the check suite's to
report for `d42c210a`.

**Mutations:** three, three caught. Forcing each server's flag argument to `false` rather than
deleting its `Use` line, which is what distinguishes "mounted" from "mounted and wired to the
config", and reverting the debug middleware's call site to `r.URL.String()`.

**Deviations from the plan**, five, all recorded with their reasoning:

- The two `initMiddleware` comments are deliberately not identical. The auth server's names the
  `id_token_hint` JWT that leaked; the admin console never issues or accepts one, so it says "any
  credential a client put in a query string" instead. The **code** still differs by exactly one line.
- Two cases per wiring file rather than three. "The request is still served" has no life of its own:
  it is the assertion that stops the other two passing against a chain that broke, so it lives inside
  both of them.
- The admin console's test serves its own settings endpoint from a four-line `httptest` server.
  Pointing the settings cache at an unreachable address, as `server_csrf_test.go` does, makes
  `MiddlewareSettingsCache` answer 500 and return, so the request never reaches the handler, which
  destroys exactly the assertion this test needs.
- `security.mdx` gained two bullets rather than one. Decision 3 accepted the stdout-to-stderr move on
  condition the cost is documented rather than hidden, and burying "capture stderr too" inside a
  paragraph about parameter names would have hidden it.
- Seam 4's tests enter at `APIDebugMiddleware` rather than `debugLog`, for the reason above. Mutation
  3 is what shows that was necessary rather than stylistic.

### After the stages: the final review, round 1 (`c668df00`)

This run reviews the whole change once at the end rather than each stage on its own, so this commit
is the first reading the code got. Three findings, all acted on, none refuted.

**The caps ran after the rendering rather than during it, which was reachable by anyone.** All three
bounds in the request logger, 128 bytes for a scalar attribute, 512 for a query component, 4096 for
the whole target, clipped a string that had already been escaped and assembled in full. chi's
`RequestID` middleware is mounted first in both servers and adopts an inbound `X-Request-Id` header
verbatim, and `net/http` admits bytes above `0x7e` in a header value, each costing three bytes to
percent-escape. So an unauthenticated request carrying 900000 such bytes cost **31.2 ms of CPU and
10,301,512 bytes of allocation to render 128 bytes of log**. It is now 0.35 ms and 778 bytes.

Every escaper involved maps each byte independently, so the escaped form of a prefix is a prefix of
the escaped form: escaping only as far as the clip reaches yields exactly the bytes that are kept,
and the discarded suffix is counted rather than built. The output is byte-identical, pinned by tests
that assert against the exact expressions the file used before.

| a 900000-byte pathological request | before | after |
|---|---|---|
| the scalar field, via `X-Request-Id` | 31.2 ms, 10,301,512 B | **0.35 ms, 778 B** |
| one 300000-byte allowlisted value | 1.05 ms, 2,260,642 B | **0.88 ms, 308,128 B** |
| 9000 high-byte parameter names | 2.30 ms, 8,243,823 B | **1.47 ms, 1,171,089 B** |
| a 900000-byte high-byte path | 1.12 ms, 5,512,892 B | 1.21 ms, 5,417,170 B |

The two middle rows are now at their floor: `url.ParseQuery` alone costs 308,128 B and 922,984 B on
those same queries. **The path row deliberately did not move**, on the grounds that all of it is
`url.URL.EscapedPath()` and that `url.ParseRequestURI` allocates 5,406,864 B on the same bytes inside
`net/http` before any code here runs, making it a constant factor on an inherent cost rather than
amplification. **Round 2 overturned that, and the last sentence of this section as first written,
"cheaper than what it replaced on every measured axis", was wrong**: this table's own last row is
5,417,170 B against chi's 2,708,478 B. See the round 2 section below.

**Nothing pinned that the target is rendered before `StripSlashes` rewrites it.** Both servers
register chi's `StripSlashes` immediately after the logger and it edits `r.URL.Path` in place, so the
target is the one logged attribute whose value depends on when it is read. The implementation was
already correct; moving the render into the deferred closure left the whole module suite green. One
test now composes the real logger with `StripSlashes` and is the only thing that fails under that
mutation.

**Three documentation claims were absolutes the code does not hold.** The pages said one record per
request, when `/health`, `/static/` and `/favicon.ico` are never logged; that parameter names are
always kept, when a parse error keeps none and the 4096-byte cap can cut later ones; and that a
retained value is written as it arrived, when the query is parsed and re-encoded so `%20` comes back
as `+`. All three corrected, and `reference/security.mdx` gained a bullet stating the lossiness
positively, including the `?[unparsable query, N bytes]` shape that an operator meeting it would
otherwise read as a bug.

**Tiers:** unit 3075 pass 0 fail across all three modules (3064 before this round), integration 1221
pass 0 fail on sqlite, site build clean at 42 pages, all 18 anchors resolve. Data not applicable,
nothing here touches the data layer. The four-engine tiers are the check suite's to report for
`c668df00`.

**Mutations:** seven, seven caught. Two of them restore the unbounded rendering verbatim, which every
output assertion in the file passes and only the allocation tests catch; one is the reviewer's own
surviving mutation from the timing finding.

### After the stages: the final review, round 2 (`66603294`)

One finding, blocking, on the sibling round 1 named and left. It was confirmed before anything was
changed by writing the allocation guard first and watching it fail: **5,416,840 bytes allocated to
emit 4096 bytes of log**, for any unauthenticated request carrying a long path.

Round 1's reasoning was wrong in two places. "A constant factor on an inherent cost" is true of the
arithmetic and not a reason: `net/http`'s parse and the logger's escape are two independent
allocations, so turning logging on doubled it, which is exactly the amplification the bound exists to
remove. And the claim that the change was cheaper than chi's logger on every measured axis was
contradicted by round 1's own table, which put this request at 5,417,170 B against chi's 2,708,478 B.
The third ground, that bounding it means reproducing `EscapedPath`'s `RawPath` semantics, was a real
cost honestly stated and still the wrong call: the file already reproduces two standard-library
escapers this way, each pinned per byte against the thing it replaces.

`escapedPathForLog` now reproduces both of `url.URL.EscapedPath`'s branches and returns a clipped
head plus the true rendered length, which the target's truncation marker still reports. Where the
client's own encoding is valid it is returned verbatim and nothing is rendered at all, the validity
decided by walking `RawPath` and `Path` together rather than decoding one to compare it. Otherwise
the path is escaped only as far as the clip reaches. `net/url` exports no whole-path escaper, so the
rule is written out here: `url.PathEscape` uses the path-segment rule and escapes `/` as well.

| the same 900000-byte path | before | after |
|---|---|---|
| the renderer, escaping branch | 5,416,840 B | **0.33 ms, 17,266 B** |
| the renderer, already-encoded branch (300000 `%8a`) | | **0.42 ms, 4,945 B** |
| the whole middleware, output discarded | 1.13 ms, 5,433,150 B | **0.35 ms, 17,826 B** |
| chi's `middleware.Logger`, the code this PR replaces | 0.20 ms, 2,708,817 B, and it wrote 900,075 bytes of it to the log | |
| `url.ParseRequestURI`, inside `net/http` before any of this | 1.78 ms, 5,406,882 B | unchanged, inherent |

Allocation is 313 times lower than before this round and 152 times lower than chi's. **CPU is not
lower than chi's and this does not claim it is:** 0.35 ms against 0.20 ms, on a request where
`net/http` has already spent 1.78 ms parsing. Reporting the true length in the truncation marker
means looking at every byte, which chi never did because it copied the whole target and wrote all
900 KB of it to the log. The logger adds 20% to a cost the server pays whether or not logging is on,
against chi's 11%. The input bound is still `http.Server.MaxHeaderBytes`, which neither server sets;
that stays a drafted follow-up rather than a change here, because it changes what the server accepts.

The output is byte-identical, which is the whole claim, so no documentation moved.

**Tiers:** unit 3083 pass 0 fail across all three modules (3075 before this round), integration 1221
pass 0 fail on sqlite, all 18 anchors resolve. Data not applicable, nothing here touches the data
layer. The four-engine tiers are the check suite's to report for `66603294`.

**Mutations:** eight, eight caught. One restores the pre-fix unbounded form verbatim and is caught by
exactly one test, the allocation guard, because every output assertion in the file passes against it.

### After the stages: the final review, round 3 (`717113bf`)

One finding, and it is about the tests rather than the code. Round 2's fix reproduces
`url.URL.EscapedPath` inside this repository, because `net/url` exports no whole-path escaper. That
is only acceptable while a test pins it against the standard library, and three of the decisions it
copies in were not pinned by one: which bytes may stand for themselves, which bytes are hex digits,
and what each hex digit is worth. The reviewer demonstrated it by making three wrong edits and
watching the whole suite stay green, and all three were reproduced here before anything was written.

**The renderer was correct and is unchanged.** No production code moved in this round. What was
missing was the ability to notice if it ever stopped being correct, which is the entire justification
for reproducing a standard-library function instead of calling it. A divergence would be silent: the
log would report a path the request never carried, with a truncation marker giving a length that was
never true.

Sampling is the wrong instrument for the decoder in particular, and this is the part worth knowing.
When it gets a byte wrong, the client's encoding is rejected and the renderer falls back to escaping
the decoded path, which writes the same escape back in upper case, so **a wrong answer rejoins the
right one**. The two only differ if the same path is non-canonical somewhere else, which a random
alphabet produces rarely and never on purpose. Every escaped-byte row therefore carries a `/%61`
canary, an escape `net/url` accepts but would never itself produce.

Each domain is 256 wide and costs milliseconds to walk, so each is now covered whole rather than
sampled: every literal byte, every escape in both hex cases, and every candidate hex digit against
every byte it could be taken to mean, that last one because a classifier admitting a byte `net/url`
refuses decodes it to *something* and which byte that is depends on the defect. Three of the four
tables assert on output, against `url.URL.EscapedPath` itself; the fourth states the rule against
`strconv`, which is what the two length tables beside it already do. The shared assertion now runs
every table at seven truncation limits rather than one.

**Tiers:** unit 3184 pass 0 fail across all three modules, exactly four more than the same tier on
the reviewed tree. Integration not re-run: one test file moved, no handler, route or schema, and both
the production and integration-test tree hashes are unmoved from round 2, so its 1221 pass stands.
All 18 anchors resolve. The four-engine tiers are the check suite's to report for `717113bf`.

**Mutations:** ten, ten caught, one per member of the family rather than one per test, including the
three that survived when the reviewer ran them.

## How this was verified, end to end

**The check suite is green on the final commit** `717113bf`,
[run 31492804002](https://github.com/leodip/goiabada/actions/runs/31492804002). All eleven jobs pass.
That run is where the data and integration tiers executed on **mysql, postgres and mssql**; nothing
local ever ran them, and no claim in this PR says otherwise.

Locally, each stage and each review round ran **unit on all three modules** and, where the change
could reach an endpoint, **integration on sqlite**. Final counts: unit 3184 pass 0 fail across
authserver, adminconsole and core; integration 1221 pass 0 fail on sqlite. The data tier is not
applicable anywhere in this change, which touches no schema, no migration and no data-layer code.
Integration was not re-run for round 3, which moved one test file and no handler, route or schema.

**Thirty-seven mutations were run across the two stages and the three review rounds, and all
thirty-seven were caught.** Each one exists to show that a specific claimed behaviour is actually
observable by a test rather than merely accompanied by a green suite. Two of them restore an earlier
unbounded rendering verbatim and are caught by exactly one test each, the allocation guards, because
every output assertion in the file passes against them. Three more are the reviewer's own mutations
that survived when it ran them, reproduced here and now caught.

The site builds clean at 42 pages and all 18 code anchors cited during the run still resolve.

### Deviations worth knowing

- **Stage 1 renders the request target before calling the next handler, not in the deferred write.**
  chi's `StripSlashes` sits immediately below this middleware and edits `r.URL.Path` in place, so
  reading it afterwards would log the rewritten target rather than the one that arrived. A test
  composes the real logger with `StripSlashes` and is the only thing that fails if this moves.
- **The two `initMiddleware` comments deliberately differ** while the code differs by exactly one
  line. The auth server's names the `id_token_hint` JWT that leaked; the admin console never issues
  or accepts one.
- **The debug middleware's tests enter at `APIDebugMiddleware`, not `debugLog`.** `debugLog` takes
  the URL as a `string` parameter, so a test calling it directly passes with the call site unchanged.
- **`status` is logged raw**, which means 0 for a panicking request, exactly as chi logged it.
  Normalising it to 200 would make the line say 200 for a request that failed. That quirk is #203.
- **Round 1 got the path bound wrong and round 2 overturned it.** The reasoning is kept above rather
  than tidied away, because the corrected claim is narrower than the original: allocation is 152
  times lower than chi's, CPU is not lower than chi's, 0.35 ms against 0.20 ms, and this PR does not
  claim it is.

### Three comments on this thread

- **[The deferred decision](https://github.com/leodip/goiabada/pull/204#issuecomment-5253418865)**,
  one question answered by assumption rather than by asking: `maxLoggedField = 128` clips
  `request_id`. It states the assumption, the evidence for it, and what reversing it costs, which is
  one constant, one test row and one sentence of documentation.
- **[The code review](https://github.com/leodip/goiabada/pull/204#issuecomment-5253473361)**: every finding from all three rounds, what confirmed it, and what
  answered it, including the reviewers' own mutations and the `unchecked` lists.
- **[The follow-ups](https://github.com/leodip/goiabada/pull/204#issuecomment-5253419373)**: #201,
  #202 and #203 were filed during the run and are still open, and a fourth is drafted ready to file,
  on neither server setting `http.Server.MaxHeaderBytes`.


leodip:
## Decision 9: the allowlist bounds which parameters keep a value, and nothing about the bytes inside them. How completely should the query component be closed?

The plan review raised this before any code was written, so nothing is committed against it. It needs
your answer because it revisits two decisions you sealed and because the answer decides what stage 1
builds.

### What the design actually does

The redactor parses the query, writes every parameter **name**, and writes a parameter's **value**
only when the name is one of twelve assessed-safe ones (`client_id`, `response_type`, `response_mode`,
`scope`, `prompt`, `max_age`, `acr_values`, `code_challenge_method`, `ui_locales`, `error`, `page`,
`size`). Everything else becomes `[redacted]`. That closes the reported defect completely: an
`id_token_hint` is redacted, and so is any parameter nobody has thought about yet.

What it does not bound is the **bytes inside** a name or inside one of those twelve values. Both are
copied at whatever length arrived, up to the 4096-byte cap on the whole target.

### Executed, not argued

I extracted the reference renderer verbatim out of the probe that the agreement was written against,
so this is that implementation and not a retyping of it. Four shapes put a complete three-segment JWT
into the log line, while the `id_token_hint` control stayed redacted in all four:

```
?client_id=<JWT>        -> /auth/authorize?client_id=eyJhbGciOi...U0lHTkFUVVJF
?scope=<JWT>            -> /auth/authorize?client_id=c&scope=eyJhbGciOi...U0lHTkFUVVJF
?<JWT>=x                -> /auth/authorize?eyJhbGciOi...U0lHTkFUVVJF=[redacted]
?client_id=c&<JWT>      -> /auth/authorize?client_id=c&eyJhbGciOi...U0lHTkFUVVJF=
control: ?id_token_hint=<JWT> -> /auth/authorize?client_id=c&id_token_hint=[redacted]
```

A 6000-byte parameter name renders to 4127 bytes of caller-chosen printable ASCII.

The last shape is the one I would not have predicted: a parameter with no `=` at all becomes a
**name**, so a URL built wrongly can land a token in the name position without anyone intending a
name. And `scope` shows this is not peculiar to `client_id`.

### Why the premise slipped

The reasoning for the twelve was that each is "drawn from a fixed vocabulary or is a number, so none
can carry a credential". That is true of what the **server accepts**. It is not true of what a client
can **send**: nothing validates the value before it is logged, so the log records the bytes that
arrived rather than the bytes the server honoured.

### What this is and is not

An attacker cannot steer **your users'** credentials into these positions. The query string is built
by the relying party, so reaching them takes a misbuilt integration or someone pasting a URL. What an
unauthenticated caller can do is write their own bounded, printable-ASCII bytes into your log. So I
would not call this an active vulnerability, but it is the gap that made an allowlist preferable to a
denylist in the first place, and a thirteenth name added to the list later reopens it silently.

**No specification decides this.** Nothing in OIDC Core 1.0, RP-Initiated Logout 1.0, RFC 6819,
RFC 6750 or RFC 9700 requires an authorization server to redact its own request log; the relevant
RFC 9700 MUST (section 4.9.3) is addressed to resource servers about access tokens. Every option below
is permitted, so this is genuinely your call rather than a conformance question.

### The options

**A. Leave the code as designed, and state the guarantee accurately.** The goal becomes "no value of a
parameter outside the twelve is written", which is what the allowlist delivers. Costs nothing. Closes
nothing further, and a misbuilt RP's token stays in the log in full.

**B. Cap every rendered query component.** Each name and each allowlisted value clipped at its own
bound, using the truncation marker the design already has for the whole target. At 256 bytes every
legitimate value fits comfortably (a long `scope` of several resource permissions is well under it)
and a JWT is cut to a prefix that cannot be replayed, since the signature is gone. Honest limitation:
this is not an absence guarantee. A short secret still fits whole, and how much of a clipped JWT
payload stays readable depends on claim ordering.

**C. Validate each of the twelve values against a shape and redact anything that fails.** Makes the
original premise true by enforcement rather than assumption. Costs twelve patterns that must stay
true as the server's own validators change, and it will sometimes redact a legitimate but unusual
value, which is the way you rejected the minimal list for making the log worse.

**D. Redact unrecognised names too**, or render them as a count like `[3 other parameters]`. Closes
the name position outright. Reverses "every parameter name is logged", which is what tells an
operator that `code_challenge` was missing or that a stray parameter arrived.

B and D compose, as do C and D.

### My recommendation: B at 256 bytes, plus A's honest restatement

B is one bound and one marker rather than twelve patterns to maintain, it degrades gracefully instead
of redacting a value that was fine, and it is the same structural move already made once when the
whole target was capped, applied one level down. A on its own leaves the gap that an allowlist exists
to avoid. D buys mostly log-noise reduction in exchange for the parameter inventory, which is real
diagnostic value.

### What it costs to get wrong

A and B are cheap either way: both live inside the one redactor function and its test table, which is
stage 1's own deliverable, so changing your mind later is a contained edit. C and D reshape that table
and D reverses a sealed decision, which is why both are much better decided now than after stage 1
lands.

Reply with **A**, **B**, **C**, **D**, or a combination, with a different byte bound if you want one.
The run is stopped until then.

leodip:
## Decision 9: **B**, at **512** bytes, together with **A**'s restatement

Take B, and change the recommended bound from 256 to 512.

**Why B.** The premise behind the twelve is true of what the server accepts and not of what a client
can send, and the probe settles that. The shape I care about is not the attacker writing their own
bytes into my log, it is the relying party that misbuilds its authorize URL and puts a real token
where `client_id` goes, because then a real user's real credential is in my log in full and neither
of us finds out. B bounds that to a prefix with no signature attached. It is one constant and one
marker instead of twelve patterns that have to stay true as the validators move, it truncates rather
than redacts so an unusual but legitimate value is still readable, and it is decision 4's move applied
one level down, which keeps the design coherent rather than adding a second idea. It also covers the
name position, which is where the two shapes I would not have predicted landed.

**Why 512 and not 256.** Eleven of the twelve are comfortable at either bound. `scope` is the
outlier here: a client granted twenty or so `resource:permission` pairs goes past 256, and `scope` is
the field I actually read when I am diagnosing why consent or scope filtering did what it did.
Truncation degrades gracefully there and the marker still reports the true length, so 256 would not
be wrong, but 512 costs one character in a constant and keeps a real scope intact. A JWT is still cut
well before its signature. One bound for every component, not a per-name table.

**And A.** Restate §2 to what the implementation actually guarantees: no value of a parameter outside
the twelve is written, and every rendered name and value is bounded. Do not leave a sentence in the
agreement that promises more than the code does, because the next person to add a thirteenth name
will read that sentence and not the code.

**Not C**, for the reason decision 2 already gave: twelve patterns that drift out of sync with the
server's own validators, redacting values that were fine. **Not D**: the parameter inventory is the
part of this log line I would actually miss, seeing that `code_challenge` never arrived is worth more
than the noise D removes.

Resume the run. Stage 1 builds `RequestTargetForLog` with a per-component cap of 512 and the
truncation marker, the whole-target cap of 4096 stays as decided, and seam 1's table grows the
component-truncation cases: a name over the bound, an allowlisted value over the bound, and one of
each sitting under a whole-target truncation so the two caps are shown composing.

leodip:
## Decision 9 applied: B at 512 bytes, plus A's restatement

Answer received and applied. The run is unblocked and stage 1 is next.

**What changed in the design, before any code.** A third constant, `maxLoggedQueryComponent = 512`.
Every rendered parameter name and every retained value becomes `truncate(url.QueryEscape(s), 512)`,
in that order: escape first so the bytes counted are the bytes the log receives, clip second.
Decision 4's 4096-byte whole-target cap is untouched, and the two compose, a component clipped first
and the assembled target second. A `[redacted]` value needs no clip, being eleven bytes whatever
arrived.

One thing worth flagging, since it is a choice the answer implied rather than stated. The cap is
measured on the **escaped** form rather than the raw one, so all three caps in this change measure
and report output bytes and their markers are comparable. Clipping the raw value first would have
bounded the output at 3 x 512 rather than at 512, since escaping expands. The cost is that a cut can
land inside a `%XX` and leave a trailing `%E`, which the field cap already accepts for the same
reason: the marker immediately after states the true count, so nothing reads as a value it is not.

**§2 restated**, and it now carries a paragraph saying what the change does *not* close: this is a
bound and not an absence guarantee, a secret under 512 bytes sent as `client_id` is still written
whole, and adding a thirteenth name to the twelve is a decision to trust that parameter's contents to
this degree. That sentence exists for exactly the reader you named, the next person to add a name.

**Seam 1's table** gains the three cases you asked for, a name over the bound, an allowlisted value
over it, and one of each under a whole-target truncation so the two caps are shown composing. I added
a fourth: a `[redacted]` value built from a 900-byte input, asserting the output is exactly
`[redacted]` with no marker. Without it nothing distinguishes "the redaction path never reaches a cap"
from "the clip was applied before the allowlist check instead of after", and the second is a real way
to get this wrong. Step 6 gains a ninth mutation, raising the constant beyond any input.

512 does not change the measured cost in the plan's table: `QueryEscape` allocates the full escaped
copy before the clip sees it. That is deliberate, per the ordering above, and it is written down so
whoever transcribes the function does not reverse it.

leodip:
## Deferred decision, for the reviewer to confirm or overturn

One question came up during the run that was answered by assumption rather than by asking. It is
recorded here because the assumption is the kind that is cheap to reverse now and awkward to revisit
later.

### `maxLoggedField = 128` clips `request_id`, and clipping a correlation identifier is lossy

Raised by the plan reviewer, marked `needs_human` with no forced answer. Deferred rather than
resolved.

**The assumption taken.** Keep 128, pin the collision with a test, and document that the field is
clipped. The branch is built on that.

**What reversing it costs.** One constant in `src/core/middleware/middleware_request_logger.go`, one
table row in its test, and one sentence in `site/src/content/docs/reference/security.mdx`. Nothing
else reads the value. No interface, wire format or persisted shape depends on it, and raising it does
not meaningfully change the log line's worst case, because the 4096-byte target bound dominates the
~4.6 KB total.

**The evidence for taking it.** The collision was executed rather than argued. Two identifiers of 129
bytes differing only in their final byte do render identically, truncation marker included. But the
shape a chaining proxy actually produces does not collide: four hops of appended UUIDs come to 147
bytes with the differing hop beginning at byte 111, so the retained 128-byte prefix still separates
them. A real collision needs two identifiers equal in length **and** identical for their first 128
bytes, which means a proxy emitting a long constant prefix with its entropy at the tail. Every
clipped value also carries `[truncated, 128 of N bytes]`, so no record presents itself as a complete
identifier when it is not. The failure mode is an operator seeing two records they cannot tell apart,
never one they wrongly believe they can.

**The question, if you want to answer it.** Should a proxy-supplied `X-Request-Id` longer than 128
bytes be:

- **A.** preserved in full, so a 900 KB header becomes a 900 KB log line,
- **B.** clipped at 128 as it is now, or
- **C.** clipped at 128 with a short digest of the whole value appended, so long identifiers stay
  distinguishable at a fixed cost?

The branch ships **B**. **C** is the only one that costs more than a constant, and it is a handful of
lines in the same function if you want it.

leodip:
## Follow-ups

Four real defects turned up beside this one. Three were filed during the run and are still open; the
fourth is drafted below, ready to file as it stands. None of them is this change's to fix, and the
reasoning for keeping each out is in each entry.

### Already filed

- **#201: activation and password reset codes travel in the URL query string.** `security`, `go`.
  `HandleAccountActivateGet` and `HandleResetPasswordGet` both read their one-time code from
  `r.URL.Query()`. This PR stops Goiabada's own log recording them; it does not move them out of the
  URL, which touches the email templates, the handlers and the links already sitting in people's
  inboxes.
- **#202: password and OTP form handlers accept their credential from the query string.** `security`,
  `go`. `r.FormValue` merges the URL query into `r.Form` behind the body, so
  `POST /auth/pwd?password=SECRET&otp=123456` with an empty body is accepted. Verified by running it.
  This PR stops the log recording such a request and does not change what the handlers accept.
- **#203: a request that panics is logged with status 0, not 500.** `bug`, `go`. `Recoverer` is
  registered before the request logger, so the logger's deferred write runs while the stack is still
  unwinding. Out of scope here, and the obvious fix would put the logger outside the only middleware
  that catches its own panics.

### Drafted, not filed

**Title:** Neither server sets `http.Server.MaxHeaderBytes`, so one request may carry about a
megabyte of headers and request line

**Labels:** `security`, `go`

**Body:**

> Both servers construct `http.Server` without setting `MaxHeaderBytes`, so Go's 1 MB default
> applies: an unauthenticated client can send about a megabyte of request line and headers on every
> request, and `net/http` allocates several times that parsing it before any handler or middleware
> runs.
>
> Four `&http.Server{}` literals across `src/authserver/internal/server/server.go` and
> `src/adminconsole/internal/server/server.go` set `Handler`, `Addr` and the TLS config and leave
> `MaxHeaderBytes` alone. That default is the input bound behind every cost measured at #159's final
> review: `url.ParseRequestURI` alone allocates 5.41 MB for a 900000-byte high-byte path, inside
> `net/http`, before any Goiabada code runs.
>
> #159 bounded what its own request logger does with such a request and deliberately did not change
> what the server accepts, because that is a behaviour change with its own compatibility question: a
> long `redirect_uri` alongside a large `Cookie` header is a legitimate request, and the bound is a
> number somebody has to choose.
>
> An authorization server serves short URLs and modest cookies, so a much smaller bound is likely
> safe. Pick a value (64 KB is a common choice, but a long `redirect_uri` alongside session cookies
> wants headroom), set it on all four `http.Server` literals, and consider whether it is worth an
> environment variable.

Duplicate search, re-run at the finish: `gh issue list --state open --search "MaxHeaderBytes"` and
`--search "request size limit header"` both return nothing.

**To file it:**

```bash
  --title "Neither server sets http.Server.MaxHeaderBytes, so one request may carry about a megabyte of headers and request line" \
  --label security --label go \
  --body 'Both servers construct `http.Server` without setting `MaxHeaderBytes`, so Go'"'"'s 1 MB default applies: an unauthenticated client can send about a megabyte of request line and headers on every request, and `net/http` allocates several times that parsing it before any handler or middleware runs.

Four `&http.Server{}` literals across `src/authserver/internal/server/server.go` and `src/adminconsole/internal/server/server.go` set `Handler`, `Addr` and the TLS config and leave `MaxHeaderBytes` alone. That default is the input bound behind every cost measured at #159'"'"'s final review: `url.ParseRequestURI` alone allocates 5.41 MB for a 900000-byte high-byte path, inside `net/http`, before any Goiabada code runs.

#159 bounded what its own request logger does with such a request and deliberately did not change what the server accepts, because that is a behaviour change with its own compatibility question: a long `redirect_uri` alongside a large `Cookie` header is a legitimate request, and the bound is a number somebody has to choose.

An authorization server serves short URLs and modest cookies, so a much smaller bound is likely safe. Pick a value (64 KB is a common choice, but a long `redirect_uri` alongside session cookies wants headroom), set it on all four `http.Server` literals, and consider whether it is worth an environment variable.'
```

leodip:
## The code review, round by round

The audit trail behind the three review commits on this branch: which reviewer read what, every finding it raised, and whether that finding was confirmed and fixed, confirmed and deferred, or refuted and on what evidence. It also records each round's **unchecked** list, which was empty every time.

The narrative version, with the benchmark tables, is in the pull request body under "What has landed". This is the accounting.

---

This run was driven with `LEO_REVIEW_MODE=end`, so no stage was reviewed on its own and the whole
branch was read once, at the end.

### Round 1

Reviewer `gpt-5.6-sol` at effort `xhigh`, fresh session, request `159-20260811T105341Z`. Verdict
`findings`: three, all acted on, none refuted. `unchecked` was empty. Fixed on the branch as a
commit of its own rather than folded back into a stage, since both stages were already pushed.

**Finding 1, blocking, security: the output caps ran after the rendering, not during it.**
CONFIRMED, and reachable by an unauthenticated client in the shipped configuration. Verified rather
than accepted: chi's `RequestID` is mounted first in both servers
(`src/*/internal/server/server.go`) and adopts an inbound `X-Request-Id` verbatim
(`~/go/pkg/mod/github.com/go-chi/chi/v5@v5.3.1/middleware/request_id.go`, `requestID :=
r.Header.Get(RequestIDHeader)`), `net/http` admits bytes above `0x7e` in a header value, and each
one costs three bytes to escape. Measured on this machine, 900000 bytes of `0x80`:

| what | before | after |
|---|---|---|
| `fieldForLog`, the reachable one, via `X-Request-Id` | 31.2 ms, 10,301,512 B | **0.35 ms, 778 B** |
| `RequestTargetForLog`, one 300000-byte allowlisted value | 1.05 ms, 2,260,642 B | **0.88 ms, 308,128 B** |
| `RequestTargetForLog`, 9000 high-byte parameter names | 2.30 ms, 8,243,823 B | **1.47 ms, 1,171,089 B** |
| `RequestTargetForLog`, a 900000-byte high-byte path | 1.12 ms, 5,512,892 B | 1.21 ms, 5,417,170 B |

The two rows that moved are now at their floor: `url.ParseQuery` alone costs 308,128 B and
922,984 B on those same queries, so the rendering adds nothing and 248 KB respectively.

**The path row deliberately did not move, and this is the boundary the fix stops at.** All of it is
`url.URL.EscapedPath()`, measured alone at 5,406,720 B. `url.ParseRequestURI` on the same bytes
allocates 5,406,864 B inside `net/http`, before any code here runs, so calling `EscapedPath` roughly
doubles a cost the server pays for that request regardless. That is a constant factor on an
inherent cost rather than the amplification the finding is about, and bounding it would mean
reimplementing `EscapedPath`'s `RawPath` semantics, which is a correctness risk taken for a factor
of two. The real control is the request size itself, drafted as follow-up 4.

Worth stating alongside: chi's `middleware.Logger`, the code this change replaced, spent 2,708,478 B
on that same request and **wrote 900,075 bytes of it to the log**. After this fix the change is
cheaper than what it replaced on every measured axis.

**How it was fixed, keeping the output byte-identical.** Every escaper involved maps each byte
independently, so the escaped form of a prefix is a prefix of the escaped form. Escaping only as far
as the clip reaches therefore yields exactly the bytes that are kept, and the discarded suffix is
counted by `queryEscapedLen` / `safeLogValueLen` rather than built. `clipped` does the same for the
whole-target assembly, and once it is full it stops escaping components altogether and measures them
with `queryComponentLen`. `truncationMarker` was factored out so the measuring and rendering paths
cannot disagree about the marker's own width.

**Finding 2, significant, quality: no test pinned target capture before `StripSlashes`.** CONFIRMED
by reproducing the reviewer's mutation: moving `target := RequestTargetForLog(r.URL)` into the
deferred closure left the whole module suite green. `TestMiddlewareRequestLogger_RendersTheTargetBeforeDownstreamRewritesIt`
composes the real logger with chi's `StripSlashes` over `/auth/authorize/?client_id=c`, and it is
now the only test that fails under that mutation.

**Finding 3, minor, conformance: the documentation stated absolutes the code does not hold.**
CONFIRMED against the branches themselves. Three corrections, each forced by implemented behaviour
rather than reopening a decision: the record is written per request *apart from* `/health`,
`/static/` and `/favicon.ico`; names are "kept" rather than "always kept", since a `ParseQuery`
error retains none and the 4096-byte cap can cut later ones; and a retained value is not "written as
it arrived", since the query is parsed and re-encoded. A second bullet in `reference/security.mdx`
now states the lossiness positively, including the `?[unparsable query, N bytes]` shape, which was
previously undocumented and is what an operator meeting it would otherwise read as a bug.

### Tiers, at `production 84ce97d30239`

| Tier | Command | Scope | Result |
|---|---|---|---|
| Unit | `where.sh test --type modules` | all three modules | **3075 pass, 0 fail**, 47 packages (3064 before this round) |
| Integration | `where.sh test --type integration --db sqlite` | sqlite only | **1221 pass, 0 fail** |
| Data | not run | | nothing here touches `data.Database`, a query, a migration or the schema; `data-tests feaf2dad58c8` is unmoved from stage 2 |
| Anchors | `check-anchors.sh` | all 18 rows | PASS |
| Site | `npm run build` in `site/` | | clean, 42 pages |

`tree-hash.sh --split`: `production 84ce97d30239`, `unit-tests d4515f8098c5`,
`data-tests feaf2dad58c8`, `integration-tests c7279c9e9c3a`, `docs f3584cb70ebd`, `all 9d20466092cb`.

### Mutations

Seven, one per seam this round introduced or repaired, each through `where.sh mutate`, which
restores from a trap and compares by hash. All **CAUGHT**.

|---|---|---|---|
| 1 | `middleware_request_logger.go` | `queryEscapedLen`: drop the space rule (`if c == ' ' \|\| 'a' <= c` → `if 'a' <= c`) | CAUGHT |
| 2 | same | `safeLogValueLen`: escape width `n += 3` → `n += 4` | CAUGHT |
| 3 | same | `queryComponentForLog`: escape a shorter prefix than the clip keeps (`head[:maxLoggedQueryComponent]` → `[:maxLoggedQueryComponent/2]`) | CAUGHT |
| 4 | same | `clipped.writeString`: stop counting dropped bytes (`c.n += len(s)` → `c.n += 0`) | CAUGHT |
| 5 | same | `clipped.writeQueryComponent`: escape everything again, the pre-fix unbounded form | CAUGHT |
| 6 | same | `fieldForLog`: `truncate(safeLogValue(s), maxLoggedField)`, the pre-fix unbounded form verbatim | CAUGHT |
| 7 | same | the target rendered inside the deferred closure instead of before the handler, the reviewer's own mutation from finding 2 | CAUGHT |

5 and 6 are the load-bearing pair: they restore the exact code the reviewer measured, and the
allocation tests catch them where every output assertion in the file passes. 7 survived the whole
module suite before this round and is caught now by one test.

`production` was re-hashed after the last mutation and matches the value the tiers ran at, so
nothing was left behind.

### Round 2

Reviewer `gpt-5.6-sol` at effort `xhigh`, resumed session, request `159-20260811T113500Z`. Verdict
`findings`: one, blocking, security, acted on and not refuted. `unchecked` was empty, so there is
nothing to cover, dismiss or file from it. Fixed on the branch as `66603294`.

**Finding 1, blocking, security: the path was still escaped in full before the target cap.**
CONFIRMED, and confirmed independently before anything was changed: the new allocation guard was
written first and failed at **5,416,840 B allocated to emit 4096 bytes** of log.

**This overturns round 1's boundary, and round 1 was wrong twice.** Round 1 saw this sibling, named
it, and deliberately left it, on two grounds. The first was that `url.ParseRequestURI` pays a
comparable cost inside `net/http` regardless, so `EscapedPath` is "a constant factor on an inherent
cost". That is true and it is not a reason: parsing and logging are two independent allocations, so
turning logging on doubled it, which is the amplification the finding was always about. The second
was the sentence "after this fix the change is cheaper than what it replaced on every measured
axis", and round 1's own table contradicts it: 5,417,170 B against chi's 2,708,478 B on this exact
request. It held only if you read past the path row. The reviewer had already put both in writing
and the run's summary did not.

The third ground, that bounding it "would mean reimplementing `EscapedPath`'s `RawPath` semantics,
which is a correctness risk taken for a factor of two", was a real cost honestly stated and it was
still the wrong call: the file already reimplements two standard-library escapers this way
(`queryEscapedLen`, `safeLogValueLen`), each pinned per byte against the thing it replaces, and the
reviewer had built and validated a prototype over 5000 randomized cases before reporting. The risk
was measurable rather than hypothetical, and measuring it is what this workflow is for.

**How it was fixed, keeping the output byte-identical.** `escapedPathForLog` reproduces both of
`EscapedPath`'s branches and returns a clipped head plus the true rendered length, which
`clipped.writeCounted` then carries into the target's marker:

- The `RawPath` branch, where `net/url` returns the client's own encoding verbatim, needs no
  rendering at all. `rawPathEncodes` walks `RawPath` and `Path` together to decide validity and
  equality in one pass, where `net/url` builds the decoded string only to compare it and discard it.
- The escaping branch escapes only as far as the clip reaches. Each byte escapes independently to
  one or three bytes, so the first `limit` input bytes always yield at least `limit` output bytes
  and they are exactly the ones kept. `pathEscapedLen` counts the rest.
- `net/url` exports no whole-path escaper, which is why this is written out rather than called:
  `url.PathEscape` uses the path-segment rule and escapes `/` as well.

**The one remaining scan is inherent and is named as such.** `pathEscapedLen` is proportional to the
whole path however little survives, because decision 4's marker reports the true rendered length and
that cannot be known without looking at every byte. Turning the thirteen-case switch into a table
derived from it took that scan from 505 to 317 microseconds on 900000 bytes.

**Measured on this machine**, a 900000-byte path of `0x80`, at `-benchtime 300x`:

| what | before | after |
|---|---|---|
| `RequestTargetForLog`, the escaping branch | 5,416,840 B | **0.33 ms, 17,266 B** |
| `url.URL.EscapedPath` alone, which the before column is all of | 1.06 ms, 5,406,721 B | not called on this path any more |
| `RequestTargetForLog`, the `RawPath` branch, 300000 `%8a` | | **0.42 ms, 4,945 B** |
| the whole middleware, output to `io.Discard` | 1.13 ms, 5,433,150 B (the reviewer's) | **0.35 ms, 17,826 B** |
| chi's `middleware.Logger`, the code this replaced | 0.20 ms, 2,708,817 B, and it wrote 900,075 bytes of it to the log | |
| `url.ParseRequestURI`, inside `net/http` before any of this runs | 1.78 ms, 5,406,882 B | unchanged, inherent |

So allocation is 313 times lower than before this round and 152 times lower than chi's. **CPU is not
lower than chi's and this round does not claim it is**: 0.35 ms against 0.20 ms, on a request where
`net/http` has already spent 1.78 ms parsing. The logger adds 20% to a cost the server pays whether
or not logging is on, against chi's 11%, and chi bought that 11% by writing 900 KB to the log. The
input bound is still `MaxHeaderBytes`, follow-up 4, unchanged by this round.

**Documentation:** none owed. The output is byte-identical, which is the whole claim, so no page
that describes what is logged is falsified. `docs f3584cb70ebd` is unmoved from round 1.

### Tiers, at `production 82a060e53322`

| Tier | Command | Scope | Result |
|---|---|---|---|
| Unit | `where.sh test --type modules` | all three modules | **3083 pass, 0 fail**, 47 packages (3075 before this round) |
| Integration | `where.sh test --type integration --db sqlite` | sqlite only | **1221 pass, 0 fail** |
| Data | not run | | nothing here touches `data.Database`, a query, a migration or the schema; `data-tests feaf2dad58c8` is unmoved from stage 2 |
| Anchors | `check-anchors.sh` | all 18 rows | PASS |
| Site | not run | | `docs f3584cb70ebd` unmoved, no page changed |

`tree-hash.sh --split`: `production 82a060e53322`, `unit-tests 0dc740ea3d73`,
`data-tests feaf2dad58c8`, `integration-tests c7279c9e9c3a`, `docs f3584cb70ebd`,
`all b157123793e6`.

The integration result was taken at `production 82a060e53322` and
`integration-tests c7279c9e9c3a`, both unmoved since, so it stands. The unit tier was re-run after
the last three table rows landed, since `unit-tests` had moved.

### Mutations

Eight, each through `where.sh mutate`, which restores from a trap and compares by hash. All
**CAUGHT**.

| # | Replacement in `middleware_request_logger.go` | Result |
|---|---|---|
| 1 | `pathByteIsSafe`: drop `/` from the unescaped set | CAUGHT |
| 2 | `pathEscapedLen`: escape width `n += 3` → `n += 2` | CAUGHT |
| 3 | `pathSafeBytes`: derive the table from nothing, `safe[c] = true` | CAUGHT |
| 4 | `escapedPathForLog`: the pre-fix unbounded form verbatim, `clip(u.EscapedPath(), limit), len(u.EscapedPath())` | CAUGHT |
| 5 | `escapedPathForLog`: escape a shorter prefix than the clip keeps, `clip(u.Path, limit/2)` | CAUGHT |
| 6 | `escapedPathForLog`: take the `RawPath` branch without validating it | CAUGHT |
| 7 | `clipped.writeCounted`: count what it kept rather than what it stood for, `c.n += len(head)` | CAUGHT |
| 8 | `rawPathEncodes`: `return decoded == len(path)` → `return true`, so a `RawPath` encoding a prefix of `Path` is accepted | CAUGHT |

4 is the load-bearing one and it is worth stating what caught it: **only**
`TestRequestTargetForLog_CountsWhatItDiscardsRatherThanBuildingIt/an_enormous_path,_escaped`. Every
output assertion in the file passes against the unbounded code, which is what makes the bound
invisible to review and is the reason it is measured rather than asserted structurally. A ninth
attempt, neutralising `rawPathEncodes`'s comparison with `if false`, is not counted: it left `b`
declared and unused, so it did not compile and the tool refused to score it.

`production` was re-hashed after the last mutation and matches the value the tiers ran at, so
nothing was left behind.

### Round 3

Reviewer `gpt-5.6-sol` at effort `xhigh`, resumed session, request `159-20260811T120022Z`. Verdict
`findings`: one, significant, quality, `needs_human: false`, acted on and not refuted. `unchecked`
was empty, so there is nothing to cover, dismiss or file from it. Fixed on the branch as the commit
this entry lands in.

**This round changed no production code, so it is the gate's last.** `production 82a060e53322` is
the same value round 2 was reviewed at, and the round budget funds another round only on a moved
`production` hash. That is the right answer here rather than a budget cutting the review short: the
finding is about the tests and the reviewer says so itself.

**Finding 1, significant, quality: three of the `net/url` decisions round 2 copied in were not
pinned.** CONFIRMED, and confirmed by re-running the reviewer's own mutations here before writing
anything. All three survived the submitted suite:

| Mutation in `middleware_request_logger.go` | Before this round |
|---|---|
| `rawPathByteIsLiteral` admits `0x22`, a byte `net/url` escapes | SURVIVED |
| `isHexDigit` admits `G` | SURVIVED |
| `unhexDigit` decodes `A`-`F` as 11-16 rather than 10-15 | SURVIVED |

**The production code is correct and was correct.** The reviewer established that separately, over
an overlay covering every literal byte, every valid `%XX` in both cases, all 256 classifier inputs,
mixed encodings, varied limits and 10,000 randomized pairs, and the tables below now say the same
thing inside the repository. What was missing was the ability to *notice* if it stopped being
correct, which is exactly the guarantee round 2 spent itself buying: the file reproduces `net/url`
rather than calling it, and the only thing that makes that acceptable is a test pinning it against
the standard library.

**Why the existing tests could not see it.** `TestEscapedPathForLog_MatchesEscapedPathOnRandomInput`
draws from a fixed 38-byte alphabet, so it covers the shapes nobody thought of but only the ones
that alphabet can spell, and `0x22` is not in it. The uppercase decoder is subtler and worth
stating, because it is the reason a sampling test is the wrong instrument here: when the decoder gets
a byte wrong the `RawPath` is rejected and the renderer falls back to escaping `Path`, which writes
the same escape back in upper case. **A wrong answer rejoins the right one.** An escape has to be
non-canonical somewhere else in the same path for the two branches to differ at all, and a random
alphabet produces that pairing rarely and never deliberately.

**What was added**, all four pinned against the standard library over a whole finite domain rather
than a sample, since the domains are 256 wide and cost milliseconds to walk:

| Test | Domain | Closes |
|---|---|---|
| `..._OnEveryLiteralByte` | 256, `Path` == `RawPath` | which bytes may stand for themselves |
| `..._OnEveryEscapedByte` | 256 × upper and lower case hex | what each escape decodes to, in both digit positions |
| `..._OnEveryHexCandidate` | 256 candidate digits × 256 bytes each could be taken to mean | which bytes are hex digits at all |
| `TestHexDigits_MatchTheStandardLibrary` | 256, against `strconv.ParseUint` | the same, stated as the rule rather than as an output |

Three of those four assert on output, through `escapedPathForLog` against `url.URL.EscapedPath`, so
they pin the observable behaviour and not the shape of the implementation. The fourth is the file's
existing idiom, the one `TestQueryEscapedLen_MatchesTheStandardLibrary` and
`TestPathEscapedLen_MatchesTheStandardLibrary` already use: say what the right answer is, per byte,
against the standard library.

**Two details worth keeping**, because both are the difference between a test that looks thorough
and one that works:

- **The canary.** Every escaped-byte row appends `/%61` to the `RawPath`, a non-canonical escape of
  `a`, so that accepting the `RawPath` and rejecting it can never render the same string. Without it
  the whole escaped-byte table is green against a broken decoder, for the fallback reason above.
- **The classifier needs the decoded byte varied too.** A classifier that admits a byte `net/url`
  refuses decodes it to *something*, and which byte that is depends on the defect, so no single
  `Path` exposes it. Varying both positions, 65,536 rows in 40 ms, contains the exposing row
  whatever the defect turns out to be.

`assertRendersLikeEscapedPath` now carries the equivalence for every table including the two that
existed, and runs each at seven limits (`0, 1, 3, 4, 5, 17, 4096`) rather than at `maxLoggedTarget`
alone, so a cut landing inside a `%XX` escape is covered by every row rather than by the corpus. Two
non-canonical rows were added to `pathCorpus` as well, including the reviewer's own
`{Path: "/A/\x8a", RawPath: "/%41/%8a"}`.

**Documentation:** none owed. No production code moved, so no page that describes what is logged is
falsified. `docs f3584cb70ebd` is unmoved from round 1.

### Tiers, at `production 82a060e53322`

| Tier | Command | Scope | Result |
|---|---|---|---|
| Unit | `where.sh test --type modules` | all three modules | **3184 pass, 0 fail**, 47 packages |
| Integration | not re-run | | one test file in `src/core/middleware` moved; no handler, route or schema. `production 82a060e53322` and `integration-tests c7279c9e9c3a` are both unmoved from round 2, so round 2's 1221 pass stands |
| Data | not run | | nothing here touches `data.Database`, a query, a migration or the schema; `data-tests feaf2dad58c8` unmoved from stage 2 |
| Anchors | `check-anchors.sh` | all 18 rows | PASS |
| Site | not run | | `docs f3584cb70ebd` unmoved, no page changed |

`tree-hash.sh --split`: `production 82a060e53322`, `unit-tests d563849ca064`,
`data-tests feaf2dad58c8`, `integration-tests c7279c9e9c3a`, `docs f3584cb70ebd`,
`all 3dec5bfe7c20`. Only `unit-tests` moved, which is the whole claim of this round.

**On the unit count, because 3083 in the round 3 request and 3184 here are not comparable.** They
were counted by different sessions and evidently by different expressions. Re-run on the reviewed
tree in this session, with the test file stashed and `grep -c -- "--- PASS"`, the same tier reports
**3180**, and with the new tables **3184**: exactly the four new tests, with the delta confirmed
independently in the one package that moved (`go test ./middleware/ -v`, 274 → 278). The absolute
figure is only meaningful alongside the expression that produced it.

### Mutations

Ten, each through `where.sh mutate` on `--type modules`, which restores from a trap and compares by
hash. All **CAUGHT**. The first three are the reviewer's survivors; the rest enumerate the remaining
members of the family rather than the instance reported, since a family the last round called
exhausted had just been contradicted.

| # | Replacement in `middleware_request_logger.go` | Member | Result |
|---|---|---|---|
| 1 | `rawPathByteIsLiteral` also admits `'"'` | literal admission, false positive | CAUGHT |
| 2 | `isHexDigit`: `'A' <= c && c <= 'F'` → `<= 'G'` | hex classification, false positive | CAUGHT |
| 3 | `unhexDigit`: `c - 'A' + 10` → `+ 11` | decoder, upper case | CAUGHT |
| 4 | `unhexDigit`: `c - 'a' + 10` → `+ 11` | decoder, lower case | CAUGHT |
| 5 | `unhexDigit`: `c - '0'` → `c - '0' + 1` | decoder, decimal | CAUGHT |
| 6 | `rawPathEncodes`: `i+2 >= len(rawPath)` → `i+2 >` | percent syntax, truncated escape | CAUGHT |
| 7 | `rawPathEncodes`: `path[decoded] != b` → `path[decoded] > b` | decoded-byte equality | CAUGHT |
| 8 | `rawPathEncodes`: `return decoded == len(path)` → `return true` | decoded-length equality | CAUGHT |
| 9 | `rawPathByteIsLiteral` drops `'*'` | literal admission, false negative | CAUGHT |
| 10 | `escapedPathForLog`: `if u.Path == "*"` → `if false` | the asterisk-form target | CAUGHT |

8 and 10 were caught before this round too and are re-run rather than assumed; 9 is the false-negative
direction of 1, which the same table closes only because `!'()*[]` are bytes the escaper would
rewrite. A false negative on a byte that is *already* safe to leave alone is unobservable in the
output, because both branches then render the same string, so there is nothing there to catch.

`production` was re-hashed after the last mutation and still reads `82a060e53322`, the value round 2
was reviewed at and the value the tiers above ran at, so nothing was left behind.

Issue #209, opened 2026-08-12
Nothing checks that every audit constant is registered in AuditEventTypes


leodip:
## Summary

`AuditEventTypes` in `src/core/constants/constants.go` is hand-maintained alongside the `Audit*` constants, and nothing checks the two agree. An event added to the const block but forgotten in the slice silently disappears from the admin UI's filter dropdown, with no test failing.

## Details

The slice describes itself as canonical:

```go
// AuditEventTypes is the canonical list of all audit event type strings.
// Used by the admin UI filter dropdown.
var AuditEventTypes = []string{
```

`constants_test.go` has `TestAuditEventTypes_Uniqueness` and `TestAuditEventTypes_NonEmpty`. Both check properties of the slice's contents; neither checks that the slice is complete with respect to the constants.

The two agree today, 97 and 97:

```bash
diff <(grep -oE '^\tAudit[A-Za-z0-9]+ +=' src/core/constants/constants.go | tr -d '\t =' | sort) \
     <(awk '/^var AuditEventTypes/,/^}/' src/core/constants/constants.go | grep -o '^\tAudit[A-Za-z0-9]*' | tr -d '\t' | sort)
```

Note the `[A-Za-z0-9]` character class matters: `AuditUpdatedClientOAuth2Flows` contains a digit, and a letters-only pattern truncates it and reports a spurious difference.

## Impact

Low and latent. The audit event is still written and still persisted; it is only missing from the filter dropdown, so an operator filtering the audit log viewer cannot select it and may conclude the event was never recorded. The failure is silent, which is the part worth fixing: every new audit event is an opportunity for it.

## Suggested fix

Add a completeness test beside the two existing ones. Parsing `constants.go` for `Audit*` declarations and asserting each appears in the slice would do it, as would generating the slice instead of maintaining it.

_Found while sizing the fanout of #79, which adds an audit event and therefore has to touch both places._

leodip:
Correction found while working #79. "Nothing checks that the two agree" is not accurate: `TestAuditEventTypes_MatchesConstants` and `TestAuditEventTypes_Count` both failed the moment a new `Audit*` constant was added, so the suite is not silent.

[...]

Issue #208, opened 2026-08-12
Authorize error page renders a hardcoded English heading beside a localized page title


leodip:
## Summary

The authorize error page mixes languages: its page title comes from the message catalog, its heading is a hardcoded English string. A pt-BR user meeting an authorize error sees a localized tab title and an English heading.

## Details

`renderErrorUi`, the closure inside `HandleAuthorizeGet` in `src/authserver/internal/handlers/handler_authorize.go`, binds the heading as a Go string literal:

```go
renderErrorUi := func(message string) {
    bind := map[string]interface{}{
        "title": "Unable to authorize",
        "error": message,
    }

    err := httpHelper.RenderTemplate(w, r, "/layouts/no_menu_layout.html", "/auth_error.html", bind)
    ...
}
```

`auth_error.html` renders `{{.title}}` as its `<h1>`, and separately renders its `<title>` from the catalog key `auth_error.page_title`, which exists in both locales:

```
$ grep -n "auth_error" src/core/i18n/catalogs/active.en.toml src/core/i18n/catalogs/active.pt-BR.toml
src/core/i18n/catalogs/active.en.toml:92:"auth_error.page_title" = "Error"
src/core/i18n/catalogs/active.pt-BR.toml:90:"auth_error.page_title" = "Erro"
```

There is no catalog key for the heading, so nothing to translate it with.

## Impact

Cosmetic, and only on the error path, but it is a user-visible language mix on a page a non-English user can reach through an ordinary bad authorize request.

## Suggested fix

Add a key for the heading to `active.en.toml` and `active.pt-BR.toml` and render it through `i18n.T`, the way the page title already is. Worth checking the other render sites of `auth_error.html` for the same pattern while there.

_Found while grounding #79, which adds its own strings to both catalogs and does not touch this call site._

Issue #112, opened 2026-07-25
Reset and activation links break for email addresses containing '+' (not URL-encoded)


leodip:
## Summary

Email addresses are concatenated raw into the query string of password-reset and account-activation links. Because `+` decodes to a space in a query string, any user whose address contains a `+` receives a link that cannot work. The email validator explicitly permits `+`, so these are addresses the system happily accepts at registration.

This is a **functional bug, not a security vulnerability**. Parameter injection is not possible (see below), but it locks a class of users out of account recovery and self-registration.

## Affected code

Three sites, all with the same shape:

`src/authserver/internal/handlers/handler_forgot_password.go:90` (forgot-password reset link)

```go
"link": config.GetAuthServer().BaseURL + "/reset-password?email=" + user.Email + "&code=" + verificationCode,
```

`src/authserver/internal/handlers/apihandlers/handler_api_users_crud.go:429` (admin-created user, reset link)

```go
"link": config.GetAuthServer().BaseURL + "/reset-password?email=" + createdUser.Email + "&code=" + verificationCode,
```

`src/authserver/internal/handlers/accounthandlers/handler_account_register.go:187` (self-registration activation link)

```go
"link": config.GetAuthServer().BaseURL + "/account/activate?email=" + email + "&code=" + verificationCode,
```

## Why `+` addresses break

`ValidateEmailAddress` (`core/validators/email_validator.go:29`) allows `+` in the local part:

```go
pattern := `^[a-zA-Z0-9._%+-]+@[a-zA-Z0-9.-]+\.[a-zA-Z]{2,}$`
```

So `user+tag@example.com` registers successfully. The generated link is then:

```
https://auth.example.com/reset-password?email=user+tag@example.com&code=...
```

On the way back in, `r.URL.Query().Get("email")` decodes `+` to a space, yielding `user tag@example.com`. `GetUserByEmail` finds nothing, and:

- `handler_reset_password.go:129` and `:223` treat the unknown address exactly as an invalid code, so the user sees "code invalid or expired" with no indication of the real cause. That indistinguishability is deliberate and correct for enumeration resistance, but here it also hides a bug.
- `handler_account_activate.go` fails the same way, so a `+` user cannot complete self-registration at all.

The verification code itself is unaffected: `GenerateSecurityRandomString` uses the alphabet `0-9a-zA-Z-_.`, all URL-safe.

## Why this is not an injection vector

[...]

PR #210, opened 2026-08-12
fix(auth): grant only the scopes the user checked, and bind every auth form to its ceremony

### LINKED ISSUE #79: Consent checkbox substring match over-grants scopes (>=11 scopes)
## Summary

The consent POST handler matches submitted checkbox names with `strings.Contains`, which does a substring match. With 11 or more requested scopes this over-grants: a scope the user unchecked can still be granted.

Severity: **Low**

## Details

`src/authserver/internal/handlers/handler_consent.go:248`:

```go
if strings.Contains(consented, fmt.Sprintf("consent%v", idx)) {
    consent.Scope = consent.Scope + " " + scope.Scope
}
```

Checkbox names are `consent0`, `consent1`, ... `consentN` (`src/authserver/web/template/consent.html:36`). Unchecked checkboxes are not submitted, so `consented` is the space-joined set of checked names.

`strings.Contains(consented, "consent1")` also matches `consent10`, `consent11`, `consent12`, ... So if a user unchecks scope index 1 (to deny it) but leaves scope index 10 checked, `consented` contains `consent10`, the substring check for `consent1` returns true, and scope index 1 is granted anyway — overriding the user's denial.

## Impact

Consent-integrity bug: the persisted consent can include scopes the user explicitly declined. Scopes are still bounded to what was requested and what the user is authorized for, and it requires >= 11 requested scopes to trigger, hence Low.

## Suggested fix

Match exact tokens instead of a substring, e.g.:

```go
if r.Form.Has(fmt.Sprintf("consent%d", idx)) { ... }
```

or build a `map[string]bool` of submitted `consent*` keys and look up the exact key.

## Location

- `src/authserver/internal/handlers/handler_consent.go:248`

_Found during a structured security review (Area A: OAuth2/OIDC protocol correctness)._

leodip:
## Issue 79: the consent screen grants scopes the user unchecked

Closes #79.

`POST /auth/consent` decided which boxes were ticked with a substring match. The checkbox names are
positional, `consent0` to `consentN`, and every submitted name was joined into one string that each
index was then searched for with `strings.Contains`. `consent1` is a substring of `consent10`, so
from 11 scopes upwards a scope the user deliberately unchecked was granted anyway.

Sweeping every scope count from 1 to 40 with exactly one box unchecked, 820 cases, the smallest
count where a denial is over-granted is **11**. With 11 scopes and `profile` unchecked, the user
submits `[0 2 3 4 5 6 7 8 9 10]` and the server grants `[0 1 2 3 4 5 6 7 8 9 10]`. The same rule
generalises: index `n` is over-granted whenever a ticked index has `n` as a prefix of its decimal
form, so `consent2` is caught by `consent20..29` and `consent3` by `consent30..39`.

That is not a display bug. The granted set is written to `user_consents`, copied into the auth
context, and becomes the authorization code's scope and then the access token's. **A denial
over-granted once is remembered for every later authorization by that client.**

Eleven scopes is ordinary. Nothing caps the count, the validator accepts 100, and the consent screen
lists up to 8 built-in OIDC scopes plus one row per resource-permission the client asked for and the
user holds.

### Two more gaps found while grounding the fix

**No form in the authorization flow names the request it was rendered for.** A browser holds one auth
context, under one session key, replaced wholesale by the next `/auth/authorize`. So a second
authorization in another tab silently takes over the first, and every form already on screen still
posts to the same URL. Two reachable consequences, both with an ordinary browser:

- The user reaches client A's consent screen, client B opens in another tab, and the user unchecks a
  scope on A's screen and submits. The remaining ticks resolve against **B's** scope list, so B is
  granted everything it asked for and a consent row is persisted for it. The user approved a screen
  naming A and never saw B's.
- The user is at A's password screen when B takes over. Submitting the password finishes **B's**
  authorization, and where B requires no consent the code is issued without the user seeing any
  screen at all. The OTP step has the same shape.

**An empty consent meant the full requested scope.** Refusal was decided by whether any key starting
with `consent` arrived, not by whether any of them matched a scope. A submitted key matching no index
passed that gate, persisted an empty consent, and both issuers read an empty consented scope as "no
consent screen was shown" and fall back to everything the client requested.

### The rule this establishes

A grant carries exactly the scopes the user left ticked, on the screen they were actually looking at.
Checkbox names are matched exactly and only from the request body, refusal is decided by what was
consented to rather than by what arrived, and every authorization-flow form carries the id of the
ceremony that rendered it. A submission naming a different ceremony is refused with a 400 and
**leaves the live ceremony untouched**, so a forgotten tab cannot cancel an authorization the user
still wants. That last mechanism is the one #112 established for emailed links, reused here.

### Stage 1: exact-key consent selection in `HandleConsentPost`
Landed in `bfc40b91`.

The `btn == "submit"` branch of `HandleConsentPost` now resolves the selection from
`authContext.Scope` alone, before anything else happens, looking each index up exactly with
`r.PostForm.Has(fmt.Sprintf("consent%d", idx))`. The joined string and its `strings.HasPrefix`
collection loop are deleted: an exact lookup has nothing left to use them for.

Two behaviours change with it, both of them gaps this closes rather than side effects:

- **`r.PostForm`, not `r.Form`.** `r.Form` merges the URL query and the consent form posts to
  `action=""`, so `/auth/consent?consent5=on` counted as a tick regardless of the box.
- **Refusal is decided by matched scopes**, `len(grantedScopes) == 0`, rather than by whether a key
  named `consent...` arrived. A submitted key naming no index used to pass that gate, persist an
  empty consent row and leave both issuers to fall back to the full requested scope. That fallback
  is correct and stays, since it is how a client with `ConsentRequired = false` gets its scope at
  all, which is why the gap closes in the handler and never at the issuers.

The `consent.Scope = ""` blanking on the existing-consent path is gone too: `strings.Join` assigns
the whole selection, which is what the blanking existed to ensure. `buildScopeInfoArray` moved above
the refusal gate and takes `nil` for the consent, which is behaviourally identical (its `consent`
argument only sets `AlreadyConsented`, which this path never reads) and costs no database call, so
the gate keeps its position ahead of the client and user lookups and #141's clear-then-redirect
ordering is untouched.

**Tests.** 24 table-driven unit cases at `POST /auth/consent`: the collision boundary from both
sides (10 scopes and 11), every index of an 11-scope request denied in turn, 21-scope requests
catching `consent2` inside `consent20`, indices submitted out of order, a key naming no index, a
non-numeric key, the query-alone and query-beside-body paths, and a body Go will not parse. Plus one
case pinning that an existing consent row is replaced rather than appended to. One integration case,
`..._ElevenScopes_DeniedScopeIsNotGranted`, walks an 11-scope authorization end to end with index 1
unchecked and asserts the denial survives into both the issued code's scope and the `user_consents`
row.

Unit 3298 pass across all three modules, integration 1230 pass on sqlite. The data tier is not
applicable: no query, interface method, migration or schema change, and nothing under
`src/core/data/` is touched. Four-engine data and integration are the check suite's on `bfc40b91`.

Five mutations were run against the new tests. Four were caught. The fifth, `idx` to `idx%10` at the
integration tier, **survived**, and that is a fact about the mutation rather than about the test: for
that one submission index 1 looks up an absent `consent1` and index 10 looks up a present `consent0`,
so both land on the right answer by coincidence. Replaced by one that reinstates the exact
`consent1`-inside-`consent10` collision the issue reports, which the integration case catches.

### Stage 2: the ceremony id, and the consent form bound to it

Landed in `e11d74c4`. The mechanism, plus the one form whose content is meaningless outside its own
ceremony.

A browser holds **one** auth context, under a single session key, created fresh at every
`/auth/authorize`. So a second authorization request replaces the first while the first client's
consent screen is still on screen, both screens post to the same URL, and the checkbox names are
positional. The boxes ticked on client A's screen therefore resolved against client B's scope list:
B was granted the scopes at those positions and a consent row was persisted for B, off a screen
naming A that the user was never shown for B. Where the submitted indices exceeded the new list, the
same gap instead collapsed the submission into a refusal and cancelled an authorization the user
meant to approve.

No rule about *writing* the auth context can close that, because the retarget happens to a page that
is already rendered. The page has to name the ceremony it belongs to and the POST has to check it,
which is the shape #112 established for the emailed-link flows.

- `AuthContext` gains `CeremonyId`, generated with `GenerateSecurityRandomString(32)` in
  `HandleAuthorizeGet`, the only place an auth context is created, and refused as a 500 if the
  CSPRNG is unavailable rather than saved empty.
- `consent.html` renders it as a hidden input **inside** its `<form>`, `HandleConsentGet` binds it,
  and `HandleConsentPost` checks it immediately after `GetAuthContext`: **before** the `AuthState`
  check, so a stale form gets a 400 rather than a 500 naming an internal invariant; **before** the
  `btnSubmit`/`btnCancel` dispatch, so a stale cancel cannot clear the live ceremony's context; and
  from `r.PostFormValue` rather than `r.FormValue`, since the form posts to `action=""` and a query
  parameter is not a submission.
- A mismatch audits a new `auth_ceremony_mismatch` event and renders `auth_error.html` at 400, with
  two new catalog strings in both locales. **The auth context is deliberately left untouched**: the
  client that would receive an error is the current one, whose authorization the user still wants,
  so a forgotten tab must not be able to cancel it.
- `ceremonyMatches` refuses an **empty stored id** rather than matching it against an empty
  submission. An auth context written by the previous binary carries none, so a user mid-flow across
  the deploy is refused once and starts the authorization again. That is the fail-closed direction
  and it is bounded by the session cookie's life.

Tiers: unit (3315 subtests) and integration on sqlite (1231) locally, plus the data tier on all four
engines. Five mutations were run against the new seams and all five were caught. The most useful was
on the template: moving the hidden input **outside** the `<form>` element, which no browser would
submit but which a document-wide scrape in a test would happily find anyway. The integration helper
scopes its selector to `form input[name="ceremonyId"]` and requires exactly one match, so it fails,
which is what makes the end-to-end case evidence about a browser rather than about a parser.

Five integration tests that hand-built a consent body were converted to the shared helper. They had
been invisible to the fanout count, and each began failing the moment the binding landed.

### Stage 3: the password and OTP forms bound to the ceremony

Landed in `54fe826b`. The same binding on the other three forms in a ceremony, which is what
decision 6 settled and what the password case is the reason for.

On the login form the defect is more direct than on the consent screen. The user is at client A's
login screen when client B opens in another tab and replaces the auth context. They type their
password into A's screen: before this change that authenticated them against **B's** request, and
because B requires no consent the flow ran straight through to `/auth/issue` and B was handed a code
with no screen ever shown for it. The OTP prompt has the same shape one factor up.

- `auth_pwd.html`, `auth_otp.html` and `auth_otp_enrollment.html` each render the ceremony id as a
  hidden input inside their `<form>`, and all five remaining renders bind it: `HandleAuthPwdGet`,
  `HandleAuthPwdPost`'s `renderError`, both branches of `HandleAuthOtpGet`, and
  `HandleAuthOtpPost`'s `renderError`, which picks between the two OTP templates at run time.
- **The error re-renders matter as much as the first render.** Without them one mistyped password or
  code would end the ceremony, because the retry would name no ceremony and be refused. Three
  existing replay tests now resubmit from the refusal's own re-render, which is the page a user who
  mistyped a code is actually looking at, so that path is exercised by tests that already existed.
- Both handlers check the id immediately after `GetAuthContext`, from `r.PostFormValue`. That is
  before the `AuthState` check, and **before any credential is looked at**: no `GetUserByEmail` and
  no password verification on one path, and no `otp.MatchStep` and therefore no
  `TryConsumeUserOTPStep` on the other. The second one is a promise in its own right: a stale
  submission that reached the claim would consume the step of a passcode the live ceremony still
  needs, so a forgotten tab could make a correct code stop working in the tab the user is on.
- The two strict `len(bind)` assertions in the OTP tests moved from 7 to 8 and from 9 to 10, and
  both now assert the id's value rather than only the map's size. The two integration helpers take
  the rendered page and read the id off it, as `postConsent` already did, across 123 call sites.

Tiers: unit (3327 subtests) and integration on sqlite (1233) locally, plus the data tier on all four
engines. **Ten mutations** were run against the new seams and all ten were caught, including moving
the hidden input outside the `<form>` in each of the three templates. Three further claims are
pinned by cases rather than mutations, because moving a check is a structural edit: both stale-form
tables submit **correct** credentials against a database mock with no expectations at all, so a
check below the credential lookup fails by being called, and the end-to-end OTP case resubmits the
refused code on the live prompt and requires it to work.

### After the review: two fixes on top of stage 1

Landed in `0330db93` and `77ebb1ce`. The whole branch was reviewed once after the last stage, and
the reading found a defect in stage 1's work that the stage's own tests had not reached.

**`0330db93` closes a second query-parameter hole, this one in the action controls.**
`HandleConsentPost` decided approve-versus-cancel with `r.FormValue`, which merges the URL query into
the body. The consent form posts to `action=""`, and a button that was not clicked sends no key at
all. So on a consent screen reached at `/auth/consent?btnSubmit=submit`, clicking **Cancel** posted
`btnCancel=cancel` in the body while `FormValue("btnSubmit")` found `submit` in the query, and the
approval branch ran. Every checkbox is `checked` by default, so a refusal granted the **full**
requested scope, persisted a `user_consents` row and issued a code.

Both controls now read from the body alone, and approval requires the submit control on its own:
`r.PostForm.Get("btnSubmit") == "submit" && !r.PostForm.Has("btnCancel")`. A body carrying both keys,
which no browser sends, is refused rather than resolved in favour of granting.

This is the same family decision 2 settled for the checkboxes, only the body counts as consent, left
unfixed on the action controls. The family was then swept: presence-based reads are the vulnerable
ones, because an absent body key lets a query value win, while value reads on always-submitted
inputs (`email`, `password`, `otp`) cannot be overridden. The consent buttons were the only
remaining member; the other three forms have unnamed buttons.

**`77ebb1ce` pins the predicate.** Three deliberately wrong versions of that one line left the suite
green: reading presence rather than the exact value, approving when neither action key is present,
and reading the cancel control from merged `r.Form`. The first two would let a hand-built request
grant scopes and the third would cost a legitimate grant. Three cases now pin exactly those, and all
three mutations are caught.

## Review

Reviewed by `gpt-5.6-sol` at effort `xhigh`, three rounds: one at the plan gate before any code
existed, and two over the whole branch after the last stage landed. **No stage was reviewed on its
own.**

|---|---|---|---|
| plan gate | 2 | 0 | both fixed in the plan, before stage 1 started |
| final 1 | 1 | 1 | fixed, `0330db93` |
| final 2 | 2 | 1 | one fixed, `77ebb1ce`; one confirmed as fact, refuted as this change's defect, filed |

- **blocking, security:** a query parameter turned Cancel into consent, granting the full requested
  scope off a refusal. Fixed in `0330db93`, as described above. Not escalated, because decision 2
  had already settled the invariant and the reviewer's `forced_answer` named the same repair.
- **blocking, security:** unauthenticated callers can drive unbounded audit writes through ceremony
  mismatches, with rate limiting off by default. Every mechanical step re-verified and true. Refuted
  as *this change's* defect: the path this branch adds is the cheapest row in a table whose worst
  member, `POST /account/register`, predates the branch and carries no limiter any setting can
  switch on. Filed as a follow-up with the full table. The review was right that section 2's stated
  reason for excluding rate limiting, "the refusal is cheap and stateless", is wrong, since decision
  5 requires the refusal to be audited.
- **significant, quality:** three branches of the new action predicate were unpinned. Fixed in
  `77ebb1ce`.
- **significant, quality (plan gate):** the hand-run command guarding the audit-event registry could
  never fail. GNU grep does not read `\t` in an extended regular expression, so both sides of the
  `diff` extracted zero entries and it exited 0 whatever the file contained. Replaced with an awk
  comparator that asserts a non-empty extraction at an expected count.
- **minor, conformance (plan gate):** the change falsifies a sentence in the documentation site
  promising "nothing to go stale in a page left open". A documentation edit answering it landed with
  stages 2 and 3 and was then reverted in `76db9b0f`, so the sentence stands and this finding is
  open. The branch carries no documentation change.

Full account, with the evidence and the refutations, in
[this comment](https://github.com/leodip/goiabada/pull/210#issuecomment-5268670229).

## Tests

**Locally**, on the final tree: unit **3335 pass, 0 fail** across 47 packages; data **1615 pass, 0
fail** on sqlite, mysql, postgres and mssql; integration **1233 pass, 0 fail** on sqlite. All 28 code
anchors resolve.

**In CI**, green on `77ebb1ce`, **11 of 11 jobs**
([run](https://github.com/leodip/goiabada/actions/runs/31610021522)): Build validation, Lint,
Vulnerabilities, Versions, the three Unit jobs, and Tests on sqlite, mysql, postgres and mssql, each
running the data **and integration** tiers on its engine. The four-engine integration result comes
from there; nothing but sqlite was run locally at that tier.

The data tier is not applicable to the change itself: no query, interface method, migration or schema
change, and nothing under `src/core/data/` is touched. It is reported because it runs anyway.

**28 mutations** were run across the branch, one per behaviour the tests claim to have made
observable: 5 at stage 1, 5 at stage 2, 10 at stage 3, 5 answering the final review's round 1 and 3
answering its round 2. All were caught except one at stage 1, which survived for a reason that was a
fact about the mutation rather than about the test (`idx` to `idx%10` made index 1 and index 10 land
on the right answer by coincidence) and was replaced by one reinstating the exact
`consent1`-inside-`consent10` collision the issue reports. Eight of the final eleven had survived for
the reviewer before the cases that now catch them were written.

## Worth knowing

- **The upgrade is user-visible, once.** An auth context created by the previous binary carries no
  ceremony id, so a user mid-flow across the deploy is refused and starts the authorization again.
  That is the fail-closed direction, chosen over accepting an empty id, and it is bounded by the
  session cookie's life.
- **Two authorizations at once now have to be finished one at a time**, and the older tab errors
  rather than silently acting on the newer request. That is the same cost #112 accepted for the
  emailed-link flows.
- **The branch changes no documentation.** The ceremony binding is not described on the
  documentation site, which still says a page left open has nothing to go stale in it.
- **Five integration tests that hand-built a consent body were converted to the shared helper** in
  stage 2. They were invisible to the fanout count and each began failing the moment the binding
  landed.
- **No deferred decisions and no blocking escalations.** Details in
  [this comment](https://github.com/leodip/goiabada/pull/210#issuecomment-5268667928).
- **Four follow-ups, all now filed**: #208, #209, #211 and #212. The correction owed to #209's
  opening claim has been posted as a comment on it. The reasoning behind each, and the duplicate
  searches, are in
  [this comment](https://github.com/leodip/goiabada/pull/210#issuecomment-5268669242).

## Status

**Complete and ready for review.** Six commits: three stages, `bfc40b91`, `e11d74c4` and
`54fe826b`, then two fixes answering the final review, `0330db93` and `77ebb1ce`, then `76db9b0f`
reverting the documentation edit.

A grant now carries exactly the scopes the user left ticked, on the screen they were actually
looking at. The check suite is green across all 11 jobs on `77ebb1ce`, the last commit to touch
code, and the code review is closed: three rounds, five findings, three resolved on the branch, one
filed as belonging elsewhere and one left open as documentation.


leodip:
## Deferred decisions: none

Nothing on this branch is waiting on an answer from you.

Every question the run met was either settled in the seven decisions sealed before it started, or
forced by evidence at the gate that raised it. The run posted **zero blocking escalations** and
deferred **zero** decisions, so there is no assumption to confirm before merging.

Two findings did arrive after the seal, both from the final review, and both were resolved inside
the run rather than deferred. The reasoning is in the review comment, and in short:

- The **round 1 blocking defect** (a query parameter turning Cancel into consent) was an unfixed
  member of a family decision 2 had already settled in those words: only the submitted body counts
  as consent. Nothing was being chosen, and the reviewer's own `forced_answer` named the same
  repair, so it was fixed rather than asked about.
- The **round 2 blocking finding** (unbounded unauthenticated audit writes) is real, and on
  re-verification it is a pre-existing property of the audit subsystem rather than something this
  branch introduces: the path this change adds is the cheapest row in a table whose worst member,
  `POST /account/register`, predates the branch and carries no limiter that any setting can switch
  on. It is drafted as follow-up 4 with the full table, not fixed here.

The one judgement call worth your eye is that second one, because it means this branch closes a
security defect while leaving a neighbouring one open on purpose. If you disagree that it belongs to
a different change, say so and it comes back here rather than to the tracker.

leodip:
## Follow-ups: two already filed, two drafted, one correction owed

Four real problems were found while doing this work, each verified against the code and each
belonging to a different change. Two were filed while the run was in progress. Two are drafted below
and are yours to file or drop.

Duplicate searches were re-run just now against every issue state, and the tracker has not moved
under any of them. Related but distinct: #113 and #107 are brute-force limits on particular
endpoints, #202 is credentials accepted from the query string, #195 is rate-limiter test coverage,
and #85 (closed, completed) fixed modulo bias in the same generator follow-up 3 is about without
touching its error contract.

---

### 1. Filed as #208: the authorize error page renders a hardcoded English heading

Found while grounding the mismatch page. No action needed.

### 2. Filed as #209, and the issue as filed is wrong about one thing

`gh issue comment 209` is owed before anyone works on it. The issue opens with "nothing checks the
two agree", and that is not true: adding `AuditAuthCeremonyMismatch` in stage 2 made
`TestAuditEventTypes_MatchesConstants` and `TestAuditEventTypes_Count` fail immediately. The survey
behind the draft read the test file for `Uniqueness` and `NonEmpty` and stopped.

The gap is still real, but narrower and differently shaped. `TestAuditEventTypes_MatchesConstants`
compares `AuditEventTypes` against `allAuditConstants`, a **third** hand-maintained copy living
inside the test, so what it enforces is that two of the three lists agree. A constant added to the
`const` block alone, and to neither list, still passes every test in the file: the guard is against
forgetting the slice, not against forgetting both. `TestAuditEventTypes_Count` is a bare number that
has to be bumped by hand as well.

The fix #209 asks for, deriving the list rather than maintaining it, is still the right one. Only its
opening sentence needs amending, or whoever picks it up will find two tests that look like the thing
being requested and reasonably close it as done.

```bash
gh issue comment 209 --body 'Correction found while working #79. "Nothing checks that the two agree" is not accurate: `TestAuditEventTypes_MatchesConstants` and `TestAuditEventTypes_Count` both failed the moment a new `Audit*` constant was added, so the suite is not silent.

The gap is real but narrower. `TestAuditEventTypes_MatchesConstants` compares `AuditEventTypes` against `allAuditConstants`, a third hand-maintained copy living inside the test file, so it enforces that two of the three lists agree. A constant added to the const block alone, and to neither list, still passes every test in the file. `TestAuditEventTypes_Count` is a bare number bumped by hand.

The fix this issue asks for, deriving the list rather than maintaining it, is unchanged. Only the framing needs it: someone picking this up today will find two tests that look like what is being asked for and reasonably close it as done.'
```

---

### 3. `GenerateSecurityRandomString` swallows CSPRNG failure, so no caller's guard can be tested

`go`. Found while building stage 2. **Drafted.**

`stringutil.GenerateSecurityRandomString` answers `""` rather than an error when the system CSPRNG is
unavailable, so every caller that needs a real value has to check for `""` itself. Two do:
`SaveLinkMarker` in `link_marker.go` (#112) and `HandleAuthorizeGet` (this change, `handler_authorize.go`).
Neither check is reached by a test, because the failure cannot be provoked without either swapping
`crypto/rand.Reader` globally from another package's test or adding a function-var seam to production
code, and this codebase does neither anywhere.

Deliberately not solved on this branch. Both guards are three lines and correct by inspection, and
the two ways to test them as things stand are worse than the gap: a global mutation of another
package's variable is a hazard to every test that runs after it, and a mutable package-level function
var exists only to be swapped. What would make them testable is a change to the generator itself,
which is a `core` change belonging to neither issue.

Verified by reading both call sites and by
`grep -rn "unable to generate a link marker" --include=*_test.go src/`, which finds nothing. The
other 17 call sites of the generator do not guard at all, which is the same problem seen from the
other side.

```bash
gh issue create --label go --title 'GenerateSecurityRandomString swallows CSPRNG failure, so no caller can propagate or test it' --body 'stringutil.GenerateSecurityRandomString returns "" on CSPRNG failure instead of an error, so each caller needing a usable value repeats an `if id == ""` guard, and none of those guards can be exercised by a test without swapping crypto/rand.Reader globally or adding a test-only seam to production code. Two guards exist today, in SaveLinkMarker (src/authserver/internal/handlers/link_marker.go) and HandleAuthorizeGet (src/authserver/internal/handlers/handler_authorize.go), and both are uncovered. The remaining call sites do not guard at all.

Give the package an error-returning variant, or an unexported reader the tests can substitute within stringutil, so callers can propagate a real error and the failure path can be covered where it is written.

The seam is already present one level down: randomStringFromReader takes an io.Reader and is tested with a failing one. Only the exported wrapper, randomStringFromAlphabet, hides it, and its comment says so: "It preserves the package'"'"'s historical contract of returning \"\" when the system CSPRNG is unavailable."'
```

---

### 4. Unauthenticated callers can drive unbounded audit writes, and rate limiting is off by default

`bug`, `security`, `go`. Raised by the final review's round 2 against this change's ceremony-mismatch
path, and found on re-verification to be a pre-existing property of the audit logger. **Drafted.**

`AuditLogger.Log` in `src/authserver/internal/audit/audit.go` does a synchronous
`GetSettingsById(nil, 1)` on **every** event, then a `CreateAuditLog` insert when
`AuditLogsInDatabaseEnabled`. `database_seeder.go` seeds both `AuditLogsInConsoleEnabled` and
`AuditLogsInDatabaseEnabled` to `true`, and `config.go` reads
`GOIABADA_AUTHSERVER_RATELIMITER_ENABLED` through `getEnvAsBool` with no default, so **rate limiting
is off unless a deployment turns it on**. Every route limiter in `middleware_ratelimiter.go` begins
`if !m.enabled { next.ServeHTTP(...); return }`.

The reachable paths, each verified by reading it, worst first:

| Path | Precondition | Work per unauthenticated request | Limiter |
|---|---|---|---|
| `POST /account/register` | self-registration, seeded `true` | a row in a business table (`pre_registrations`, or `users` when SMTP is off), a settings read, an audit insert, and on the SMTP branch an outbound email | **none, on any setting** |
| `POST /auth/pwd`, unknown email | one auth context at level1 | user select, a **full bcrypt** (`VerifyPasswordHash(DummyPasswordHash, ...)`, run deliberately for timing safety), settings read, audit insert, render | `LimitPwd`, off by default |
| `POST /auth/otp`, wrong code | one level1 context | settings read, audit insert, render | `LimitOtp`, off by default |
| `GET`/`POST /reset-password`, bad code | none | settings read, audit insert, render | `LimitResetPwd`, off by default |
| any bound form, ceremony mismatch (**this branch**) | one auth context | settings read, audit insert, render | none on consent; pwd and otp as above |

**Why it is not this change's to fix.** The member this branch adds is the *cheapest* row in that
table: strictly less work per request than `POST /auth/pwd` on the same route behind the same absent
limiter, and `POST /account/register` writes to a business table and sends mail with no limiter that
can ever be switched on. An attacker maximising damage per request would use a path that predates
this branch, so the new member adds no capability and closing it alone would leave the family
untouched. No cheap narrowing remedy was available either: a limiter on the consent route is gated on
the same `m.enabled` the finding is about, and dropping the mismatch audit contradicts decision 5.

The review was right about one thing in the contract, though. Section 2 excluded rate limiting on
ceremony mismatches because "the refusal is cheap and stateless", and the second half of that is
wrong: the refusal is audited, and an audit write is neither cheap nor stateless. The exclusion holds
for the better reason above.

```bash
gh issue create --label bug --label security --label go --title 'Unauthenticated callers can drive unbounded audit writes; rate limiting is off by default' --body 'AuditLogger.Log performs a synchronous settings read and an audit-table insert on every event, and several routes reach it without authentication and without a bound in the configuration Goiabada ships.

GOIABADA_AUTHSERVER_RATELIMITER_ENABLED is read through getEnvAsBool with no default, so it is false unless a deployment sets it, and every limiter opens with `if !m.enabled { next.ServeHTTP(...); return }`. LimitPwd, LimitOtp and LimitResetPwd therefore pass every request straight through. POST /account/register carries no limiter at all, while self-registration is seeded enabled. A caller varying the email per request can make the server write audit rows, and on the registration path business rows and outbound mail, for as long as it likes.

The cost is audit-storage growth, audit-trail integrity (real events buried in noise), and database and log I/O per request.

Reachable paths, worst first:

| Path | Precondition | Work per unauthenticated request | Limiter |
|---|---|---|---|
| POST /account/register | self-registration, seeded true | business-table row, settings read, audit insert, outbound email on the SMTP branch | none, on any setting |
| POST /auth/pwd, unknown email | one auth context at level1 | user select, a full bcrypt (deliberate, for timing safety), settings read, audit insert, render | LimitPwd, off by default |
| POST /auth/otp, wrong code | one level1 context | settings read, audit insert, render | LimitOtp, off by default |
| GET/POST /reset-password, bad code | none | settings read, audit insert, render | LimitResetPwd, off by default |

Three remedies would each help and they are independent:

1. Bound the audit write itself: deduplicate or sample repeated identical events from one source, or queue them so a request is not held behind an insert.
2. Stop re-reading settings on every event. It is a full row fetch per audit line.
3. Consider defaulting GOIABADA_AUTHSERVER_RATELIMITER_ENABLED to true. A security control that is off unless configured protects nobody who has not read the reference page.

Add a limiter to POST /account/register regardless, since that path writes more than an audit row.

Anchors: Log in src/authserver/internal/audit/audit.go; the seeded settings in src/core/data/database_seeder.go; NewRateLimiterMiddleware and the `if !m.enabled` guard in src/core/middleware/middleware_ratelimiter.go; the route table in src/authserver/internal/server/routes.go.

Found while reviewing #79, which added one more (and the cheapest) member of this family and does not introduce the property. Related but distinct: #113 and #107 are brute-force limits on specific endpoints; #195 is rate-limiter test coverage.'
```

leodip:
## The code review

This run reviewed the whole branch once, at the end, rather than each stage on its own, so this is
the only independent reading the code got. It is recorded here because `.review/` is throwaway and
gitignored: without this comment you would see the fix commits with no way to learn what prompted
them, what was refuted, or what was left deliberately.

Three rounds in total: one at the plan gate before any code existed, and two over the whole branch.
Five findings, three of them blocking or significant. Two blocking findings were fixed, one blocking
finding was confirmed as fact and refuted as this change's defect, and two plan findings were fixed
in the plan before stage 1 started.

---


Reviewed by `gpt-5.6-sol` at effort `xhigh` throughout. The plan gate and the final gate each got a
fresh Codex session; the final gate's round 2 resumed round 1's.

### Round 1, plan gate, 2026-08-12

The implementation plan, before any code was written. Verdict `findings`: two, one `significant`,
one `minor`, neither needing a human. All three axes (security, conformance, quality) reviewed. The
reviewer also ran the issue's 820-case probe, confirmed all 28 code anchors resolve, and verified the
quoted RFC 6749 §3.3 and OIDC Core §3.1.2.4 text against the publishers.

Both findings confirmed and **both fixed in the plan**, before stage 1 started. No production code
existed at this gate; the tree hash was byte-identical before and after, which is the check that only
planning documents moved.

**1. `significant`, quality: the audit-registry check could never fail.** Goiabada keeps its audit
event names in two hand-maintained halves of `src/core/constants/constants.go`, a `const` block and
an `AuditEventTypes` slice that feeds the admin filter dropdown. Nothing in the test suite checks
they agree, so the plan carried a shell command to check it by hand before adding this change's new
event. That command was:

```
diff <(grep -oE '^\tAudit[A-Za-z0-9]+ +=' ... ) <(awk '/^var AuditEventTypes/,/^}/' ... | grep -o '^\tAudit[A-Za-z0-9]*' ...)
```

GNU grep does not read `\t` as a tab in an extended regular expression. It warns `stray \ before t`
and matches a literal `t`, so `^\tAudit` becomes `^tAudit` and matches nothing. Both sides extract
zero entries, `diff` compares two empty streams, and **the command exits 0 whatever the file
contains**: it could not have caught the one thing it existed for, an event added to one half and not
the other.

Worth recording because it is a trap for anyone repeating it: the reviewer measured `0,0` and the
first re-run measured `97,97`, and the disagreement is the environment. Claude Code shims `grep` to
`ugrep`, which does honour `\t`.

| Where | `grep` | Extracts |
|---|---|---|
| the devcontainer, where tests run | GNU grep 3.12 | **0 and 0**, exits 0 |
| the host, `/usr/bin/grep` | GNU grep 3.12-modified | **0 and 0**, exits 0, with the warning |
| a Claude Code session | `ugrep` 7.5.0 | 97 and 97, exits 0 |

So it was vacuous in both environments the work actually happens in, and the earlier "identical, 97"
recount was a correct number produced by a command that could not have reported anything else.

Fixed by replacing it with an awk field-split comparator that needs no tab escaping, works
identically under both greps, and **asserts each extraction is non-empty at an expected count passed
as an argument** so that another extraction regression cannot make equality vacuous again. Proven
able to fail, three mutations: the new constant added to the `const` block only, expecting 98, gives
COUNT MISMATCH; added to both, OK; the real 97-entry file expecting 98, COUNT MISMATCH.

That nothing in CI enforces this at all is a separate defect, already filed as
[#209](https://github.com/leodip/goiabada/issues/209) before the review. This finding is more
evidence for it: the hand-run substitute was not merely manual, it was broken.

**2. `minor`, conformance: the change falsifies a claim in the security documentation.**
`site/src/content/docs/reference/security.mdx`, under **Cross-site request forgery**, tells readers:

> **No token and no cookie** - There is no hidden CSRF token in the forms and no CSRF cookie. The
> origin is the whole check, so there is nothing to expire, nothing to go stale in a page left open,
> and nothing to configure

The second half of this change puts a hidden, per-authorization value into four authorization forms
whose entire purpose is to **make a page left open fail** once another authorization replaces it.
Calling it a ceremony id rather than a CSRF token does not rescue the sentence: a user who has been
told nothing goes stale in a page left open, and who then meets a 400 on a form they left open, has
been told the wrong thing by this page. The planning sweep had concluded no documentation page was
owed, having searched for `consent`; this sentence is about CSRF, which is why it was missed.

Swept independently rather than taken on the reviewer's word. `go stale`, `nothing to expire`,
`hidden token`, `hidden input`, `hidden field` return that line and nothing else across the docs
site, and `another tab`, `second tab`, `multiple tabs`, `concurrent authoriz`, `two authoriz` return
nothing at all. One sentence, one page.

Fixed by making it owed work, **split across the two stages that make it false** so each edit is true
when it lands and a run that stops early still ships honest documentation: the stage that binds the
consent screen qualifies the bullet, and the stage that binds the login form and OTP prompt widens
the same sentence to all three. The edit preserves the real distinction, which is that CSRF stays
entirely origin-based with no token and no cookie, and that the admin console forms named in that
same section are deliberately outside the ceremony binding.

**One `unchecked` entry turned into a fix.** The reviewer noted it had performed no interactive
browser pass and had traced hidden-input behaviour through the templates and the Go test harness
instead. Rather than noting that back, the gap behind it was worth looking for, and there is a real
one: the existing helper the new page-parsing helper was going to copy,
`getOtpSecretFromEnrollmentPage`, searches the whole document. A ceremony-id helper in that shape
would find a hidden input placed **outside** the `<form>` element and post it by hand, passing green,
while no browser submits such a field and every real submission would fail with a 400. Closed for
free by scoping the selector to `form input[name="ceremonyId"]` and requiring exactly one match, now
written into the plan. The other two `unchecked` entries were dismissed with reasons: production
mutations cannot run at a plan gate where no implementation exists (they are scheduled against the
code that stage writes), and token validation, PKCE, revocation, redirect matching and SQL were out
of scope for a change that touches none of them.

**No round 2 at this gate.** Both findings were resolved in the document the round reviewed, neither
needed a human, and no production code existed for a second reading to look at. The code itself gets
its review once the branch is complete.

### Round 1, final gate (the whole branch), 2026-08-12

`gpt-5.6-sol`, effort `xhigh`, fresh Codex session, request `79-20260812T140724Z`. Scope was the
whole branch against `ff84992c`, four commits, 41 files. Verdict `findings`: **one, `blocking`,
security**, `needs_human: true` with a populated `forced_answer`. All three axes reviewed.

The reviewer verified HEAD and a clean worktree, matched all six recorded tree hashes, passed all 28
anchors, confirmed the 11 green CI checks, ran the focused handler and sqlite integration tests, ran
the 820-case consent probe, and executed five mutations of its own. It did not re-run the full tiers,
correctly, since the hashes matched and CI was green on that exact commit.

**1. `blocking`, security: a query parameter turned Cancel into consent.** Status: **Resolved**

`HandleConsentPost` read the two action controls with `r.FormValue`, which merges the URL query into
the submitted body. `consent.html` posts to `action=""`, so the browser posts back to whatever URL it
is on, and a button that was not clicked sends no key at all. On a consent screen reached at
`/auth/consent?btnSubmit=submit`, a click on **Cancel** posts `btnCancel=cancel` in the body while
`FormValue("btnSubmit")` finds `submit` in the query, so the approval branch ran and granted every
box the form still had checked. Every box is `checked` by default, so the refusal granted the full
requested scope, persisted a `user_consents` row and issued a code.

Confirmed independently rather than taken on the reviewer's word, at the parsing level:

```
POST /auth/consent?btnSubmit=submit   body: ceremonyId=abc&consent0=on&consent1=on&btnCancel=cancel
  FormValue(btnSubmit)     = "submit"      PostFormValue(btnSubmit) = ""
  FormValue(btnCancel)     = "cancel"      PostFormValue(btnCancel) = "cancel"
  today's dispatch takes   = "submit"  (approve = true)
  body-only dispatch takes = "cancel"  (approve = false)
```

**Not escalated, and this is the reasoning.** Decision 2 already decided the invariant, in those
words: only the submitted body counts as consent, because `r.Form` merges the query and the form
posts to `action=""`. Stage 1 applied it to the checkbox selection and left the action controls on
`r.FormValue`, so this is an unfixed member of a family the agreement had already settled rather than
a question anybody has to answer. The reviewer's `forced_answer` was populated and named the same
repair, which is the second independent demonstration `decisions.md` section 1 asks for, and the
remedy only narrows what is accepted, changes no interface and costs nothing at runtime.

Fixed in `0330db93`. Both controls now read from `r.PostForm`, and approval requires the submit
control alone: `r.PostForm.Get("btnSubmit") == "submit" && !r.PostForm.Has("btnCancel")`. A body
carrying both, which no browser sends, is refused rather than resolved in favour of granting.

**The family was swept independently, and it has one other kind of member.** Presence-based reads are
the vulnerable ones, because a key absent from the body lets a query value win; value reads on
always-submitted inputs cannot be overridden, since the body value comes first in `r.Form` and a text
input submits its key even when empty. Sweeping every form read in the four authorization-flow
handlers:

| Read | Kind | Verdict |
|---|---|---|
| `btnSubmit` / `btnCancel`, consent | presence | **the defect**, fixed |
| `consent0..N`, consent | presence | already `r.PostForm.Has` |
| `ceremonyId`, consent, pwd, otp | value | already `r.PostFormValue`, but unpinned: see below |
| `email`, `password`, `otp` | value | not overridable; the body always carries the key |
| `r.Form.Has("request")` and `request_uri`, authorize | presence | correct: the query is the legitimate source at `/auth/authorize`, and the effect is to reject |

`auth_pwd.html`, `auth_otp.html` and `auth_otp_enrollment.html` have unnamed buttons, so consent is
the only form with an action control at all.

**The reviewer's three surviving mutations were the real second half.** It reported that switching
the ceremony-id read from `r.PostFormValue` to `r.FormValue` in consent, pwd and otp left the suite
green: the code was right and nothing held it there, so `why_exhaustive` correctly called this a
regression-coverage gap in the same family. Each handler's `Stale ceremony` table now carries a case
posting the current ceremony id **in the query alone**, expecting the 400 mismatch page. All three
mutations are now caught.

**Mutations, five, all `MUTATION CAUGHT`:**

| File | Replacement | Result |
|---|---|---|
| `handler_consent.go` | the body-only dispatch back to `r.FormValue("btnSubmit") == "submit"` | caught |
| `handler_consent.go` | drop the `&& !r.PostForm.Has("btnCancel")` term | caught |
| `handler_consent.go` | ceremony read `r.PostFormValue` to `r.FormValue` | caught (survived for the reviewer) |
| `handler_auth_pwd.go` | ceremony read `r.PostFormValue` to `r.FormValue` | caught (survived for the reviewer) |
| `handler_auth_otp.go` | ceremony read `r.PostFormValue` to `r.FormValue` | caught (survived for the reviewer) |

**Tiers**, re-run in full after the fix, against `production cc7ea6a20d60`, `unit-tests
13143e708527`, `data-tests 6b3fbe9c559e`, `integration-tests 0fb04525aa07`, `all 7099aa626f0e`.
Unit 3332 ok 0 fail, ~40s, five subtests up from 3327. Data 1615 ok 0 fail on sqlite, mysql, postgres
and mssql, ~35s. Integration 1233 ok 0 fail on sqlite, ~50s. Anchors: all 28 still resolve.
`production` was re-read after the last mutation and matches the value the tiers ran against. The
four-engine integration result is the check suite's, on `0330db93`.

**No integration case was added**, deliberately. Seam 1 owns the exhaustive selection table and its
scope explicitly includes "a key in the URL query"; this is a member of that table, and decision 4
fixed the integration cases at four end-to-end scenarios. The unit case is distinguishing rather than
thin: the database mock carries no expectations, so persisting anything fails it outright, which is
exactly what the old code did.

**The `unchecked` array, all five answered.**

1. *No interactive browser run.* **Dismissed.** The behaviour is server-side dispatch and was
   reproduced at the HTTP parsing level; the two facts a browser would add, that the form posts to
   `action=""` and that every box is `checked`, are read directly out of `consent.html`, and seams 1
   and 2 are the boundaries the user confirmed. Neither is a browser.
2. *No mixed-version rolling deployment simulated.* **Dismissed.** Both directions fail safe and both
   are pinned: an old auth context on a new binary carries no ceremony id and is refused, which is
   the "an auth context from before the ceremony id existed" case in all three tables; a new context
   on an old binary is simply the pre-change behaviour.
3. *Full tiers not re-run locally.* **Covered.** All three re-run this session at the hashes above,
   after the fix.
4. *Token validation, refresh rotation, PKCE, redirect-URI matching, database authorization and
   session lifetime not re-audited.* **Dismissed.** Out of scope per section 2 and untouched by the
   diff.
5. *The CSPRNG failure branch was inspected but not fault-injected.* **Already filed**, as follow-up
   3 in section 9 above, which is what the reviewer was referring to. No new action.

**No follow-ups.** The reviewer raised none, and nothing in the sweep above belongs to a different
change.

### Round 2, final gate (the whole branch), 2026-08-12

`gpt-5.6-sol`, effort `xhigh`, resumed Codex session, request `79-20260812T143445Z`. Same scope, the
whole branch against `ff84992c`, now five commits and 41 files. Verdict `findings`: **two, one
`blocking` security and one `significant` quality**, both `needs_human: true`. All three axes
reviewed.

The reviewer verified HEAD and a clean tracked tree, matched all six recorded tree hashes, passed all
28 anchors, confirmed **all 11 CI jobs green on `0330db93`** including the four database jobs, ran
the focused handler tests and the three sqlite integration ceremony cases, ran three consent-action
mutations of its own, and wrote an independent URL-encoded and multipart form-parser probe.

**1. `blocking`, security: unbounded unauthenticated audit writes through ceremony mismatches.**
Status: **Confirmed as a fact, refuted as this change's defect. Filed** as follow-up 4 in section 9.

The claim is that an unauthenticated caller can obtain one auth-context cookie from the public
authorize endpoint, then repeatedly make any of the three bound POST handlers persist an
`auth_ceremony_mismatch` event: `rejectCeremonyMismatch` deliberately leaves the context reusable
(decision 5), the mismatch check runs before each handler's `AuthState` check, `AuditLogger.Log` does
a synchronous settings read and an audit insert, both audit targets are seeded on, and rate limiting
is off by default. The reviewer named this new evidence against section 2's exclusion of rate
limiting, whose stated reason is that "the refusal is cheap and stateless".

**Every mechanical step of that is true and was re-verified independently**, by reading `Log` in
`src/authserver/internal/audit/audit.go`, the seeded `AuditLogsInConsoleEnabled` and
`AuditLogsInDatabaseEnabled` in `database_seeder.go`, `getEnvAsBool("GOIABADA_AUTHSERVER_RATELIMITER_ENABLED")`
with no default in `config.go`, the `if !m.enabled { next.ServeHTTP(...); return }` opening of every
limiter, and the route table. The reviewer is also right that the exclusion's stated reason is wrong:
decision 5 requires the refusal to be audited, so it is neither cheap nor stateless.

**What the reviewer did not check is whether the family predates the branch, and it does.** Its
`siblings_checked` enumerated the three ceremony-mismatch callers and their limiters; it never
compared them against the audit-writing paths already on the same server. Sweeping those:

| Path | Precondition | Work per unauthenticated request | Limiter |
|---|---|---|---|
| `POST /account/register` | self-registration, seeded `true` | business-table row, settings read, audit insert, outbound email on the SMTP branch | **none, on any setting** |
| `POST /auth/pwd`, unknown email | one auth context at level1 | user select, a **full bcrypt**, settings read, audit insert, render | `LimitPwd`, off by default |
| `POST /auth/otp`, wrong code | one level1 context | settings read, audit insert, render | `LimitOtp`, off by default |
| `GET`/`POST /reset-password` | none | settings read, audit insert, render | `LimitResetPwd`, off by default |
| ceremony mismatch (**this branch**) | one auth context | settings read, audit insert, render | none on consent |

The member this change adds is the **cheapest row in that table**. `POST /auth/pwd` with an unknown
email runs `hashutil.VerifyPasswordHash(hashutil.DummyPasswordHash, password)`, a deliberate full
bcrypt for timing safety, plus a user select, before the same settings read and audit insert, behind
the same limiter that is off by default, on the same route. An attacker maximising damage per request
would use that path, not this one. `POST /account/register` is worse still and cannot be limited by
any setting.

So the finding is real, the exposure is real, and **this change neither introduces it nor widens its
worst case**. Closing it properly means bounding audit persistence across the subsystem, or
defaulting the rate limiter on, which is a change to `audit.go`, the seeder and the route table and
belongs to neither of this issue's two defects. Filed with the full table and three independent
remedies rather than fixed here. Per `leo-run` section 5b, real but not this change's to fix is a
follow-up.

**No cheap narrowing remedy was available**, which is the test that would have made it fixable here
regardless. Adding a limiter to the consent route is not free (a legitimate consent post could be
refused) and would not touch the default configuration the finding is about, since it too is gated on
`m.enabled`. Dropping the mismatch audit contradicts decision 5, which the user sealed.

**2. `significant`, quality: three branches of the action predicate were not pinned.** Status:
**Resolved.**

The round 1 fix, `approved := r.PostForm.Get("btnSubmit") == "submit" && !r.PostForm.Has("btnCancel")`,
is correct, and the reviewer confirmed it against its own parser probe across urlencoded and
multipart bodies. But three deliberately wrong versions of it left the suite green, which the
reviewer demonstrated by running them:

- reading presence rather than the exact value, so any `btnSubmit=<anything>` approves;
- approving when **neither** action key is present;
- reading the cancel control from merged `r.Form`, so `?btnCancel=cancel` in the query turns a real
  click on Submit into a refusal.

The first two would let a hand-built request grant scopes; the third costs a legitimate grant. The
repair was forced and the reviewer's `forced_answer` named the same three rows, so this was fixed
rather than escalated. `Exact-key consent selection` gains exactly those three cases, and the table
now runs 30 rows plus the 11 denied-in-turn cases.

**Mutations, three, all `MUTATION CAUGHT`** where all three had survived for the reviewer:

| File | Replacement | Result |
|---|---|---|
| `handler_consent.go` | `r.PostForm.Get("btnSubmit") == "submit"` to `r.PostForm.Has("btnSubmit")` | caught (survived for the reviewer) |
| `handler_consent.go` | approve when neither action key is present | caught (survived for the reviewer) |
| `handler_consent.go` | `!r.PostForm.Has("btnCancel")` to `!r.Form.Has("btnCancel")` | caught (survived for the reviewer) |

**Tiers.** Only the consent unit test file changed this round, so `production` is unchanged at
`cc7ea6a20d60`, re-read after the last mutation and matching the value round 1's tiers ran against.
`unit-tests` moves to `6db8afd524fd`; `data-tests 6b3fbe9c559e`, `integration-tests 0fb04525aa07` and
`docs aa7f74398809` are all untouched. Unit re-run in full: **3335 pass, 0 fail**, ~55s. (Recorded as 3187 at the time, which was an
undercount by the counting command rather than a different suite: stage 3 ended at 3327, round 1
added 5 and round 2 added 3. Re-run at the closing pass on the same tree: 3335 pass, 0 fail across
47 packages.) Data and
integration not re-run and not owed: no production code, no query, route, handler or schema moved,
and their input hashes are identical to the run CI proved green on `0330db93`. The four-engine result
stands from that commit's suite, 11 of 11 jobs.

**The `unchecked` array, both entries answered.**

1. *No interactive browser pass.* **Dismissed**, on the same grounds as round 1 and with more
   evidence behind it: the reviewer's own urlencoded and multipart parser probe covered the native
   form encodings, and what a browser adds beyond that (layout, focus, JavaScript) has no bearing on
   a server-side dispatch predicate. `consent.html` carries no JavaScript.
2. *No sustained hostile-load benchmark.* **Filed**, inside follow-up 4. The reviewer says itself
   that load testing would quantify capacity rather than decide reachability, and reachability is not
   contested. Measuring the rate belongs with the fix, not with a change that does not own the path.

**No round 3.** The round changed no production code, which is the budget's own test, and both
findings are closed: one fixed with its mutations proven caught, one filed with its evidence. A third
reading would re-raise a scope judgement the user is about to see in the pull request thread and
answer for themselves.

Issue #212, opened 2026-08-12
AuditLogger.Log does a settings read and an insert per event, with no bound of its own


leodip:
AuditLogger.Log performs a synchronous settings read and an audit-table insert on every event, and several routes reach it without authentication and without a bound in the configuration Goiabada ships.

GOIABADA_AUTHSERVER_RATELIMITER_ENABLED is read through getEnvAsBool with no default, so it is false unless a deployment sets it, and every limiter opens with `if !m.enabled { next.ServeHTTP(...); return }`. LimitPwd, LimitOtp and LimitResetPwd therefore pass every request straight through. POST /account/register carries no limiter at all, while self-registration is seeded enabled. A caller varying the email per request can make the server write audit rows, and on the registration path business rows and outbound mail, for as long as it likes.

The cost is audit-storage growth, audit-trail integrity (real events buried in noise), and database and log I/O per request.

Reachable paths, worst first:

| Path | Precondition | Work per unauthenticated request | Limiter |
|---|---|---|---|
| POST /account/register | self-registration, seeded true | business-table row, settings read, audit insert, outbound email on the SMTP branch | none, on any setting |
| POST /auth/pwd, unknown email | one auth context at level1 | user select, a full bcrypt (deliberate, for timing safety), settings read, audit insert, render | LimitPwd, off by default |
| POST /auth/otp, wrong code | one level1 context | settings read, audit insert, render | LimitOtp, off by default |
| GET/POST /reset-password, bad code | none | settings read, audit insert, render | LimitResetPwd, off by default |

Three remedies would each help and they are independent:

1. Bound the audit write itself: deduplicate or sample repeated identical events from one source, or queue them so a request is not held behind an insert.
2. Stop re-reading settings on every event. It is a full row fetch per audit line.
3. Consider defaulting GOIABADA_AUTHSERVER_RATELIMITER_ENABLED to true. A security control that is off unless configured protects nobody who has not read the reference page.

Add a limiter to POST /account/register regardless, since that path writes more than an audit row.

Anchors: Log in src/authserver/internal/audit/audit.go; the seeded settings in src/core/data/database_seeder.go; NewRateLimiterMiddleware and the `if !m.enabled` guard in src/core/middleware/middleware_ratelimiter.go; the route table in src/authserver/internal/server/routes.go.

[...]

Issue #98, opened 2026-07-05
Drop the plaintext users.otp_secret column (follow-up to #82)


leodip:
Follow-up to #82, which encrypted TOTP secrets at rest into \`users.otp_secret_encrypted\` and added a fail-closed startup backfill that migrates and **blanks** the legacy plaintext \`users.otp_secret\` column.

The plaintext column was intentionally retained in the #82 release: the Go backfill runs *after* SQL migrations and needs the plaintext to read from, so the column could not be dropped in the same release.

## To do (a release after #82 ships)
Once installs have upgraded and run the backfill at least once:

- [ ] Migration to drop \`users.otp_secret\` (all 4 DBs: sqlite, mysql, postgres, mssql), + update the \`schema.sql\` snapshots.
- [ ] Remove the \`OTPSecret\` field from \`models.User\`.
- [ ] Simplify the OTP helpers: \`SetOTPSecret\`/\`ClearOTPSecret\` no longer touch a plaintext field; \`GetOTPSecret\` unchanged.
- [ ] Remove \`BackfillEncryptedOTPSecrets\` and its call in \`data.NewDatabase()\` (and the \`Database\` interface entry + per-DB forwarders + mock), OR keep one more release as a safety net — decide at the time.
- [ ] Update tests that still reference \`OTPSecret\` (the integration dual-write setup can drop the plaintext half).

Severity: low (hardening / cleanup).

PR #215, opened 2026-08-14
fix(security): read credential form fields from the request body only

### LINKED ISSUE #202: Security: password and OTP form handlers accept their credential from the query string

leodip:
Closes #202 when merged.

## What this is for

Every credential-bearing HTML form field in Goiabada is read with `r.FormValue`, which falls back to
the query string when the request body does not carry the name. So a POST to `/auth/pwd` with an
empty body and `?password=hunter2` authenticates, and the same holds for the OTP passcode, the
password reset and registration forms, an OAuth client secret, a TOTP enrollment secret, the SMTP
password, and an admin setting another user's password.

This is not a remotely triggerable authentication bypass: there is no request an attacker can make a
victim's browser send that authenticates anybody, and cross-site POSTs to these paths are already
refused by the origin check. It is a footgun. The server accepting a credential in a request target
is what makes a mistaken `curl`, a copied integration snippet or a crafted link work, and a request
that works is a request that gets repeated, at which point the credential is in the browser history
and in the access log of every proxy, CDN and WAF on the path. Goiabada's own request log already
redacts these names, and `Referrer-Policy: same-origin` already stops the cross-origin Referer leak.
Neither reaches the places above.

## The rule this establishes

**The only way a handler in either module obtains the value of a credential-named form field is
`formcredentials.Value`, which reads `r.PostForm` and nothing else.** One package holds the read and
the list of what counts as a credential; a lint test in each module turns "the only way" from a
convention into a checked property of the source, which matters most in the admin console, where a
wrong read would otherwise be invisible for want of any handler test at all.

No browser form submission changes behaviour. Every affected field is a text or hidden input inside
a POST form, so the browser always submits it, and a body field that is present but empty already
beat the query. What stops working is the three non-browser shapes: an empty body, a body that does
not parse, and a body sent with no `Content-Type`.

Three security bindings in this repo already read body-only for exactly this reason, with the
reasoning written above them, and the credentials two lines below them did not.

## What else is in here

Grounding the fix turned up three adjacent things, folded in deliberately rather than left:

- **The admin console's first handler test tier.** Eleven of the seventeen credential reads live in a
  module with no handler tests, no mocks and seven stub `TestMain` files. Resting those eleven on a
  lint and the helper's own tests would leave them with no test that runs their code.
- **A rate-limit bucket keyed on the empty string.** `LimitPwd` and `LimitForgotPwd` key their
  per-account bucket on the submitted address, and a blank field, an unparseable body or a missing
  `Content-Type` all read as `""`, so every such request in the deployment shares one bucket. At
  five per five minutes, the sixth person to submit forgot-password with the field empty gets a 429
  instead of "email is required".
- **Email identity that differs per database.** The login lookup is a plain equality, so which
  spellings resolve to one account is decided by the engine's collation: MySQL and SQL Server fold
  case and accents, PostgreSQL and SQLite fold nothing. On the first two that means the per-account
  brute-force ceiling does not hold, since each case variant is a fresh budget against the same
  account. On the other two it means one capital letter and login just fails. This change moves the
  fold into the application, behind a canonical column, so all four agree.

### Stage 1: `formcredentials`, the vocabulary and the body-only read
_Not started._

### Stage 2: the auth server's credential reads, its continuation binding and its lint guard
_Not started._

### Stage 3: the admin console's first handler test tier
_Not started._

### Stage 4: the admin console's credential reads, its lint guard and the body-only cases
_Not started._

### Stage 5: the empty rate-limit key and the multipart guard
_Not started._

### Stage 6: `emailutil.Fold`, the folded limiter keys and the seeder
_Not started._

### Stage 7: the canonical column, migration 000030, the backfill and the folded lookups
_Not started._

## Status

Run in progress. The plan is written and no stage has landed yet.


leodip:
## Decision 13: What bounds the new "credential in the query" warning at `POST /account/register`?

The plan review (Codex, `xhigh`) raised this and I confirmed it, then found it is broader than
reported. Nothing is committed; the branch is still just the empty commit that opened this PR.

### Background, in one paragraph

Issue #202 moves every credential-bearing form read to the request body. When a handler reads a
credential and the *query string* also carries that name, the agreed design emits one `slog.Warn`
naming the parameter, the path and the client IP, never the value, so a caller sending credentials
in a request target can be found and fixed. We deliberately rejected a global middleware that
scanned every request's query, because that hands an unauthenticated caller one log line per
request on unlimited routes, which is the unbounded-log-write class #159 closed. The design was
judged safe because **every route that reaches the read is rate limited or authenticated.**

### That justification is false

Two separate ways, both checked against the code rather than reasoned:

1. **`POST /account/register` has no limiter at all.** `src/authserver/internal/server/routes.go`
   registers it as a bare `r.Post("/register", ...)`, directly above `/activate` which carries
   `LimitActivate`. `/auth/pwd`, `/auth/otp`, `/forgot-password`, `/reset-password` and
   `/connect/register` all carry theirs.
2. **The limiters that exist are off by default.** `config.go` builds `RateLimiterEnabled` from
   `getEnvAsBool("GOIABADA_AUTHSERVER_RATELIMITER_ENABLED")` with no default, and every limiter
   opens `if !m.enabled { next.ServeHTTP(w, r); return }`. So on a deployment that has not opted in,
   the bound fails on the other three auth server routes too.

`HandleAccountRegisterPost` reads `password` and `passwordConfirmation` immediately after the
settings check, before the email is validated, so an unauthenticated caller sending
`POST /account/register?password=x&passwordConfirmation=y` with an empty body gets **two warning
records per request, unbounded**, whenever self-registration is on (it is seeded enabled). No
browser is needed: `MiddlewareCsrf` is an origin check and admits a request carrying neither
browser-origin header.

### What that route already costs per request, which is the proportion argument

The same unauthenticated request, with a valid email and password, already causes a bcrypt hash, an
encryption, a `pre_registrations` row, an audit-log row and an **outbound email** to an
attacker-chosen address. All of that is already unbounded, and all of it is already tracked as
**#212** ("Unauthenticated callers can drive unbounded audit writes; rate limiting is off by
default"), which names this route explicitly and ends "Add a limiter to POST /account/register
regardless."

So the marginal harm this change adds is two log lines against a database insert and an SMTP send.
That is real but small. The reason it is still your call is that it cuts against the stated reason
we rejected the global-middleware design in the first place.

### Options

- **A. Leave the warning as designed, and correct the agreement's claim.** Rest on the proportion
  above and on #212 owning the route's boundedness. Cost: on a default deployment, an
  unauthenticated caller can write two log lines per request, which is a narrower version of the
  thing we rejected.
- **B. Read the two register credentials through a non-warning variant.** Keeps the diagnostic
  everywhere it is bounded and drops it where it is not. Cost: the one public, unauthenticated form
  is exactly where a third-party integration is most likely to make this mistake, and it would go
  undiagnosed.
- **C. Bound the warning inside the helper**, at most one record per parameter name per interval per
  process. Keeps the diagnostic everywhere and removes the unboundedness structurally. Cost: rate
  limiting inside a logging helper, with its own state and its own tests, and a genuine burst gets
  sampled away.
- **D. Add a limiter to `POST /account/register` in this change.** Closes part of #212 here. Cost:
  choosing a budget for a route this issue is not about, and it does not help a deployment that has
  not enabled rate limiting at all.

### Recommendation

**A**, with the agreement's sentence corrected to say what is actually true. The executed proportion
is what decides it for me: on the request that triggers the warning, the server is already doing a
DB insert and possibly an SMTP send, unbounded, so two log lines do not change that route's risk in
any way an operator would notice. And #212 already owns making the route bounded, where it can be
done once for the audit row, the business row and the mail rather than once for the log line.

If you would rather not ship any new unauthenticated log write, **C** is the honest version of that
preference and **B** is the cheap one. I would not pick D inside this change.

Answer with the letter. Nothing is built yet, so any of the four costs the same from here.

leodip:
## Decision 14: What comparison should the new `email_canonical` column use on MySQL and SQL Server?

Raised by the plan review and confirmed against the schema snapshots. This blocks the last stage of
issue #202, not the first, but nothing is built yet so it is cheap to settle now.

### Background

Issue #202 makes login treat case and accents as insensitive, identically on all four databases.
`José@X.com`, `jose@x.com` and `JOSE@x.com` become one account. We chose to do the folding **in the
application**, because SQLite has no accent-insensitive collation without an ICU extension, so no
set of collation changes makes the four engines agree. A new `email_canonical` column stores the
folded form and `email` keeps the address as the user typed it, because password-reset mail is sent
to `user.Email` and folding it in place could deliver somebody's reset link to a different mailbox.

The whole premise of doing it this way was written down as: *once the fold is ours, the database
compares exactly against a value we computed and has no competing opinion.* And we drew an explicit
line about what is an accent: **combining marks only**, so `ø`, `đ`, `ł` and `æ` are letters and stay
distinct. That was recorded as a deliberate narrowing, in these words: "`jøse@x.com` and
`jose@x.com` become two accounts there where today they cannot coexist."

### The premise does not survive the migration as specified

The migration was specified as adding a nullable column and nothing else. **A column with no
explicit collation inherits the table's**, and:

- `src/core/data/mysqldb/schema.sql`: every table ends `ENGINE=InnoDB DEFAULT CHARSET=utf8mb4
  COLLATE=utf8mb4_0900_ai_ci`. Neither `users.email` nor `pre_registrations.email` overrides it.
  `_ai_ci` is accent-insensitive and case-insensitive.
- `src/core/data/mssqldb/schema.sql`: both columns are bare `NVARCHAR(64)`, under the database
  collation `Latin1_General_100_CI_AI_SC_UTF8` that `NewMsSQLDatabase` creates. Same story.

So `WHERE email_canonical = Fold(email)` would still equate `jøse@x.com` with `jose@x.com` on those
two engines. The reviewer ran both live: each returned equal under the inherited collation and
unequal under an exact one.

Three things follow. The four engines still disagree, which is the finding this part of the change
exists to fix. The `ø` line above is simply not delivered. And the rate limiter, which keys on our
`Fold` output, hands `jøse@x.com` a separate per-account bucket while the database resolves it to
the same user, which is the per-account brute-force ceiling bypass we are trying to close.

### Options

- **A. An exact text collation.** `utf8mb4_0900_as_cs` on MySQL, `Latin1_General_100_CS_AS_SC_UTF8`
  on SQL Server. Accent- and case-sensitive. Reads naturally next to the existing columns.
- **B. A binary collation.** `utf8mb4_bin` on MySQL, `Latin1_General_100_BIN2` on SQL Server. This is
  the only option that provably has no Unicode collation opinion of its own. Since `Fold` already
  emits NFC-normalised output, two canonical values should be equal exactly when their bytes are
  equal, and binary is the comparison that says that and nothing more.
- **C. Leave the columns as specified**, and narrow the identity rule to "case and accents, except
  where MySQL and SQL Server fold further". This is coherent, but it puts back the engine
  disagreement that was the reason for moving the fold into the application.

Whichever you pick also decides what a unique index on that column would mean, which is decision 15.

### Recommendation

**B.** The design's stated goal is that the database has no opinion, and binary comparison is the
only one that literally has none: an exact text collation is still a UCA collation and can still
equate byte sequences we intended to keep apart. Nothing sorts or range-scans this column, it is an
equality lookup on a value we computed, so binary costs nothing.

The cost is the same one already accepted when the `ø` line was drawn: on MySQL and SQL Server,
addresses that differ only by `ø`, `đ`, `ł` or `æ` stop resolving to each other. No existing pair can
break, because the loose collation that folds them is also what stopped them ever being stored
together, so this narrows nothing that exists today.

I would not pick C: it gives up the one property that justified the application-level fold.

Answer with the letter.

leodip:
## Decision 15: Should canonical email uniqueness be claimed atomically, or stay a check before a write?

This one reverses something we put out of scope, so it is squarely your call. It arrives with new
evidence rather than a different opinion, which is why I am asking rather than letting it stand.

### What we decided, and why

Issue #202 adds `email_canonical` to `users` and `pre_registrations`, filled by a **Go** backfill
that runs after the SQL migrations (no engine offers a portable NFD, so the fold cannot be SQL). We
put the unique index on that column out of scope for this release, with this reason: migrations run
before the Go backfill, so an index created in migration 000030 would sit on an empty column and the
backfill would then violate it. That is the same shape as #82 and #98, where the TOTP secrets were
encrypted in one release and the plaintext column dropped in a later one. Meanwhile uniqueness was
to be "enforced at the application level and from day one", because registration and admin user
creation both call `GetUserByEmail` before inserting, and after this change that lookup folds.

### The evidence against both halves of that

**The mechanical reason is weaker than it reads.** A *nullable* unique index admits many NULL rows
on SQLite, PostgreSQL and MySQL, so it can be created before the column has any values; SQL Server
expresses the same boundary with a filtered unique index. And the only way the backfill could
violate such an index is two existing rows folding together, which we already decided must stop the
upgrade and name the addresses. So the index would turn a case we already refuse into a case we
refuse slightly earlier, with a worse message unless the backfill preflights, which it can.

**The application-level guarantee is not atomic.** The lookup and the write are two separate
operations, and neither `commondb.CreateUser` nor `commondb.UpdateUser` makes a conditional claim.
Two concurrent requests can both see no canonical match and both commit two raw-distinct rows for
one identity. `GetUserByEmail` then returns whichever row the engine yields first, and login,
password reset and duplicate detection all rest on that arbitrary pick. It is reachable
unauthenticated by racing two self-registrations while registration is enabled, and by
authenticated users or admins racing equivalent email changes. The reviewer reproduced it: on a
scratch SQLite schedule both prechecks saw zero and both `José@example.com` and `jose@example.com`
inserted with the same canonical value; adding a unique canonical index made the second insert fail.

The sharpest member is account activation, whose own comment says the raw `users.email` unique index
is what makes two racing activations produce one account. That index does not reject two
raw-distinct spellings on SQLite or PostgreSQL, and after this change those spellings are one
identity.

### Options

- **A. Keep the deferral.** Ship the canonical column with no index, record the race as a known
  deferred decision, and file the index for a later release as planned. Cost: a window where two
  rows can name one account, closed later.
- **B. Add the nullable unique index to migration 000030**, filtered on SQL Server, with the backfill
  preflighting collisions first so it still stops and names the addresses rather than dying on a
  constraint. Cost: the migration grows, and the four-engine data tier gains cases for concurrent
  equivalent creates, an update into an occupied identity, pre-registration duplicates, and a user
  versus pre-registration collision followed by activation.
- **C. Claim it in Go instead**, inside a transaction that both reads and writes. Cost: a transaction
  boundary the current create and update methods do not have, on four engines, and it is the option
  most likely to grow the stage.

Note this interacts with decision 14: the index's collation is what "unique" would mean, so if you
pick B here, 14's answer decides whether `jøse` and `jose` can coexist under it.

### Recommendation

**B**, on the reasoning that the deferral's stated blocker turns out not to block. The thing that
made #82's index wait a release was that the column would hold wrong data until the backfill ran; a
nullable index does not care about empty rows, so the analogy does not carry. And the invariant it
protects, one canonical address names one account, is the one this whole part of the change is
about.

That said, **A is a perfectly defensible ship-it answer**: the race needs concurrent requests
carrying two spellings of one address, it is not something an attacker gains anything from beyond
confusing an account, and the index was always coming in a later release. If you want this change to
stay the size it was scoped at, take A and I will write the deferred decision up with what reversing
costs.

I would not pick C.

Answer with the letter.

leodip:
## Decision 16: Is the email identity rule lowercase mapping, or Unicode case folding?

Confirmed by running it. The probe is `docs/issue-202-postformvalue-credentials/probe/fold_probe.go`
with its output beside it, so the table below is executed rather than argued.

### Background

Issue #202 makes an email address name one account whatever its case or accents, identically on all
four databases. The folding function was specified as: lowercase, trim, decompose with NFD, drop
every combining mark, recompose. And "accent" was defined as **a combining mark**, so `ø`, `đ`, `ł`
and `æ` are letters and stay distinct, on the grounds that a table of letter equivalences would have
to be maintained against no authority and differs per locale, while "combining marks" is a line a
reader can check.

### `strings.ToLower` does not deliver "every case spelling is one account"

Unicode *lowercase mapping* and Unicode *case folding* are different operations, and they disagree
wherever a character has a contextual or expanding case mapping. Running the specified algorithm
against a `cases.Fold` variant:

| pair | intended | `strings.ToLower` | `cases.Fold` |
|---|---|---|---|
| `JOSE@x.com` vs `jose@x.com` | one account | one | one |
| `José@X.com` vs `jose@x.com` | one account | one | one |
| `jøse` / `đose` / `łose` / `æose` vs plain | two accounts | two | two |
| `ΟΣ@x.com` vs `ος@x.com` | one account | **two** | one |
| `Straße@x.com` vs `STRASSE@x.com` | one account | **two** | one |

The mechanism: capital sigma `Σ` lowercases to `σ`, but Greek final sigma `ς` lowercases to itself,
so `ΟΣ@x.com` gives `οσ@x.com` while `ος@x.com` gives `ος@x.com`. And `ß` lowercases to `ß` while
`SS` lowercases to `ss`, so `Straße@x.com` and `STRASSE@x.com` never meet. Case folding maps `ς` to
`σ` and `ß` to `ss`, and both pairs land together.

**The part that does not change either way:** `ø`, `đ`, `ł` and `æ` stay distinct under both
algorithms. So the line drawn about letters versus accents survives whichever answer you give, and
this question is only about case.

### Options

- **A. Keep `strings.ToLower`**, and narrow the promise to match it. The documentation would say the
  match ignores case for the Latin alphabet and accents generally, rather than claiming every case
  spelling is one account, and Greek final sigma and German sharp S spellings would be separate
  accounts.
- **B. Use `golang.org/x/text/cases.Fold`** before the mark stripping. Makes the promise true.
  `golang.org/x/text v0.40.0` is already a direct dependency of `core`, so it costs no new
  dependency and nothing at runtime.

### The operational difference worth weighing

It changes **which existing deployments get a blocked upgrade.** The backfill stops and names the
addresses when two existing rows fold together. `cases.Fold` folds strictly more pairs, so strictly
more deployments hit that stop. In practice this affects SQLite and PostgreSQL only: on MySQL and
SQL Server the loose collation already prevents such a pair from ever having been stored.

Concretely, a PostgreSQL deployment holding both `Straße@x.com` and `strasse@x.com` as two real
accounts today upgrades fine under A and is stopped under B until an operator merges them. That is
the design working as intended, but it is a stop that A would not produce.

### Recommendation

**B.** The promise made to users is "case does not matter", and A ships a rule that is true for
Latin and quietly false for Greek and German. `cases.Fold` is the Unicode authority for exactly this
question, so it is not the hand-maintained table that was rejected when the letters-versus-accents
line was drawn: it is the standard's own answer, and it leaves that line intact.

The extra blocked upgrades are the honest consequence of a correct rule, and they surface with a
message naming the two addresses, which is the outcome already chosen for collisions.

Take **A** if you would rather not widen who counts as one account at this point, in which case the
documentation must say plainly what the limit is.

Answer with the letter.

leodip:
Closing unmerged. The run for #202 is being discarded and restarted from scratch; nothing here landed beyond the placeholder commit that opened this thread. The four decisions above stay readable on this closed PR for whoever picks the issue up next.

Issue #122, opened 2026-07-26
Redirect URIs are not required to be absolute: public-client policy bypass and RFC 6749 compliance defect


leodip:
## Summary

Nothing in Goiabada requires a redirect URI to be an absolute URI. A client registered with the scheme-relative value `//evil.example/cb` is accepted, matches at authorization time by exact string comparison, and causes the authorization server to emit:

```
Location: //evil.example/cb?code=THECODE&state=st
```

A protocol-relative `Location` is resolved by the user agent against the current scheme, so the authorization code is delivered to `https://evil.example/cb`.

Severity: **Medium.** This is a **public-client policy bypass and an RFC compliance defect**, and it is worth being precise about what it is not.

The public-client branch of `validateRedirectURI` exists to confine public DCR clients to loopback `http` and custom schemes, so that a self-registered client with no secret cannot receive codes at a remote host. A scheme-relative value defeats that restriction. But it does not grant an attacker a capability they lack otherwise: anonymous DCR already permits registering a **confidential** client with an arbitrary attacker-controlled `https` redirect URI, and the registration response hands back the client secret. So an attacker with DCR enabled can already receive codes on a host they control through an entirely permitted path.

What follows from that:

- The **compliance defect** is unconditional. RFC 6749 section 3.1.2 requires an absolute URI, and nothing enforces it, on either the DCR path or the admin path.
- The **policy bypass** is real: the public-client scheme restriction is circumventable, so the restriction does not mean what the code intends.
- Any **phishing or code-interception consequence** depends on #108, not on this issue. Consent is the control that binds client identity to user authority, and while `ConsentRequired: false` is the default for DCR clients, a code reaching an attacker-controlled host is reachable with or without this bug. Fixing this does not close that chain; #108 does.

The missing validation is universal rather than DCR-specific: the admin redirect URI endpoint applies no scheme validation either, so the value is storable in any deployment and any already-stored row is live today.

## Normative requirements, and where each stands

| Requirement | Source | Status |
|---|---|---|
| "The redirection endpoint URI MUST be an absolute URI as defined by [RFC3986] Section 4.3." | RFC 6749 section 3.1.2 | Enforced at **DCR registration** since #105. **Not enforced at authorization time**, which is this issue |
| `absolute-URI = scheme ":" hier-part [ "?" query ]` | RFC 3986 section 4.3 | Partially. The shipped predicate checks the form, not the full grammar, see below |
[...]

PR #227, opened 2026-08-16
fix(oauth): emit exactly one state, and leave the registered redirect URI's query untouched

### LINKED ISSUE #146: Bug: redirToClientWithError emits two state parameters when the redirect URI already carries one

leodip:
Closes #146 when merged.

## What this is

The authorization endpoint builds its responses by reading the registered redirect URI's query into
`url.Values`, adding the response parameters, and re-encoding. That construction has four defects,
and the issue names one of them.

1. **Two `state` parameters.** `values.Add("state", ...)` on top of a registered `?state=fixed`
   emits both. Go's own `url.Values.Get` reads the first, so an RP comparing the CSRF token from
   RFC 9700 section 2.1 compares against a value it never generated. The success path has the same
   defect with an unconditional `Add`, and that is the redirect carrying the authorization code.
   RFC 6749 section 3.1: "Request and response parameters MUST NOT be included more than once."
2. **The registered query does not survive the round trip.** RFC 6749 section 3.1.2 says a
   registered query "MUST be retained when adding additional query parameters". `Query()` then
   `Encode()` deletes a semicolon-separated field outright, sorts the fields, adds `=` to a valueless
   one, normalises percent-escapes and collapses an empty field. The redirect URI was matched byte
   for byte at authorization time, so altering it sends the RP somewhere it did not register.
3. **`TrimSpace` alters the exact value.** RFC 6749 sections 4.1.2 and 4.1.2.1 require "the exact
   value received from the client". A whitespace-only `state` is dropped entirely, and space is a
   valid `state` character: Appendix A.5 defines `state = 1*VSCHAR` and `VSCHAR = %x20-7E`.
4. **form_post emits a `state` field even when the request carried none.** The template renders the
   input unconditionally, whatever the Go guard decided, so the RP receives an empty `state` on a
   request that never sent one. An empty `state` is not a legal response value under the same
   Appendix A.5 grammar.

Not attacker-reachable: the redirect URI is matched against the client's registered set, so nobody
can supply one carrying a `state` of their choosing. It needs an operator or a self-registered
client to have registered a redirect URI with a query, which RFC 6749 section 3.1.2 expressly
permits and this project's own client documentation says is allowed. Defect 2 fires for any
registered query at all.

## The rule this establishes

Every byte of a registered redirect URI that this server did not write is emitted exactly as
registered, and every byte the client sent as `state` is emitted exactly as sent. The only fields
the server touches are the ones it is answering with, and each appears exactly once. `state` is
emitted if and only if it is non-empty, which is what presence means at this endpoint: RFC 6749
section 3.1 says "Parameters sent without a value MUST be treated as if they were omitted from the
request."

That holds for all three response modes (query, fragment, form_post) on both the error and the
success path. One construction does it, in one place, and the RP-initiated logout redirect (fixed
separately in #109) is refactored onto the same function, because a second implementation of this
is the condition that produced this issue.

Two smaller things land in branches this change was already rewriting. The successful form_post
render is buffered, so a template that parses and then fails part way cannot leave a partial body
behind an implicit 200, which is what #141 fixed for its twin. And both form_post responses get
`Cache-Control: no-store` and `Pragma: no-cache`, which OAuth 2.0 Form Post Response Mode section 2
requires and neither branch had: "the Authorization Server MUST instruct the User Agent (and any
intermediaries) not to store or reuse the content of the response."

### Stage 1: the shared construction, and the logout emitter refactored onto it

Landed in `37b3e94`. Adds `src/authserver/internal/handlers/authorization_response.go`:
`writeResponseParams`, which writes an ordered list of response parameters into a client's redirect
URI, and `encodeResponseParams`, which renders that list in declaration order rather than
`url.Values.Encode`'s alphabetical sort. It is the logout endpoint's construction from #109
generalised from one replaced parameter name to a set: `url.Parse`, copy every registered field
verbatim and in order except those being replaced, append the new ones escaped, rejoin.

`buildPostLogoutRedirect` now delegates to it, with its signature and its 23 table rows unchanged.
That refactor is the point rather than tidying beside it: #109 fixed this in one place and left the
authorization emitters running the broken shape, so a second implementation is the condition that
produced #146.

No authorization response changes yet, and the logout redirect's behaviour is unchanged, including
its deliberate divergence on a valueless `?state=` (RFC 6749 section 3.1 makes a valueless parameter
omitted; RP-Initiated Logout 1.0 has no such rule).

Tiers: unit on all three modules, 3423 passing; integration on sqlite, 1251 passing. Four mutations
against the new helper, one per behaviour its table claims, all caught. No data tier: nothing here
touches a model, a migration or the data layer.

### Stage 2: the error emitter

Landed in `78c30db`. `redirToClientWithError`, all three response modes, moved onto the stage 1
construction.

The **query branch** is the defect the issue is named for. It seeded `url.Values` from the
registered redirect URI's own query and `Add`ed the response parameters on top, so a client that had
registered `?state=fixed` got back two `state` parameters and `url.Values.Get` answered with the
registered one. It now calls `writeResponseParams`, which keeps every registered field byte for byte
and replaces only the ones being answered with. That also fixes the wider retention defect, which
has nothing to do with `state`: decoding and re-encoding the query deleted a semicolon-separated
field outright, sorted the rest by key, added `=` to a valueless field, normalised percent-escapes
and collapsed an empty field, against RFC 6749 3.1.2's "MUST be retained when adding additional
query parameters".

The **fragment branch** appends `encodeResponseParams`, and keeps appending rather than sharing the
query path: a redirect URI cannot carry a fragment of its own for the response to collide with, so
there is no registered field list there to preserve. Its registered query is left exactly as it
stands, which a case now asserts.

The **form_post branch** gains `Cache-Control: no-store` and `Pragma: no-cache`, which the Form Post
Response Mode spec section 2 requires and which neither renderer set. They are written after the
render succeeds, so a failed render still leaves the whole response untouched for the caller's 500.

`TrimSpace` is gone from all three. RFC 6749 section 3.1 makes a parameter sent without a value an
omitted one, so `?state=` and `?state` get nothing back; but space is `%x20` and so is `VSCHAR`, so a
whitespace-only `state` is a value the client chose and is now echoed byte for byte instead of being
dropped. The `end_session_endpoint` deliberately keeps its own rule, because RP-Initiated Logout 1.0
carries no valueless-parameter rule at all. No signature changed anywhere.

One planned test could not be written, and the code says so where it would have gone: to a template,
an absent bind-map key and a key holding `""` are the same value, so the form_post branch's Go-side
`state` guard is unobservable through any template. The guard is kept for the reasons stated at it,
and what form_post really does with an empty `state` is `form_post.html`'s, landing in stage 3.

Tiers: unit 3434 passing, data 1619 passing on all four engines, integration 1251 passing on sqlite.
Seven mutations attempted against the stage's cases, six run and all six caught; the seventh did not
compile and was replaced.

**Corrected by stage 3:** the paragraph above is wrong about the guard being untestable. It is
untestable through `{{.state}}` and `{{if .state}}` only, not through a template that enumerates the
bind map. The guard is unchanged and is now covered; see stage 3's entry.

### Stage 3: the success emitters, the template, and the docs

Commit `f6cde265`. The same defect at the emitter the issue does not mention, and the one that
matters most: `issueAuthCode` ran the identical construction with an **unconditional**
`values.Add("state", code.State)` on top of it, so a client registering `?state=fixed` got two
`state` parameters on the redirect that carries the authorization code. That is the response an RP's
CSRF check exists to guard, and it fired on every successful authorization rather than only on a
refusal. Both success emitters, `issueAuthCode` in all three response modes and `issueImplicitTokens`
in the fragment, now build one ordered `[]responseParam` and hand it to the shared construction, so
the registered query survives byte for byte and each parameter appears exactly once. `TrimSpace` is
gone from both, on the same reading of RFC 6749 section 3.1 and Appendix A.5 as stage 2.

`form_post.html` gets `{{if .state}}` around the state input, matching the `code`, `error` and
`error_description` guards already in the file. That input was emitted unconditionally until now, so
a request carrying no `state` reached the client as `state=""` whatever either emitter had decided:
this is what made the two Go-side guards observable, and it is why the template waited for the stage
where both emitters could be covered together against the real template rather than a stub.
`issueAuthCode`'s form_post render is also buffered and gains `Cache-Control: no-store` and
`Pragma: no-cache`, closing the last two asymmetries with its twin on the error path.

`concepts/clients.mdx` now answers the question a client registering `?state=fixed` arrives with: a
query component is permitted and comes back unchanged, and a registered `state`, `code`, `error` or
`error_description` is replaced by the server's own value.

**One correction to what stage 2 recorded**, below. Its account said the Go-side `state` bind-map
guard was unobservable through any template. That holds for `{{.state}}` and `{{if .state}}` and is
false in general: a template that ranges over the bind map yields the key only when it is present,
and `len` counts it. Since `form_post.html` is operator supplied whenever
`GOIABADA_AUTHSERVER_TEMPLATEDIR` is set, that is a real reader, and a guard nothing can observe is
one the next reader deletes as dead. Both guards now carry a corrected comment and are pinned by a
single test covering both emitters. Neither guard's behaviour changed, so this is a coverage and
record correction rather than a fix.

Tiers: unit 3460 passing, data 1619 passing on all four engines, integration 1251 passing on sqlite.
Eight mutations against the stage's cases: seven caught outright, and the eighth is the survival
that produced the correction above, caught by the test it prompted.

### Stage 4: the two end-to-end cases

Commit `445f4167`. One new file,
`src/authserver/tests/integration/authorize_registered_query_state_test.go`. No production code
moved: this stage is entirely the coverage every unit seam above was assuming.

Those unit cases hand-build a `models.Code` or a `redirectErrorInput`, so all of them take for
granted something none of them shows: that a redirect URI carrying a query of its own can be
registered, requested back, and matched by `ValidateClientAndRedirectURI`'s byte-for-byte
comparison. Both cases here drive the real `/auth/authorize` endpoint against a client registering
`https://app.example.com/callback?state=fixed&lang=en`, where the two fields do different jobs:
`state=fixed` is a field the server answers with and must **replace**, `lang=en` is a field the
server never writes and must arrive **byte for byte**. The client's own state is `csrf a+b/c=`, so
the space, `+`, `/` and `=` all have to survive the request parse, the session cookie, the `codes`
row and the escape on the way out.

- **Success:** the whole level1 ceremony through to `/auth/issue`, asserting the `Location` equals
  `https://app.example.com/callback?lang=en&code=<code>&state=csrf+a%2Bb%2Fc%3D`.
- **Error:** a fresh cookie jar and `prompt=none`, the cheapest of the sixteen error redirects since
  `handlePromptNone` answers `login_required` on its first check, asserting the `Location` equals
  `https://app.example.com/callback?lang=en&error=login_required&error_description=User+authentication+is+required&state=csrf+a%2Bb%2Fc%3D`.

Both compare the **whole raw `Location` string**, and both read `code` out of the parsed query but
never `state`. Parsing is what hid this defect in the first place: `url.Values.Get` returns the first
of two, which is the registered `fixed`, so a parsed read of the field under test would have passed
against the bug these cases exist to catch.

Tiers: unit 3460 pass; data 1619 pass on all four engines; integration 1253 pass on sqlite, up from
1251, which is exactly these two cases. Three mutations of `authorization_response.go`, one per
property claimed, all caught by these two cases alone: the replacement filter made never to match,
the registered query dropped, and the parameter value left unescaped.

### After the stages: decision 13, the names an authorization response owns

Commit `0209f8a3`. Not a stage. The whole-branch review found a neighbouring case the goal covers
and the code did not, it became a question for the repository owner rather than a repair the run
could make on its own, and this is the answer applied.

The shared construction dropped a registered redirect-URI field only when the selected response was
itself appending that name. Every authorization-response name the chosen path did not emit therefore
survived from the registered query into the response:

| Registered redirect URI | Request `state` | Before | After |
|---|---|---|---|
| `?state=fixed&lang=en` | absent | `?state=fixed&lang=en&code=fresh` | `?lang=en&code=fresh` |
| `?state=fixed&lang=en` | absent, refused | `?state=fixed&lang=en&error=access_denied&…` | `?lang=en&error=access_denied&…` |
| `?error=stale&error_description=…&lang=en` | sent, success | `?error=stale&error_description=…&lang=en&code=fresh&state=…` | `?lang=en&code=fresh&state=…` |
| `?code=stale&lang=en` | sent, refused | `?code=stale&lang=en&error=access_denied&…&state=…` | `?lang=en&error=access_denied&…&state=…` |
| `?error_uri=…&lang=en` | any | survives | survives, unchanged |

Rows 3 and 4 shipped a response that reads as a success and a failure at once: an RP that checks for
`error` before reading `code`, which is the usual order, treated a granted authorization as a
refusal. Rows 1 and 2 put a `state` in front of the client on the redirect carrying the authorization
code that the client never sent and cannot have bound to its user agent.

`writeResponseParams` gained a `reservedNames []string` parameter, and the two query-mode
authorization emitters pass `code`, `state`, `error`, `error_description`. A registered field
carrying one of those four is now dropped whether or not the response emits it. Nothing else about
the construction moved: every other registered field is still copied byte for byte and in order, and
the response parameters are still appended in declaration order.

Three boundaries, each deliberate and each held by a test:

- **`error_uri` is not in the set.** RFC 6749 4.1.2.1 defines it and this server never emits it, so
  reserving it would delete a registered field no response of ours can collide with.
- **Query mode only.** In `fragment` and `form_post` mode the response travels in the fragment or the
  POST body, and the registered query is left exactly as registered.
- **`end_session_endpoint` reserves nothing.** `buildPostLogoutRedirect` passes `nil` and keeps
  retaining a registered post-logout `state`, because RP-Initiated Logout 1.0 says only that the OP
  passes a state back "if included in the logout request". Its 23 table rows pass unedited.

No `MUST` in RFC 6749 forces this either way, which is why it was the owner's call: 3.1.2 requires
the registered query to "be retained when adding additional query parameters", while 4.1.2 says only
that "The client MUST ignore unrecognized response parameters". It was settled on the grounds that
the two shapes above are worth more than a registered fixed value, which a client keeps by choosing a
name this server does not emit.

`site/src/content/docs/concepts/clients.mdx` was corrected with it: the sentence stage 3 shipped
promised the four names are "replaced by the server's own value", which was false in exactly these
rows and did not qualify itself by response mode. It now says the four are removed even when the
server has nothing of its own to put there, names `error_uri` as surviving, and scopes the whole rule
to the query component.

Tiers: unit 3471 pass, 0 fail; data green on all four engines; integration 1253 pass, 0 fail on
sqlite. Four mutations, all caught: the reserved filter disabled, and the reserved set removed from
the error emitter and the success emitter, plus an inverted one adding it to the logout emitter,
which proves that endpoint's retention is pinned rather than merely currently true.

### After the stages: the one shape the helper's contract did not keep

Commit `2dd0ec06`. Also not a stage: the second whole-branch review round found it and it is fixed
here.

`writeResponseParams` assigns `strings.Join(fields, "&")` to `RawQuery`, and `url.URL.String` writes
the `?` only when `RawQuery` is non-empty or `ForceQuery` is set. A field list of exactly one empty
field joins to `""`, so the delimiter went with it. A registered
`https://rp.example/cb?state=fixed&`, with the four reserved names applied and no response parameter
appended, therefore came back as `https://rp.example/cb`: a different target from the one the client
registered and `ValidateClientAndRedirectURI` had just matched byte for byte. That is the RFC 6749
section 3.1.2 retention requirement this whole change exists to satisfy, failing on a shape no row
happened to cover.

```go
if len(fields) > 0 && redirUrl.RawQuery == "" {
    redirUrl.ForceQuery = true
}
```

No shipped flow reaches it, and that was checked rather than assumed: the combination needs an empty
parameter list with a non-empty reserved set, and all three production callers are excluded
(`redirToClientWithError` always carries `error`, `issueAuthCode` always carries `code`,
`buildPostLogoutRedirect` reserves nothing). It was fixed rather than deferred because this helper is
the one construction every emitter's correctness now reduces to, and #146 exists precisely because a
second copy of this logic was left alone; shipping it with a contract its own doc comment states and
its body does not keep is the same mistake one level down.

Both boundaries are pinned, because the interesting part is which of the two empty-query cases keeps
the `?`. Zero surviving fields still loses it (`?state=fixed` filtered to nothing gives
`https://app.example.com/cb`), which an existing row already asserted and which is right, since
nothing is left for the delimiter to delimit. Two rows were added: the lone-empty-field case, and its
sibling with a parameter appended, `?state=fixed&` plus `code=fresh-code` giving `?&code=fresh-code`.
Two mutations against the new guard, both caught, including `> 0` to `>= 0`, which is the mutation a
careless fix actually is and which only the pre-existing zero-fields row catches.

## Tiers

Final state of the branch, at `2dd0ec06`:

| Tier | Engines | Result | Where it ran |
|---|---|---|---|
| unit, all three modules | n/a | 3473 passed, 0 failed | local and CI |
| data | sqlite, mysql, postgres, mssql | 1619 passed, 0 failed | local and CI |
| integration | sqlite | 1253 passed, 0 failed | local |
| integration | mysql, postgres, mssql | passed | **CI only** |
| Lint, Vulnerabilities, Versions, Build validation | n/a | passed | **CI only** |

**The four-engine integration claim is the check suite's, not a local run's.** It is green across all
11 jobs on the head commit `2dd0ec06`:
[run 31972352431](https://github.com/leodip/goiabada/actions/runs/31972352431).

Unit went 3423 → 3434 → 3460 → 3471 → 3473 across the four stages and the two fix commits.
Integration went 1251 → 1253, which is stage 4's two end-to-end cases. Data is unchanged at 1619
throughout: nothing here touches a model, a migration or the data layer, so that tier ran on all four
engines purely as a regression check.

**21 mutations were run against this branch's own tests**, one per behaviour each stage claims to
have made observable: 4 in stage 1, 6 in stage 2 (a seventh did not compile and was replaced), 8 in
stage 3, 3 in stage 4 (all caught by the two integration cases alone), 4 for decision 13 and 2 for
the fix above. All were caught except one, in stage 3, and that survival is what produced the stage 2
correction noted below; it is caught by the test it prompted. The reviewer ran a further 3 of its own
across rounds 1 and 3, all caught.

### Review

Reviewed by `gpt-5.6-sol` at effort `xhigh`: one round over the plan before stage 1, then **3 rounds
over the whole branch** against `origin/main`. `LEO_REVIEW_MODE=end` was in force for the run, so **no
stage was reviewed on its own**.

| Gate | Round | Findings | Blocking / significant | Outcome |
|---|---|---|---|---|
| plan | 1 | 2 | 2 blocking | both applied before stage 1, as decisions 11 and 12 |
| final | 1 | 1 | 1 significant | escalated as decision 13, answered by the owner, landed in `0209f8a3` |
| final | 2 | 1 | 0 | fixed in `2dd0ec06` |
| final | 3 | 0 | 0 | clean: zero findings, zero unchecked, zero follow-ups |

- **blocking, conformance:** an empty `state` is an omitted one under RFC 6749 section 3.1, so the
  planned presence flag implemented a distinction the protocol does not have. Applied before stage 1:
  the flag, two emitter parameters and a template bind key were deleted from the design, and `state`
  is emitted iff non-empty.
- **blocking, conformance:** both form_post branches returned a cacheable 200 carrying an
  authorization code or a state. Applied in stages 2 and 3: `Cache-Control: no-store` and
  `Pragma: no-cache`, which Form Post Response Mode section 2 requires.
- **significant, conformance:** a registered response name survived whenever the selected path did not
  emit it, so a success response could carry a stale `error` and an error response a stale `code`.
  Became decision 13, because two sealed statements of the agreement pointed opposite ways and no
  `MUST` in RFC 6749 settles it. Fixed in `0209f8a3` per the owner's answer.
- **minor, conformance:** a lone surviving empty query field lost its `?`. Fixed in `2dd0ec06`, above.

Round 3 was clean against an independent raw-query oracle over **161,020 cases** and two fresh
mutations aimed at round 2's fix from the two sides the run's own mutations did not cover, both
caught. It reported nothing, so no fourth round was fundable and the review closes at three of four.

Full account, with the evidence for every finding,
[in this comment](https://github.com/leodip/goiabada/pull/227#issuecomment-5309760704).

## Deviations worth knowing

- **No function signature changes anywhere in this branch.** The sealed plan carried a separate
  presence flag to distinguish "sent empty" from "not sent"; the plan review showed RFC 6749 has no
  such distinction, so it was deleted from the design before being written.
- **Stage 2 recorded something false and stage 3 corrected it on the record.** Stage 2 concluded the
  Go-side form_post `state` guard was unobservable through any template, deleted its planned case and
  said so in a comment. That holds for `{{.state}}` and `{{if .state}}` but not in general: a template
  that ranges over the bind map yields the key only when present. Since `form_post.html` is operator
  supplied whenever `GOIABADA_AUTHSERVER_TEMPLATEDIR` is set, that is a real reader, and a guard
  nothing can observe is one the next reader deletes as dead. Both guards now carry a corrected
  comment and are pinned by a test covering both emitters. No behaviour changed.
- **Two commits sit on top of the four stages rather than inside them.** Decision 13 (`0209f8a3`) and
  the helper fix (`2dd0ec06`) both arrived from the whole-branch review after every stage was
  committed and pushed, so each is a fix commit naming the work it corrects rather than a rewritten
  stage.
- **`end_session_endpoint` is deliberately left diverging** on two points: it retains a registered
  `state`, and it answers a valueless `?state=` where the authorization endpoint treats it as absent.
  Both follow from RP-Initiated Logout 1.0 carrying no valueless-parameter rule at all. Its 23
  regression rows passed unedited through every stage, and one mutation exists specifically to prove
  that retention is pinned rather than merely currently true.

## Status

**Complete.** Seven commits, ten files, 1599 insertions and 116 deletions. All four stages landed,
both post-stage fixes landed, the whole-branch review closed clean at round 3, and the check suite is
green on the head commit `2dd0ec06` across all 11 jobs, which is where the four-engine data and
integration result comes from.

The shared construction exists in `authorization_response.go`, the logout emitter is refactored onto
it, and the error redirect, the authorization code redirect and the implicit redirect each emit
exactly one `state` holding the client's exact bytes, in every response mode. **The duplicate-`state`
defect this issue is about is fixed**, along with the three neighbouring defects in the same
construction: the registered query is retained byte for byte, `TrimSpace` no longer alters the exact
value the client sent, and form_post no longer emits an empty `state` field on a request that carried
none. A query-mode response also no longer carries any of the four names it owns in from the
registered URI, so a client cannot receive a `state` it never sent or an `error` beside a fresh
`code`.

Three comments carry what does not belong in the body:

- [Decisions](https://github.com/leodip/goiabada/pull/227#issuecomment-5309760535): 13 settled, none
  deferred, with the three the owner did not settle during the interview written out.
- [Follow-ups](https://github.com/leodip/goiabada/pull/227#issuecomment-5309760602): one, drafted
  ready to file, plus two things raised and deliberately not filed.
- [The review account](https://github.com/leodip/goiabada/pull/227#issuecomment-5309760704): all four
  rounds, every finding, and the evidence for each disposition.

The question behind decision 13, the measured behaviour, the three options and the recommendation are
in [this comment](https://github.com/leodip/goiabada/pull/227#issuecomment-5309515515); the owner
answered **A**. The plan review that amended the design before stage 1 is
[in this one](https://github.com/leodip/goiabada/pull/227#issuecomment-5309106987).


leodip:
## Plan review, round 1: two blocking findings, both applied

Codex (`gpt-5.6-sol`, effort `xhigh`) reviewed the plan before any code was written. Two blocking
findings, both confirmed here by fetching the specifications rather than by taking the verdict's
word, and both applied. Neither was escalated, because in each case the normative text leaves one
answer. Flagging them here anyway: **the first one supersedes half of a decision you made in the
interview**, and this thread is where you can say so if you disagree.

### 1. An empty `state` is an absent `state`, so the presence flag is not built

The interview settled that `state` would be echoed if the request *carried the parameter*, whatever
its value, and rejected emitting it on non-emptiness. The reasoning was that RFC 6749 conditions on
presence, and that the alternative would leave this server answering `state=` on the logout redirect
and nothing on the authorization redirect for the same request shape.

RFC 6749 section 3.1, the authorization endpoint, defines what presence means:

> Parameters sent without a value MUST be treated as if they were omitted from the request. The
> authorization server MUST ignore unrecognized request parameters. Request and response parameters
> MUST NOT be included more than once.

And Appendix A.5, whose section list covers the response as well as the request:

> The "state" element is defined in Sections 4.1.1, 4.1.2, 4.1.2.1, 4.2.1, 4.2.2, and 4.2.2.1:
> `state = 1*VSCHAR`

`VSCHAR = %x20-7E` (Appendix A). So `?state` and `?state=` must be treated as omitted, and `state=`
is not a legal value in a response either. Echoing it is out, and refusing the request is out too,
since the parameter is omitted rather than malformed. Presence and non-emptiness coincide at this
endpoint, and `state` is emitted iff it is non-empty.

The consistency argument does not survive either, and I checked it because it was the only thing
that could have kept this an open question. RP-Initiated Logout 1.0 section 2 has no
valueless-parameter rule anywhere in the document, and says of `state` only "If included in the
logout request, the OP passes this value back to the RP". Two endpoints, two specifications,
differing on exactly this point. So #109's logout behaviour is correct as it stands and is not
touched: same signature, same presence flag, same 23 table rows.

What this deletes from the plan, before any of it was written:

- `AuthContext.StatePresent`, and with it the upgrade window. Nothing new is serialised into the
  session cookie, so an authorization ceremony spanning a restart is no longer affected and the
  release note is not owed.
- `redirectErrorInput.statePresent` and its plumbing through `redirectErrorFromAuthContext`.
- The new parameters on `issueAuthCode` and `issueImplicitTokens`, so **no function signature
  changes at all** and the 14 + 7 + 7 test call sites stay as they are.
- The template's `statePresent` bind key. The guard becomes `{{if .state}}`, matching the `code`,
  `error` and `error_description` guards already in that file.

What is unchanged: a whitespace-only `state` is still a valid `state`, space being `%x20`, so
dropping `TrimSpace` and round-tripping the exact bytes stands exactly as agreed. The duplicate
`state` fix and the byte-preserving registered query, which are the issue itself, are untouched.

The same RFC paragraph turns out to carry the direct requirement for the defect this issue exists
for, which the agreement had not quoted: "Request and response parameters MUST NOT be included more
than once."

**If you would rather keep the presence flag and answer `state=`, say so here and it goes back in.**
It would be a knowing departure from both quoted sentences, for a request shape that carries no CSRF
token in the first place.

### 2. Both form_post responses are cacheable, and the response mode forbids that

OAuth 2.0 Form Post Response Mode, section 2:

> Because the Authorization Response is intended to be used only once, the Authorization Server MUST
> instruct the User Agent (and any intermediaries) not to store or reuse the content of the response.

The two production renderers of `form_post.html`, the error emitter in `handler_authorize.go` and
the success emitter in `handler_auth_issue.go`, write the rendered page with no cache headers, and
nothing on the `/auth` routes supplies any. The success page carries an authorization code and the
state and comes back as a plain cacheable 200.

Both branches get `Cache-Control: no-store` and `Pragma: no-cache` before the page is written, which
is the pair the token endpoint, dynamic client registration, the health check and
`handlerhelpers/http_helper.go` already write. Two lines each, in branches this change was already
rewriting, so it lands here rather than as a follow-up, on the same reasoning that admitted the
buffered form_post render.

### Where the plan stands

Four stages still. Stage 1 (the shared construction plus the logout refactor) is unchanged. Stage 2
is now just the error emitter and is smaller. Stage 3 keeps the success emitters, the template and
the docs. Stage 4 keeps the two end-to-end cases. One follow-up was drafted for later, not filed:
the authorization endpoint reads every request parameter through `r.FormValue`, so a parameter sent
twice is accepted with one occurrence silently chosen, against the MUST NOT quoted above.

leodip:
## Decision 13: which names in a registered redirect URI's query does an authorization response own, and in which response modes?

The whole-branch review found one thing, and it needs your call before this can leave draft. All four
stages have landed and the check suite is green on `445f416`; nothing is committed that rests on this
question.

### What the code does today

The shared helper drops a registered query field only when that field's name is one the selected
response is **appending**. So any authorization-response name the chosen path does not itself emit
survives from the registered redirect URI into the response. Driving the real emitters:

| Registered redirect URI | Request `state` | What the client receives today |
|---|---|---|
| `?state=fixed&lang=en` | absent | `?state=fixed&lang=en&code=fresh-code` |
| `?state=fixed&lang=en` | absent, refused | `?state=fixed&lang=en&error=access_denied&error_description=denied` |
| `?error=stale&error_description=stale-detail&lang=en` | `client-csrf`, success | `?error=stale&error_description=stale-detail&lang=en&code=fresh-code&state=client-csrf` |
| `?code=stale&lang=en` | `client-csrf`, refused | `?code=stale&lang=en&error=access_denied&error_description=denied&state=client-csrf` |
| `?error_uri=https%3A%2F%2Fstale&lang=en` | any | survives; this server never emits `error_uri` |

Rows 3 and 4 ship a response that is a success and a failure at once. An RP that checks for `error`
before reading `code`, which is the usual order, treats row 3's successful authorization as a refusal.
Rows 1 and 2 put a `state` in front of the client that the client never sent, on the redirect carrying
the authorization code.

The duplicate-`state` defect this PR is actually about is fixed and stays fixed: when the request
carries a `state`, exactly one reaches the client and it is the client's own. This is the neighbouring
case where the server has no value of its own to write.

### Why I stopped instead of fixing it

Two things in the sealed agreement point opposite ways about this exact shape.

The goal says `state` is "present when and only when the client's request carried the parameter", and
the documentation this PR ships says a client registering `?state=fixed` "will not see `state=fixed`
in the callback". Rows 1 and 2 falsify both.

But `buildPostLogoutRedirect` already does exactly what row 1 does, **on purpose**. Its table row
"A registered state survives when none is supplied" asserts that retention, and the function's comment
calls it "preserved deliberately". It is one of the 23 rows the agreement seals as
behaviour-preserving, so `end_session_endpoint` is settled and stays as it is whatever you decide here.
That is why this is a choice rather than a bug I should quietly repair.

### The specification permits all three options

Fetched from [RFC 6749](https://www.rfc-editor.org/rfc/rfc6749.txt), not recalled:

- **3.1.2**: the registered query component "**MUST be retained** when adding additional query
  parameters". That argues for keeping every registered field.
- **4.1.2.1**: `state` is "REQUIRED if a "state" parameter was present in the client authorization
  request". There is **no sentence in 3.1.2, 4.1.2, 4.1.2.1, 4.2.2 or 4.2.2.1 forbidding `state` in a
  response to a request that carried none**, and no rule about response parameters beyond those
  defined. 4.1.2 says only "The client MUST ignore unrecognized response parameters", and these names
  are recognized ones.
- **3.1**: "Request and response parameters MUST NOT be included more than once" bites only where the
  server also emits the name, which is the case this PR already closed.

So no MUST forces an answer. The choice is genuinely yours.

### The options

**A. The response owns the four documented names, on the query path.** Filter `code`, `state`, `error`
and `error_description` out of the registered query on every query-mode authorization response,
whether or not the selected path emits them, and qualify the documentation by response mode.
*Price:* a client that deliberately registered `?state=fixed` and sends no `state` stops receiving it.

**B. Keep what is shipped.** Closest to 3.1.2's retention MUST, and consistent with
`end_session_endpoint`. *Price:* the goal sentence and the documentation sentence are both false as
written and have to be narrowed, and a success response can carry `error=stale`.

**C. Split by name.** Filter `code`, `error` and `error_description` unconditionally; leave `state`
emitted-only. Removes the success-that-also-says-error shape while keeping `state` retention aligned
with logout. *Price:* the "if and only if" clause stays false and the documentation needs qualifying
name by name.

Two sub-questions that one answer settles either way, so please say if you want them differently:

- **`error_uri`.** RFC 6749 4.1.2.1 defines it, this server never emits it, and the shipped
  documentation names four fields rather than five. I would leave it out of the set and say so on the
  page, since adding it removes a field no response of ours can collide with.
- **Container.** In `fragment` and `form_post` mode the response travels in the fragment or the POST
  body, and the registered query is left exactly as registered, which is what the goal asks for per
  mode. Measured: `?state=fixed&lang=en#code=fresh-code&state=client-csrf`. Nothing is duplicated
  inside one container. I would change no code there and qualify the documentation sentence by mode.

### My recommendation: A, the four documented names, query mode only

It is the only option under which both sealed sentences are true without editing either, and it removes
the response that says success and failure simultaneously, which is the row most likely to break a real
client. The divergence it creates from `end_session_endpoint` is the same kind the agreement already
documents and the code already comments: two endpoints, two specifications, differing on one point.

Reversing A later costs one entry in a reserved-name list and one documentation sentence.

### What it costs to be wrong either way

Toward **A**: a deployment relying on a registered `?state=fixed` coming back silently loses that
field. Contained to one function plus a docs sentence.

Toward **B**: this PR ships documentation stating a promise the code does not keep, and a conforming
RP reading `error` first treats a successful authorization as a refusal.

Both are reversible in small work. Both are visible at the client's callback, which is why I am asking
rather than picking.

### Where the run is

Stopped, waiting on this. Four stages landed and pushed, check suite green on `445f416` across all 11
jobs including the data and integration tiers on SQLite, MySQL, PostgreSQL and SQL Server. Nothing is
committed that depends on the answer. Reply with **A**, **B** or **C** and I will apply it, land the
change, and take the PR out of draft.

leodip:
A

leodip:
## Decisions: 13 settled, none deferred

**Nothing was deferred.** All 13 decisions in the agreement are `Decided`, so no part of this branch
rides on an assumption waiting to be confirmed after merge. This comment exists so that statement is
on the record rather than inferred from silence, and so the three decisions the user did not settle
during the interview are visible without opening anything.

Ten were settled by the repository owner during the interview, before the agreement was sealed. The
other three are below: two the reviewer raised at the plan gate and the run applied without asking,
and one the run raised mid-review and the owner answered on this thread.

---

### 11. Is a `state` sent without a value "present" for the purposes of the response?

Raised by the reviewer at the plan review, `blocking`. **Applied without escalating**, because the
normative text admits exactly one answer and so there was no question to put to anyone.

**Decided: no. A valueless parameter is an absent one at this endpoint, so `state` is emitted if and
only if its value is non-empty.** RFC 6749 section 3.1: "Parameters sent without a value **MUST** be
treated as if they were omitted from the request." Appendix A defines `VSCHAR = %x20-7E`, and
Appendix A.5 gives `state = 1*VSCHAR` while naming sections 4.1.2, 4.1.2.1 and 4.2.2, which are the
*response* sections, so `state=` is not a legal response value either. Both alternatives are closed
by that text: echoing `state=` contradicts it, and refusing the request is ruled out because section
3.1 makes the parameter omitted rather than malformed.

This is the decision that changed the design before a line was written. The sealed plan carried a
separate presence flag (`AuthContext.StatePresent`, `redirectErrorInput.statePresent`, two new
emitter parameters and a `statePresent` template bind key) to distinguish "sent empty" from "not
sent". Under the fetched text there is no such distinction, so all of it was deleted from the design
and **no function signature changed anywhere in this branch**.

Two things it deliberately did **not** change: a whitespace-only `state` is still echoed byte for
byte, because space is `%x20` and so is VSCHAR; and `end_session_endpoint` keeps its own
presence-based behaviour, because RP-Initiated Logout 1.0 carries no valueless-parameter rule
anywhere in the document. Two endpoints, two specifications, differing on one point.

**Reversing costs:** re-introducing the flag through four call paths and a bind key, against the
quoted text. Not a knob, a reading of the RFC.

### 12. Do the two form_post branches get the cache instruction the response mode requires?

Raised by the reviewer at the plan review, `blocking`. **Applied without escalating**: it only
narrows what is accepted, changes no interface, costs nothing at runtime, and it lands in the two
branches this change was already rewriting.

**Decided: yes. `Cache-Control: no-store` and `Pragma: no-cache` on both form_post renderers.**
OAuth 2.0 Form Post Response Mode section 2: "Because the Authorization Response is intended to be
used only once, the Authorization Server **MUST** instruct the User Agent (and any intermediaries)
not to store or reuse the content of the response."

Verified against the tree rather than assumed: the two production renderers of `form_post.html` are
the only ones, neither set a cache header, and nothing on the `/auth` routes supplied one. The
success page carries an authorization code and the state and was being returned as a plain cacheable
200. The header pair is exactly what the token endpoint, DCR, the health check and
`handlerhelpers/http_helper.go` already write.

**Reversing costs:** two lines per branch, and a documented `MUST` unmet.

### 13. Which names in a registered redirect URI's query does an authorization response own, and in which response modes?

Raised by the run, from the reviewer's round 1 finding at the final gate, after every stage had
landed. **Escalated to this thread and answered by the owner: "A".** The full question, the measured
behaviour, the three options and the recommendation are in
[this comment](https://github.com/leodip/goiabada/pull/227#issuecomment-5309515515).

In short: the shared construction dropped a registered field only when the selected response was
itself emitting that name, so every response name the chosen path did not emit survived from the
registered query into the response. Two of the five measured rows shipped a response that reads as a
success and a failure at once (a registered `error=stale` surviving beside a fresh `code`), and two
put a `state` in front of the client, on the redirect carrying the authorization code, that the
client never sent.

It was the owner's call rather than the run's for two reasons, both stated in the escalation. No
`MUST` in RFC 6749 forces either answer: 3.1.2 requires the registered query to "be retained when
adding additional query parameters", while 4.1.2 says only "The client MUST ignore unrecognized
response parameters". And the same codebase already answers this exact shape the opposite way on
purpose, at `end_session_endpoint`, which is sealed by decision 3 and out of scope.

**Decided: A, the four documented names, query mode only.** A registered `code`, `state`, `error` or
`error_description` is dropped from a query-mode authorization response whether or not the response
emits it. `error_uri` is not in the set (this server never emits one, so reserving it would delete a
field nothing can collide with), `fragment` and `form_post` leave the registered query exactly as
registered, and `end_session_endpoint` passes `nil` and is byte for byte unaffected.

**Reversing costs, stated at the time and unchanged:** one entry in the reserved set and one
documentation sentence. What it costs a deployment: a client that deliberately registered
`?state=fixed` and sends no `state` stops receiving it, and keeps it by choosing a name this server
does not emit.

---

Applied in `0209f8a3`, with `site/src/content/docs/concepts/clients.mdx` corrected in the same
commit, because the sentence stage 3 shipped promised the four names are "replaced by the server's
own value", which was false in exactly the rows this decision is about.

leodip:
## Follow-ups: one, ready to file

One real defect was found during this run that belongs to a different change. It is drafted here to
be filed as it stands, not summarised. The run does not file issues; that is yours.

Duplicate search re-run just now, over every state, since the tracker may have moved while the run
was working:

```
gh issue list --state all --search "repeated parameters"        -> #202 (open, credentials in the query string)
gh issue list --state all --search "duplicate request parameter" -> #109 (closed), #122 (closed)
gh issue list --state all --search "parameter more than once"    -> #219 (closed, rate limiting)
gh issue list --state all --search "FormValue duplicate"         -> nothing
```

Nothing covers it, open or closed. The nearest neighbours are #202 (credentials in the query string,
open), #122 (redirect URI absoluteness, closed) and #109 (the logout redirect, closed).

---

### Reject repeated authorization endpoint parameters instead of silently selecting one

**Labels:** `bug`, `go`. Worth deciding whether it also deserves `security`, see the last paragraph
of the body.

**Where it came from.** Raised by the reviewer during the plan review as an out-of-scope observation,
then verified against the code here before being drafted.

**Why it is out of scope for #146.** This issue is bounded to how the authorization *response* is
constructed. That is request validation, it touches every parameter rather than `state`, and it
changes which requests are refused, which is its own risk to assess.

**Title:**

```
Authorization endpoint accepts a repeated request parameter and silently uses the first occurrence
```

**Body:**

> The authorization endpoint reads every request parameter through `r.FormValue`
> (`src/authserver/internal/handlers/handler_authorize.go`, `HandleAuthorizeGet`), which returns the
> first occurrence and drops any others. There are 12 such reads, covering `client_id`,
> `redirect_uri`, `response_type`, `response_mode`, `code_challenge`, `code_challenge_method`,
> `max_age`, `acr_values`, `state`, `nonce`, `scope` and `ui_locales`, and the validators receive
> scalars from them. A request carrying a parameter twice is therefore accepted, with one value
> silently chosen and no audit trace: `?client_id=a&client_id=b` is not an error today.
>
> RFC 6749 section 3.1: "Request and response parameters MUST NOT be included more than once."
> Section 4.1.2.1 lists exactly this shape under `invalid_request`, a request that "includes a
> parameter more than once", so the error to answer with is already named by the specification.
>
> Refuse such a request instead, at the same point the other request validation runs. One thing to
> decide as part of it: whether the refusal is emitted to the client as a redirect or shown as the
> interstitial, since a duplicated `redirect_uri` or `client_id` is precisely the case where the
> response cannot be trusted to a redirect.
>
> Worth deciding whether this deserves the `security` label. The concrete risk is parameter
> pollution, where this server and any proxy, gateway or logging layer in front of it read different
> occurrences of the same parameter.

**File it with:**

```bash
gh issue create \
  --title "Authorization endpoint accepts a repeated request parameter and silently uses the first occurrence" \
  --label bug --label go \
  --body 'The authorization endpoint reads every request parameter through `r.FormValue` (`src/authserver/internal/handlers/handler_authorize.go`, `HandleAuthorizeGet`), which returns the first occurrence and drops any others. There are 12 such reads, covering `client_id`, `redirect_uri`, `response_type`, `response_mode`, `code_challenge`, `code_challenge_method`, `max_age`, `acr_values`, `state`, `nonce`, `scope` and `ui_locales`, and the validators receive scalars from them. A request carrying a parameter twice is therefore accepted, with one value silently chosen and no audit trace: `?client_id=a&client_id=b` is not an error today.

RFC 6749 section 3.1: "Request and response parameters MUST NOT be included more than once." Section 4.1.2.1 lists exactly this shape under `invalid_request`, a request that "includes a parameter more than once", so the error to answer with is already named by the specification.

Refuse such a request instead, at the same point the other request validation runs. One thing to decide as part of it: whether the refusal is emitted to the client as a redirect or shown as the interstitial, since a duplicated `redirect_uri` or `client_id` is precisely the case where the response cannot be trusted to a redirect.

Worth deciding whether this deserves the `security` label. The concrete risk is parameter pollution, where this server and any proxy, gateway or logging layer in front of it read different occurrences of the same parameter.

Found during the work on #146, which fixed the response side of the same RFC 6749 section 3.1 sentence.'
```

---

### Two things that were raised and are deliberately *not* follow-ups

Recorded so they are not rediscovered as gaps.

**A browser was never launched for the form_post response mode.** Raised as `unchecked` by the
reviewer at rounds 1 and 2, dismissed both times. The branch's whole change to `form_post.html` is
four lines wrapping the `state` input in `{{if .state}}`, which is the same construct the `code`,
`error` and `error_description` inputs above it already use and which ships on `main` today. The
auto-submit body and the `<form>` element are untouched, and the form submits whichever hidden inputs
are present without enumerating them. No assertion in this branch depends on browser behaviour, so
there is no action to file.

**RP parsers that still treat a literal `;` as a query separator were not surveyed.** Raised as
`unchecked` at round 1, dismissed with the boundary stated, because the survey cannot change what
this server writes. `encodeResponseParams` escapes every value with `url.QueryEscape`, which renders
`;` as `%3B`, so only the registered bytes can carry one, and reproducing those exactly is RFC 6749
section 3.1.2's retention `MUST`. Splitting on `;` to filter inside such a field would mean
re-implementing a parse Go itself removed in 1.17, in order to alter bytes an operator registered and
`ValidateClientAndRedirectURI` had just matched exactly. The claim this change makes, that exactly
one `state` reaches the client, holds under the normative parse and is not claimed beyond it.

leodip:
## The review account

This is the whole independent reading this code got, pasted rather than linked, because the verdicts
live in a gitignored directory that the next issue's first round overwrites.

Reviewer throughout: **`gpt-5.6-sol` at effort `xhigh`**, taken from `.review/reviewer.env` rather
than inferred. `LEO_REVIEW_MODE=end` was in force for the whole run, so **no stage was reviewed on
its own**: one round read the plan before stage 1, and three rounds read the whole branch against
`origin/main` after every stage had landed.

| Gate | Round | Findings | Blocking / significant | Outcome |
|---|---|---|---|---|
| plan | 1 | 2 | 2 blocking, both `needs_human: true` | both confirmed against the specs and applied, as decisions 11 and 12, before stage 1 |
| final | 1 | 1 | 1 significant, `needs_human: true`, `forced_answer` empty | escalated as decision 13, answered by the owner, landed as `0209f8a3` |
| final | 2 | 1 | 1 minor | fixed on the branch, `2dd0ec06` |
| final | 3 | 0 | none | clean: zero findings, zero `unchecked`, zero follow-ups |

Round 3 reported nothing, so it was unproductive by the round budget's own measure and no round 4 was
fundable. The review closes at three of four.

---

## Plan review, round 1

Request `146-20260816T182846Z`, gate `plan`, fresh Codex session. Verdict: **2 findings, both
`blocking` at confidence `high`, both `needs_human: true` with a populated `forced_answer`.** All
three axes (`conformance`, `quality`, `security`) reported `reviewed`.

What it ran: the contract, the plan, all three probes and the emitter, validator, template and test
seams; it re-ran the three probes and its own scratch model of the planned helper; `check-anchors.sh`
(16 anchors, passed); and it read RFC 6749 sections 3.1, 3.1.2, 4.1.2, 4.1.2.1, 4.2.2 and Appendix
A.5, RFC 9700 sections 2.1 and 4.7.1, and Form Post Response Mode section 2 from their publishers.

**Finding 1, blocking, conformance: an empty `state` is omitted by RFC 6749, so the planned presence
flag implements a distinction the protocol does not have.** **Confirmed and applied**, as decision 11.

Confirmed by fetching the RFC rather than by accepting the verdict. Section 3.1: "Parameters sent
without a value MUST be treated as if they were omitted from the request." Appendix A:
`VSCHAR = %x20-7E`. Appendix A.5: `state = 1*VSCHAR`, defined for sections 4.1.2, 4.1.2.1 and 4.2.2
among others, which are the response sections, so `state=` is not a legal response value either. Both
alternatives are closed by that text, which is what made it a correction to apply rather than a
question to escalate.

The one part of the superseded decision's reasoning the finding did not address was checked too,
since it was the only thing that could have kept the question open: consistency with #109's logout
redirect. RP-Initiated Logout 1.0 section 2 contains no valueless-parameter rule anywhere in the
document and says of `state` only "If included in the logout request, the OP passes this value back
to the RP". Two endpoints governed by two documents that differ here, so the logout behaviour is
correct as it stands and was not touched.

Effect: `state` is emitted iff non-empty. `AuthContext.StatePresent`, `redirectErrorInput.statePresent`,
the two planned emitter parameters and the template's `statePresent` bind key were deleted from the
design before being written, the template guard became `{{if .state}}`, and **no function signature
changes in this branch at all**.

**Finding 2, blocking, conformance: both form_post branches return a cacheable 200 carrying an
authorization code or a state, with no instruction not to store it.** **Confirmed and applied**, as
decision 12.

Form Post Response Mode section 2: "Because the Authorization Response is intended to be used only
once, the Authorization Server MUST instruct the User Agent (and any intermediaries) not to store or
reuse the content of the response." Confirmed against the tree: the two renderers of `form_post.html`
are the only ones, neither set a header, and nothing on the `/auth` routes supplies one.
`Cache-Control: no-store` and `Pragma: no-cache` went into the error emitter in stage 2 and the
success emitter in stage 3.

**Nothing was refuted.** One recommendation was declined with its reason: a third integration case
asserting the cache headers on a real route, where the sealed agreement fixed the count at two and no
middleware on those routes writes cache headers at all. One follow-up was drafted (see the follow-ups
comment). Its two `unchecked` entries were answered in the plan log: no suite exists to run at a plan
gate, and no live caching-proxy reproduction was made.

Detail at the time: [this comment](https://github.com/leodip/goiabada/pull/227#issuecomment-5309106987).

---

## Final review, round 1

Request `146-20260816T200111Z`, gate `final`, fresh session. Scope: the whole branch against
`origin/main`, ten files, four commits `37b3e942` through `445f4167`. **One finding, `significant`,
`needs_human: true`, `forced_answer` empty. Not fixed: turned into decision 13 and escalated. No
production code changed this round.**

**Finding 1, significant, conformance: registered response names survive whenever the selected path
does not emit that name.** **Verified against the code and extended, then escalated.**

`writeResponseParams` filtered a registered field only when its decoded name appeared in the `params`
slice being appended, and both emitters build `params` with `state` conditional on a non-empty value.
So a request without `state` plus a registered `?state=fixed` left `state=fixed` in the response on
both paths; a success response retained a registered `error` and `error_description`; an error
response retained a registered `code`.

Reproduced by driving the real emitters rather than by reading, in
`probe/registered_response_names.go`. That probe extended the reviewer's account in three ways it did
not reach:

- **`error_uri` is a fifth member.** RFC 6749 4.1.2.1 defines it, this server never emits it, and a
  registered one therefore always survives. The shipped documentation named four fields, not five, so
  the reserved set's membership was part of the question rather than a detail.
- **Fragment mode, measured:** `?state=fixed&lang=en#code=fresh-code&state=client-csrf`. Nothing is
  duplicated inside one container, so this is a documentation-qualification question rather than a
  code defect.
- **`end_session_endpoint` already answers the same shape the opposite way, deliberately.** Its table
  row "A registered state survives when none is supplied" asserts exactly the retention row 1 shows
  at `/auth/authorize`, and `buildPostLogoutRedirect`'s comment calls it "preserved deliberately". It
  is one of the 23 rows the agreement seals, so it is out of scope and must not move whatever is
  decided.

That last point settled the disposition. Two sealed statements of the agreement pointed opposite ways
about the same shape, so this was not a finding the run could fix in passing. The reviewer said the
same independently by leaving `forced_answer` empty.

**The spec does not force it either, and that was fetched rather than recalled.** Section 3.1.2's
registered query "MUST be retained when adding additional query parameters" argues for keeping every
registered field. There is no sentence in 3.1.2, 4.1.2, 4.1.2.1, 4.2.2 or 4.2.2.1 forbidding `state`
in a response to a request that carried none; 4.1.2 says only "The client MUST ignore unrecognized
response parameters", and these names are recognized. Section 3.1's "MUST NOT be included more than
once" bites only where the server also emits the name, which is the case this change had already
closed. So the specification permitted every option on offer.

**Escalated as decision 13** with the class stated rather than the instance, three options with their
prices, and a recommendation of A. The owner answered "A" at 2026-08-16T20:30:05Z. Applied as
`0209f8a3`, 210 insertions across 8 files, with four mutations all caught. See the decisions comment.

**The verdict's two `unchecked` entries, both answered rather than left to be rediscovered:** no
browser was launched for form_post (**dismissed**, the branch introduces no new template construct and
no JavaScript, so no assertion depends on browser behaviour); RP parsers beyond Go's current parser
were not surveyed (**dismissed with the boundary stated**, the server never writes a semicolon, and
filtering inside a registered field would mean re-implementing a parse the platform retired in order
to alter bytes the operator registered).

**Coverage this round did establish**, recorded because a clean axis is evidence too. The reviewer
read the whole agreement, plan, all four stage logs, the closing record and every probe before the
diff; verified the five split tree hashes against the tree, all 17 anchors, the final HEAD, all 11 CI
jobs green on `445f4167`, a clean tracked status and a clean `git diff --check`; enumerated every
production caller of `writeResponseParams`, `encodeResponseParams` and `form_post.html`; ran an
old-versus-new URI builder differential across HTTPS, authority, IPv6, custom-scheme, opaque, Unicode,
escaped-path and bare-query shapes and found no component change outside the deliberately preserved
query; and confirmed both form_post renderers buffer before writing and set the cache pair only after
a complete render. Its one mutation, restricting the decoded-name filter to an identical raw spelling,
was caught by the existing percent-encoded state test and the logout regression table. It recorded one
mutation that did not compile and correctly declined to count it as caught.

**Security found nothing:** no route, credential comparison, database predicate, token validation,
session transition, permission boundary or audit outcome changes, and no state, code, token or secret
is newly logged.

---

## Final review, round 2

Request `146-20260816T204338Z`, gate `final`, round 2 of 4, over the whole branch against
`origin/main`. All three axes `reviewed`. **One finding, `minor`, `needs_human: false`. Fixed.**

**Finding 1, minor, conformance: a lone surviving empty query field lost its delimiter.** Family: raw
query delimiter preservation after filtered fields collapse. Target `writeResponseParams`.

**Not contested. Verified by writing the case before touching the source**, which is the evidence that
matters here: the reviewer's claim is that the current code was already the failing variant, so the
test that pins it had to fail against the tree as committed. It did, naming the defect exactly:

```
--- FAIL: TestWriteResponseParams/A_lone_surviving_empty_field_keeps_the_query_delimiter
    expected: "https://app.example.com/cb?"
    actual  : "https://app.example.com/cb"
```

The mechanism: the function assigns `strings.Join(fields, "&")` to `RawQuery`, and `url.URL.String`
writes the `?` only when `RawQuery` is non-empty or `ForceQuery` is set. A field list of exactly one
empty field joins to `""`, so the delimiter went with it. A registered
`https://rp.example/cb?state=fixed&` with the reserved set applied and no response parameter appended
therefore came back as `https://rp.example/cb`, a different target from the one the client registered
and `ValidateClientAndRedirectURI` had just matched byte for byte. That is the RFC 6749 section 3.1.2
retention requirement this whole change exists to satisfy, failing on a shape no row happened to
cover.

**No shipped flow reaches it, and that was checked rather than accepted.** The combination needs an
empty `params` with a non-empty `reservedNames`. Both authorization query-mode emitters always append
at least one parameter (`redirToClientWithError` always carries `error`, `issueAuthCode` always
carries `code`), and `buildPostLogoutRedirect` passes `nil` reserved names, so with empty params it
filters nothing and the join is the identity. The three production callers exhaust the reachable
combinations, so this is a helper-contract defect and a trap for the next caller, not a live one.

**Fixed rather than deferred**, as `2dd0ec06`. The helper is the one construction every emitter's
correctness now reduces to, and #146 exists because a second copy of this logic was left alone;
shipping it with a contract its own doc comment states and its body does not keep is the same mistake
one level down. Four lines and one condition:

```go
if len(fields) > 0 && redirUrl.RawQuery == "" {
    redirUrl.ForceQuery = true
}
```

**Both boundaries are pinned**, because the interesting part is which of the two empty-query cases
keeps the `?`. Zero surviving fields still loses the query: `?state=fixed` filtered to nothing returns
`https://app.example.com/cb`, which an existing row already asserted and which is right, since there
is nothing left for the delimiter to delimit. Two new rows: the lone-empty-field case, and its sibling
with a parameter appended (`?state=fixed&` plus `code=fresh-code` gives `?&code=fresh-code`), which
passed before the fix and pins that the empty field's separator survives in front of the params.

**Two mutations against the new guard, both caught:**

| Mutation | What it neutralises | Result |
|---|---|---|
| `if len(fields) > 0 && redirUrl.RawQuery == "" {` → `if false {` | the delimiter restoration entirely | **CAUGHT** |
| `> 0` → `>= 0` | the boundary in the other direction: it would force a `?` onto the zero-surviving-fields case too | **CAUGHT** |

The second is the one worth having. It is the mutation a careless fix actually is, and only the
pre-existing zero-fields row catches it. The reviewer's own `mutations_run` was empty for this
finding, with the stated reason that the current source was already the failing variant, which is
sound for establishing the defect and says nothing about the fix. Hence both of the above.

**Tiers after the fix:** unit 3473 passed / 0 failed (3471 before, the two new rows), data 1619
passed / 0 failed on sqlite, mysql, postgres and mssql, integration 1253 passed / 0 failed on sqlite.
Of the six split tree hashes only `production` and `unit-tests` moved from the values the request
recorded, which is the fix and its two rows and nothing else. All 17 anchors resolve,
`git diff --check` clean, and `git status --porcelain` after both mutations showed only the two
intended files, so neither mutation left residue.

**The verdict's two `unchecked` entries:** browser automation not repeated (**dismissed**, nothing in
round 2 touched a template, and the `docs` and `integration-tests` hashes are unmoved); external RP
parsers treating `;` as a separator (**covered, unchanged** from round 1's stated residual).

---

## Final review, round 3

Request `146-20260816T211500Z`, gate `final`, round 3 of 4, Codex session resumed, over the whole
branch against `origin/main`. All three axes `reviewed`. **Clean: zero findings, zero `unchecked`
entries, zero follow-ups.**

**Why this is clean rather than narrow.** Zero findings in a narrow scope reads identically to zero
findings in a thorough one, so the scope that produced it is on the record. `findings` is empty,
`conformance`, `quality` and `security` all read `reviewed`, `unchecked` is empty rather than carrying
anything the gate depended on, and the response's "Not checked" section reads "Nothing in the
requested branch scope remains unchecked". What it ran:

- The complete agreement record, every probe, all seven commits, and the ten-file branch diff.
- **An independent raw-query oracle over 161,020 cases**: field sequences of length one to four across
  empty, valueless, ordinary, all four reserved, percent-encoded reserved, `error_uri`,
  malformed-escape and literal-semicolon atoms, plus no query and a bare `?`, five emitted parameter
  sets, and both reserved-set choices. Every case matched, including round 2's lone-empty-field
  boundary.
- The real success, error, fragment, form_post, logout and injection overlays driven again.
- **Two fresh mutations, both caught**, aimed at round 2's fix from the two sides the run's own
  mutations did not: discarding empty registered fields failed the helper and logout empty-field rows
  including both of round 2's new cases, and clearing `ForceQuery` before return failed the
  bare-question-mark and lone-empty-survivor rows.
- Publisher checks against RFC 6749, RFC 9700 and OAuth 2.0 Form Post Response Mode.
- Current-head CI across all 11 jobs, the 17 anchors, split hashes, tracked status and diff hygiene.

**Corroborated by the run rather than taken on trust.** Every checkable fact the response asserts was
re-established before the verdict was accepted:

| Claim | Checked with | Result |
|---|---|---|
| Mutations restored | `tree-hash.sh --split` | all six categories identical to the values round 2 recorded (`production 68be6ebd0a45`, `unit-tests 3508cfe9914d`, `data-tests cc52c7f6ac48`, `integration-tests 3cb3f1d61021`, `docs 409c34f64a4f`, `all 31371ecc1886`) |
| Tracked tree clean | `git status --porcelain` | empty |
| Citations resolve | `check-anchors.sh` | all 17 rows resolve to exactly one line |
| CI green at HEAD | `where.sh ci` | green, all 11 jobs, at `2dd0ec06` |

The `production` hash matching is the one that matters most: the reviewer ran two mutations against
production code this round, and an unchanged `production` is the evidence that neither survived.

**The one observation it raised and dismissed, verified independently.** Under Quality, the response
records that a raw non-ASCII path is serialized to its percent-encoded URI form by `net/url`, and
dismisses it as not a branch regression because the base implementation already crosses the same
`URL.String` boundary. Verified rather than accepted, since it is the only substantive technical
judgement in an otherwise empty verdict:

```
origin/main:src/authserver/internal/handlers/handler_authorize.go
  810:  redirUrl, err := url.ParseRequestURI(input.redirectURI)
  820:  redirUrl.RawQuery = values.Encode()
  822:  http.Redirect(w, r, redirUrl.String(), http.StatusFound)
```

The base sets `RawQuery` on a parsed `url.URL` and serialises with `URL.String()`, which is exactly
what the shared helper does, so path serialisation is unchanged by this branch and sits on both sides
of the diff. The branch's switch to `url.Parse` only widens which URIs parse; it does not touch how a
parsed one is written back. The dismissal stands.

**Nothing moved this round**, so round 2's tier results stand: all six split hashes are byte-identical
to the values recorded after its fix, and re-running would measure the same tree.

**Why the review closes here.** Round 3 reported nothing, so it is unproductive by the budget's own
measure and no round 4 is fundable. That is the budget working rather than a gate cut short: rounds 1
and 2 each found a real defect in the shared construction and each was answered, which is what bought
round 3, and round 3 was requested for a specific reason, that round 2's fix added a conditional to
the function every emitter reduces to and nothing but the run had read it. That reason is now
answered, by an oracle over 161,020 cases and two mutations aimed straight at the new guard.

Issue #108, opened 2026-07-25
Security: DCR clients default to ConsentRequired: false, allowing silent token issuance to self-registered clients


leodip:
## Summary

Clients created through the public `POST /connect/register` endpoint are stamped with `ConsentRequired: false`. Because the authorization-code flow applies no client-side scope restriction, the consent screen is the only control that binds *which client* is receiving *the user's authority*. Removing it means an anonymously self-registered client can obtain a code, and then tokens, for any user with a live SSO session without that user seeing a single rendered page.

Severity: **Medium** (OWASP A06:2025 Insecure Design). Bounded by DCR being off by default (`Settings.DynamicClientRegistrationEnabled`).

## Affected code

`src/authserver/internal/handlers/handler_dynamic_client_registration.go:88`

```go
client := &models.Client{
    ClientIdentifier: clientIdentifier,
    ...
    ConsentRequired:  false,
    ...
}
```

`src/authserver/internal/handlers/handler_auth_completed.go:222` is the only consent gate in the flow:

```go
// we must redirect to consent if the client requires it or if there's an offline_access scope
if client.ConsentRequired || authContext.HasScope(oidc.OfflineAccessScope) {
    ... redirect to /auth/consent
}
// if there's no need for consent, we're ready to issue the code
authContext.AuthState = oauth.AuthStateReadyToIssueCode
```

A DCR client never satisfies the first disjunct, and an attacker simply does not request `offline_access` to avoid the second.

## Why this gate carries more weight than it appears to

Scope granting in the authorization-code flow is filtered **only** against the user's permissions:

```go
// handler_auth_completed.go:186
effectiveScope, err := permissionChecker.FilterOutScopesWhereUserIsNotAuthorized(authContext.Scope, user)
```

There is no equivalent client-side check. `ClientLoadPermissions` has exactly three non-test call sites: `token_validator.go:285` (the `client_credentials` branch) and two in the admin client-permissions API. Nothing in the authorize path consults the client's own permission grants.

That is normal OAuth design. Granted scope is effectively `requested ∩ user's permissions`, and **consent is the control that binds client identity to that authority**. It is a reasonable tradeoff to skip it for a client an administrator deliberately registered. It is not reasonable for a client an anonymous caller registered seconds earlier.

## Chain

With DCR enabled:

1. Attacker calls `POST /connect/register`. Unauthenticated, rate limited only. Receives `client_id` (and `client_secret` for a confidential registration).
[...]

PR #229, opened 2026-08-16
fix(oauth): authenticate the user before redirecting an authorization error to a client

### LINKED ISSUE #208: Authorize error page renders a hardcoded English heading beside a localized page title

### LINKED ISSUE #213: Security: /auth/authorize validates the request before authenticating, so an unauthenticated user agent can be redirected to a client's URI
## Summary

[RFC 9700](https://www.rfc-editor.org/rfc/rfc9700.txt) section 4.11.2 carries a MUST that this server does not currently satisfy:

> The authorization server MUST take precautions to prevent these threats. The authorization server MUST always authenticate the user first and, with the exception of the silent authentication use case, prompt the user for credentials when needed, before redirecting the user.

`HandleAuthorizeGet` in `src/authserver/internal/handlers/handler_authorize.go` calls `authorizeValidator.ValidateRequest` and then `authorizeValidator.ValidateScopes`. Both deliver their errors to the client by redirecting to the registered redirect URI, and both run **before** the session check and before any redirect to `/auth/level1`. So a link carrying an invalid scope redirects a logged-out visitor to a host the client chose, with nothing rendered and nobody authenticated.

Severity: **Low to Medium**, depending on how a client's redirect URI came to be registered. See "What bounds it" below.

## Reproduction

Verified against a running server. A browser holding no cookies, sending an invalid scope for a client whose redirect URI is `https://attacker.example.com/callback`:

```
GET /auth/authorize/?client_id=<client>&redirect_uri=https%3A%2F%2Fattacker.example.com%2Fcallback
    &response_type=code&code_challenge_method=S256&code_challenge=...
    &scope=openid+nonsense&state=probe&nonce=...
```

```
302 Found
Location: https://attacker.example.com/callback?error=invalid_scope&error_description=Invalid+scope+format%3A+%27nonsense%27....&state=probe
```

Same result for an unknown `resource:permission`:

```
Location: https://attacker.example.com/callback?error=invalid_scope&error_description=Invalid+scope%3A+%27nosuchresource%3Anosuchperm%27....&state=probe
```

No login page, no consent screen, no user interaction beyond loading the URL.

## Why the ordering matters

This is the first of the three attacks RFC 9700 section 4.11.2 describes for an authorization server used as a redirector:

> 1. Intentionally send an erroneous authorization request, e.g., by using an invalid scope value, thus instructing the authorization server to redirect the user agent to its phishing site.

The other two are a declined consent (the server redirects anyway, regardless of what the user chose) and a `prompt=none` request. The spec's remedy for the first is the ordering: authenticate before redirecting, so an anonymous drive-by cannot use the endpoint at all.

## What bounds it

Stating these plainly, since they are why this is not High:

- The redirect URI must be registered on that client, so this is only interesting where an attacker can get one registered. With dynamic client registration disabled (the default) that means persuading an administrator.
- The destination is the client's exact, pre-registered redirect URI. Matching is plain string equality in `src/core/validators/authorize_validator.go`, with no wildcards or prefixes, so this does not let anyone reach an arbitrary URL.
- The response carries only an OAuth error, no code and no token.

The value to an attacker is the redirect itself: a link that appears to point at the authorization server's own domain and lands somewhere else, which is the phishing property RFC 9700 is describing.

## Relationship to #108

#108 mitigates the consequence for one class of client without changing the ordering. It adds an interstitial for clients created through dynamic client registration: rather than redirecting, the server shows the user where the client wants to send them and lets them choose. That is RFC 9700's SHOULD-level answer:

> The authorization server SHOULD only automatically redirect the user agent if it trusts the redirection URI. If the URI is not trusted, the authorization server MAY inform the user and rely on the user to make the correct decision.

The MUST quoted at the top asks for something different and broader: that **no unauthenticated user is redirected at all, for any client**, self-registered or not. #108 leaves that untouched, deliberately.

## Suggested fix

Move request and scope validation to after authentication in `HandleAuthorizeGet`, so the flow becomes:

1. validate `client_id` and `redirect_uri` (unchanged, and already required to come first by RFC 6749 4.1.2.1, which says the server MUST NOT redirect to an invalid redirect URI)
2. authenticate the user
3. validate the rest of the request and its scopes
4. deliver any error by redirect

`prompt=none` keeps its current ordering, per the silent authentication carve-out in the sentence quoted above and per OIDC Core 3.1.2.1.

**Worth sizing carefully before starting.** `/auth/authorize` is the busiest path in the server, and a large number of existing tests assert the current ordering by driving a request that fails validation and expecting a redirect. Expect the change to be mostly test churn rather than handler code, and expect at least one case where an error currently reported to the client would now require a login first, which is a behaviour change for clients that rely on fast failure.

## References

- RFC 9700 section 4.11.2 (Authorization Server as Open Redirector), BCP 240
- RFC 6749 section 4.1.2.1 (Error Response)
- OpenID Connect Core 1.0 section 3.1.2.1 (`prompt=none`)
- #108, which adds the interstitial for dynamically registered clients

leodip:
Closes #213 when merged. Closes #208 as well: that issue reports the hardcoded
`"Unable to authorize"` heading beside a localized page title, and stage 5 replaces it with a
catalog key rendered in the visitor's locale.

## What this is

`GET`/`POST /auth/authorize` runs five validations that answer the client with a 302 to its
registered `redirect_uri`, and every one of them runs before the handler has looked at the session.
A logged-out browser that loads one crafted link is therefore redirected to a host the client chose,
with nothing rendered and nobody authenticated. That is attack 1 from RFC 9700 (BCP 240) section
4.11.2, verbatim:

> Intentionally send an erroneous authorization request, e.g., by using an invalid scope value, thus
> instructing the authorization server to redirect the user agent to its phishing site.

The same section states the obligation this change implements:

> The authorization server MUST always authenticate the user first and, with the exception of the
> silent authentication use case, prompt the user for credentials when needed, before redirecting
> the user.

#108 implemented the neighbouring **SHOULD** from that section, withholding the error redirect from
self-registered clients. This is the **MUST**, and it is wider: it turns on who is at the keyboard
rather than on where the redirect URI came from.

### The rule this establishes

> **Authentication is required before a redirect, and only before a redirect.** Where the answer is
> a page, it is rendered at once. Where the answer is a redirect, the ceremony authenticates first,
> unless the request asked for silence or the visitor is already authenticated.

Concretely: a validation failure with nobody authenticated parks the error code and description on
the auth context, sends the visitor to `/auth/level1`, and delivers the error redirect from
`/auth/level1completed` once credentials are verified. A browser already holding a valid session is
answered immediately, exactly as today.

The error still reaches the client, which is what keeps OIDC Core 1.0 3.1.2.2 ("If the Authorization
Server encounters any error, it MUST return an error response, per Section 3.1.2.6") satisfied at
the same time. Withholding the redirect entirely would have satisfied RFC 9700 and departed from
OIDC Core for the majority of traffic, since logged out is the common case for a fresh login rather
than an edge.

Two carve-outs, both deliberate:

- **`prompt=none` keeps today's ordering completely.** RFC 9700's exception clause ("with the
  exception of the silent authentication use case") and OIDC Core 3.1.2.3's "The Authorization
  Server MUST NOT interact with the End-User ... the Authentication Request contains the prompt
  parameter with the value none" both require it. Silence is read from the raw `prompt` parameter,
  case-sensitively, because a prompt the validator rejects is never stored, and `prompt=none login`
  still contains the value `none`.
- **`prompt=login` defers even with a valid session.** It is the client saying not to answer on the
  strength of an existing session, and that request was going to show the login page regardless.

Where no redirect was going to be emitted anyway, because #108's provenance guard or #122's
emittability guard will withhold it, the blocked page is rendered at once with no login: no redirect
leaves the server, so the requirement that governs redirects has nothing to say.

### A second, independent defect

The refusal page at `/auth/authorize` is the only page in this flow that ignores the visitor's
locale: its title is a Go literal and its message is the validator's English sentence, while the two
other renderers of the same template localise both. Seven messages reach that page, all from
`ValidateClientAndRedirectURI`. They become `*i18n.LocalizedError` with catalog entries in both
locales, English values byte-identical to today's text.

`error_description` on the redirect path stays English, per RFC 6749 4.1.2.1, which names the client
developer as its audience. That line is now carried by a comment so a later reader does not "fix"
the inconsistency by putting non-ASCII into a redirect parameter.

### Third: no `error_description` carries a character RFC 6749 forbids

Found by the plan review, before any code landed, and settled by the repository owner on this thread.

RFC 6749 Appendix A.8 gives one production for the authorization redirect, the implicit redirect and
the token endpoint alike: `error-description = 1*NQSCHAR`, where NQSCHAR is
`%x20-21 / %x23-5B / %x5D-7E`. That admits printable ASCII except the double quote and the
backslash, and excludes every control character, DEL and every byte above 0x7E.

Descriptions interpolate request text. `ValidateScopes` puts a rejected scope into three of its
descriptions and `ValidatePrompt` puts a rejected prompt value into one, and `token_validator.go` has
twelve more of the same shape. Nothing filtered them, so `scope=💣` put byte 0xf0 into a protocol
parameter. This is pre-existing behaviour rather than something the ordering change introduces, and
there is no injection: query and fragment go through `url.QueryEscape` and `form_post` through an
escaped template. It is a conformance defect.

The fix is one pure function applied at every point where a description becomes an
`error_description`: the authorization endpoint's single emitter, the token endpoint's error exits,
and the new parking site, which also bounds the string on its way into a size-limited cookie. It is
deliberately not applied in the shared JSON error writer, whose other 177 call sites are the admin
console's AJAX API and `/userinfo`: RFC 6749 5.2 does not govern those, and this server's
localisation commitment points the other way for an admin screen.

## Status

**Complete and ready for review.** All five stages have landed, the review has run and closed, and
the check suite is green on the final commit.

The security defect this issue is about is fixed and proven end to end: `/auth/authorize` no longer
redirects an unauthenticated browser to a client's redirect URI, and the deferral is driven through
the real login ceremony by integration tests. The two defects added to this change alongside it have
landed too: every `error_description` this server emits now conforms to RFC 6749's character set at
both endpoints, and the refusal page renders in the visitor's locale.

**Both escalated decisions are answered and built.** Decision 10 came from the plan review, before
any code landed, and its answer `C` inserted the conformance fix as stage 2, because stage 3 parks a
description in a bounded cookie and must bound it on the way in. Decision 11 came from round 1 of the
review of the finished branch, and its answer `C` landed as `83985610`, on top of the five stages
rather than as a rewrite of one. **No decision is deferred and none is open**, which the
[decisions comment](https://github.com/leodip/goiabada/pull/229#issuecomment-5315624803) sets out
along with the three things that were deliberately left alone and why.

### Tests

| Tier | Engines | Result | Run by |
|---|---|---|---|
| unit | n/a | 3528 pass, 0 fail | locally, per stage |
| data | sqlite, mysql, postgres, mssql | 1619 pass, 0 fail | locally, per stage |
| integration | sqlite | 1266 pass, 0 fail | locally, per stage |
| integration | mysql, postgres, mssql | pass | the check suite only |
| `Lint`, `Vulnerabilities`, `Versions` | n/a | pass | the check suite only |

Green on `cd41ee60`, the final commit:
[run 32025896369](https://github.com/leodip/goiabada/actions/runs/32025896369). **Ten of the eleven
jobs ran and all ten passed**, including the data and integration tiers on all four database engines
as parallel jobs.

The eleventh, `Build validation`, is **skipped on a pull request by design**, not by anything about
this branch: `check.yml` gates it on `github.event_name == 'push' || inputs.full`, and the workflow's
push trigger is `branches: [main]`. It is the only job that exercises the release path, five
cross-compiles plus both Docker images, and it runs on the push to `main` after merge and again in
`release.yml` against the tagged commit. Nothing in this change touches build tags, the Dockerfiles or
the module graph, so there is no reason to expect it to behave differently here, but it has not run on
this code and this line says so rather than counting it as green.

**Every new test was proven able to fail.** Twenty-seven mutations were run across the five stages and
the two review fixes (2, 4, 7, 4, 4, then 4 and 2), one per behaviour the change claims to have made
observable, and **all twenty-seven were caught**. The ones worth naming are the four that a green suite could never have
found on its own: a page never handed the request's locale, a status code nothing asserts, a filter
that exists and is correct and is never called, and a delivery point wired without the template
filesystem it needs for one of three response modes. Each passes every other assertion in the tree.

### Review

Reviewed by `gpt-5.6-sol` at effort `xhigh`, three rounds: one over the plan before any code landed,
then two over the whole branch against `origin/main`. No stage was reviewed on its own.

|---|---|---|---|---|
| 1 | plan | 3 | 1 | 1 escalated as decision 10, 1 folded into it, 1 resolved in-plan |
| 2 | final | 2 | 2 | both fixed: `7a97baff`, and `83985610` after decision 11 was answered |
| 3 | final | 1 | 0 | verified true, pre-existing, filed as a follow-up; two record corrections in `cd41ee60` |

- **blocking, conformance:** the token endpoint's error boundary conformed a description only when the
  error was an `*ErrorDetail`. Everything else, which is what a malformed request body produces,
  reached the shared writer unchanged, and that writer builds its description around chi's request id,
  which chi adopts verbatim from the caller's own `X-Request-Id`. So an unauthenticated caller could
  put `"`, `\`, a tab or any byte above 0x7E into `error_description`, which is precisely what the
  third part of this change set out to make impossible. Fixed in `7a97baff`: both branches are
  conformed, and a test runs the two paths over the same error so the duplicated sentence cannot
  drift. A raw-socket probe written to check the finding showed four of the five forbidden shapes
  survive a real connection, so it landed stronger than the evidence offered for it.
- **blocking, conformance:** OpenID Connect Core 3.1.2.6 ends with an exception this change did not
  account for. An unsupported `response_mode` is answered with a local HTTP 400 carrying no error
  parameters, because a server that cannot encode the requested mode cannot return parameters in it.
  This server redirected it to the client instead, and on this branch a logged-out visitor was asked
  for a password first, which put a credential prompt ahead of a redirect that should not exist.
  Inherited from `main` rather than introduced here, and the scope was a genuine choice, so it was
  escalated as **decision 11** and answered **C**: a local 400 for every authorization request, not
  only for an OIDC Authentication Request. Fixed in `83985610`, described in
  [What decision 11 changed](#what-decision-11-changed) below.
- **significant, conformance:** response-mode capability is modelled as one global set, but the real
  capability depends on the response type, so `form_post` is advertised for implicit requests and
  refused for them, and an explicit `response_mode=query` on an implicit request is refused in query.
  Verified true in all three of its parts, and **entirely pre-existing**: the two files carrying most
  of it are byte-identical to `origin/main` on this branch. Dispositioned as a follow-up rather than
  fixed here, because closing it means deciding how implicit tokens are delivered. `cd41ee60` corrects
  the two durable places where this branch had written a too-loose justification for it, one docs
  sentence and one test comment, and touches no production code.

**No security defect was found at any round.** The full account, with the spec quotes, the probes, the
refutations and the mutation tables, is in
[this comment](https://github.com/leodip/goiabada/pull/229#issuecomment-5315625938).

### Follow-ups

Two, neither belonging to this change, both now filed: **#230** (six validators' catalog messages are
never asserted, only their codes) and **#231** (implicit flow response-mode support is inconsistent).
[The follow-ups comment](https://github.com/leodip/goiabada/pull/229#issuecomment-5315625399) carries
the reasoning behind each, plus three candidates that should **not** be filed and the reason for each.

### Stage 1: the emission predicate, the clear-then-answer helper, and the parked-error fields

`36731d5d`. Nothing a client or a browser receives moves in this stage; it exists so the ordering
change in stage 3 is the ordering change alone.

- **`redirectWillBeEmitted(client, redirectURI, site)`** answers whether an error response to a
  client would actually leave the server as a redirect, or whether it would be withheld and
  replaced by the refusal interstitial. It is #108's provenance gate and #122's emittability last
  resort asked as one question. The emitter enforces the answer at the last moment; stage 3's
  handler asks the same question earlier, to decide whether anybody has to authenticate first, and
  one definition is what stops the two drifting. It takes a `site` label because
  `checkRedirectURIEmittable` logs where a refusal happened and deliberately never logs the URI
  (#159), so the existing log line is unchanged.
- **`answerClientWithError`** performs #141's clear-then-answer sequence once, replacing the two
  hand-written copies in `HandleAuthorizeGet` and `handlePromptNone`. Its `server_error` fallback is
  derived from the caller's own input with the code and description swapped, so the authorize
  closure keeps answering from the request and the silent path from the stored ceremony, exactly as
  each did before. `redirectErrorFromRequest` names the request-reading twin of
  `redirectErrorFromAuthContext` that the closure was inlining.
- **`AuthContext.DeferredErrorCode` and `DeferredErrorDescription`**, unread so far.
  `DeferredErrorCode != ""` is the sentinel for a parked error, and it is sound because the five
  deferrable validations have 23 error returns between them and none carries an empty code, while
  the empty-code constructor is used only by `ValidateClientAndRedirectURI`, which answers a
  rendered page and never a redirect.

Tiers: unit 1063 pass / 0 fail; integration on sqlite 1253 pass / 0 fail, which here is the whole
existing suite proving the refactor changed nothing. `TestRedirectWillBeEmitted` owns the
provenance and emittability table (four rows) and caught both mutations run against the new
predicate, one per clause. The four-engine data and integration tiers are the check suite's.

### Stage 2: the description filter, at both boundaries

`d848643d`. Decision 10, answered **C**. Conformance rather than a security fix, and pre-existing
rather than introduced here: every emitter already escapes, so nothing injected before this change
either. What it closes is that no boundary in this server filtered a description on its way to
becoming a protocol parameter, so every validator written in future inherited the same gap.

- **`ConformErrorDescription`** (`src/core/customerrors`) replaces each rune outside RFC 6749
  Appendix A.8's `NQSCHAR` (`%x20-21 / %x23-5B / %x5D-7E`) with one `?`, then bounds the result to
  512 bytes ending in `...`. Replacement rather than dropping, because dropping turns `'a"b'` into
  `'ab'` and tells an integrator their scope was a string they never sent. Runes rather than bytes,
  so a four-byte emoji costs one `?`. The bound second, so it operates on single-byte characters and
  cannot split a rune. 512 is measured against the tree: the longest description this server
  produces with nothing interpolated is 214 characters. Conforming input is returned byte-identical,
  which is why no existing assertion in the tree moved.
- **The authorization boundary** is `redirToClientWithError`, which conforms once before the
  response-mode dispatch, so the query slice, the fragment built from it and the `form_post` bind
  map all carry the filtered value. It sits *below* the redirect guard on purpose: the refusal
  interstitial renders the description onto an HTML page, which is a user interface rather than a
  protocol parameter, and keeps the validator's own text.
- **The token boundary** is a package-local `jsonErrorConformed` wrapping `JsonError` at all nine of
  `handler_token.go`'s exits, rather than `JsonError` itself. 175 of that writer's 186 call sites
  are the admin console's AJAX API, which section 5.2 does not govern and which this server answers
  in the administrator's language.
- **`ErrorDetail.WithDescription`** clones the detail rather than round-tripping through four
  getters and the four-argument constructor, which reads correct today and would silently drop any
  detail key added later.

The filter is idempotent, and that is load-bearing rather than tidy: stage 3 filters again at the
parking site, and the description a client receives on the deferred path has to be byte-identical to
the one it receives immediately.

Tiers: unit 1065 pass / 0 fail; integration on sqlite 1255 pass / 0 fail. Data is not applicable,
since the stage adds no query, interface method, migration or schema change. Four mutations, all
caught: the replacement made a no-op and the bound neutralised, both caught by the filter's own
twelve-row table; then the call removed from each boundary in turn, caught by the two integration
rows, because a filter that exists, is correct and is never called passes every row of the unit
table. The four-engine data and integration tiers are the check suite's.

### Stage 3: authenticate before the redirect

Landed in `d91a61f0`. The ordering change itself.

`HandleAuthorizeGet` now computes a single boolean before the five validations that answer a client
by redirect, and it decides how all five are answered:

```go
answerClientNow := requestsSilence ||
    !redirectWillBeEmitted(client, authContext.RedirectURI, "authorize") ||
    (!requestsLogin && hasValidUserSession())
```

When it is false the error is parked on the auth context and the visitor is sent to `/auth/level1`;
`HandleAuthLevel1CompletedGet` emits the redirect once the password is verified, from a new branch
directly below its state gate. The client receives the same error code, description, response mode
and `state` either way, so only the timing changes. That ceremony ends at the delivery point and
creates no session, since the session row is written at `/auth/completed`.

`requestsSilence` and `requestsLogin` are read from the **raw** `prompt` parameter, case-sensitively,
on whitespace-separated tokens. A prompt the validator rejects is never assigned to
`AuthContext.Prompt`, and OIDC Core 3.1.2.3 forbids interacting with a request that "contains the
prompt parameter with the value `none`" whether or not the rest of it parsed, so `none login` stays
silent while `NONE` does not.

The session load is lazy and memoized, and the two pure clauses are evaluated first. That is a
deviation from the design sketch, settled at the plan-review gate: an eager load would add a query
to every `prompt=login` request and a second one to every `prompt=none` request, and would let a
database fault preempt three outcomes that answer today without touching the session. An unreadable
session still fails closed with a 500 for the one outcome that needs it, because reading it as
absent would send a signed-in user back through the login page.

The parked description goes through stage 2's `ConformErrorDescription` on the way into the cookie
as well as on the way out of the emitter. The filter is idempotent, so the two paths stay
byte-identical, and a bound applied only at emission would leave `ChunkedCookieStore`'s 50-chunk cap
reachable through an interpolated scope.

Two closures where there was one. The disabled-account `access_denied` calls the immediate one
explicitly and never defers: a disabled user sent to the login page cannot complete it, so the error
their client is owed would never arrive.

All existing assertions are kept and made to pass by giving each browser a session, rather than
being retargeted at the login page. They exist to prove the error response is correct, which is
orthogonal to who is at the browser. **44 tests, not the 43 the sweep predicted**: the integration
run found `TestAuthorize_ValidateRequest_FormPostResponseMode`, which asserts its error through the
rendered `form_post` page rather than a URL and so escaped the classifier. The remedy was the same
one-line change as the other 30.

Documentation landed with it: a `Deferred error redirects` section in
`concepts/authorization-lifecycle.mdx` with the routing table, the precondition added to
`oauth2-flows/authorization-code.mdx`, a sibling bullet to #108's in `reference/security.mdx` (that
one turns on where the redirect URI came from, this one on who is at the browser), and the flow
handler table in `CLAUDE.md`. `concepts/prompt-parameter.mdx` was confirmed unaffected rather than
edited, since the silent path is unchanged.

Tiers: unit 3523 pass / 0 fail; data 1619 pass / 0 fail on sqlite, mysql, postgres and mssql;
integration on sqlite 1255 pass / 0 fail. The four-engine integration tier is the check suite's.
Seven mutations, all caught: each of the three clauses in turn, the deferral itself, the laziness
made eager, the delivery gate at `/auth/level1completed`, and the filter at the parking site.

### Stage 4: the deferral, end to end

Landed in `18f7a474`. Seven integration tests in one new file,
`authorize_deferred_error_test.go`, and no production change: stage 3 built the deferral, this is
the tier that drives it through the real ceremony rather than through a recorder.

The reproduction inverted first, and as two assertions rather than one. That the cookieless browser
reaches `/auth/level1` is the ceremony working; that the client's URI was not emitted is the
security requirement, and an emitter answering both would satisfy the first while breaking the
second.

Then delivery after the login in all three response modes, with the `Location` compared as a whole
string so the parameter order is part of the assertion. `form_post` is the case that matters: it is
the only mode that parses a template, so it is the only one that needs the `templateFS` parameter
stage 3 added to `HandleAuthLevel1CompletedGet`. A delivery point wired without it redirects
perfectly well in the other two modes and answers 500 there.

All three modes send `state` = `csrf a+b/c=`, the value that pinned #146, and require
`csrf+a%2Bb%2Fc%3D` on the wire from a **literal** rather than from a call to `url.QueryEscape`, so
the encoding is pinned by the test file and not by the function the emitter uses. RFC 6749 4.1.2
requires "the exact value received from the client", and carrying an error through a second handler
is how that property gets lost.

One case belongs to the conformance fix rather than to the deferral: the same emoji-bearing scope
is sent both ways and the two answers must be the same bytes, and NQSCHAR. The filter runs at the
parking site and again at the emitter and is idempotent so the paths cannot diverge; a replacement
or bound applied at only one of them passes every single-path test in the tree and fails here.

Also `POST /auth/authorize`, since §2's property names it and `r.FormValue` resolves from the body
there, with the foreign `Origin` that keeps the endpoint's CSRF exemption (#67) on the path; and
that the ceremony leaves no SSO session, asserted behaviourally by requiring the next authorization
request to ask for the password again.

`prompt=none` is deliberately not re-asserted here. It is unchanged and already has 13 passing
integration tests, one of which drives a cookieless silent request all the way to the client's URI.

Tiers: unit 3523 pass / 0 fail; data 1619 pass / 0 fail on sqlite, mysql, postgres and mssql;
integration on sqlite 1262 pass / 0 fail, seven of them new. The four-engine integration tier is the
check suite's. Four mutations against stage 3's handler, all caught: forcing `answerClientNow` true,
blanking the `state` and then the `responseMode` carried on the auth context, and dropping the
emitter's conform call, which only the parity case above can see. The parking-site conform call is
deliberately not mutated: it exists for the cookie bound, which no integration test can observe.

### Stage 5: the refusal page in the visitor's locale

`fb010ee3`. The second defect the user added to this change, and the last stage.

`/auth/authorize` was the only page in the login flow that ignored the visitor's language. It passed
`"Unable to authorize"` as a Go literal and put the validator's own English sentence beneath it,
while both other renderers of the same template localise title and message alike, so the page
arrived with its chrome translated and its content not.

`ValidateClientAndRedirectURI` now returns `*i18n.LocalizedError` at all seven of its sites: a
missing or unresolvable `client_id`, a disabled client, a client without the authorization code
flow, and a `redirect_uri` that is missing, not absolute, or not registered. Seven codes join the
registry under `validator.authorize.*`, eight key pairs (those seven plus the title) go into both
catalogs, and the handler renders `Localize(r.Context())` against the locale it has already refined
from `ui_locales` a few lines earlier.

**Why these seven and not the five above them.** It is not a line drawn by taste. RFC 6749 4.1.2.1
forbids redirecting to an invalid redirection URI, so each of these seven renders a page and none of
them can ever become an `error_description`, which is the parameter the ASCII rule binds and which
that section addresses to the client developer. The five validations that run afterwards do become
one. A comment on the closure they share now says so, and says what a translation added anyway would
actually meet: stage 2's filter, which would deliver an accented sentence as a row of question
marks.

**The English text did not move.** The catalog values are byte-identical to the literals they
replaced, so the seven integration assertions on `p#errorMsg` stand exactly as written, and they are
now the regression guard proving nothing was quietly reworded on the way through the catalog.

Two adjacent corrections came with it. The comment on the registry's email group claimed no authorize
error is represented there, which this stage makes false; it now states the rule that holds, about
`error_description` rather than about endpoints. And the validators package never loaded the i18n
bundle in `TestMain`, so `EnglishFallback()` returned message ids instead of sentences and no test in
that package could assert a rendered string at all.

Unit 1069 pass / 0 fail, integration on sqlite 952 pass / 0 fail, one of them new. Four mutations,
all caught: restoring the hardcoded title, swapping `Localize` for `EnglishFallback`, giving one
condition another condition's code, and rewording an English catalog value. The first two are caught
only by the new `ui_locales=pt-BR` case, which is the whole reason it exists: a page wired to the
catalog but never handed the request's locale passes every English assertion in the file.

Locale coverage is deliberately one condition rather than seven. Catalog parity across locales and
the English fallback for an unknown locale are each proved once in `src/core/i18n`; what neither can
prove is that the request's locale reaches this page, which was the defect.

## What decision 11 changed

`83985610`. Not a stage: it came out of the review of the finished branch, so it is a commit on top
of the five rather than a rewrite of one of them. The code it sits above is stage 3's.

OpenID Connect Core 1.0 section 3.1.2.6 ends with an exception to the rule that returns authorization
errors to the client:

> If the Response Mode value is not supported, the Authorization Server returns an HTTP response code
> of 400 (Bad Request) without Error Response parameters, since understanding the Response Mode is
> necessary to know how to return those parameters.

This server did the opposite. `ValidateRequest` rejected any `response_mode` outside `query`,
`fragment` and `form_post` with `invalid_request`, and the emitter has no branch for a mode it does
not recognise, so the error fell through to the query default and answered the client in a mode it
had not asked for and may not be able to read.

`HandleAuthorizeGet` now answers it directly, on the refusal page at 400 with no `error`,
`error_description` or `state`. The branch sits above the client load, the session lookup and all
five redirecting validations, and that position is the substance rather than a detail: an error that
cannot be encoded in the requested mode can be neither redirected nor parked, so whichever other
failure the request also carries, this is the answer. The refusal page's renderer took a status
parameter to serve it; the client and `redirect_uri` conditions keep the 200 they have always
answered.

**Applied to every authorization request**, not only to an OIDC Authentication Request, which is what
answer `C` chose. The sentence's reasoning does not turn on the scope, and OAuth 2.0 Multiple Response
Type Encoding Practices section 2.1, which defines `response_mode`, states no behaviour for an
unsupported value, so extending the rule contradicts nothing. Splitting on whether `openid` is in the
scope would leave this server answering one malformed parameter two ways for a reason no later reader
could infer.

Three things the answer did not settle, decided against the code:

- The supported set has **one definition**, `validators.IsSupportedResponseMode`, called by the
  handler and by `ValidateRequest`, so a mode added to one cannot be refused by the other.
  `ValidateRequest` keeps its own branch even though the handler now answers first: the rule belongs
  to the validator, and a second caller would not inherit the handler's.
- **The implicit-flow rule is untouched.** `response_mode=query` on an implicit request asks for a mode
  this server does implement, so 3.1.2.6's unsupported-mode exception does not reach it and the new
  branch correctly leaves it alone: the request is refused by the ordinary validation and the refusal
  reaches the client. It has its own unit case, because it is the case a reader is most likely to think
  the new branch covers. What that case pins is inherited behaviour rather than correct behaviour:
  Multiple Response Type Encoding Practices section 3 says the query encoding MUST NOT be used for
  these response types, and that prohibition attaches to the response type's encoding rather than to
  tokens specifically, so it reaches the error response too and this refusal should be in the fragment.
  The deviation predates this change and this change does not widen it; it is
  [follow-up 2](https://github.com/leodip/goiabada/pull/229#issuecomment-5315625399), and the case's
  comment carries the citation.
- **The page message is localised**, like the seven this same page already renders, with the English
  value byte-identical to the `error_description` the condition used to carry. A new English literal
  here would have reintroduced the defect stage 5 closed.

One existing assertion was deliberately replaced rather than kept, and it is the only one in the
change: `TestAuthorize_ValidateRequest_InvalidResponseMode` asserted the 302 and its
`error_description`, which pinned behaviour the specification contradicts. It now asserts the 400 and
the page, still driven by a browser that holds a session, because that is what shows the answer does
not depend on who is at the keyboard. Four cases were added around it: a logged-out visitor is
answered rather than sent to log in, the message renders in the request's locale, the branch precedes
everything that can redirect, and a supported-but-unusable mode still reaches the client.

Unit 3528 pass / 0 fail, data 1619 pass / 0 fail on all four engines, integration on sqlite 1266 pass
/ 0 fail. Four mutations, all caught: the branch neutralised, the predicate forced true, the page
status forced to 200, and the localised message replaced with an English literal. The last two matter
most, because a status nothing asserts and a page never handed the request's locale both pass every
other assertion in the tree.


leodip:
## Decision 10: untrusted request text reaches `error_description` unbounded and unfiltered. How far does this change close it?

The plan review for #213 found this before any code landed, and it needs your call because the fix is
a genuine choice rather than something the code forces. The branch is still `main` plus one empty
commit, so nothing has been built on either answer.

### What the run was assuming, and why that was wrong

The agreed design left `error_description` on the redirect path alone, on the grounds that it is
already plain ASCII and addressed to the client developer rather than the end user. It also planned to
write a source comment on the five authorize validators saying exactly that, so the next reader would
not "helpfully" localise them and put non-ASCII into a redirect parameter.

The scoping was fine. **The assumption underneath it was not: the descriptions are not ASCII when the
input is not.**

### The rule

RFC 6749 section 4.1.2.1, for the authorization error response:

> Values for the "error_description" parameter MUST NOT include characters outside the set
> %x20-21 / %x23-5B / %x5D-7E.

Section 5.2 repeats it verbatim for the token endpoint, and Appendix A.8 gives one production,
`error-description = 1*NQSCHAR`, covering sections 4.1.2.1, 4.2.2.1, 5.2 and 7.2. The permitted set
excludes every byte above 0x7E (so all non-ASCII), every control below 0x20, DEL, the double quote
and the backslash.

### What the code actually does

`ValidateScopes` interpolates the rejected scope into three of its descriptions, `ValidatePrompt`
interpolates the rejected token into one, and nothing filters them on the way out. Run against the
real validators:

| Input | Description produced | Verdict |
|---|---|---|
| `scope=bogus` | `Invalid scope format: 'bogus'. ...` | conforming |
| `scope=💣` | `Invalid scope format: '💣'. ...` | violates, byte 0xf0 |
| `scope=абв` | `Invalid scope format: 'абв'. ...` | violates, byte 0xd0 |
| `scope=a"b` | `Invalid scope format: 'a"b'. ...` | violates, byte 0x22 |
| `scope=a\b` | `Invalid scope format: 'a\b'. ...` | violates, byte 0x5c |
| `scope=a<NUL>b` | `Invalid scope format: 'a\x00b'. ...` | violates, byte 0x00 |
| `prompt=💣` | `Invalid prompt value: 💣` | violates, byte 0xf0 |

Two things that bound how alarming this is, both checked rather than assumed:

- **It is pre-existing, and this change does not widen it.** The error closure passes the description
  through untouched today, and deferring the redirect until after login carries the same bytes to the
  same place. Every row above is current `main` behaviour.
- **There is no injection.** Query and fragment values go through `url.QueryEscape` and `form_post`
  through an escaped template. This is a conformance defect, not a security one, which is why it is a
  question rather than something I fixed and reported.

### The class is wider than the authorize endpoint

The reviewer scoped this to the four authorize sites. Checking the rest of the tree,
`src/core/validators/token_validator.go` has **thirteen more** of the identical shape, reaching the
token endpoint's `error_description` under section 5.2. The real fact is that no boundary in this
server filters a description on its way to becoming a protocol parameter, so every validator written
later inherits the gap. That is why the question below is about the boundary rather than about four
call sites.

### A second, smaller thing that shares the remedy

The same text is now unbounded on its way into a bounded cookie. The plan parks the error code and
description on the auth context so the redirect can be delivered after login. The design called that
size-safe because the cookie store disables securecookie's own 4096-byte check, which is true, but
`ChunkedCookieStore.Save` imposes a separate hard cap of 50 chunks, about 190 KB, and returns
`session too large: requires N chunks (max 50)` above it.

The handler already saves the context once before the validations, so an oversized request fails at
that first save today. What the change adds is a band where the first save fits and the second, now
carrying the description, does not. In that band a logged-out visitor gets a 500 instead of the
deferral. It fails closed, no redirect is emitted, and the caller can only break their own request.
A length bound on the description closes the band, which is why it is folded in here rather than
asked separately: one answer settles both.

### Options

- **A. Out of scope.** File the whole class as a follow-up covering both endpoints. This change then
  corrects only its own claims, so nothing false ships: the goal's "stays English and ASCII" becomes
  "stays English", and the planned comment says these descriptions are not localised without
  asserting the character set is enforced.
- **B. Fix the authorize boundary, file the token endpoint.** Filter and length-bound the description
  inside the clear-then-answer helper that stage 1 is already extracting, which is the single point
  every authorize error redirect passes through, including the new deferred one. Covers all five
  deferrable validators and every future one, and closes the cookie band in the same step. Roughly one
  step plus a table test.
- **C. Fix both boundaries.** B, plus the token endpoint's JSON error writer, closing the whole class
  here. Thirteen more sites, on a handler and a test surface this change never scoped.
- **D. Drop the interpolated detail.** Replace the dynamic descriptions with fixed text. Fully
  conforming, and it removes the diagnostic that tells an integrator *which* scope was wrong.

### Recommendation: B

It is the smallest change that makes the run's own claims true rather than softening them. It acts at
a boundary this change is already extracting, so it costs one filter and not a sweep of the
validators, and it keeps the diagnostic that D throws away. C does the same work twice on a surface no
seam covers, and the token endpoint is genuinely a different change: different handler, different
tier, thirteen sites. A is defensible precisely because the defect predates this change, and its
honest cost is that the ASCII sentence in the goal has to be softened rather than satisfied.

### What getting it wrong costs

Choosing **A** when you wanted B means the four stages land and the emitter is reopened afterwards,
with the comment asserting a property the code lacks shipped in between. Choosing **B** when you
wanted A means deleting one filter and its test. That asymmetry is why this is asked now rather than
after stage 4.

Reply with the letter, with or without reasoning. The run is stopped until then and no code has been
committed.

leodip:
C

leodip:
## Decision 11: OIDC Core answers an unsupported `response_mode` with a local 400. We redirect it to the client, and this branch now makes the visitor log in first. Does #213 close that, and for which requests?

Raised by the run, from the final review of this branch (round 1, finding 2). The run is stopped on
your answer.

### The text

OpenID Connect Core 1.0, section 3.1.2.6, closes with an exception to the redirect rule stated
immediately above it:

> If the Response Mode value is not supported, the Authorization Server returns an HTTP response
> code of 400 (Bad Request) without Error Response parameters, since understanding the Response Mode
> is necessary to know how to return those parameters.

No RFC 2119 keyword, so it is not a literal MUST. The reasoning is self-justifying: a server that
does not understand the requested mode does not know how to encode a response in it.

The other document in play says nothing. `response_mode` is defined by OAuth 2.0 Multiple Response
Type Encoding Practices section 2.1, "OPTIONAL. Informs the Authorization Server of the mechanism to
be used for returning Authorization Response parameters from the Authorization Endpoint", and it
states no behaviour for an unsupported value. So the sentence above is the only normative statement
there is, and it sits in a document that governs OIDC Authentication Requests rather than plain
OAuth ones. That is what makes the scope a real question rather than a transcription.

The spec **requires** the local 400 for an OIDC Authentication Request, **permits** anything at all
for a plain OAuth request, and **exists to enable** a client that asked for a mode this server
cannot speak getting back an answer it can actually parse.

### What the code does today

`ValidateRequest` (`src/core/validators/authorize_validator.go`) rejects any `response_mode` outside
`query`, `fragment`, `form_post` with `invalid_request` and "Invalid response_mode parameter.
Supported values are: query, fragment, form_post." That is one of the five validations `#213`
reroutes. `redirToClientWithError` has no branch for an unrecognised mode, so it falls through to
the query default and answers the client in a mode the client did not ask for.

### What this branch did to it

For a visitor holding a valid session, nothing: the immediate redirect shipped before this branch
and ships now. For a logged-out visitor the error is parked and a password form is shown first, so
this branch put a credential prompt in front of a redirect the spec says should not be emitted at
all. The redirect itself is inherited, not introduced by #213.

### Options

**A. Follow-up.** File it and ship #213 as it stands. The defect predates the branch and nothing the
change claims is falsified: its goal is about *when* a redirect is emitted, not about whether this
particular one should exist. Cost: the branch merges carrying a gap the reviewer called blocking,
and the new credential prompt in front of it ships too.

**B. Local 400, OIDC requests only.** An Authentication Request (`scope=openid`) with an unsupported
`response_mode` is answered locally with 400 and no `error`, `error_description` or `state`. Matches
the governing document exactly and changes nothing it does not govern. Cost: one parameter, two
behaviours, split by whether `openid` is in the scope.

**C. Local 400 for every request** carrying an unsupported `response_mode`. One behaviour to
document and test, and the spec's own reasoning applies word for word to a plain OAuth request.
Extends a rule past the document that states it, which nothing forbids because the OAuth document is
silent. Cost: a behaviour change for non-OIDC clients that no specification compelled.

**Rejected before it is offered**, in case it is the first instinct: keep the redirect and merely
stop deferring it, so at least no password is asked for. That is incoherent. Answering a logged-out
browser immediately with a redirect to a client-chosen host is exactly the RFC 9700 section 4.11.2
requirement this whole change exists to satisfy, so it would trade a conformance deviation for a
security one.

### Recommendation: C

The sentence's reasoning does not turn on OIDC at all, and B leaves this server answering the same
malformed parameter two different ways depending on a scope value, which is a seam that gets got
wrong later by whoever does not know why it is there. It also matches how you settled the
`error_description` character-set question earlier on this branch: the whole class closed inside the
change that found it, rather than half closing and the rest being filed.

B is the strictly-conformant minimum and is a perfectly good answer if you would rather this change
touched nothing OIDC Core does not reach.

### What it costs to be wrong

Choosing **A** when B or C was wanted means merging a known blocking finding and reopening
`HandleAuthorizeGet` afterwards. Choosing **B** or **C** when A was wanted means removing one
handler branch, one test and one docs line, all contained.

### What either fix touches

A branch in `HandleAuthorizeGet` above the deferral routing, since the answer must not reach the
parking closure at all; the existing local refusal page rendered with status 400; and
`TestAuthorize_ValidateRequest_InvalidResponseMode` in
`src/authserver/tests/integration/authorize_request_scope_validation_test.go`, which asserts the
current 302 and its `error_description` and would assert the 400 instead.

That last one is why this is a decision rather than a repair. The plan for this change says every
existing assertion is kept and made to pass, and this would be the single assertion deliberately
replaced, because it pins behaviour the specification contradicts.

### Meanwhile

The other blocking finding from the same review, the token endpoint emitting a caller-chosen
`X-Request-Id` verbatim into `error_description`, is independent of this and is already fixed and
pushed in `7a97baff`.

leodip:
C

leodip:
## Deferred decisions: none

Eleven decisions were taken on this change and **all eleven are `Decided`**. Nothing on this branch
rests on an assumption waiting to be confirmed, and nothing is waiting on you.

Two of the eleven were raised by the run rather than at the interview, and both were answered on this
thread:

| # | Question | Answered | Where it landed |
|---|---|---|---|
| 10 | Untrusted request text reaches `error_description` unfiltered, and this change now also parks it in a bounded cookie. How far does this change close that? | **C**, fix both boundaries, 2026-08-16 | stage 2 (`d848643d`), plus the token endpoint's generic branch in `7a97baff` |
| 11 | An unsupported `response_mode` is redirected to the client rather than answered with a local 400, against OIDC Core 3.1.2.6. Fix it here, and for which requests? | **C**, a local 400 for every authorization request, 2026-08-17 | `83985610` |

Decision 10's answer widened the change by one stage. Decision 11's arrived after all five stages had
landed, which is why it is a commit on top of them rather than a rewrite of one.

### Three things were deliberately left, and none of them is a deferral

A deferral is a question answered by assumption. These are answered, with reasons, and they are here
so the absence of a deferral list is not read as "nothing was left alone":

- **`HttpHelper.EncodeJson` falls back to `JsonError` at four success-path exits in
  `handler_token.go`**, bypassing the conforming boundary stage 2 added. Not fixed, and it cannot
  matter: `EncodeJson` has already written a 200 and a `Content-Type` before `Encode` can fail, so
  that path cannot produce a well-formed RFC 6749 5.2 error response whatever its description says.
  Conforming it would dress up a response that is already malformed at the HTTP level.
- **The shared JSON error writer is not conformed**, only the token endpoint's own exits. 175 of that
  writer's 186 call sites are the admin console's AJAX API and `/userinfo`. RFC 6749 5.2 does not
  govern those, and stripping every byte above 0x7E there would be this change quietly forbidding a
  Portuguese sentence on an admin screen.
- **Discovery builds `response_modes_supported` from its own literal** rather than from the shared
  `validators.IsSupportedResponseMode` this change introduced. Wiring them together would change no
  behaviour today, since the two lists are identical, and both drift directions already fail safe: a
  mode added to the predicate only goes unadvertised, and a mode added to discovery is still refused
  by the validator. It is named in follow-up 2 as part of that family rather than treated as closed.

leodip:
## Follow-ups: two filed, three that should not be

**Both are now filed, verbatim from the drafts below: #230 and #231.** Each draft's claims were
re-verified against the tree at filing time, and the deviations from what is written here are
recorded under each one. Duplicate searches were run over **every** state, so a closed issue that
already settled one would have been found.

---

## Follow-up 1: six validators' catalog messages are never asserted, only their codes

Found while building stage 5. Not this change's work: #213 touched only `authorize_validator.go`, and
widening to the other six would have meant rewriting test tables this change has no reason to open.

Duplicate search, `--state all`, over `i18n bundle validators test`, `LocalizedError catalog assert`,
`localization test coverage` and `catalog message asserted`: **nothing tracked**.

**Filed as #230.**

**Title:** `Six validators' catalog messages are never asserted, only their codes`
**Labels:** `go`, `tests`

**Body:**

> `src/core/validators` has seven validators returning `*i18n.LocalizedError`: authorize, address,
> email, identifier, password, phone and profile. Their tests assert `locErr.Code` and nothing else,
> for example in `email_validator_test.go` and `address_validator_test.go`. Nothing in the package
> asserts the sentence a user actually reads.
>
> The reason is mechanical rather than a choice anyone made. Until #213 the package's `TestMain`
> (`src/core/validators/test_main_test.go`) did not call `i18n.LoadBundle`, and `EnglishFallback()`
> returns the message id when `defaultBundle` is nil, which is i18n's deliberate visible-miss policy.
> Any assertion on rendered text would therefore have compared `validator.email.required` against
> "Please enter an email address." and failed, so nobody wrote one. #213 added the `LoadBundle` call,
> because the seven authorize messages are asserted on their sentences: keeping those sentences
> byte-identical through the move into the catalog is exactly what proves the refusal page still says
> what it always said.
>
> So the mechanism is now in place for the other six and unused by them.
>
> **Why it is worth closing.** A code assertion proves a validator picked the right branch. It cannot
> see a catalog entry that is wrong, empty of meaning, reworded by accident, or mapped to the wrong
> condition, and `TestCatalog_ParityEnPtBR` and `TestCatalog_NoEmptyValues` cannot see those either:
> parity proves a key exists in both locales and non-emptiness proves it is not blank. Nothing proves
> `validator.password.too_short` is about a password being too short. The concrete failure is a
> catalog edit that swaps two entries in a group: every test stays green and every affected form tells
> the user to fix the wrong field.
>
> **What closing it takes.** Each of the six test files has a table with an `expectedCode` column. Add
> an `expectedMessage` column and one `assert.Equal(t, tt.expectedMessage, locErr.EnglishFallback())`
> per table, mirroring what `authorize_validator_test.go` does now. Roughly six tables. It needs no
> production change: `TestMain` already loads the bundle since #213.
>
> Worth doing per group rather than all at once, and worth a mutation per group afterwards (reword one
> English catalog value, confirm the group's table catches it), because that mutation is the only
> thing that distinguishes an assertion that reads the catalog from one that reads a duplicate of it.
>
> **Anchor:** `src/core/validators/test_main_test.go`, the `i18n.LoadBundle` call in `TestMain`.

**Filed as #230 with** the body above, unmodified:

```bash
gh issue create \
  --title "Six validators' catalog messages are never asserted, only their codes" \
  --label go \
  --body-file followup-1.md
```

**One deviation from this draft.** The `tests` label does not exist in this repository, so #230
carries `go` alone. The closest precedent, #195 ("Four of the seven rate limiters have no tests"), is
labelled `bug, go`, so either that pair or a new `tests` label would match convention.

**One thing the body says that is true only on merge.** "`TestMain` already loads the bundle since
#213" describes this branch, which was still open when #230 was filed.

---

## Follow-up 2: implicit response-mode support is inconsistent across five surfaces

Found by round 2 of this branch's review. Verified against the code and against fetched spec text,
and **entirely pre-existing**: the two files carrying most of it are byte-identical to `origin/main`
on this branch. It ships identically whether this branch merges or not, which is why it is a follow-up
rather than a blocking finding.

Duplicate search, `--state all`, over `response_mode`, `form_post`, `implicit fragment` and
`response_modes_supported`: returns #121 (custom-scheme redirect URIs with `form_post`), #160
(`issueAuthCode` partial 200 on a `form_post` template failure), #228 (repeated request parameters)
and #197 (implicit ceremony and an ended session). **None covers this.**

**Filed as #231.**

**Title:** `Implicit flow response-mode support is inconsistent: form_post is advertised but rejected, and an error for an explicit response_mode=query is emitted in query`
**Labels:** `bug`, `go`

**Body:**

> Goiabada treats response-mode support as a single global set, but its actual capability depends on
> the response type. The two facts do not line up, in both directions, and the seam runs through
> request validation, discovery metadata, the success emitter, the error emitter, the tests and the
> docs.
>
> **1. `form_post` is advertised for implicit requests and refused for them.**
>
> `HandleWellKnownOIDCConfigGet` advertises `response_modes_supported` as `["query", "fragment",
> "form_post"]` unconditionally, and `validators.IsSupportedResponseMode` accepts the same three. But
> `AuthorizeValidator.ValidateRequest` contains:
>
> ```go
> // Per RFC 6749 4.2.2 and OIDC Core 3.2.2.5: implicit grant tokens MUST be in fragment
> // If response_mode is explicitly set for implicit flow, it must be fragment
> if isImplicitFlow && len(input.ResponseMode) > 0 && input.ResponseMode != "fragment" {
>     return customerrors.NewErrorDetailWithHttpStatusCode("invalid_request",
>         "Implicit flow requires response_mode=fragment or no response_mode (fragment is the default for implicit flow).",
>         http.StatusBadRequest)
> }
> ```
>
> So `response_type=token&response_mode=form_post`, and the same for `id_token`, `id_token token` and
> `token id_token`, is refused although the server advertises the mode. That refusal is not required by
> either document the comment cites. OAuth 2.0 Form Post Response Mode section 4 says:
>
> > "In particular, it is safe to return Authorization Response parameters whose default Response
> > Modes are the query encoding or the fragment encoding using the form_post Response Mode."
>
> and Appendix A of the same document demonstrates exactly
> `response_type=id_token&response_mode=form_post`. OpenID Connect Core 3.2.2.5 uses the fragment
> "unless a different Response Mode was specified", and OAuth 2.0 Multiple Response Type Encoding
> Practices prohibits only the *query* encoding for these response types, not `form_post`.
>
> **2. Even with that validation bypassed, the success emitter cannot honour the mode.**
> `issueImplicitTokens` in `handler_auth_issue.go` has no response-mode dispatch. It ends in:
>
> ```go
> http.Redirect(w, r, redirectURI+"#"+encodeResponseParams(params), http.StatusFound)
> ```
>
> so an implicit success is always a fragment regardless of `AuthContext.ResponseMode`. The error
> emitter `redirToClientWithError`, by contrast, does implement a `form_post` branch and renders it
> correctly, so success and error capabilities disagree with each other as well as with discovery.
>
> **3. In the other direction, an explicit `response_mode=query` on an implicit request is refused,
> and the refusal is emitted in query.** `redirToClientWithError` defaults an implicit error to the
> fragment only when no mode was supplied:
>
> ```go
> effectiveResponseMode := input.responseMode
> if isImplicitFlow && effectiveResponseMode == "" {
>     effectiveResponseMode = "fragment"
> }
> ```
>
> An explicitly supplied `query` therefore falls through to the query branch. For the plain implicit
> grant, RFC 6749 section 4.2.2.1 says the authorization server
>
> > "informs the client by adding the following parameters to the fragment component of the
> > redirection URI using the 'application/x-www-form-urlencoded' format"
>
> and for the OIDC response types, Multiple Response Type Encoding Practices section 3 (`id_token`)
> and section 5 (`id_token token` and the hybrid types) say
>
> > "The default Response Mode for this Response Type is the fragment encoding and the query encoding
> > MUST NOT be used."
>
> The same document says "Both successful and error responses SHOULD be returned using the supplied
> Response Mode", which collides with that MUST NOT precisely here; the MUST NOT governs. So this
> refusal should be in the fragment.
>
> **Why it is defensible to have left it.** No security impact: the implicit flow is disabled by
> default, no token is present in any of these responses, and every emitter already escapes, so
> nothing injects. Every part of it predates #213, which touched none of the code above:
> `handler_auth_issue.go` and `handler_well_known_oidc_config.go` are unchanged on that branch, and
> the implicit rule in `ValidateRequest` and the fragment default in `redirToClientWithError` appear in
> its diff only as unchanged context. #213 added `IsSupportedResponseMode` as a shared definition for
> the *unsupported*-mode set and a unit case pinning item 3's current behaviour, whose comment records
> the deviation and cites the sentences above.
>
> **What closing it would take.** Model response-mode capability per response type rather than
> globally, and make one source of truth govern all five surfaces:
>
> - decide the support contract for implicit `form_post`: either implement it in
>   `issueImplicitTokens` for every advertised implicit response type, with all token fields and the
>   `no-store` / `no-cache` headers the error emitter's `form_post` branch already sets, or narrow what
>   discovery advertises and document the narrower contract with the reason.
> - keep `query` prohibited for implicit responses and return the rejection of an explicit
>   `response_mode=query` through a permitted mechanism, which is the fragment.
> - have discovery derive `response_modes_supported` from the same definition the validator uses
>   instead of its own literal `[]string{"query", "fragment", "form_post"}` in
>   `handler_well_known_oidc_config.go`. A review mutation that appended a mode to
>   `validators.supportedResponseModes` alone left the validator, handler, emitter and discovery tests
>   all green, so nothing currently pins the two together. Both drift directions fail safe today,
>   which is why #213 left this alone, but it is the seam that lets the set diverge.
> - add one invariant test joining request validation, discovery, success emission, error emission and
>   the documentation, so a mode cannot be advertised without being emittable.
>
> **Anchors.** `src/core/validators/authorize_validator.go` `ValidateRequest`, at
> `if isImplicitFlow && len(input.ResponseMode) > 0`.
> `src/authserver/internal/handlers/handler_auth_issue.go` `issueImplicitTokens`.
> `src/authserver/internal/handlers/handler_well_known_oidc_config.go`
> `HandleWellKnownOIDCConfigGet`, at `responseModes := []string{`.
> `src/authserver/internal/handlers/handler_authorize.go` `redirToClientWithError`, at
> `if isImplicitFlow && effectiveResponseMode == ""`.

**Filed as #231 with** the body above, unmodified:

```bash
gh issue create \
  --title "Implicit flow response-mode support is inconsistent: form_post is advertised but rejected, and an error for an explicit response_mode=query is emitted in query" \
  --label bug --label go \
  --body-file followup-2.md
```

**Re-verified at filing time**, since the draft was written at review round 2 and the branch moved
after it: the discovery literal, `IsSupportedResponseMode`, the implicit rule in `ValidateRequest`,
and the fragment default in `redirToClientWithError` all still read as quoted, and
`issueImplicitTokens` still ends in the unconditional fragment redirect.

---

## Three candidates that should not be filed, and why

Recorded so a later run does not raise them again as new.

1. **The refusal page renders in English only.** Real, and **fixed in this change**: you pulled it in
   during the interview, it became decision 9 and the second half of the goal, and it landed as stage
   5 (`fb010ee3`).
2. **`/auth/authorize` has no rate limiter.** Dropped on your instruction, and already settled
   deliberately: #219 is closed as completed and its out-of-scope list names this endpoint explicitly,
   with the reason ("all are unauthenticated work amplifiers rather than credential oracles ...
   `/auth/authorize` is the heaviest public endpoint"). Filing it fresh would be a duplicate. One
   thing has changed and is worth keeping: decision 7 records that the missing limiter is what rules
   out an audit event on the new deferral path, since an unauthenticated unlimited endpoint writing to
   `audit_logs` is a log-flood primitive. If a limiter ever lands on this route, that event becomes
   available and is worth adding with it.
3. **The token endpoint interpolates untrusted scope text into `error_description`.** Drafted at the
   plan gate as out of scope, then **pulled into this change** by decision 10's answer `C`. It landed
   in stage 2 (`d848643d`), covering all nine of `handler_token.go`'s error exits, and the generic
   non-`*ErrorDetail` branch that stage missed was closed by review round 1 in `7a97baff`.

leodip:
## The review account

Every round this change was given, what it found, and what happened to each finding. The verdicts
themselves live in a gitignored working directory that the next round overwrites, so this comment is
the durable copy.

Reviewer throughout: **`gpt-5.6-sol` at effort `xhigh`**, a fresh Codex session per round, reviewing
correctness, security and conformance. Under `LEO_REVIEW_MODE=end` **no stage was reviewed on its
own**: one round read the plan before any code landed, then the whole branch was read against
`origin/main` twice after all five stages had landed.

|---|---|---|---|---|---|
| plan | 1 | 3 | 1 | 1 | 1 escalated as decision 10, 1 folded into it, 1 resolved in-plan |
| final | 1 | 2 | 2 | 0 | 1 fixed (`7a97baff`), 1 escalated as decision 11 then fixed (`83985610`) |
| final | 2 | 1 | 0 | 1 | verified true, pre-existing, dispositioned as follow-up 2; two record corrections in `cd41ee60` |

**No security defect was found at any round.** Every finding was a conformance or quality defect, and
the three that were verified as real were all pre-existing behaviour this change did not introduce.

---

## Plan gate, round 1

`type: plan-review`, against an unchanged production tree (`git diff origin/main..HEAD` empty). Three
findings, all `needs_human: true` with an empty `forced_answer`. No tier was owed or run: no
production or test file had been modified yet.

| # | Severity | Axis | Claim | Outcome |
|---|---|---|---|---|
| 1 | blocking | conformance | `ValidateScopes` and `ValidatePrompt` copy untrusted text into `error_description` with no character-set gate, against RFC 6749 4.1.2.1 | **Confirmed by execution, escalated as decision 10** |
| 2 | significant | quality | the second auth-context save can exceed `ChunkedCookieStore`'s 50-chunk cap after the first succeeded, giving a 500 instead of the deferral | **Confirmed from source, folded into decision 10** |
| 3 | minor | quality | the eager session lookup queries where the raw prompt or the emission predicate already decides, and lets a lookup fault preempt them | **Confirmed from source, resolved in-plan** |

**Finding 1** was verified independently rather than accepted, by a probe calling the real validators
and applying the RFC's own predicate to what they returned:

| Input | Verdict |
|---|---|
| `scope=bogus` | conforming |
| `scope=💣` | violates, byte 0xf0 |
| `scope=абв` | violates, byte 0xd0 |
| `scope=a"b` | violates, byte 0x22 |
| `scope=a\b` | violates, byte 0x5c |
| `scope=a<NUL>b` | violates, byte 0x00 |
| `scope=a<DEL>b` | violates, byte 0x7f |
| `prompt=💣` | violates, byte 0xf0 |

Two corrections to the reviewer's framing, both from the code: it is **pre-existing and not widened by
this change**, since the emitter passes the description through untouched today and the deferral
carries the same bytes to the same place; and it carries **no security impact**, since every emitter
escapes (`url.QueryEscape` for query and fragment, an escaped template for `form_post`). One
correction that widened it: the reviewer's family stopped at the authorize validators, and
`token_validator.go` holds twelve more of the identical shape under RFC 6749 5.2, which is why the
escalation asked about the *boundary* rather than about the four sites.

**Finding 2.** `DefaultMaxChunks = 50` at `DefaultChunkSize = 3800`, about 190 KB, enforced in `Save`
independently of the `MaxLength(0)` line the design had relied on. Bounded from the handler: it already
saves once above the validations, so an oversized request 500s today at the first save, and what the
change would add is the band where the first fits and the second does not. Fails closed, no redirect
emitted, self-inflicted only. Folded into decision 10 because a length bound at finding 1's boundary
closes it as a side effect.

**Finding 3.** The reviewer's reading of the control flow was exact. Resolved as local and reversible
by a memoized closure with the pure clauses first: `||` short-circuits and the emission predicate is
pure, so the predicate's value is unchanged on all 19 rows of the routing table probe and only the side
effects move. Recorded as an as-built deviation in stage 2, with three session-fault unit rows added.

**`unchecked`, all three answered.** No implementation exists at a plan gate, so unrunnable suites and
mutations were dismissed as structural, with stage 1 already owning the predicate's two mutations.
Portuguese wording was dismissed here and covered by stage 4's rendered pt-BR assertion.

---

## Final gate, round 1: the whole branch

Round 1 of at most 4, over the whole branch against `origin/main`. Verdict: **two findings, both
`blocking`, both `conformance`, both `confidence: high`**. The reviewer verified all six tree hashes
and all 49 anchors before starting, and ran two scratch overlay probes that failed in the production
branches it named.

### Finding 1, blocking, conformance: the token endpoint's generic error path emits forbidden bytes

**The claim.** The conforming boundary stage 2 added filtered only a direct `*customerrors.ErrorDetail`.
Every other error reached the shared writer unchanged, and that writer builds
`"An unexpected server error has occurred. ... Request Id: %v"` around chi's request id. chi adopts an
inbound `X-Request-Id` verbatim, so an unauthenticated caller chooses part of a parameter RFC 6749
Appendix A.8 confines to `%x20-21 / %x23-5B / %x5D-7E`.

**Verified, and the mechanism is real at every link.** `server.go` installs `middleware.RequestID`;
chi v5.3.1 reads `r.Header.Get("X-Request-Id")` and generates an id only when that is empty;
`r.ParseForm()` fails on a body of `%` before any client is identified, and its error is not an
`*ErrorDetail`.

**The finding came out stronger than the evidence offered for it.** The reviewer's reproduction set the
header on a hand-built `httptest` request, which does not answer whether the bytes survive a real
connection. A new probe drove `net/http`'s server over a raw socket: `"` (0x22), `\` (0x5C), TAB
(0x09), U+1F4A3 and Cyrillic are all accepted and handed to the handler verbatim; only DEL (0x7F) is
refused with a 400 at parse. Four of the five forbidden shapes are reachable over the wire.

The code comment asserting the opposite ("a request id that no request text reaches") was the defect:
it was written from the shape of the Go error rather than from where the string comes from.

**Fixed in `7a97baff`.** A non-`*ErrorDetail` is rebuilt at the boundary as the generic server error
the shared writer would have produced, and the single conform call moved onto the final description, so
neither branch reaches the wire unfiltered and a future third branch has to come through it to be
written at all. The shared writer is untouched, per decision 10's reading. The log line that writer
emits on the generic branch is emitted at the boundary instead, keeping the **raw** request id so an
operator correlating on the header still finds it. The sentence is now repeated in two places, which is
the cost of not filtering in the shared writer, so
`TestJsonErrorConformed_GenericDescriptionMatchesSharedWriter` runs both paths over the same error and
fails if the bodies differ.

**One existing assertion was replaced, deliberately.** `TestHandleTokenPost/ParseForm gives error`
asserted `JsonError` receives a `*errors.errorString`, which pinned the pass-through this finding says
is wrong. It now asserts the boundary hands over a `server_error` `*ErrorDetail` at 500 whose
description is already conformed.

**The family, resolved rather than the instance.** The boundary has exactly two runtime branches and
both are now conformed, so all nine error exits in `handler_token.go` are covered. One sibling the
reviewer did not name, `EncodeJson`'s fallback at four success-path exits, was dismissed with reasoning
rather than fixed: it has already written a 200 and a `Content-Type` before `Encode` can fail, so that
path cannot produce a well-formed RFC 6749 5.2 error response whatever its description says.

### Finding 2, blocking, conformance: an unsupported `response_mode` is redirected, not answered 400

**The claim.** OIDC Core 3.1.2.6 requires an unsupported Response Mode to be answered with a local HTTP
400 carrying no Error Response parameters. This server returned `invalid_request` through the client's
redirect URI instead, and on this branch a logged-out visitor was asked for a password first.

**Verified, and the reviewer is right about the text.** The section's closing sentence, fetched: "If the
Response Mode value is not supported, the Authorization Server returns an HTTP response code of 400
(Bad Request) without Error Response parameters, since understanding the Response Mode is necessary to
know how to return those parameters." The agreement quoted the paragraph directly above it and stopped
one sentence short. Also fetched: OAuth 2.0 Multiple Response Type Encoding Practices section 2.1
defines `response_mode` and states no behaviour for an unsupported value, so the OIDC sentence is the
only normative statement in play.

**Escalated as decision 11 rather than fixed**, not because the finding was doubted (it was confirmed)
but because it was not the run's to settle: the sentence carries no RFC 2119 keyword, it changes
observable wire behaviour at a public interface, nothing was being restored (the agreement decided
nothing about response modes), the redirect is inherited from `main` rather than introduced here, and a
genuine choice sat inside it that the text does not settle, namely whether the 400 applies to
`scope=openid` only or to every request. You answered **C** (every request), and it landed in
`83985610`.

1. **"Full MySQL, PostgreSQL and SQL Server integration tiers were not independently rerun."**
   **Covered.** They are the check suite's, by design, and it runs all four engines as parallel jobs on
   every push. Confirmed green on the final commit; see the Status section of the description.
2. **"No manual browser rendering pass was performed."** **Dismissed for this round**, whose change is
   a JSON body at the token endpoint: no template, no locale, no rendering. For the branch, the property
   a browser pass would check is pinned by `TestAuthorize_ValidateClientAndRedirectURI_LocalizedPage`,
   which drives the real endpoint with `ui_locales=pt-BR` and reads the rendered title and `p#errorMsg`
   back out of the response.
3. **"Untouched token signing, audience validation, session binding, refresh-token rotation, revocation
   and authorization-code one-shot algorithms were not re-audited."** **Dismissed.** Every error path
   that already runs with a session present is out of scope by design, and the diff reaches none of
   those algorithms.

### Tiers and mutations for the fix

Unit 1066 top-level / 0 fail, data 301 top-level per engine on all four / 0 fail, integration on sqlite
953 top-level / 0 fail. `production` moved from the reviewed hash by finding 1's fix and by nothing
else, checked by hash rather than assumed.

| File | Replacement | Test | Result |
|---|---|---|---|
| `handler_token.go` | the conform call removed from the final description | `TestJsonErrorConformed_GenericErrorCarriesNoForbiddenByte` | **caught** |
| `handler_token.go` | the generic sentence shortened | `TestJsonErrorConformed_GenericDescriptionMatchesSharedWriter` | **caught** |

The first proves the new test reaches the filter rather than passing on the shape of the response; the
second proves the drift pin is load-bearing rather than decorative.

---

## Final gate, round 2: the whole branch again

One finding, `significant`, conformance, `confidence: high`, `needs_human: true` with no forced answer.
All three axes `reviewed`, `unchecked` **empty**, so no work list was owed. The reviewer states plainly
that no separate security defect was found and that both round 1 repairs are sound within their
intended families.

### Finding 1: the global response-mode model mishandles both non-default implicit modes

**Not refuted. Verified true in all three of its parts**, by reading the code: `form_post` is advertised
and refused for every implicit response type; `issueImplicitTokens` has no response-mode dispatch at all
so an implicit success is always a fragment; and an explicit `response_mode=query` on an implicit request
is refused, with the refusal itself emitted in query.

**The spec text was fetched rather than recalled, and one citation came out sharper than the
reviewer's.** The reviewer cited two OIDC-facing documents, but the unit case the finding points at uses
`response_type=token`, plain RFC 6749 implicit, so the directly governing sentence is RFC 6749 4.2.2.1's
own: the server "informs the client by adding the following parameters to the fragment component of the
redirection URI". That sentence carries no RFC 2119 keyword. For the OIDC response types the prohibition
*is* a MUST NOT, from Multiple Response Type Encoding Practices sections 3 and 5: "The default Response
Mode for this Response Type is the fragment encoding and the query encoding MUST NOT be used." The same
document's "Both successful and error responses SHOULD be returned using the supplied Response Mode"
collides with that MUST NOT precisely here, and the MUST NOT governs.

### Disposition: a follow-up, not another round and not a decision

**Every part of the defect is inherited, in code this branch does not touch**, and this was checked
rather than assumed: `git diff origin/main...HEAD --stat` is **empty** for `handler_auth_issue.go` and
`handler_well_known_oidc_config.go`; the implicit rule in `ValidateRequest` appears in the diff only as
unchanged context (what the branch changed there is the *unsupported*-mode branch above it); and
`grep` for `isImplicitFlow` and `effectiveResponseMode` in the emitter's diff returns nothing.

So the remedy rewrites no work this change did, which is the test for blocking. It is also not free and
stricter: the reviewer's own recommendation is to implement implicit `form_post` success for every
advertised response type, or to have the owner explicitly choose and document a narrower support
contract. That is a change to how access tokens and ID tokens are **delivered**, which no seam in this
change covers, and the reviewer agrees, marking it `needs_human` with no forced answer and saying "there
is a real design choice". Filed as follow-up 2 in the follow-ups comment.

### What this branch did own, and corrected in `cd41ee60`

Two durable artifacts this branch wrote stated the protocol premise too loosely. The premise justified
leaving the implicit rule alone in decision 11, and **the answer to decision 11 is unaffected**:
`query` genuinely is a supported mode, so OIDC Core 3.1.2.6's unsupported-mode exception genuinely does
not reach it. What was wrong is the clause "merely may not use for tokens": the prohibition attaches to
the Response Type's encoding, not to tokens specifically, so it reaches the error response too.

- `site/src/content/docs/concepts/authorization-lifecycle.mdx` loses the false justification and states
  what happens instead, that both modes are implemented so the request is refused by the ordinary
  validation and the refusal reaches the client. User-facing and durable, which is why it was fixed
  rather than noted.
- `handler_authorize_test.go`'s new case had a comment reading that the query emission was "asking for
  something encodable, so its error is delivered as a redirect the client can read", which reads as
  endorsement. It now says it pins inherited behaviour, quotes RFC 6749 4.2.2.1 and Multiple Response
  Type Encoding Practices section 3, and says correcting it belongs to its own change.

Neither correction touches production code: `production` is byte-identical to the hash round 2 was
reviewed at, which is the evidence that the finding was dispositioned rather than quietly patched. Unit
3528 pass / 0 fail and data 1619 pass / 0 fail on all four engines were re-run and are unchanged, as
expected for a docs sentence and a comment. No mutation is owed, since neither edit adds or changes an
assertion, and the case the comment sits on was already mutation-covered when stage 5 landed it.

By the round budget, a round that moves no production code funds no successor, and nothing was left to
raise, so the review closed here.

Issue #228, opened 2026-08-16
Authorization endpoint accepts a repeated request parameter and silently uses the first occurrence


leodip:
The authorization endpoint reads every request parameter through `r.FormValue` (`HandleAuthorizeGet` in `src/authserver/internal/handlers/handler_authorize.go`), which returns the first occurrence and drops any others. A request carrying a parameter twice is therefore accepted, with one value silently chosen and no audit trace: `?client_id=a&client_id=b` is not an error today.

There are 14 such reads, covering `client_id`, `redirect_uri`, `response_type`, `response_mode`, `code_challenge`, `code_challenge_method`, `max_age`, `acr_values`, `state`, `nonce`, `scope`, `ui_locales`, `prompt` and `id_token_hint`. The validators all receive scalars from them, so nothing downstream can see that a second value ever arrived.

## Why this is wrong

RFC 6749 section 3.1, on the authorization endpoint:

> Request and response parameters MUST NOT be included more than once.

Section 4.1.2.1 lists exactly this shape under `invalid_request`, so the error to answer with is already named by the specification:

> The request is missing a required parameter, includes an invalid parameter value, includes a parameter more than once, or is otherwise malformed.

The concrete risk is parameter pollution: this server and any proxy, gateway, WAF or logging layer in front of it may read different occurrences of the same parameter. A gateway that authorizes on the last `client_id` while this server acts on the first is the general shape of the problem, and it is invisible in the audit log because only the chosen value is ever recorded.

### 1. Where the check goes, and why it has to be in two places

A refusal cannot be delivered through a redirect built out of the very parameters that are ambiguous. So the check splits at the trust boundary that already exists in `HandleAuthorizeGet`:

**Phase A, before `ValidateClientAndRedirectURI`, answered with the interstitial (`renderErrorUi`).** For `client_id`, `redirect_uri`, `response_type` and `response_mode`. The first two decide *where* an error redirect would go, and the second two decide *how* it would be delivered (`redirToClientWithError` selects query, fragment or form_post from them). If any of those is repeated, there is no single trustworthy answer to "where does this response go", so it must not go to a client at all.

**Phase B, beside the existing `ValidateUnsupportedRequestParameters` call, answered with `redirToClientWithError`.** For everything else. By that point `ValidateClientAndRedirectURI` has passed, so the redirect URI is registered and matched, and RFC 6749 4.1.2.1 says the refusal belongs at the client. Answer `invalid_request`.

[...]

Issue #155, opened 2026-08-08
Security: gorilla/csrf TrustedOrigins ignores the scheme (CVE-2025-47909), so a MitM on http://<trusted-host> passes origin validation


leodip:
## Problem

`govulncheck` reports **GO-2025-3884 / CVE-2025-47909** (`github.com/gorilla/csrf`) as *reachable* from three of our four modules. This is not a "present in the dependency graph" finding — the vulnerable symbol is on a live call path in each:

```
core:         middleware/middleware_csrf.go:225:22  MiddlewareCsrf → csrf.TrustedOrigins
authserver:   internal/server/server.go:270:47      Server.initMiddleware → MiddlewareCsrf → csrf.TrustedOrigins
adminconsole: internal/server/server.go:212:47      Server.initMiddleware → MiddlewareCsrf → csrf.TrustedOrigins
```

**There is no fixed version.** The advisory records `Fixed in: N/A` for all versions of `github.com/gorilla/csrf`, so this cannot be closed by a version bump. Dependabot has it open as alerts #47, #48 and #49 (one per `go.mod`), and they will stay open until we change libraries.

We are on `v1.7.3`, which is itself the fix for the *previous* CSRF advisory (CVE-2025-24358, alerts #30–#32, now `fixed`). This new one is a defect in that fix.

## Root cause

`TrustedOrigins` compares **only the host**, discarding the scheme. From `csrf.go` in v1.7.3:

```go
// line 288
if !sameOrigin(&requestURL, parsedOrigin) && !slices.Contains(cs.opts.TrustedOrigins, parsedOrigin.Host) {
    r = envError(r, ErrBadOrigin)
    ...
}
```

`parsedOrigin.Host` drops `http://` vs `https://`, so trusting a host implicitly trusts **both** of its origins.

What makes this worse than it first looks: about twenty lines below, the library deliberately rejects cleartext referers when serving over TLS —

```go
// disallow cleartext HTTP referers when serving via TLS
if referer.Scheme == "http" {
    r = envError(r, ErrBadReferer)
```

— but that whole branch is guarded by `if origin == "" && !isPlaintext`. A request that carries an `Origin` header never reaches it. So the scheme-blind check at line 288 silently bypasses the scheme check the library goes out of its way to perform 20 lines later, and the comment above that branch describes precisely the attack this leaves open ("prevent against CSRF via HTTP machine in the middle attacks").

## Our exposure

[...]

Issue #219, opened 2026-08-15
Rate limiting: fix the keys that void whole tiers, the budgets, and the missing endpoints


leodip:
## Summary

A full review of every rate limiter in the tree. The foundations are sound: `GetClientIPFromRequest` reads only `r.RemoteAddr` after `MiddlewareRealIP` has resolved it, so no key is spoofable, and the two-tier IP-plus-account shape on `LimitPwd` and `LimitForgotPwd` is the right pattern. What is wrong sits in the keys, the budgets, and the coverage.

Two defects void a whole tier each, measurably. Three endpoints that verify a credential or send mail have no limiter at all. The OTP budget is two orders of magnitude looser than the guidance it is trying to meet. None of this is visible today because the limiters are off unless a deployment turns them on, which is the point of the issue: when someone does turn them on, they should get what the code claims to give them.

**Scope decision, recorded so it is not re-litigated:** `GOIABADA_AUTHSERVER_RATELIMITER_ENABLED` stays defaulted to `false`. Many deployments terminate at Cloudflare, a WAF or a reverse proxy that already does this better, and the reference page already points there. This issue is only about the enabled path behaving well. It deliberately does not propose flipping the default, and it deliberately does not propose anything that only pays off at a scale an external layer should be handling.

## Current state

Nine limiter instances across seven middlewares, all in `core/middleware/middleware_ratelimiter.go`, budgets set in `NewRateLimiterMiddleware`:

| Middleware | Route | Budget | Key |
|---|---|---|---|
| `LimitPwd` | `POST /auth/pwd` | 30/min + 15/min | client IP, then raw `email` form value |
| `LimitOtp` | `POST /auth/otp` | 10/min | `user_<id>` from AuthContext |
| `LimitForgotPwd` | `POST /forgot-password` | 20/5min + 5/5min | client IP, then raw `email` form value |
| `LimitResetPwd` | `GET`+`POST /reset-password` | 30/5min | client IP |
| `LimitActivate` | `GET /account/activate` | 20/5min | client IP |
| `LimitDCR` | `POST /connect/register` | 10/min | client IP |
| `LimitROPC` | `POST /auth/token`, `grant_type=password` only | 5/min | `clientId_username_IP` |

## A. Key defects that void a tier

Both were reproduced against the real middleware with throwaway probes; the numbers below are measured, not reasoned.

### A1. Email case variation defeats both per-account tiers

[...]

PR #236, opened 2026-08-22
fix(security): read form credentials from the request body only

### LINKED ISSUE #202: Security: password and OTP form handlers accept their credential from the query string

leodip:
Closes #202.

## What this is for

Goiabada's login form, 2FA prompt, password reset, self-registration and every admin console form
read their credential with `r.FormValue`. Go's `ParseForm` merges the URL query into `r.Form` behind
the body, so `POST /auth/pwd?password=hunter2` with an empty body authenticates. Verified against
`net/http` on go1.26.6: `r.FormValue("password")` returns the query value there, and
`r.PostFormValue` returns `""`.

That is not a bypass of anything. The credential still has to be correct. The harm is where the
credential ends up: a URL reaches the browser address bar, browser history, the `Referer` of
anything the page loads, and the access log of every proxy, gateway or CDN in front of the
deployment. Goiabada's own request log is already clean, since #159 made unrecognised parameter
values `[redacted]`, so what remains is entirely outside it.

One case the issue did not name inverts the precedence and is the reason to fix this by
construction rather than site by site: with a `multipart/form-data` body, `ParseMultipartForm`
appends the body's values *after* the query, so `r.FormValue` returns the **query** value even when
the body carries the field. No credential-bearing form is multipart today, so it is latent rather
than live, but no per-site reasoning about which source wins survives a template gaining an
`enctype`.

## The rule this establishes

**A credential submitted to a Goiabada form is read from the request body and nowhere else.** For
each name in the agreed set, a POST whose credential is only in the query behaves as though the
field were absent.

24 read sites in 12 files become `r.PostFormValue`. Nothing else about any handler changes: no new
branch, no new string, no new audit event, no signature change.

**Why that is safe rather than a behaviour change.** Every one of the 24 reads is in a `...Post`
handler whose route is registered POST-only, with its GET counterpart a separate handler that
renders the form. Go populates `r.PostForm` from the body alone and `r.Form` from the body followed
by the query, so for every request a legitimate caller makes the two accessors return the same
value. They diverge only when the credential arrived somewhere no legitimate caller puts it. The
only requests whose outcome changes are ones that were already wrong, which is why no existing test
moves.

Two things that could have broken that property, both checked. No middleware on any route in scope
consumes the body. And a middleware that parses first does not break the handler: `LimitPwd` calls
`r.FormValue("email")` before the handler runs, `ParseForm` populates `r.PostForm` from the body at
the same time, and Go caches it, so the handler's `r.PostFormValue` reads that cache.

**Deliberately not changed.** `/auth/authorize` keeps its 19 query-capable reads: OIDC Core 3.1.2.1
requires the authorization endpoint to accept GET and POST, and the documented endpoint contract
says both bindings take the same names. `/auth/token` already reads `r.PostForm.Get` everywhere.
The `email` read at `/auth/pwd` is not a credential and moving it without the rate limiter's
matching `accountRateLimitKey` read would buy an attacker a fresh bucket. Repeated parameters across
the query and body are #228, for two endpoints this does not touch.

## How it is guarded

The admin console has no handler tests, so 16 of the 24 reads have no unit tier available. Two
instruments, and they cover different amounts on purpose:

- **Behavioural cases** in the four authserver handler test files that exist, each building a POST
  with the credential in the query alone and asserting the handler answers as though the field were
  absent, with the mock set proving no credential check ran.
- **One source-text lint per module**, in the shape of the existing `csrf_lint_test.go` that #155
  produced for exactly this gap. Each forbids the literal `.FormValue("<name>")` for the names that
  have no legitimate query source *in that module*, which is why the two lists differ: the admin
  console's can forbid `code` and `state` and the authserver's cannot.

The lints' honest limit, stated because it decides what they are worth: they assert a spelling is
absent, not that its replacement is right. In the authserver the behavioural cases cover the
replacement. In the admin console nothing does.

### Stage 1: the authserver's forms read their credentials from the body
Landed in `c0cabee5`.

Eight reads in four files became `r.PostFormValue`: `password` at `/auth/pwd`, `otp` at `/auth/otp`,
`password` and `passwordConfirmation` at `/reset-password` and `/account/register`, and both
`continuationId` reads in the reset handler, the gate and the re-render echo. This extends a pattern
already in the tree rather than introducing one: `/auth/pwd`, `/auth/otp` and `/auth/consent` have
read their ceremony id body-only since #79, for the same reason and with the reason in the comment.
After this, the two reads in `HandleAuthPwdPost` agree about what a submission is, where before the
ceremony id was body-only and the password beside it was not.

New `src/authserver/internal/handlers/credential_read_lint_test.go`,
`TestHandlers_NoCredentialQueryFallback`. It walks every non-test Go source under `internal/` and
fails on `.FormValue(` plus any of the ten credential names or the two marker constants. It
deliberately omits `state` and `code`, because `handler_authorize.go` reads `.FormValue("state")`
legitimately. Its floor asserts a nonzero non-test file count under `internal/handlers` and
`internal/middleware`, so a walk that silently covers nothing cannot pass.

Four behavioural cases, one per existing handler test file. The OTP case uses a **live** passcode,
correct for the user's own secret, because the failure there is worse than a leak: with `r.FormValue`
the code verifies and `TryConsumeUserOTPStep` burns the step, so the passcode the user is about to
type into the real form is then refused as a replay. The reset handler took two cases, the second
submitting the marker's own live continuation id, read back through `GetLinkMarker` so the test
cannot invent one the handler would never have rendered.

`CLAUDE.md` and `AGENTS.md` gained item 5 under `## Important Patterns`, identical in both, naming
the field set, the mechanism, the reason, the lint and `handler_authorize.go`'s exemption.

Tiers: unit 1073 pass / 0 fail across 47 packages; integration on sqlite 1266 pass / 0 fail;
`go vet` clean in all three modules. Data tier not applicable, nothing here reaches the data layer.
Seven mutations, one per seam plus one on the lint's own floor, all `MUTATION CAUGHT`.

One deviation worth knowing: the reset handler carries its explanatory comment on the `password`
read rather than on the lexically-first `continuationId` echo, which sits inside a closure that
already has a comment explaining why it echoes at all. The comment says explicitly that it covers
both `continuationId` reads, so the closure is accounted for without repeating the mechanism.

### Stage 2: the admin console's forms read their credentials from the body

Landed in `c2388417`. Sixteen reads across eight handlers, `r.FormValue` to `r.PostFormValue`: the
three password fields on the account change-password form; the four on the account OTP form, being
the password plus the TOTP shared secret, the QR image encoding it and the enrolment code; the email
verification code that is the whole gate on an email change; the SMTP password; the client secret;
an admin-set password on another user's account; and a new user's password. Plus the four
`form_post` arrivals at `/auth/callback`, `code`, `state`, `error` and `error_description`.

The callback is the sharpest of them. `auth_helper.go` asks the authorization endpoint for
`response_mode=form_post` precisely so the code arrives in a request body, and the handler then read
it with an accessor that also accepted it from the query, undoing the exact leak that choice was made
to prevent, on the one path in scope that is CSRF-exempt.

Scope was recounted rather than trusted: `probe/inventory.sh` reported 12 credential reads in 7 files
plus the callback's 4, matching the agreed fanout with no drift, and afterwards reports zero. The
safety property was re-checked too, all eight routes are registered `r.Post` alone with a separate
GET handler rendering the form, so the two accessors return the same value for every request a
legitimate caller makes.

`credential_read_lint_test.go` is this module's half of the guard, forbidding `.FormValue("<name>")`
for the ten credential names plus `code`, `state`, `error` and `error_description`. Those four are
safe to forbid here and deliberately absent from the authserver's list, which needs them at
`/auth/authorize` because OIDC Core 3.1.2.1 requires that endpoint to accept `GET` as well as `POST`.
This module has no handler test harness, so the lint is the only instrument over these sixteen reads,
and it is negative-only: it asserts a spelling is absent, not that the replacement is right. That
limit is agreed rather than discovered, and building the module's first handler harness is drafted as
a follow-up.

`site/src/content/docs/reference/security.mdx` gained a bullet under `### Authentication security`
saying a credential placed in a URL is answered exactly as though the field had been left blank. Its
first draft also claimed nothing is saved in that case, which is false for the SMTP password and the
client secret, both of which feed an overwriting request so that a blank submission blanks the stored
value. That is pre-existing and explicitly out of scope, unchanged here, and the bullet now claims
only what holds for all eight forms.

Tiers: unit 3535 pass / 0 fail across all three modules. Data and integration not applicable, nothing
reaches the data layer, no route moves, and this module has no integration coverage at all. Four
mutations, all `MUTATION CAUGHT`: one credential name, one callback name, the SMTP password that sits
among eight non-credential reads of the same accessor, and one pointing the lint's own walk at a tree
holding no handlers to prove its file-count floor fires.

Two deviations worth knowing. The client secret's comment sits above its struct literal rather than
inside it, because inside it gofmt regrouped three neighbouring fields and put pure whitespace churn
in the diff. And this commit corrects a sentence in the authserver lint that stage 1 added: its
per-tree comment said `internal/middleware` holds the rate limiter, and the rate limiter is in
`src/core/middleware`, outside the walk of either module's lint.

### After the stages: `a265cdc6`, the review's one fix

The whole-branch review found two things, and this commit answers both. 3 files, +139/-15, no
production code.

Three behavioural cases were added, because three of the eight authserver conversions had no case
that could tell a body-only read from a merged one: the two existing query-only password cases put
*both* credential fields in the query, so the handler returns at `len(password) == 0` before the
confirmation is read at all. The three now pinned are `passwordConfirmation` in registration,
`passwordConfirmation` in reset, and the `continuationId` the reset form echoes back when it
re-renders, that last one asserting the re-rendered hidden value is **empty**. All eight authserver
conversions now carry a mutation that fails without them.

And the `security.mdx` bullet stopped claiming that a query-only credential is refused "without any
account being looked up". That is false on two of the five paths the sentence named:
`HandleAuthOtpPost` calls `GetUserById` before it ever reads `otp`, and `HandleAccountRegisterPost`
calls `GetUserByEmail` and `GetPreRegistrationByEmail` first. The bullet now says the value in the
URL is never compared against anything and that each handler takes the path it already takes for
that field being blank, which in some of these forms looks the account up first. The guarantee the
page exists to state is unchanged and is the one the code actually makes.

## Review

Reviewed by `gpt-5.6-sol` at effort `xhigh`. Three rounds: one over the plan before any code existed,
and two over the whole branch against its base. **No stage was reviewed on its own** — the code was
read once in full, at the end, with everything built on it in the same diff.

|---|---|---|---|---|
| plan | 1 | 0 | 0 | clean |
| final | 1 | 2 (1 significant, 1 minor) | 0 | both fixed, `a265cdc6` |
| final | 2 | 0 | 0 | clean, gate closed |

- **significant, quality:** three of the eight converted authserver reads had no case that could
  distinguish a body-only read from a merged one, proved by three surviving mutations. Fixed in
  `a265cdc6`: three cases added, and the family closed by recounting every converted read in the
  module rather than fixing the three handed over. All three mutations now `CAUGHT`.
- **minor, conformance:** the new `security.mdx` bullet claimed no account is looked up, which is
  false for the OTP and registration paths. Fixed in `a265cdc6`: the claim now says the query value
  is never compared against anything, which is what the code guarantees.

Round 2 was pointed at `a265cdc6` itself, since round 1's fix had not been read by any reviewer, and
came back clean with zero `unchecked` entries. It re-ran round 1's three mutations plus two more
against the new test machinery, independently recounted the production diff to exactly 24
substitutions, and confirmed the agreed exclusions are intact.

Full account, with the evidence, in [this comment](https://github.com/leodip/goiabada/pull/236#issuecomment-5381104460).

## Tiers

**Local, per stage:** unit green in all three modules, and integration on sqlite for stage 1 (1266
pass / 0 fail). `gofmt` and `go vet` clean. Neither stage touched the data layer, a migration, a
schema, an interface method or a query, so the data tier is not applicable; stage 2's module has no
integration coverage at all.

**Eleven mutations across the stages, plus three at the review, every one `MUTATION CAUGHT`.** One
per seam, one per case pinning a decision, and two aimed at the lints' own file-count floor so a walk
that silently covers nothing cannot pass. The review's three use `r.Form.Get` rather than reverting
to `r.FormValue`, because that is the spelling the lint cannot see, so each tests the behavioural
case alone.

**Four-engine data and integration is the check suite's**, and it has concluded green on the exact
commit this merges: `a265cdc6`,
[run 32580206086](https://github.com/leodip/goiabada/actions/runs/32580206086), eleven jobs all
passing — `Tests / sqlite`, `Tests / mysql`, `Tests / postgres`, `Tests / mssql` each running the
data tier then the integration tier, plus `Unit / core`, `Unit / authserver`, `Unit / adminconsole`,
`Lint`, `Vulnerabilities`, `Versions` and `Build validation`. Nothing local ran the other three
engines, and no claim here extrapolates from the sqlite run.

## Deviations

Each stage's are recorded under its own heading above: where an explanatory comment sits relative to
the read it covers (twice, once to keep gofmt from churning three neighbouring struct fields), and a
wrong sentence in stage 1's lint comment that stage 2 corrected. None of them changes behaviour.

The one substantive correction during the run was the `security.mdx` bullet, twice: once inside stage
2, when its first draft claimed nothing is saved on a query-only submission, which is false for the
SMTP password and client secret because both feed an overwriting request; and once at the review, for
the account-lookup claim above. Both times the fix was to narrow the claim to what holds for all
eight forms.

## Also on this pull request

- [Deferred decisions](https://github.com/leodip/goiabada/pull/236#issuecomment-5381098808) — none,
  and the six decisions that were settled instead.
- [Follow-ups](https://github.com/leodip/goiabada/pull/236#issuecomment-5381100107) — one, ready to
  file: the admin console has no handler test harness, which is why 16 of these 24 reads are guarded
  by a lint rather than by behaviour.
- [The review account](https://github.com/leodip/goiabada/pull/236#issuecomment-5381104460) — every
  round, every finding, with the evidence.

## Status

**Complete.** Both stages landed, the whole-branch review closed clean at round 2, and the check
suite is green on `a265cdc6`, the exact commit this merges.

The credential-bearing inventory this started from was 18 reads across 11 files; it is now zero. All
five form-binding markers are body-only, where three of five were before. The four `form_post`
arrivals at the admin console callback are body-only. `/auth/authorize`'s 19 query-capable reads and
`/auth/token`'s 10 `PostForm.Get` reads are untouched, as agreed, and so are the `email` reads that
the rate limiter has to keep agreeing with.

The accepted limit: the two lints assert a spelling is absent, not that its replacement is right. In
the auth server the eight conversions are each pinned by a behavioural case with a mutation behind
it. In the admin console the sixteen are not, because that module has no handler test harness. That
is the follow-up.


leodip:
## Deferred decisions

**None.** Nothing in this change was built on an assumption the reader has to accept or reverse.

All six decisions were settled before any code was written, and the run raised none of its own:

| # | Question | Answer |
|---|---|---|
| 1 | Replace the query fallback with a body-only read, or an explicit refusal? | **Body-only read.** Each site reads its name with `r.PostFormValue`, so a query-supplied credential is not seen and the handler's existing "required" path renders. |
| 2 | Which names are in the set: the four the issue grepped for, or all ten credential-bearing names? | **All ten**, so all 18 credential reads in 11 files. |
| 3 | Does `continuationId` move too? | **Yes, both reads** in `HandleResetPasswordPost`, making all five form-binding markers in the tree body-only. |
| 4 | Does the admin console callback's `code` move too? | **Yes, all four reads** (`code`, `state`, `error`, `error_description`). |
| 5 | What guards the change, given the admin console has no handler tier? | **One source-text lint per module**, plus behavioural cases in the four existing authserver handler test files. |
| 6 | Does a refused credential-in-query request get an audit event? | **No.** Moot under decision 1: a body-only read refuses nothing, so there is no event to log. A query-only submission is audited exactly as any other failed submission. |

Two rejected options are worth naming, because each is the thing a reader might expect to find and
will not:

- **An explicit refusal** (per handler, or one middleware) was rejected. It would additionally catch
  the one shape a body-only read accepts: the credential in the query *and* in the body, where the
  body wins and the request succeeds with the credential still in the URL. That single extra shape
  costs a localised string in both catalogs, an audit decision, and a name list that becomes a second
  source of truth beside the reads, where a name missing from the list reads as assessed and safe
  while the read behind it stays vulnerable. The realistic mistaken integration is the query-only one,
  a copied `curl` with no body, and a body-only read stops that outright.
- **Building the admin console's first handler test harness** was rejected as larger than the fix
  itself, and drafted as a follow-up instead. That is the one accepted limit in the change: 16 of the
  24 reads are guarded by a lint that asserts a spelling is absent, not that its replacement is
  right. See the follow-up comment.

leodip:
## Follow-ups

One, ready to file. The duplicate search was re-run just now against every state:
`gh issue list --state all --search "adminconsole handler tests"`,
`--search "test harness admin console"` and `--search "handler test harness"` return nothing that
covers it.

### 1. The admin console has no handler test harness, so its handlers cannot be tested

**Labels:** `enhancement`, `go`

**Title:** `The admin console has no handler test harness, so its handlers cannot be tested`

**Body:**

> The admin console's handlers have no unit tests. Seven of the module's test files are
> `test_main_test.go` stubs and the only real handler test is
> `handler_admin_client_redirect_uris_test.go`, so nothing can assert what an admin console handler
> actually does.
>
> Adding a harness needs mocks for `apiclient.ApiClient`, a `sessions.Store` fixture, and injection
> of `ContextKeyJwtInfo` and `ContextKeySettings`. That is the wiring the auth server's handler tests
> get from `mocks_handlerhelpers`, which is why that module has a full behavioural tier and this one
> has none.
>
> Two security changes have now worked around its absence by writing a source-text lint test instead:
>
> - **#155** added `csrf_lint_test.go` after finding that reintroducing a deleted CSRF field bind left
>   the whole module green.
> - **#202** added `credential_read_lint_test.go`, and it is the *only* instrument over the 16 read
>   sites that change moved from `r.FormValue` to `r.PostFormValue` in this module. A lexical lint
>   asserts a spelling is absent; it cannot assert the replacement behaves correctly. In the auth
>   server the same change is pinned by behavioural cases; here nothing pins it.
>
> A third change should build the harness rather than add a third lint. The value outlives any one
> change: it would let the admin console's forms be tested the way the auth server's are.

**File it with:**

```bash
gh issue create \
  --title "The admin console has no handler test harness, so its handlers cannot be tested" \
  --body 'The admin console'"'"'s handlers have no unit tests. Seven of the module'"'"'s test files are `test_main_test.go` stubs and the only real handler test is `handler_admin_client_redirect_uris_test.go`, so nothing can assert what an admin console handler actually does.

Adding a harness needs mocks for `apiclient.ApiClient`, a `sessions.Store` fixture, and injection of `ContextKeyJwtInfo` and `ContextKeySettings`. That is the wiring the auth server'"'"'s handler tests get from `mocks_handlerhelpers`, which is why that module has a full behavioural tier and this one has none.

Two security changes have now worked around its absence by writing a source-text lint test instead:

- #155 added `csrf_lint_test.go` after finding that reintroducing a deleted CSRF field bind left the whole module green.
- #202 added `credential_read_lint_test.go`, and it is the only instrument over the 16 read sites that change moved from `r.FormValue` to `r.PostFormValue` in this module. A lexical lint asserts a spelling is absent; it cannot assert the replacement behaves correctly. In the auth server the same change is pinned by behavioural cases; here nothing pins it.

A third change should build the harness rather than add a third lint. The value outlives any one change: it would let the admin console'"'"'s forms be tested the way the auth server'"'"'s are.'
```

Nothing else was found worth filing. In particular, two neighbouring defects were noticed, checked,
and deliberately not drafted because they are already tracked or already agreed as out of scope:

- **`r.ParseForm`'s discarded error**, where a malformed percent-escape in the query silently drops
  the affected field. Already named in **#228** section 5. Body-only reads are unaffected by it,
  verified by probe.
- **The SMTP password and client secret not surviving an empty submission.** Both feed an overwriting
  API request, so a blank submission blanks the stored value. Pre-existing and unchanged here; the
  `security.mdx` bullet this change adds was reworded during the run so it does not claim otherwise.

leodip:
## The review account

`LEO_REVIEW_MODE=end`, so **no stage was reviewed on its own**. The code was read once in full, at
the end, with everything built on it in the same diff. Three rounds ran in total: one over the plan
before any code existed, and two over the whole branch.

Reviewer throughout: `gpt-5.6-sol` at effort `xhigh`, fresh or resumed session, scope stated per
round, all three axes (conformance, quality, security) in scope every time.

|---|---|---|---|---|
| plan | 1 | 0 | 0 | clean |
| final | 1 | 2 (1 significant, 1 minor) | 0 | both fixed, `a265cdc6` |
| final | 2 | 0 | 0 | clean, gate closed |

---

### Plan gate, round 1: clean

Sent 2026-08-22T13:51:44Z, answered 14:01:06Z. Scope: `plan.md`, both stages, all three axes.
**Zero findings.** What the round is worth is the independent execution behind it, so that is what is
recorded:

- It **recounted the inventory rather than trusting it**: re-ran the run's own inventory probe and
  separately enumerated every literal `FormValue` and `PostFormValue` name under both modules,
  reproducing the sealed boundary of 18 credential reads, 2 remaining form-marker reads and 4
  `form_post` callback reads, with no omitted credential-bearing production read. That count is what
  the whole change is scoped by.
- It **re-derived the mechanism with a real `net/http` probe on Go 1.26.6**: `PostFormValue` returns
  empty for a query-only value, returns the body value for URL-encoded and multipart requests, and
  stays body-only *even after `FormValue` has already triggered form parsing*. That last clause is
  the operationally important one, because `LimitPwd` calls `r.FormValue("email")` before the handler
  runs, and it confirms that earlier parse does not reintroduce the query credential.
- It **executed the lint pattern**: `.FormValue("password")` matched `r`, `req` and `request`
  receivers and did not match `PostFormValue`.
- It **traced all five seams** through current handler ordering and mocks, confirming each planned
  case distinguishes the two implementations, and naming that the OTP case stops before `MatchStep`
  and `TryConsumeUserOTPStep`, and that the reset continuation case resolves the live marker and then
  fails the body-only match before `BeginTransaction`.
- It **checked RFC 6749, RFC 6750, RFC 9700 and OIDC Core against publisher copies** and endorsed the
  framing: none of them governs Goiabada's own password and OTP forms, so the body-only rule is
  correctly presented as this server's own decision rather than as conformance. That was the claim
  most at risk of being overstated, and it held.

Its two `unchecked` entries were answered: the planned edits not existing yet (the mutations are each
stage's own obligation, recorded per stage with their `CAUGHT`/`SURVIVED` results), and no test tier
having been re-run (correct: the tree under review was `origin/main` plus the empty commit that
opened this pull request, and the request claimed nothing green).

---

### Final gate, round 1: two findings, both fixed in `a265cdc6`

Started 2026-08-22T14:38:20Z, verdict 14:46:21Z. Scope: the whole branch against its base, three
commits, 21 files, +664/-27.

#### Finding 2, significant, quality: three converted reads were not behaviourally pinned

**The claim.** Of the eight authserver reads this change converted, three had no case that could tell
a body-only read from a merged one: `passwordConfirmation` in registration, `passwordConfirmation` in
reset, and the `continuationId` the reset form echoes back when it re-renders. The two existing
query-only password cases put *both* credential fields in the query, so the handler returns at the
`len(password) == 0` check before the confirmation is read at all, and the continuation cases
exercise the marker gate rather than the earlier echo in `renderError`.

**Verified, and the evidence reproduced.** The reviewer ran three mutations that survived. The
mechanism is exactly as described: with the password in the query the handler cannot reach the second
read, so no assertion downstream of it can depend on how that read was made. The lexical lint catches
only the literal `.FormValue(<name>)` spelling and does not claim to prove runtime behaviour, which
the reviewer confirmed with a fourth mutation that the lint *did* catch.

**Not escalated.** No design choice existed. Decision 5 had settled that the authserver seams get
behavioural cases in the four existing test files, and three of them were short of that, so this is
coverage the change itself owes.

**Fixed, with three cases added, and the family closed rather than the three instances.** The
family is "multi-read handler tests short-circuit before exercising later converted reads", so the
boundary is every read this change converted. Recounted from the diff: the authserver has exactly
eight converted production reads (`password` at `/auth/pwd`, `otp` at `/auth/otp`, `password` and
`passwordConfirmation` in registration, and the echo, `password`, `passwordConfirmation` and the gate
in reset). Five were already pinned by the stages' mutations, the three above were not, and all eight
now are. No ninth converted authserver read exists.

- `TestHandleAccountRegisterPost/Password_confirmation_in_the_query_alone` — body carries the email
  and `password`, the query alone carries a *matching* `passwordConfirmation`. Body-only, the
  confirmation is absent and the answer is `Password confirmation is required.`; merged, the two would
  agree and the account would be created. `passwordValidator` and `userCreator` are given no
  expectations, so reaching either fails the test on an unexpected call.
- `TestHandleResetPasswordPost_PasswordFieldRejections/the_confirmation_is_in_the_query_alone` — same
  split, expecting the existing mismatch answer with the validator unexpected.
- `TestHandleResetPasswordPost_PasswordFieldRejections/the_continuation_id_is_in_the_query_alone` — a
  submission with no password and the continuation id in the query alone, asserting the re-rendered
  form echoes an **empty** hidden value.

The reset table gained a `build` field (nil meaning the ordinary body-only request) and a
`wantEchoedContinuationId` field, replacing a single `inQuery` bool, because the last case needs a
different expected echo. Two builders were added beside the existing ones.

**Mutations, all three now caught:**

| File | Replacement | Ran against | Result |
|---|---|---|---|
| `accounthandlers/handler_account_register.go` | `passwordConfirmation := r.PostFormValue(...)` → `r.Form.Get(...)` | `TestHandleAccountRegisterPost` | **CAUGHT** |
| `handler_reset_password.go` | `passwordConfirmation := r.PostFormValue(...)` → `r.Form.Get(...)` | `TestHandleResetPasswordPost_PasswordFieldRejections` | **CAUGHT** |
| `handler_reset_password.go` | `"continuationId": r.PostFormValue(continuationIdField),` → `r.Form.Get(...)` | `TestHandleResetPasswordPost_PasswordFieldRejections` | **CAUGHT** |

`r.Form.Get` is the honest mutation rather than reverting to `r.FormValue`: it is the spelling the
lint cannot see, so it tests the behavioural case and nothing else.

#### Finding 1, minor, conformance: the security page claimed no account is looked up

**The claim.** The new bullet in `site/src/content/docs/reference/security.mdx` said a query-only
credential is refused "without any account being looked up or any credential being checked". The
first half is false on two of the five paths the sentence names.

**Verified against the code, and it stands.** `HandleAuthOtpPost` calls `database.GetUserById` before
it ever reads `otp`. `HandleAccountRegisterPost` calls `GetUserByEmail` and then
`GetPreRegistrationByEmail`, and only then reaches `if len(password) == 0`. The run's own new tests
corroborate it: both arrange those lookups in order to reach the assertion.

**Not escalated.** Nothing was being chosen: the sentence was simply untrue, and the correction only
narrows a claim. No interface, behaviour or security property moves, and making the docs promise less
than they did cannot be the wrong answer to a promise the code does not keep.

**Fixed.** The clause now says the value in the URL "is never compared against anything" and that
each handler "takes the path it already takes for that field being blank, which in some of these
forms looks the account up first, exactly as it does when the field is genuinely left empty." The
guarantee the page exists to state is untouched and is the one the code actually makes: a query value
is never checked, never authenticates, never authorizes, and is never stored. The equivalent bullets
in `CLAUDE.md` and `AGENTS.md` were checked and make no account-lookup claim, so they needed no
change.

#### Round 1's three `unchecked` entries, all answered

- *The sixteen admin console conversions were not executed through a handler harness.* **Filed**, and
  already was: that module has no handler test harness at all, which is the follow-up in the previous
  comment. The reviewer reached the same conclusion and explicitly declined to reopen it. They were
  covered here by route tracing, source inspection, the module's lint, the recorded mutations and
  green exact-head CI.
- *No interactive browser pass or new local integration run.* **Dismissed.** This round changed test
  and documentation files only, which the split tree hashes prove: `production` unchanged at
  `3e58f5758515`. No handler, route, template or schema moved.
- *Token validation, PKCE, redirect matching, session lifecycle, revocation and SQL paths were not
  dynamically re-probed.* **Dismissed.** No changed code enters those surfaces; the reviewer checked
  diff reachability and this round did not widen the diff into any of them.

#### Tiers after the fix

|---|---|---|---|
| unit | `where.sh test --type modules` | n/a | **3728 run, 3728 pass, 0 fail, 0 skip** |
| anchors | `check-anchors.sh` | n/a | pass, all 19 rows |
| fmt, vet | `gofmt -l`, `go vet ./internal/handlers/...` | n/a | clean |
| integration | not re-run | n/a | not applicable: `production` did not move |
| data | not run locally | n/a | no migration, schema, interface method or query |

Counted with `grep -cE "^ *--- PASS"`, which equals the `=== RUN` count in the same log, with zero
`--- SKIP` and zero `FAIL`. That figure is not directly comparable with the `3535 pass` recorded at
stage 2: the difference is far larger than the three subtests added here, so the two were counted by
different methods rather than the suite having grown. The self-consistent run above is the one to
trust, and the check suite is the independent arbiter either way.

---

### Final gate, round 2: clean, gate closed

Started 2026-08-22T14:59:15Z, verdict 15:06:45Z. Scope: `git diff origin/main...HEAD`, all three
axes. **Zero findings, zero `unchecked` entries.** The reviewer's own words on the residue: "Nothing
material remains unchecked for this round."

Round 2 existed because round 1's fix, `a265cdc6`, had not itself been read by any reviewer: three
new behavioural cases, two new request builders, a restructured table, and a reworded docs bullet.

- **Conformance.** Independently recounted the production diff to exactly **24 substitutions**, 8 in
  the authserver and 16 in the admin console, each now `PostFormValue`, each route still POST-only
  with a separate GET renderer. Confirmed the agreed exclusions are intact: `handler_authorize.go`'s
  `state` reads still merged per OIDC Core 3.1.2.1, `/auth/token` still `PostForm.Get`, and the
  activation and reset-link `code` reads still GET query inputs outside the form-submission boundary.
  Confirmed the repaired `security.mdx` bullet is true of the code.
- **Quality.** Re-ran round 1's three mutations with `r.Form.Get`; all three caught, each by exactly
  its new case. It then ran **two further mutations against the test machinery itself**, which no
  earlier round had done: hard-coding the reset echo expectation (failed only the new
  continuation-query case) and replacing the per-case builder with the ordinary one (failed all three
  query-shaped reset cases). That is the evidence the refactored table carries live configuration
  rather than inert fields, which is the thing a table refactor most plausibly gets wrong.
- **Security.** Every changed read narrows its trust boundary from body-plus-query to body-only. For
  an unauthenticated password, OTP, registration or reset submission, a query-only secret now reaches
  the same blank-field path as an omitted one. For the authenticated admin forms the query value is
  not forwarded to the API. At the CSRF-exempt admin callback all four `form_post` fields are read
  from the body, with the session-state comparison and PKCE exchange unchanged. No fail-open storage
  predicate, one-shot claim, revocation path, loose match, scope or audience widening, secret
  comparison, SQL, template trust change, route or authorization boundary was added. Rate-limit
  identity stays aligned because the non-credential `email` reads and their limiter keys did not move,
  which is the property #219 established and section 1 said must hold.

**Verified against the current tree before closing, rather than taken on trust:**

- The inventory probe reports **category A at 0 reads in 0 files**, down from the sealed 18. All five
  form-binding markers are `PostFormValue` (three `ceremonyId`, two `continuationId`). The callback
  category is empty. The deliberately-unchanged category is unmoved: 19 `r.FormValue` in
  `handler_authorize.go`, 10 `r.PostForm.Get` in `handler_token.go`, and the limiter's two
  `accountRateLimitKey(r.FormValue("email"))` reads still exactly as the handlers read `email`.
- `git diff --stat origin/main...HEAD`: 21 files, +793/-32. No limiter, authorize or token file in the
  set.
- `check-anchors.sh`: **PASS, all 19 rows** resolve to exactly one line in the named function.
- Split tree hashes identical to the values the round was requested at, so every mutation this round
  ran was restored. Working tree clean, `HEAD` = `origin/issue-202-credentials-from-query`.

**No new follow-up.** The reviewer explicitly declined to reopen the admin console handler harness,
which is already drafted.

---

### The check suite, on the exact merge commit

`a265cdc6db74e7131d32af2b3bcadf581350f356`, [run 32580206086](https://github.com/leodip/goiabada/actions/runs/32580206086):
**eleven jobs, all green.** `Tests / sqlite`, `Tests / mysql`, `Tests / postgres`, `Tests / mssql`
(data tier then integration tier on each), `Unit / core`, `Unit / authserver`, `Unit / adminconsole`,
`Lint`, `Vulnerabilities`, `Versions`, `Build validation`.

That is where the four-engine data and integration claim for this change comes from, on the exact
commit this pull request will merge, not extrapolated from the local sqlite run. The earlier green
suite at `c2388417` ([run 32578917579](https://github.com/leodip/goiabada/actions/runs/32578917579))
covered the two stage commits; this one covers round 1's fix on top of them.

PR #238, opened 2026-08-22
fix(auth): amr names only what the bound session verified

### LINKED ISSUE #140: Security: a session-less completion inherits the previous session's amr, claiming otp that did not happen

leodip:
Closes #140.

## What this fixes

`amr` is supposed to name the authentication methods that were actually verified for the session a
grant is bound to. Two paths break that, in opposite directions, and both are fixed here.

**Over-claiming (#140 itself).** `/auth/authorize` copies a valid session's `AuthMethods` onto the
auth context for SSO. If that session then stops being usable mid-ceremony (its row is deleted, its
idle timeout or maximum lifetime elapses, or a requested `max_age` elapses), the ceremony is sent
back to `/auth/level1` and the user re-enters a password. `AuthContext.AddAuthMethod` appends and
is a no-op on a duplicate, so `pwd otp` survives that restart untouched: the freshly created session
row and the authorization code are both stamped `pwd otp` for a ceremony that only ever verified a
password. The grant comes out self-contradictory, `acr` at `urn:goiabada:level1` beside
`amr` `["pwd","otp"]`, and a relying party gating on `amr.includes('otp')` grants access on the
strength of a second factor nobody presented.

Two restart routes reach it, `/auth/completed`'s level 1 gate and `/auth/issue`'s liveness refusal,
and the second has the wider window because a person can sit on a consent screen. The same restart
carries `AuthContext.ConsentedScope`, which nothing revisits and which `CreateAuthCode` prefers over
the freshly filtered scope, so a permission revoked while the user is back at the password form
still reaches the code.

**Under-claiming.** `BumpUserSession` replaces a session's `AuthMethods` whenever they differ, in
either direction, while refusing to lower its `AcrLevel`. So a `prompt=login` ceremony at level 1
over a live level 2 session emits `acr=level2_mandatory` with `amr=["pwd"]` and erases the row's
record of `otp`. The row is what every later grant from that session reads, so this one is wider:
it fails closed, but it keeps failing.

## How it works

Two mechanisms, one per direction.

A ceremony arriving at `GET /auth/level1` is about to perform level 1 from scratch, so it discards
`AuthMethods` and `ConsentedScope` there. That single site is enough because it is a choke point:
`routes.go` registers `/auth/level1` once and `Get` only, and `/auth/pwd` gates on
`AuthStateLevel1Password`, which only this handler sets, so no restart can route around it. Nothing
genuinely earned is lost, because `otp` is only ever verified through `/auth/level2`, which is
reached whenever the target ACR is above level 1, and the target is recomputed identically on every
hop.

`BumpUserSession` merges rather than replaces, mirroring the never-downgrade rule its `AcrLevel`
already has, and `/auth/completed` reads the merged value back off the returned session so the row
and the code cannot disagree.

### Stage 1: a restart at level 1 discards what the ceremony accumulated
_Not started._

### Stage 2: a session's amr only ever grows
_Not started._

## Status

Run in progress. The plan is written and no stage has landed yet.


leodip:
## Decision 9: stage 2's never-lower rule for `amr` is under-specified in two independent ways. How should the run proceed?

The plan review (`gpt-5.6-sol`, effort `xhigh`, one round) executed stage 2's design in a scratch
tree and found two blocking defects in it. I verified both against the code on this branch. Neither
touches stage 1, which the same review found sound, so **whichever option you pick, stage 1 proceeds
exactly as planned** — the two stages share one file in different functions.

No code has landed. The branch is still the single empty commit that opened this PR.

### (a) The merge makes an old OTP look as recent as a fresh password

`handler_auth_completed.go:HandleAuthCompletedGet` refreshes the session's `AuthTime` whenever the
ceremony entered a credential. And `handler_auth_level1.go:HandleAuthLevel1CompletedGet` decides
step-up on `targetAcrLevel.IsHigherThan(acrLevelFromSession)`, which is **false** when the session
already sits at or above the target — so a `prompt=login` request with a `level2_mandatory` target,
over a `level2_mandatory` session, skips OTP entirely and completes on the password alone.

| | `acr` | `amr` | `auth_time` | An RP gating on `amr.includes('otp')` |
|---|---|---|---|---|
| today | `level2_mandatory` | `["pwd"]` | now | refuses — **fails closed** |
| stage 2 as planned | `level2_mandatory` | `["pwd","otp"]` | now | accepts — **fails open** |

Today's grant is self-contradictory and an RP can see that. Stage 2's is coherent and false: the OTP
happened up to a full session lifetime earlier. So stage 2 as designed removes the signal the RP
currently fails closed on.

Two shipped pages make this reachable rather than theoretical. `concepts/prompt-parameter.mdx`:
"After re-authentication, the `auth_time` claim in the ID token will reflect the new authentication
time, which will be later than any previous `auth_time`." `concepts/acr-amr.mdx` hands the reader
`if (idToken.amr.includes('otp'))`. An RP following both accepts a stale second factor as a fresh
one.

The reviewer reproduced this on sqlite for both a `level1` and a `level2_mandatory` target; the same
ceremony against the unmodified tree passed, because the token omits `otp`.

**This is not me relitigating decision 1.** Decision 1 put the three claims in one class describing
the session's authentication, and we scoped out the pure SSO case, where nothing is verified and
`AuthTime` is therefore not refreshed. This is the case where a credential *is* verified and its set
is a strict subset of what the session retains. Decision 8 reasoned about the row and the code
disagreeing, never about the timestamp.

### (b) The merge is still last-writer-wins

`BumpUserSession` reads the session row *before* `BeginTransaction`, merges in memory, and
`CommonDatabase.UpdateUserSession` then writes every non-`pk` column with only `id` in the
predicate — all four engines delegate to that one implementation. Two bumps read `pwd`, a step-up
writes `pwd otp`, and a stale password-only, silent or refresh bump writes its old `pwd` last.

So §4's stated safety property, "both writers of a session's methods are monotonic", does not hold,
and the `concepts/acr-amr.mdx` sentence stage 2 owes would be published false. The reviewer executed
the interleaving against the real sqlite implementation, with the reverse write order as a passing
control.

Ordinary two-tab or two-client traffic reaches it — nothing an attacker steers — and the harm
direction is under-claiming, so it fails closed. But the invariant is the thing stage 2 exists to
state.

### On the spec

Nothing here is forced by the normative text, and that is worth saying plainly: the choice is
genuinely yours. OIDC Core defines `auth_time` as "Time when the End-User authentication occurred"
and `amr` as methods "used in the authentication", both OPTIONAL, and never says which
authentication when a session outlives a ceremony. RFC 9068 §2.2.1 is the closest thing to a
constraint, putting `auth_time`, `acr` and `amr` in one class whose "values are fixed and remain the
same across all access tokens that derive from a given authorization response" — which says the
three must describe *one* authentication, and that is exactly what (a) breaks. It does not say which
one. All four options below are permitted; the spec exists to enable an RP to reason about how
authentication was performed, which is the thing (a) costs it.

### Options

All four leave stage 1 untouched.

**A. Land stage 1 only; file the mirror case as its own issue carrying both findings.** Reverses
decision 2 on evidence decision 2 did not have. Stage 1 closes #140 as filed — the over-claiming
path, which fails open. What decision 2 refused was shipping "a stated property half enforced", and
B below is precisely that, whereas A ships one property whole and claims nothing about the other.
Cost: the mirror defect stays open, failing closed, until its own run.

**B. Keep stage 2, preserve `auth_time` when the ceremony's methods are a strict subset of the
session's, and file the concurrency gap as a follow-up.** Smallest coherent triple — all three
claims then describe the session's original authentication. Cost: falsifies the published
`prompt=login` promise for a partial re-authentication over a level 2 session (a `prompt=login` then
buys no fresh `auth_time`, so RPs using it for freshness are affected), and still states an
invariant concurrency can violate.

**C. B, plus an atomic narrow persistence operation with four-engine data coverage.** Closes both
fully. Cost: a `Database` contract change and the data tier §5 says this change does not owe, and a
materially larger stage 2.

**D. Keep stage 2 and require the retained methods again — a ceremony re-authenticating over a
session holding `otp` goes through level 2 before completing.** Makes all three claims honest *and*
fresh, and removes the subset case at its source rather than papering over it. Cost: a user-visible
extra OTP prompt on every `prompt=login` over a level 2 session, including for level 1 clients, plus
a change to when level 2 is required, which §2 does not currently scope.

### Recommendation: A

Stage 1 closes the defect the issue was actually filed for, and it is unaffected by both findings.
Stage 2 fixes a defect that fails closed, and getting it right now needs *both* an owner call on the
`(auth_time, acr, amr)` transition and an interface-level persistence change the agreement
explicitly scoped out. Doing it in its own agreement costs one more run and buys a design the
interview can settle properly; doing it here ships a stated invariant that the change does not
enforce, which is the outcome decision 2 was trying to avoid in the first place.

If you would rather not split it, **B** is the next best: it makes the triple coherent for one
grant, and the concurrency residual fails closed.

**Cost of getting it wrong either way.** Picking A when B would have done means the mirror defect
lives longer while failing closed. Picking B, C or D when A would have done means a second round of
design inside a sealed agreement, and for D a behaviour change users will notice.

Issue #199, opened 2026-08-10
Security: no way to find a grant that was cross-bound before #133, once its session row is gone


leodip:
After #133 an authorization ceremony can no longer bind to a session belonging to a different user, and four validation paths refuse a grant whose session belongs to somebody else. Every one of those comparisons needs the session row to still exist. Sessions are deleted after their idle timeout and maximum lifetime (defaults 2h and 24h) while an auth-code-derived offline grant runs to `RefreshTokenOfflineMaxLifetimeInSeconds` (default 31536000, one year), so a grant cross-bound shortly before an upgrade can outlive the only evidence of it by roughly four hundred to one and keep rotating for up to a year. After that it is byte-for-byte indistinguishable from a legitimate grant whose session merely expired.

Its tokens name the correct user. What is wrong is the authentication strength they assert: `createTokenInputFromCode` re-copies `Code.AcrLevel`, `Code.AuthMethods` and `Code.AuthenticatedAt` on every rotation, so the grant keeps presenting an `acr` taken from the other user's session while its `amr` holds only what its own holder proved. It is not access to the other account.

Investigate whether a durable incoherence signal exists in the `codes` table. The candidate, proposed during the review that raised this, is `acr_level = 'urn:goiabada:level2_mandatory'` with `auth_methods` not containing `otp`, which should be unreachable through a legitimate ceremony because a `level2_mandatory` target always ends in OTP verification or enrollment. Nothing has verified that claim, so step one is establishing whether any legitimate path can produce that pair, over the real `codes` table across the supported engines. Only then is a revocation off it defensible. Note it does not catch the `level2_optional` case at all, which is indistinguishable from a user with no OTP enrolled.

Then decide whether Goiabada should offer an operator-run revocation over the matching population, or simply a report.

Raised during #133 (PR #196) as option D of its session-ownership decision, and deliberately not taken there: running a revocation off a detector invented during a review is the wrong way round in an authorization server. #133 documents the residual in `concepts/user-sessions.mdx` so an operator can close it by hand.

Issue #239, opened 2026-08-22
Spike: review the whole authorization ceremony for RFC conformance and security


leodip:
A spike, not a fix. The deliverable is a written review of the authorization ceremony as a whole,
plus one issue per finding. No behaviour changes land under this issue.

## Why now

The ceremony has been changed one defect at a time: #79/#112 (a form must name the ceremony that
rendered it), #106 (authentication generation), #129 (durable session termination), #133 (a ceremony
may not reuse a session it does not own), #213 (an authorization error is not delivered to an
unauthenticated browser), #140/PR #238 (`amr` names only what the bound session verified). Each fix
was locally correct, and most of them added something: a field on `AuthContext`, a gate at the top of
a handler, or a rule about which handler may write which field.

`oauth.AuthContext` now carries 27 fields. Five of them (`Level1AuthCompleted`,
`AuthStateGeneration`, `CeremonyId`, `DeferredErrorCode`, `DeferredErrorDescription`) have
multi-paragraph invariant comments, and those comments are the only place the invariants are written
down. `AuthState` has 11 constants. Nothing states the full transition table, which handler is
allowed to write which field, or what a restart at level 1 is supposed to discard. PR #238 exists
because two fields (`AuthMethods` and `ConsentedScope`) survive a restart that should have cleared
them, and it was found by chasing a specific report rather than by reading the state machine. The
next field with the same shape will be found the same way unless someone reads the whole thing once.

## Surface under review

Handlers, all `src/authserver/internal/handlers`:

| route | file | writes |
| --- | --- | --- |
| `GET,POST /auth/authorize` | `handler_authorize.go` (1085 lines) | mints the auth context and `CeremonyId` |
| `GET /auth/level1` | `handler_auth_level1.go` | auth state |
| `GET /auth/level1completed` | `handler_auth_level1.go` | auth state, clears `Level2AuthConfigHasChanged` on the session row, delivers a deferred error |
| `GET /auth/level2` | `handler_auth_level2.go` | auth state |
| `GET,POST /auth/pwd` | `handler_auth_pwd.go` | `AuthMethods`, `AuthenticatedAt`, `Level1AuthCompleted`, `AuthStateGeneration` |
| `GET,POST /auth/otp` | `handler_auth_otp.go` | `AuthMethods`, `AuthenticatedAt`, OTP enrolment |
| `GET /auth/completed` | `handler_auth_completed.go` | creates or bumps the user session row |
[...]

PR #253, opened 2026-08-23
fix(core): make the client's ACR level a floor that acr_values cannot lower

### LINKED ISSUE #240: Security: acr_values lowers a client's configured ACR level, so a URL parameter turns off a mandatory second factor
Found by the #239 authorization ceremony review (findings [F1](https://github.com/leodip/goiabada/issues/239#user-content-f1) and [F15](https://github.com/leodip/goiabada/issues/239#user-content-f15)).

## 1. `acr_values` lowers a client's configured ACR level

`AuthContext.GetTargetAcrLevel` returns the first parsable `acr_values` entry outright:

```go
func (ac *AuthContext) GetTargetAcrLevel(defaultAcrLevelFromClient enums.AcrLevel) enums.AcrLevel {
	acrValuesFromAuthorizeRequest := ac.parseAcrValuesFromAuthorizeRequest()
	if len(acrValuesFromAuthorizeRequest) > 0 {
		return acrValuesFromAuthorizeRequest[0]
	}
	return defaultAcrLevelFromClient
}
```

All four call sites (`HandleAuthorizeGet` via `handlePromptNone`, `HandleAuthLevel1CompletedGet`, `HandleAuthLevel2Get`, `HandleAuthCompletedGet`) use the result directly. Nothing clamps it to the client's `DefaultAcrLevel`.

**Stated without reference to any documentation: there is currently no configuration that enforces two-factor authentication.** Set a client to `urn:goiabada:level2_mandatory`, the option the admin console labels *"ACR level 3 - password + mandatory OTP"*, and appending `&acr_values=urn:goiabada:level1` turns it off.

`acr_values` arrives on the front channel with no client authentication. Whoever composes the URL decides, and that is the end user, or anyone who can get the end user to follow a link. The client's registered configuration is the only server-side statement of policy in the exchange, and a front-channel parameter overrides it.

### The bypass runs against the seeded default configuration

The admin console client is seeded `DefaultAcrLevel: enums.AcrLevel2Optional` (`database_seeder.go`), so an administrator who has enrolled an authenticator is prompted for it.

1. The admin console redirects the browser to `/auth/authorize?...&acr_values=2` through `AuthHelper.RedirToAuthorize`. That value is unparsable and therefore discarded, so the client default applies and OTP is required.
2. An attacker holding the victim's password edits the redirect target to `acr_values=urn:goiabada:level1`. The `state`, `nonce` and `code_verifier` live in the admin console's own cookie and are untouched, so its callback still validates.
3. The password is entered. No OTP is requested.
4. `handler_auth_callback.go` validates the nonce and signs the user in. **It never reads `acr`.**
5. The account area is reachable, and the victim's OTP can be turned off from there.

So the second factor protects nothing against an attacker who has the password, which is the exact threat a second factor exists for. No operator misconfiguration is required, and it defeats the second factor of exactly the population that chose to enable one.

## Decision: `DefaultAcrLevel` becomes a floor

This is a deliberate design being reversed, not a typo being fixed, and the issue should be read that way. The current behaviour is pinned by `TestGetTargetAcrLevel_SingleValue` and `TestGetTargetAcrLevel_FirstRecognizedValueWins`, whose comment reads *"the first recognized `acr_values` entry must win regardless of the client default"*.

The documentation contradicts itself, and the code sides with the majority of it:

| Source | Says | Reading |
|---|---|---|
| `acr-amr.mdx`, "Default ACR per client" | *"determines the **minimum** authentication level required"*, with a banking `level2_mandatory` example | floor |
| `authorization-lifecycle.mdx`, "acr_values vs client default ACR" | *"allowing applications to request a **higher (or lower)** authentication level"* | default |
| `clients.mdx`, "Default ACR level" | *"You can **override** the client's default"*, step-up example only | ambiguous |
| Admin console tooltip (`active.en.toml`) | *"The **default** authentication policy for the client"* | default |
| `GetTargetAcrLevel` | first parsable entry wins | default |

### Is a floor RFC conformant? Yes, and nothing requires the alternative

- **OIDC Core 15.1**, Mandatory to Implement Features for All OpenID Providers, is the ceiling on the whole obligation: *"OPs MUST support requests for specific Authentication Context Class Reference values via the `acr_values` parameter, as defined in Section 3.1.2. (Note that the minimum level of support required for this parameter is simply to have its use not result in an error.)"* Ignoring `acr_values` outright is conformant, so clamping it is conformant a fortiori.
- **OIDC Core 5.5.1.1** holds the only MUST to honour a requested value, and it is doubly conditioned: *"If the acr Claim is requested as an Essential Claim for the ID Token with a value or values parameter requesting specific Authentication Context Class Reference values **and the implementation supports the claims parameter**, the Authorization Server MUST return an acr Claim Value that matches one of the requested values."* Goiabada supports neither the `claims` parameter nor advertises `claims_parameter_supported`, both verified. **That MUST is unreachable here.**
- **OIDC Core 5.5.1.1**, on the only path that does exist: *"Note that the RP MAY request the acr Claim as a Voluntary Claim by using the `acr_values` request parameter... If the Claim is not Essential and a requested value cannot be provided, the Authorization Server SHOULD return the session's current acr as the value of the acr Claim."* Clamping up and reporting the higher level is what that prescribes.
- **OIDC Core 3.1.2.1** frames the parameter as the values the server *"is being requested to use... with the values appearing in order of preference"*, and *"The acr Claim is requested as a Voluntary Claim by this parameter."*
- **OIDC Core 2**, defining the claim: *"String specifying an Authentication Context Class Reference value that identifies the Authentication Context Class that the authentication performed satisfied."* The claim reports what happened.
- **OIDC Core 16.20**: *"the `max_age` and `acr_values` provide more assurance about the nature of the authentication performed when delivered in signed requests."* The specification itself discounts unsigned `acr_values`, and this server supports no signed requests at all.

One sentence cuts the other way, recorded here rather than left for a reviewer to find. **OIDC Core 3.1.2.2**: *"If the acr Claim was requested, the Client SHOULD check that the asserted Claim Value is appropriate."* That is a SHOULD on the relying party, not a licence for the authorization server, and this server's own admin console does not do it.

### What other identity providers do

| Provider | Behaviour |
|---|---|
| Okta | Floor. *"The app sign-in policy is always evaluated before the `acr_values` parameter."* Unsupported values *"are ignored"*. A level the user cannot meet returns `error=unmet_authentication_requirements`. |
| Microsoft Entra ID | Floor in effect. Conditional Access is evaluated server side and the authentication-context claim is issued only when every policy assigned to it has been satisfied. |
| Keycloak | The same behaviour Goiabada has today, and it documents the limitation: *"default ACR values are used as the default level, however it cannot be reliably used to enforce login with the particular level."* An "Enforce LoA" authenticator was requested (keycloak#16884) and never merged. |
| Auth0 | Documents requesting MFA upward; step-down is not addressed either way, so nothing is claimed. |

Keycloak behaves exactly as Goiabada does and warns operators not to rely on it. Goiabada has Keycloak's behaviour with Okta's promises. The choice was never conform or do not conform, since both conform. It was whether to move the behaviour to the promise or the promise to the behaviour, and moving the promise would have meant renaming "mandatory" to something that is not a lie.

## 2. An empty `amr` array can be signed

Independent of the decision above and can land separately.

`authMethodsToArray("")` returns `[]string{}` and `claims["amr"]` is assigned unconditionally. `"amr": []` positively asserts that the authentication used no methods; absent would say nothing, and `amr` is OPTIONAL per OIDC Core 2.

Unreachable through a fresh ceremony today, reachable from a `user_sessions` row with an empty `auth_methods`, which nothing prevents at the database level.

Fix: assign `amr` only when the array is non-empty, in both `generateAccessTokenCore` and the id_token generator.

## What this pull request has to carry

- `GetTargetAcrLevel` returns `enums.AcrMax(requested[0], clientDefault)`.
- `amr` is omitted rather than emitted empty, at both claim sites.
- The two pinning tests are **rewritten, not deleted**. `TestGetTargetAcrLevel_FirstRecognizedValueWins` keeps its cases where the request is at or above the client default and gains cases asserting the clamp below it. Its comment becomes false and must be replaced rather than left to confuse the next reader.
- `AuthHelper.RedirToAuthorize` sends `acr_values=2`, which has been silently discarded since the ACR values became URNs. Fix it here or in the ceremony tidy-up issue, but not neither.
- `authorization-lifecycle.mdx` loses *"allowing applications to request a higher (or lower) authentication level"* and its table row, both of which become wrong.
- `clients.mdx` keeps its step-up example, which stays correct, and gains a sentence saying a request below the client's level is raised to it.
- `acr-amr.mdx` keeps *"minimum authentication level required"*, which becomes true for the first time, and gains the same sentence.
- **Release note.** This is a behaviour change for any deployment relying on step-down, in the safe direction. It is breaking and should be called one.

## Tests

- Unit: the two rewritten `GetTargetAcrLevel` cases, plus a clamp case per ACR pair.
- Unit: `amr` omitted when `AuthMethods` is empty, at both claim sites.
- Integration: client with `DefaultAcrLevel = level2_mandatory`, user with `OTPEnabled = true`, authorization request carrying `acr_values=urn:goiabada:level1`, assert `/auth/otp` is reached.

## Deliberately not done

A per-client "allow `acr_values` to lower the level" switch. If a deployment turns up needing step-down it can be added later, and adding it then is a loosening rather than a second breaking change. Building it now costs a migration on four engines, an admin console field, an admin API field and documentation, for a use case nobody has asked for.

## Related

- #140 / PR #238 is the sibling defect on `amr`: a session-less completion inherits the previous session's `amr`. Once it lands, the "exactly four writers of `AuthMethods`" invariant is worth pinning with a test.
- The admin console callback should check the `acr` claim per OIDC Core 3.1.2.2. A floor makes that check redundant for Goiabada's own client, and it remains the right thing for any RP to do.

leodip:
Fixes #240.

## What this is for

`acr_values` is a query parameter on the front channel. Nothing authenticates it: whoever composes
the authorization URL decides its value, which means the end user, or anyone who can get the end
user to follow a link. Today Goiabada takes the first recognized entry in it as the target
authentication level outright, without ever comparing it against the client's configured
**Default ACR Level**.

So a client configured `urn:goiabada:level2_mandatory`, the setting the admin console labels
"password + mandatory OTP", is turned off by appending `&acr_values=urn:goiabada:level1`. There is
currently no configuration that enforces a second factor.

It runs against Goiabada's own admin console, which is seeded at `level2_optional`. An attacker
holding an administrator's password rewrites the redirect target to ask for `level1`. The `state`,
`nonce` and `code_verifier` live in the admin console's own cookie and are untouched, so its callback
still validates, and that callback never reads `acr`. The account area is then reachable and the
victim's OTP can be turned off from there.

## The rule this establishes

**A client's configured ACR level becomes a floor.** The target level is
`max(first recognized acr_values entry, client's DefaultAcrLevel)`. A request can raise the
authentication level and can never lower it.

This is a deliberate design being reversed rather than a typo being fixed, and it is a breaking
change for any client that relied on stepping down. Both behaviours conform to OpenID Connect Core:
15.1 permits ignoring `acr_values` entirely ("the minimum level of support required for this
parameter is simply to have its use not result in an error"), and 5.5.1.1's MUST to honour a
requested value is conditioned on supporting the `claims` request parameter, which this server does
not. 5.5.1.1 also says that where a requested value cannot be provided and the claim is not
essential, the server SHOULD return the session's current `acr`, which is what reporting the raised
level does.

An integrator comparing the returned `acr` to the one they requested with string equality will be
surprised the first time they ask for less than the client's configured level. OIDC Core 3.1.2.2 says
"appropriate" rather than equal, so equality was never the prescribed check, and the surprise is
always in the safe direction.

A second, smaller defect ships alongside it: an access token or ID token could carry `"amr": []`,
which positively asserts that the authentication used no methods where absent would say nothing.
OIDC Core 2 makes `amr` OPTIONAL, so the claim is now omitted when no method was recorded.

## Status

**Complete and ready for review.** All three stages landed, the whole branch was reviewed four
times, every finding is resolved, and the check suite is green on the final commit `dcb2e959` across
all eleven jobs, [run 32644613327](https://github.com/leodip/goiabada/actions/runs/32644613327).
Nothing is deferred and there is no follow-up to file.

### Stage 1: the ACR floor

Commit `022e58bb`.

**The floor.** The target ACR is now `enums.AcrMax(first recognized acr_values entry, client's
DefaultAcrLevel)`. `enums.AcrMax` was already the codebase's ACR comparison, used one layer up by
`SetAcrLevel` for the same never-downgrade rule against a session's ACR, so this makes a client's
configuration a floor in exactly the way a session's ACR already was. There are four readers of the
target and every one of them takes it from the same function, so no path computes one another way.

**The target is fixed when the request is accepted.** `AuthContext` gains `TargetAcrLevel`, written
once in `HandleAuthorizeGet` after every validation has passed, and the later handlers read it.
Without this, `/auth/level1completed`, `/auth/level2` and `/auth/completed` each reloaded the client
and recomputed the target, so the level a ceremony had to reach was a live read of a mutable row at
three points in one ceremony. An administrator raising `default_acr_level` after the step-up
decision had been taken made `/auth/completed` stamp `acr: urn:goiabada:level2_mandatory` over
`amr: ["pwd"]` on a ceremony that only ever saw a password, and OIDC Core section 2 defines `acr` as
the class "the authentication performed satisfied", so that claim was false in the direction a
relying party doing 3.1.2.2's check trusts. Lowering the row before `/auth/level2` took its `switch`
to the default branch and answered 500.

That race predates this change: with no `acr_values` in the request, which is the ordinary shape of
nearly every request, the old code returned the client default verbatim, so the live re-read already
decided alone. The floor would have widened it to all requests, which is why it is closed here
rather than filed. Reaching it needs `authserver:manage`, and that actor could already set the row
to `level1` permanently, so nobody gained a capability; the realistic case is an operator tightening
policy while somebody is mid-ceremony.

**A consequence worth knowing:** a change to a client's ACR level applies to authorization requests
that start after it, not to ceremonies already in flight. `concepts/clients.mdx` says so. An absent
or unparsable snapshot, which is what a cookie written by an older binary carries, falls back to
computing from the client's current row: that is the pre-snapshot behaviour, is never below what the
request asked for, and lets a ceremony in flight across a deploy finish rather than fail.

**Also deleted:** `RedirToAuthorize` sent `acr_values=2`, unparsable since the ACR levels became
URNs in 2023 and therefore silently discarded. Removing it is behaviour-identical, and it is the
admin console's own redirect, so the seeded client this issue's worked example attacks now has its
configured level as the only thing deciding.

**Tests.** The seam owning the logic gets an explicit fifteen-row `(acr_values, client default,
expected)` grid across three unit functions, replacing a shape that could not express a client
default above the request at all. Two endpoint cases cover both consumers of the target, the silent
path through `handlePromptNone` and the interactive one through the three later handlers, and one
drives a real ceremony with the client row raised in the gap between two hops. The cross-user
fixture now takes the client's level and user A's session level as separate parameters: the floor
applies to user B's ceremony too, so raising the client to lift A's session lifted B's target with
it and the ACR assertion stopped discriminating.

**Tiers:** unit 1077 pass; data 1199 pass on all four engines; integration 957 pass on sqlite; the
docs site builds. Six mutations, all caught. Three of them are wiring mutations, and they are the
ones that matter: every unit case here would still pass if the endpoint never sent `acr_values` or
if `HandleAuthorizeGet` never took the snapshot, because they exercise the functions directly.

### Stage 2: `amr` omitted when no authentication method was recorded

Commit `c8932e4c`.

Both token claim sites, `generateAccessTokenCore` and `generateIdTokenCore` in
`src/core/oauth/token_issuer.go`, assigned `claims["amr"] = input.AuthMethods` unconditionally, so a
grant carrying no recorded authentication method signed `"amr": []`. OIDC Core 1.0 section 2 makes
`amr` OPTIONAL: absent says nothing about how the user authenticated, where an empty array
positively asserts that the authentication used no methods at all. Each site is now guarded with
`if len(input.AuthMethods) > 0`.

The state is unreachable through a ceremony and reachable from storage. `AuthMethods` is only ever
written by `AddAuthMethod`, which appends `pwd` for every password ceremony, or copied from a
`user_sessions` row, whose `auth_methods` is a plain string column with no constraint. So this stops
a false claim being signed for a row nothing currently writes, rather than changing any token a
login can mint. Which methods a ceremony is entitled to claim is a separate question and belongs to
#140.

`authMethodsToArray` is unchanged: `[]string{}` for an empty string is the right thing for a caller
to test, and moving the decision into the helper would put it away from the claim sites, where the
OPTIONAL-claim rule applies. `acr` on the adjacent line is deliberately left alone, because an empty
session ACR is refused by `enums.AcrLevelFromString` before any token is signed.

One documentation edit. The site's "Checking ACR and AMR in your application" example read
`idToken.amr.includes('otp')`, which throws a `TypeError` rather than reading false when the claim is
absent, against this OP or any other. It now checks the claim is there first. Two other pages
mention `amr` in their core-claims lists and were deliberately left: `amr` is still present in every
token a ceremony can mint, so rewording them would announce internal work no user can observe.

**Tiers:** unit 3563 pass, 0 fail. No data or integration tier: no query, no migration, no schema,
no handler and no route changes, and the empty state cannot be produced by a ceremony, so an
integration case would have to hand-write a `user_sessions` row and would then be asserting its own
fixture. Two mutations, one per claim site, each reverting the guard to an unconditional assignment;
both caught.

### Stage 3: the upgrade-note sweep

Commit `9bf6f0c9`.

A documentation rule the user settled during the interview, applied to the pages that broke it: the
documentation describes the software as it is now, and upgrade guidance belongs in the release
notes, written at release time. Six notes across five pages did not follow it. History, meaning
"this used to be different", is deleted. Live state, meaning "your deployment can be in this
condition right now", is kept and reworded to describe the condition rather than the upgrade that
caused it.

Deleted: the whole `## Upgrading from earlier versions` section in `concepts/audit-log.mdx`, about
an environment variable removed in v1.5.1, whose every current-state fact (both logging targets and
their defaults, 180 day retention, the settings page) is already stated above it; and the
`**For operators upgrading:**` paragraph in `concepts/clients.mdx`, whose two live facts are stated
three paragraphs above it.

Reworded: the unbadged self-registered client in `concepts/clients.mdx`, which an operator still has
to find and tick **Consent required** on by hand; the grants tied to a previous user's session in
`concepts/user-sessions.mdx`, whose section still carries the three-row remediation table #199's
closing line points readers at, and whose heading is now **Grants tied to somebody else's session**;
the database-held encryption key in `reference/environment-variables.mdx`, where only the bold lead
changed, since the auth server makes that check at every startup rather than only the first; and the
request log on stderr in `reference/security.mdx`, where the live warning stays and now reads "you
will not see request records at all" rather than "you will stop seeing" them.

Two anchors were at risk and both were checked in the built HTML. The renamed heading in
`user-sessions.mdx` has no inbound link. The `### Rotating the key` heading does, from the
`GOIABADA_AES_ENCRYPTION_KEY_PREVIOUS` row on the same page, which is why only the paragraph lead
beneath it changed.

**Tiers:** none applicable. Five `.mdx` pages, no Go, no catalog key, no route, no handler, no
schema, so nothing executable changed and no unit, data or integration result could differ. Checked
with `cd site && npm run build`, green, 42 pages. That build parses and renders the pages but does
not validate links, so the three anchors were read back out of `site/dist` directly. The recount
`grep -rn -i 'upgrad|earlier version' site/src/content/docs/` went from 13 lines to 6; each survivor
was read and none is an upgrade note (five use "upgraded" for an ACR level rising within a session,
which is current behaviour, and one credits a security assessment of version 0.0.3).

### After the stages: four documentation fixes from the review

`d35c9b1f`, `f3c26df6`, `9647756b`, `dcb2e959`. Every finding the four review rounds raised was a
documentation defect, and each is fixed in its own commit rather than folded back into a stage, so
the record of what the review changed survives. In order: the self-registered-client caution
described the `dcr_` identifier prefix as the live discriminator when the runtime reads
`clients.created_via_dcr`; the same caution then credited the mark for the consent default, which
nothing reads it for; three descriptions said the *first* `acr_values` value wins when the parser
discards unrecognised entries before selecting; and the lifecycle algorithm still partitioned on the
parameter's presence while its own table partitioned on recognition, so `acr_values=unknown`
satisfied both arms.

No Go file, catalog key, route or schema moved in any of the four. The split tree hashes are
byte-identical across all four rounds, `production 9fd22db15c79`, `unit-tests af675851e310`,
`data-tests cc52c7f6ac48`, `integration-tests 16e1b1da7a5d`, with only `docs` changing.

## Tiers

|---|---|---|
| Unit | locally, and the check suite | stage 1: 1077 pass. stage 2: 3563 pass, 0 fail |
| Data, four engines | the check suite | `Tests / sqlite`, `mysql`, `postgres`, `mssql`, all green |
| Integration, sqlite | locally | stage 1: 957 pass |
| Integration, four engines | **the check suite only** | green at `9647756b` and at `dcb2e959` |
| Docs site | locally | `npm run build`, 42 pages, green |
| Anchors | locally | `check-anchors.sh`, all 33 rows PASS |

The four-engine data and integration claim is the check suite's, not a local run's: nothing local
ever executed mysql, postgres or mssql. Stage 3 and the four review fixes ran no Go tier at all,
correctly, because they change only `.mdx` pages and no unit, data or integration result could
differ.

**Eight mutations against the change's own new tests, all caught.** Six in stage 1, two in stage 2.
Three of stage 1's are *wiring* mutations and they are the ones that matter: every unit case there
would still pass if the endpoint never sent `acr_values`, or if `HandleAuthorizeGet` never took the
snapshot, because they exercise the functions directly. The reviewer ran a further eleven of its own
across rounds 1 to 3, and reported every one caught.

## Review

Reviewed by `gpt-5.6-sol` at effort `xhigh`: one round at the plan gate, then **four rounds over the
whole branch**. `LEO_REVIEW_MODE=end` was in force for the entire run, so **no stage was reviewed on
its own** and the code was read as a whole with everything built on top of it in view.

|---|---|---|---|---|
| plan | 1 | 4 | 1 | 1 escalated as decision 6, 2 fixed in the plan, 1 withdrawn |
| final | 1 | 1 | 0 | fixed, `d35c9b1f` |
| final | 2 | 2 | 0 | both fixed, `f3c26df6` |
| final | 3 | 1 | 0 | fixed, `9647756b` |
| final | 4 | 1 | 0 | fixed, `dcb2e959` |

`unchecked` was empty in all four final rounds, so there was no work list outstanding. Nothing was
deferred. One finding was refuted on evidence and one was withdrawn by the reviewer itself.

- **blocking, security:** a client's `DefaultAcrLevel` was re-read at three handlers, so an
  administrator changing it mid-ceremony could stamp `acr: level2_mandatory` over `amr: ["pwd"]`.
  Mechanism confirmed, **attribution refuted**: a probe against the real functions shows the race
  already on `main` in the ordinary no-`acr_values` shape, with the floor widening it rather than
  creating it. Escalated as decision 6; you answered to close it here, and it is the `AuthContext`
  snapshot in `022e58bb`.
- **significant, quality:** the endpoint tests would have passed with their `acr_values` stimulus
  removed. Answered by a third mutation run at the integration tier, where a dropped stimulus makes
  it survive; these are stage 1's three wiring mutations.
- **significant, quality:** stage 2 called an independent defect annexed to this change. Withdrawn
  against #240's own scope, which names the `amr` omission as its second bullet.
- **four minor, conformance:** all documentation, all fixed, listed above.

**The pattern is the useful summary.** Every finding after the plan gate was a documentation defect,
and three of the four were inside the paragraph the previous round had just rewritten. The
executable change was reported sound from round 1 and **no round moved a line of it**. The budget
funds four rounds, so round 4's own one-line repair is the single edit here that no reviewer has
seen; it copies its wording verbatim from a sentence on the same page that rounds 2, 3 and 4 each
confirmed correct.

Full account with the evidence: [the review comment](https://github.com/leodip/goiabada/pull/253#issuecomment-5386475439).

## Deviations worth knowing

- **Stage 1 grew a design the agreement did not seal.** The `AuthContext.TargetAcrLevel` snapshot
  came from the plan review's blocking finding and was built only after the question was escalated
  and answered. It changes a persisted shape, since `AuthContext` rides in the session cookie, and
  it decides an operator-visible behaviour: **a change to a client's ACR level applies to
  authorization requests that start after it, not to ceremonies already in flight.**
  `concepts/clients.mdx` now says so.
- **A cookie written by an older binary carries no snapshot.** That case falls back to computing the
  target from the client's current row, which is the pre-snapshot behaviour and is never below what
  the request asked for, so ceremonies in flight across a deploy finish rather than fail. The
  residual window closes as the session cookies age out.
- **The documentation sweep in stage 3 is wider than the defect.** Five pages unrelated to ACR were
  touched, under a rule settled during the interview: the documentation describes the software as it
  is now, and upgrade guidance belongs in the release notes at release time. That cost was put and
  accepted before the stage was planned. It is also why this change adds **no** upgrade note of its
  own, and why this body is written to be the source a release note draws on.
- **Two anchors were re-swept mid-review**, `docs/acr-amr-first` and `docs/lifecycle-steps`, because
  round 2's fix changed the very sentences they cite.

## For the release notes

Lifted from `concepts/user-sessions.mdx`, which described this as live state. It is upgrade
guidance, so it belongs here and the page no longer carries it.

**Grants from before 1.6.0 can over-claim `acr`.** Until this release, a sign-in on a shared browser
could stay tied to the previous person's session. A grant like that names the right user, and its
`amr` and `auth_time` are honest, but its `acr` can be the level the previous person's session had
reached, higher than anything its holder actually presented, and it goes on claiming that through
every refresh.

Goiabada refuses these wherever it can still see the session they name. That stops once the session
is gone, and sessions are deleted when they idle out or reach their maximum lifetime, two hours and
twenty-four hours by default. An offline refresh token's maximum lifetime defaults to a year, so it
outlives the evidence by a long way.

If your applications make authorization decisions from `acr`, and shared browsers are realistic in
your deployment:

| Situation | What to do |
| --- | --- |
| The session it came from is still there | End that session, from the user page or the user's own sessions page. That revokes the refresh tokens and unredeemed codes it issued, offline ones included |
| The session has already gone | Revoke the client's consent, from the user page or the user's own consents page. An offline grant is checked against consent every time it is refreshed, so this stops it at the next one |
| Neither fits | Set a new password for the user, or disable them. Either stops every grant that user holds, on every device |

**Lowering the offline maximum lifetime on Settings → Tokens does not close these.** That setting is
read when a grant is first issued, and every refresh afterwards carries the deadline the grant was
born with. There is no token revocation endpoint either.

Deployments where every user has their own browser have nothing to do.

## Closing comments

- [Decisions](https://github.com/leodip/goiabada/pull/253#issuecomment-5386467927), six taken, none
  deferred, and the one that was escalated to you.
- [Follow-ups](https://github.com/leodip/goiabada/pull/253#issuecomment-5386470173), none to file,
  and why the one draft was discarded.
- [The review account](https://github.com/leodip/goiabada/pull/253#issuecomment-5386475439), all five
  rounds with the evidence.


leodip:
## Decision 6: a client's `DefaultAcrLevel` is re-read at every handler, so changing it mid-ceremony can stamp an `acr` the user never performed. Close it here, or file it?

The plan review raised this as a blocking security finding against the ACR floor. I verified it and
reproduced it, and the mechanism is real but the attribution is not: **the race is already on `main`.
The clamp widens it rather than creating it.** That changes the question from "is the design wrong"
to "does this issue own the fix", which is yours.

### The class

`HandleAuthLevel1CompletedGet`, `HandleAuthLevel2Get` and `HandleAuthCompletedGet` each call
`GetClientByClientIdentifier` and then `authContext.GetTargetAcrLevel(client.DefaultAcrLevel)`. The
effective authentication target is therefore a live read of a mutable row at three points in one
ceremony, and nothing carries forward what the ceremony was told to do when it started. Four members:

- **False elevation.** An administrator raises `level1` to `level2_mandatory` after
  `HandleAuthLevel1CompletedGet` has already decided no step-up is needed. `HandleAuthCompletedGet`
  passes the new value to `BumpUserSession` and `SetAcrLevel`, and the authorization code carries
  `acr: urn:goiabada:level2_mandatory` with `amr: ["pwd"]`. No OTP was performed. OIDC Core 2 defines
  `acr` as the class "the authentication performed satisfied", so the claim is false, and a relying
  party doing OIDC Core 3.1.2.2's check is misled in the unsafe direction.
- **The same, one hop later.** `level2_optional` raised to `level2_mandatory` after
  `HandleAuthLevel2Get` has already chosen to skip OTP for a user with no authenticator.
- **A 500.** Lowered to `level1` before `HandleAuthLevel2Get`, whose `switch` handles only
  `AcrLevel2Optional` and `AcrLevel2Mandatory` and answers its `default:` branch with
  `errors.New("invalid targetAcrLevel: ...")`. Availability rather than integrity.
- **Safe.** Anything after `HandleAuthCompletedGet`: `AuthContext.AcrLevel` is fixed by then.

### Executed, not argued

I replayed the two target reads through the real `GetTargetAcrLevel` and `SetAcrLevel` with the
client row changed in between, on the tree at `091fb711`:

| Tree | `acr_values` | acr stamped on a password-only ceremony | |
|---|---|---|---|
| today | absent | `urn:goiabada:level2_mandatory` | **race, already on `main`** |
| today | `level1` | `urn:goiabada:level1` | safe |
| clamp | absent | `urn:goiabada:level2_mandatory` | race, unchanged by this issue |
| clamp | `level1` | `urn:goiabada:level2_mandatory` | **race, added by the clamp** |

Row 1 is the point. With no `acr_values` in the request, which is the ordinary shape of very nearly
every real request, `GetTargetAcrLevel` returns the client default verbatim today, so the live
re-read decides on its own and the race is already reachable. The clamp adds the minority of requests
that do carry the parameter. In the opposite direction the clamp is **strictly safer**: an
administrator lowering the row mid-ceremony can no longer take the target below what the request
asked for, because `AcrMax` floors it there, where today it is unbounded.

### Who has to act

`DefaultAcrLevel` is writable from `apihandlers/handler_api_clients.go` and the admin console's
`adminclienthandlers/handler_admin_client_settings.go`, both requiring `authserver:manage`. That
actor can already set the row to `level1` permanently and turn the second factor off for every user
of the client, which needs no timing and is strictly worse. So the widening hands **nobody a
capability they did not already have**, and the unprivileged front-channel attacker this issue exists
to stop cannot reach it at all. The realistic case is benign: an operator tightening a client's
policy while somebody happens to be mid-ceremony, minting one token with a false `acr`.

### Options

- **A. File it as its own issue and build this one as sealed.** The clamp is unchanged. Reversal
  cost: the family stays open one more release, and closing it later is exactly the same work,
  because where the target is *read from* is orthogonal to what `GetTargetAcrLevel` *computes*. I
  have a complete issue drafted and ready to post.
- **B. Snapshot the effective target into `AuthContext` when the authorization request is accepted**,
  and have all four call sites read the snapshot. Closes the whole family at once. Costs:
  `AuthContext` is persisted in the session cookie, so this changes a persisted shape; and it
  decides, silently, that a policy *raise* does not apply to ceremonies already in flight. That is
  defensible, but it is not obviously what an operator tightening policy expects.
- **C. Detect that the row moved and restart or step up before stamping.** Keeps a raise effective
  and still refuses the false claim. Most work of the three, needs its own answer at each of the
  three call sites, and needs a rule for what counts as changed.

### Recommendation: A

The defect is real and worth closing, and it is not this issue's. This issue is about an unprivileged
party rewriting a URL; that is a privileged party racing a row, it predates the change, and the
increment hands nobody a new capability. B is probably the eventual answer, and it changes a
persisted shape and an operator-visible behaviour, which is an interview's worth of decisions rather
than a step appended to a sealed stage. If you want B or C, I would still give it its own issue and
its own agreement rather than an extra stage here.

**Cost of being wrong either way.** Choosing A when you wanted B: one more release carrying a
privileged-actor race that has been present for the life of the feature, and no rework of this
change. Choosing B here: a cookie-shape change and a behaviour change land inside a security fix that
did not need them, unreviewed by an interview, and in-flight ceremonies quietly stop honouring a
policy raise.

Reply with **A**, **B** or **C**. The run is stopped until then; nothing is committed.

leodip:
B

leodip:
## Decisions: nothing was deferred

Six decisions were taken over this change. **All six are `Decided`, none was deferred**, so there is
no assumption on this branch waiting for you to confirm or reverse. This comment says what they were
and where each is visible, because a run that reports "no deferred decisions" and lists nothing is
indistinguishable from one that never recorded any.

Five were settled during the interview or by demonstration during the run. One was escalated to you
mid-run and is the only one that changed the shape of the code.

### The one that was escalated

**Decision 6. A client's `DefaultAcrLevel` was re-read at every handler, so changing it mid-ceremony
retroactively redefined what that ceremony was required to do. Close it here, or file it?**

The plan review raised this as blocking and attributed it to the floor. The run verified the
mechanism and refuted the attribution: the race is **already on `main`**, reachable today in the
ordinary no-`acr_values` shape, and the floor widens it to requests that do carry the parameter
while making the opposite direction strictly safer. Reaching it needs `authserver:manage`, and that
actor can already set the row to `level1` permanently, so nobody gains a capability.

The run recommended **A**, file it as its own issue and build #240 as sealed. **You answered `B` at
2026-08-23T12:19:31Z**, and B is what was built, inside stage 1 rather than as a follow-up.

Two consequences of B that are not visible from the diff, and that a release note should carry:

- **A change to a client's ACR level applies to authorization requests that start after it**, not to
  ceremonies already in flight. This is the cost side of B that the escalation named. It is now
  stated in `concepts/clients.mdx` rather than left silent.
- **`AuthContext` gains a persisted field**, `TargetAcrLevel`, which rides in the session cookie. A
  ceremony in flight across a deploy carries a context without it; that case falls back to computing
  the target from the client's current row, which is the pre-snapshot behaviour and is never below
  what the request asked for. Those ceremonies finish rather than fail, and the residual window
  closes as the session cookies age out.

All three unsafe members of that family are closed by `022e58bb`. There is no follow-up issue for
it, because the work landed.

### The other five, in one line each

| # | Question | Answer |
|---|---|---|
| 1 | What does the clamp compute when `acr_values` lists several levels? | `AcrMax(first recognised value, client default)`. Rejected: walking the list for the first entry at or above the floor, which would push a user with no authenticator into enrolment on the ordering of a URL they never saw; and erroring, which would break every existing client sending `acr_values=level1` at a level2 client. |
| 2 | Does the admin console's `RedirToAuthorize` get fixed here? | Yes, by deleting the `acr_values=2` line. Unparsable since the levels became URNs in 2023, so deletion is behaviour-identical, and it leaves the seeded client's configured level as the only thing deciding. |
| 3 | Does `reference/security.mdx` need anything? | No change. The page already claims "Optional or mandatory 2FA using TOTP" under Authentication security; that claim is false today and this change is what makes it true. The smallest edit that makes the page true is none. |
| 4 | Where does the breaking-change note live? | Nowhere in the docs. **The rule you set:** the documentation describes the software as it is now, and upgrade guidance belongs in the release notes at release time. This PR body carries the narrative for that note to draw on. Stage 3 then swept the six notes already in the docs that broke the rule. |
| 5 | How does the cross-user ACR test keep discriminating under the floor? | Decouple the fixture into the client's level and user A's session level separately, so B's ceremony still targets below A's session and the ACR assertion still fails if the bug returns. Rejected: dropping the assertion, which would then pass with the bug reinstated. |

Decision 4 is the one worth a second look, because it is the only one that widened the change:
sweeping five otherwise unrelated documentation pages beside a second-factor bypass was put to you
as a cost and accepted, and it is stage 3.

leodip:
## Follow-ups: none to file

This run drafted **one** follow-up and then **discarded it**, because the work it described landed
inside this change instead. There is nothing here for you to file.

Recording it anyway, because "no follow-ups" and "nobody looked" read identically, and because one
earlier entry in this run's own record got this wrong and had to be corrected.

### The one draft, and why it is discarded

**"A client's `DefaultAcrLevel` is re-read at every handler, so changing it mid-ceremony can stamp
an `acr` the user never performed."** `bug`, `go`, `security`. Drafted at the plan gate, ready to
file, and explicitly **conditional on decision 6 being answered `A`**.

You answered `B`. The work moved into stage 1 and shipped in `022e58bb`, so the draft describes a
defect that no longer exists on this branch. Verified against the tree as it stands rather than
taken from the record:

- `AuthContext.TargetAcrLevel` exists (`src/core/oauth/auth_context.go`), written once by
  `SetTargetAcrLevel` at the point the authorization request is accepted.
- All four readers of the target go through `GetTargetAcrLevel`, which returns the snapshot when
  there is one: `handler_authorize.go`, `handler_auth_level1.go`, `handler_auth_level2.go`,
  `handler_auth_completed.go`. No path computes a target another way.

All three unsafe members of the family, the two false elevations and the 500, are closed by that.
The fourth member was already safe.

### One thing that was examined and deliberately left, so it does not look like an oversight

`concepts/acr-amr.mdx`'s **ACR and max_age interaction** table says the required level is "the higher
of the requested value and the client's default". With zero recognised `acr_values` entries that
reduces to the client's default, which is what the row already promises. The sentence is true as
written, so the smallest edit that makes the page true is none. Round 3 of the review swept this
family and this is the member it left.

### One correction the record carries, repeated here

Round 1's entry claimed the reviewer's surviving snapshot-writer mutation was "recorded as follow-up
3". That was wrong twice over: there is no follow-up 3, and the survival was not a defect. The
reviewer's own words put it in `ran`, not `findings`: the mutation "survived the core boundary, as
expected from the unit table's ownership, but failed the real cross-user integration path at the
missing `/auth/level2` hop", which "establishes where the test family ends". The property **is**
pinned, by `authorize_cross_user_session_test.go`, which this change added and which caught it.
Filing an issue for it would have pointed a bug report at a test suite that works.

### Duplicate search, re-run at the finish

`gh issue list --state all --search "DefaultAcrLevel mid-ceremony"` and
`--search "acr level client policy change ceremony"` return only #240 and the closed spike #239. The
tracker has not moved since the draft was written, which is what makes the discard safe rather than
merely convenient.

leodip:
## The review account

This is the whole of the reviewing this change received. The verdicts themselves live in a
gitignored directory that the next round overwrites, so this comment is the only durable copy: it is
written self-contained, with the evidence and the numbers, rather than pointing anywhere.

**Reviewer:** `gpt-5.6-sol` at effort `xhigh`, in every round.
**Rounds:** one at the plan gate, then **four over the whole branch** at the final gate.
`LEO_REVIEW_MODE=end` was in force for the entire run, so **no stage was reviewed on its own** and
the code was read as a whole, four times, with everything built on top of it in view.

| Gate | Round | Findings | Blocking | Significant | Minor | Outcome |
|---|---|---|---|---|---|---|
| plan | 1 | 4 | 1 | 2 | 1 | 1 escalated as decision 6; 1 fixed in the plan; 1 withdrawn; 1 fixed in the plan |
| final | 1 | 1 | 0 | 0 | 1 | fixed, `d35c9b1f` |
| final | 2 | 2 | 0 | 0 | 2 | both fixed, `f3c26df6` |
| final | 3 | 1 | 0 | 0 | 1 | fixed, `9647756b` |
| final | 4 | 1 | 0 | 0 | 1 | fixed, `dcb2e959` |

`unchecked` was **empty in all four final rounds**, so there was no work list to cover, dismiss or
file. Nothing was deferred. Exactly one finding was refuted, and one more was withdrawn by the
reviewer itself.

**The pattern, stated because it is the single most useful thing here.** Every one of the five
findings after the plan gate was a documentation defect, and three of the four final-gate findings
were inside the paragraph the previous round had just rewritten. **The executable change was
reported sound from round 1 and no round moved a line of it.** The split tree hashes make that
mechanical rather than a claim: `production 9fd22db15c79`, `unit-tests af675851e310`,
`data-tests cc52c7f6ac48`, `integration-tests 16e1b1da7a5d` are byte-identical across rounds 1
through 4, with only `docs` moving.

---

## Plan gate, round 1

Request `240-20260823T114144Z`, fresh session. Four findings, one blocking. All three axes
`reviewed`. The reviewer read the sealed contract, the plan, the probes, the production code, the
tests, both locale catalogs and the affected documentation; verified the branch state, the split
tree hashes, all 28 anchors and the fanout recounts; fetched OpenID Connect Core 1.0 errata 2 and
RFC 9700 from their publishers; and ran four overlay probes of its own rather than re-running suites
the run had already reported.

**1, blocking, security. A client policy change during a ceremony can stamp an ACR nobody
performed.** `/auth/level1completed`, `/auth/level2` and `/auth/completed` each reloaded the client
and recomputed the target, so the effective target was a live read of a mutable row at three points
in one ceremony. An administrator raising a client's default between two of those reads makes
`/auth/completed` stamp `acr: urn:goiabada:level2_mandatory` with `amr: ["pwd"]` on a password-only
ceremony.

**Mechanism confirmed, attribution refuted.** The reviewer reported it as introduced by the floor.
It is already on `main`. Executed against the real `GetTargetAcrLevel` and `SetAcrLevel`, not argued:

| Tree | `acr_values` | acr stamped on a password-only ceremony | |
|---|---|---|---|
| today | absent | `urn:goiabada:level2_mandatory` | **race, already on `main`** |
| today | `level1` | `urn:goiabada:level1` | safe |
| floor | absent | `urn:goiabada:level2_mandatory` | race, unchanged by this issue |
| floor | `level1` | `urn:goiabada:level2_mandatory` | **race, widened by the floor** |

The race is reachable today in the ordinary no-`acr_values` shape, which is nearly every real
request; the floor adds the minority carrying the parameter. In the opposite direction the floor is
strictly safer, because it bounds a mid-ceremony *lowering* that is unbounded today. The actor needs
`authserver:manage` and could already set the row to `level1` permanently, so the widening grants
nobody a capability they lacked.

**Escalated as decision 6 rather than settled by the run**, because a security finding the reviewer
does not force an answer on is a choice, not work. The run recommended filing it; the answer was to
build it, and it is stage 1's snapshot. See the decisions comment.

**2, significant, quality. The endpoint tests needed their own clamp-reversion mutation.** Both
endpoint cases would have passed with their `acr_values=level1` stimulus omitted, because each
fixture's own client default independently produces the asserted outcome, and the planned module
mutations run against a function that cannot observe whether the HTTP request carries the parameter.
Fixed in the plan: a third mutation, the same clamp reversion run at
`--type integration --db sqlite` over the two new case names, where a dropped stimulus makes the
mutation survive. This is why stage 1 reports three *wiring* mutations separately.

**3, significant, quality. Stage 2 is an independent defect annexed to this change. Withdrawn.**
Issue #240's own scope names the `amr` omission as its second bullet, and states the independence
the finding cited as evidence. The scope was set with that independence known.

**4, minor, conformance. The documentation sweep left claims the change falsifies.** Two residual
ACR sentences the plan had missed, both folded into the plan: `concepts/acr-amr.mdx`'s max_age table
row, and `concepts/authorization-lifecycle.mdx`'s final table. The plan had cleared the max_age
example *below* the table and never looked at the row above it.

---

## Final gate, round 1

Request `240-20260823T125813Z`. Scope: the whole branch against `fe543374`, four commits, nineteen
files, +748 / -93. Correctness, security, conformance and quality all `reviewed`; `unchecked` empty.

**What it ran**, recorded because zero-findings-per-axis means nothing without it: read the request,
the whole agreement, all three stage logs, the closing record, the probe sources with their outputs,
the full branch diff and every changed test and page; traced authorization validation, `prompt=none`,
interactive step-up, session reuse, OTP, completion, consent, code issuance, implicit issuance and
the signed token claims; verified the OIDC Core errata set 2 and RFC 9700 text bearing on
`acr_values`, `acr` and `amr`; reproduced all six split hashes, verified all 33 anchors, and
confirmed all 11 CI jobs green at `9bf6f0c9`; ran the core suite and four targeted sqlite
integration cases; and ran its own mutations rather than re-running ours.

**Its mutations.** Floor, snapshot-reader and both `amr` claim mutations were caught by the core
tests. The snapshot-**writer** mutation survived the core tests at both test boundaries and was
caught by the real cross-user integration path at the missing `/auth/level2` hop. The reviewer's own
reading, which the run initially misrecorded as a gap and then corrected: that "establishes where
the test family ends rather than leaving the surviving unit mutation unexplained". It is in `ran`,
not in `findings`.

**Finding 1, minor, conformance. Fixed in `d35c9b1f`.** Stage 3 had rewritten the `clients.mdx`
caution about a self-registered client missing its badge to read "Goiabada recognises one by the
`dcr_` prefix on its identifier". Runtime behaviour uses `clients.created_via_dcr`; the prefix was
only the one-time backfill's matching signal.

Verified before acting, and it stands. Four independent sites say the column is what everything
reads: `models/client.go` on `CreatedViaDCR`, `generateDCRClientIdentifier` ("the `dcr_` prefix is
cosmetic ... testing the prefix answers 'was this named like a self-registered client', not 'did
this client register itself'"), the badge template gating on `{{if .CreatedViaDCR}}`, and
`auth_redirect_blocked.go`. The prefix appears in exactly one runtime expression, the string that
builds a new identifier, plus migration 000029's `LIKE 'dcr!_%'` backfill.

The old wording had a practical cost, not just an inaccurate one: a reader following it literally
would hunt the condition by scanning identifiers, which finds administrator-created `dcr_*` clients
and **misses the renamed self-registered one the caution exists for**.

**Family: "upgrade-history documentation rewritten as current-state behaviour".** The reviewer
enumerated it properly and the enumeration is closed by construction, because decision 4 is a fixed
list of six edits. It checked all six against production behaviour and cleared five; only one
substituted a reconstruction rule for the live one. `mutations_run` empty, which is the honest
answer for a prose family.

Re-verified after the fix: anchors PASS on all 33 rows, `npm run build` green at 42 pages, and the
upgrade-wording recount still returns the same 6 intended survivors, so the fix reintroduced no
upgrade note.

---

## Final gate, round 2

Request `240-20260823T131918Z`. Five commits, nineteen files. Two findings, both minor conformance,
both `confidence: high`, both fixed in **`f3c26df6`**. `unchecked` empty. No production, unit-test,
data-test or integration-test byte moved.

**Finding 1. The consent default does not read the self-registration mark.** The caution round 1 had
just repaired said "the badge, the consent default and the refusal behaviour above all read that
mark". Verified by enumerating every non-test reader of each field:

- `CreatedViaDCR` is read by the badge template, `handler_authorize.go`'s refusal redirect,
  `client_display.go`'s unverified-name label, and the identifier-immutability guard. **None of them
  decides consent.**
- `ConsentRequired` is read by `handler_authorize.go`, `handler_auth_completed.go` and
  `token_validator.go`. **None of them reads the mark.**

Registration sets `ConsentRequired: true` and `CreatedViaDCR: true` as two separate assignments
three lines apart, and the clients API assigns `ConsentRequired` unconditionally. So the false clause
was worse than imprecise: it told an administrator that consent on a self-registered client is a
*consequence of the mark*, when unticking it is the documented remedy.

The second half of the claim is right for a different reason than the reviewer gave. The legacy
state is *known*, not merely editable: migration `000029` states every client that registered itself
before it was stamped `ConsentRequired: false`, and both relevant commits first appear in tag
`v1.6.0-rc1`, so no release ever stamped consent true without also carrying the column. What the old
wording actually missed is that an operator may have ticked it since, which makes "tick **Consent
required** by hand" wrong when the box is already ticked. The remediation now reads "if it is not
already ticked".

**Finding 2. The first *recognised* value wins, not the first value.** Three changed descriptions
said the first raw `acr_values` value wins. `parseAcrValuesFromAuthorizeRequest` appends only where
`enums.AcrLevelFromString` returns no error, and selection then takes element zero, so
`acr_values=nonsense urn:goiabada:level2_mandatory` selects `level2_mandatory`, the *second* raw
value. The change's own unit case was already named
`TestGetTargetAcrLevel_FirstRecognizedValueWins`, so the executable coverage was precise all along
and only the prose lagged.

| Where | Was | Now |
|---|---|---|
| `concepts/acr-amr.mdx`, multiple-values tip | "uses the first one in the list" | "uses the first one it recognises, ignoring any it does not" |
| `concepts/authorization-lifecycle.mdx`, selection step 1 | "take the first value" | "take the first value it recognises" |
| `concepts/authorization-lifecycle.mdx`, summary-table row | "the first value from the `acr_values` parameter" | "the first recognised value from the `acr_values` parameter" |

Anchors failed on two rows immediately after these edits, because both cite the very sentences that
changed; section 0 was re-swept and all 33 resolve again.

The reviewer ran seven fresh mutations this round covering ACR ordering and recognition, snapshot
storage, both AMR writers and consent-marker independence. **Every one caught.** Its note: "The
OAuth implementation, security behavior, and executable coverage are sound."

---

## Final gate, round 3

Request `240-20260823T134024Z`. One finding, minor, conformance. `unchecked` empty. Nothing refuted,
nothing deferred. The reviewer traced the ACR snapshot through validation, session serialization,
the silent and interactive consumers, session updates, code issuance, implicit issuance, refresh and
the token claim writers; checked every `CreatedViaDCR` reader and writer, consent configurability,
the migrations and the release history; verified all 33 anchors, the split hashes, a clean worktree
and green CI at `f3c26df6`; and ran four mutations covering the client-default fallback, snapshot
presence, access-token `amr` and id-token `amr`. **All four caught, all restored byte-identically.**

**Finding 1, fixed in `9647756b`. The repaired wording still gave no correct outcome when *every*
`acr_values` entry is unrecognised.** Production falls back to the client default; the lifecycle
algorithm and its table reserved that fallback for an *absent* parameter, and the ACR tip never
stated the zero-recognised outcome.

Verified rather than accepted: `AcrLevelFromString` is a three-arm switch over the three full URNs,
so recognition is exact-string and nothing else. With zero survivors the selector takes its
`len(...) > 0` branch false and returns the client default, which is the absent-parameter arm. So
production had a defined outcome for a case both pages left undefined.

**Resolved as the family, not the instance.** The reviewer's `siblings_checked` named three members;
the sweep found a fourth it did not: `concepts/acr-amr.mdx`'s `max_age` interaction example asked
for `acr_values=level1`, and bare `level1` is not one of the three URNs, so **the example was itself
a zero-recognised request riding the undocumented fallback while appearing to demonstrate a level1
request**. It now sends `urn:goiabada:level1`. Its conclusion was unaffected either way, which is
why no assertion and no other sentence moved with it.

---

## Final gate, round 4

Request `240-20260823T105600Z`, 13:58:32Z to 14:05:15Z. One finding, minor, conformance. `unchecked`
"None". Follow-ups: none. **This was the last round the budget funds** (`max_rounds: 4`).

It re-read the request, the agreement, the plan, all three stage logs, the closing log, all three
prior final-gate reviews, the preserved probes and the full branch diff; traced target snapshotting
and AMR emission through every production consumer and every changed test; reproduced all five split
hashes and all 33 anchors; and confirmed **all eleven CI jobs green at the exact reviewed HEAD
`9647756b`**, which is the four-engine data and integration result for every executable line in this
change. It ran no mutation and said why: the executable and test hashes were unchanged from round 3,
whose four mutations were all caught, so repeating them would add no evidence about the changed
wording. Its note: "The executable implementation, coverage, and security behavior are sound."

**Finding 1, fixed in `dcb2e959`.** Round 3's repair stated the correct zero-recognised fallback, but
the numbered algorithm in `concepts/authorization-lifecycle.mdx` still opened step 1 with "If
`acr_values` is in the authorization request" while step 2's `Otherwise` was now defined to include
a present parameter naming no recognised value. **A request carrying `acr_values=unknown` satisfied
both arms**, and the table further down the same page partitioned the same input correctly, so the
page held two incompatible versions of one algorithm.

The fix is one line, and it copies the partition wording from the table on the same page verbatim
rather than composing new prose: step 1 now reads "If `acr_values` names at least one recognised
value". Step 2 was already correct and did not move, and is now genuinely the otherwise arm.

**Swept as the family independently rather than taken on trust.** `grep -rln acr_values
site/src/content/docs/` gives six pages; a sweep for partitioning sentences across the whole docs
tree finds every member. The algorithm was the only overlap. The table was already correct and is
the wording the fix adopts. `concepts/acr-amr.mdx` states the absent case only and claims no
exhaustive partition, and its tip states the zero-recognised outcome explicitly. The other four
pages assert no selection order or branch partition.

### One thing about round 4 worth knowing before you read the diff

The budget funds no fifth reading, so **round 4's own repair is the one edit in this change that no
reviewer has seen.** Two things bound that, offered as reasons rather than reassurance: it changes a
single conditional clause instead of rewriting a paragraph, and it copies its wording from a
sentence on the same page that rounds 2, 3 and 4 each read and each confirmed correct. What it does
not have is an independent reading.

Issue #249, opened 2026-08-22
user_session_clients and user_consents have no unique constraint on their natural keys, so a concurrent write duplicates the row


leodip:
Found by the #239 authorization ceremony review (finding [F20](https://github.com/leodip/goiabada/issues/239#user-content-f20)).

Two tables, one shape: a read-then-insert with nothing serialising it, over a natural key the schema does not constrain.

## `user_session_clients(user_session_id, client_id)`

`BumpUserSession` loops the loaded clients, concludes the current one is absent, and appends. Two concurrent `/auth/completed` requests both conclude that and both insert.

## `user_consents(user_id, client_id)`

`HandleConsentPost` calls `GetConsentByUserIdAndClientId` and creates when it finds nothing. Two concurrent submissions of the same consent form both find nothing and both call `CreateUserConsent`.

**The lasting consequence is on the read side.** `GetConsentByUserIdAndClientId` builds a plain `WHERE user_id = ? AND client_id = ?` with no `ORDER BY` and no `LIMIT`, and `getUserConsentCommon` takes the first row the driver yields. With two rows carrying different scopes, **which one governs is whatever the engine happens to return**, and it can differ between the consent screen, `handlePromptNone`'s consent check, and the account consents page. A later `UpdateUserConsent` then edits one row and leaves the other stale, which is also the territory #115 is in.

## Verification notes

Both were checked against the **migrations** rather than the `schema.sql` snapshots, since the snapshots are documentation only, and across all four engines. `user_sessions` does carry `UNIQUE (session_identifier)`, so the omission is specific to these two tables rather than a project-wide habit.

## Fix

A unique index on each natural key across all four engines, which turns a silent duplicate into a caught error, plus an upsert or a retry at the two insert sites.

**Both tables need a duplicate sweep in the same migration**, because existing deployments may already hold pairs. For `user_consents` the sweep has to decide which row survives; the most recently granted is the defensible choice, and #115 is relevant because `granted_at` is not refreshed on update today.

### Practical notes for whoever writes it

- **This is a migration, so it costs more than its Low severity suggests.** Four engines, a data-layer test each, and the sweep.
- Next available migration number is **000030** (verify with `ls src/core/data/sqlitedb/migrations | tail` rather than trusting this line).
- Only MySQL declares indexes inline; PostgreSQL, SQLite and SQL Server use separate `CREATE UNIQUE INDEX` statements. Reading the `CREATE TABLE` block alone will find a constraint on one engine and miss it on three.
[...]

Issue #54, opened 2026-01-23
Return Email Verification Code


Joebeurg:
When using API to create an acount while the instance settings do not allow Self Registration nor the OIDC Registration, the API does not have an option to return the Verification Code when the user account has been created, let'ssay I am using my own methods to send emails with custom templates hosted on an external email provider without relying or activating the Goiabada SMTP config.
I could not find a way to return the Verification Code when an account is created, or even when the email has been updated.
The response for user creation is as this based on the docs and API test:
```json
{
  "email": "newuser@example.com",
  "emailVerified": false,
  "givenName": "New",
  "familyName": "User",
  "setPasswordType": "now",
  "password": "SecurePassword123!"
}
```

Maybe add a new optional property "returnCode" in the request that would return the code in the response JSON body.
I am guessing this is the generated code line: [L82](https://github.com/leodip/goiabada/blob/c0882e7817184662928a976103db6b74e47f2b09/src/authserver/internal/handlers/apihandlers/handler_api_account_email_verification.go#L82)

Issue #43, opened 2025-11-29
Feature: Implement granular REST API scopes for fine-grained authorization


leodip:
## Summary

Currently the REST API uses a coarse-grained "all or nothing" authorization model with just 3 scopes. This limits the ability to delegate specific administrative functions without granting full access.

## Current State

| Scope | Access |
|-------|--------|
| `authserver:manage` | Full access to ALL 70+ admin endpoints |
| `authserver:manage-account` | User self-service (own account) |
| `authserver:userinfo` | OIDC userinfo endpoint |

**Problem:** Granting `authserver:manage` allows the client/user to perform any admin operation - there's no way to grant read-only access or limit to specific resource types.

### New Admin Scopes

| Scope | Access |
|-------|--------|
| `authserver:admin:read` | Read-only access to all admin endpoints (users, groups, clients, settings, keys) |
| `authserver:users` | Full access to users, groups, permissions, sessions |
| `authserver:clients` | Full access to clients and their configuration |
| `authserver:settings` | Full access to system settings and signing keys |
| `authserver:manage` | Full admin access (backwards compatible, aggregate of all above) |

### Unchanged Scopes

| Scope | Access |
|-------|--------|
| `authserver:manage-account` | Self-service account management |
| `authserver:userinfo` | OIDC userinfo endpoint |

## Use Cases

| Use Case | Scopes Needed |
|----------|---------------|
| Read-only admin dashboard / audit | `authserver:admin:read` |
| User support team | `authserver:users` |
| Client registration service | `authserver:clients` |
| Ops/security team | `authserver:settings` |
| Full admin (legacy behavior) | `authserver:manage` |

### `authserver:admin:read`
All GET endpoints under `/api/v1/admin/*`

### `authserver:users`
- `/api/v1/admin/users/*` - User CRUD, search
- `/api/v1/admin/user-attributes/*` - User attributes
- `/api/v1/admin/user-sessions/*` - User sessions
- `/api/v1/admin/user-consents/*` - User consents
- `/api/v1/admin/groups/*` - Group CRUD, membership
- `/api/v1/admin/group-attributes/*` - Group attributes
- `/api/v1/admin/permissions/*/users` - Permission-user mappings
- `/api/v1/admin/phone-countries` - Reference data

### `authserver:clients`
- `/api/v1/admin/clients/*` - Client CRUD
- `/api/v1/admin/clients/*/sessions` - Client sessions
- `/api/v1/admin/clients/*/authentication` - Client auth settings
[...]

PR #254, opened 2026-08-23
fix(core): rotate signing keys in one transaction, refusing before anything is destroyed

### LINKED ISSUE #251: Outside the ceremony: signing-key rotation is four unsynchronised writes and can leave the server with no current key
Found by the #239 authorization ceremony review (finding [F40](https://github.com/leodip/goiabada/issues/239#user-content-f40)). **Well outside the ceremony**, filed separately so it can be scheduled on its own terms. The ceremony does depend on it (`/auth/issue` cannot mint anything without a current signing key), but this is administrative key lifecycle.

## The sequence

`HandleAPISettingsKeysRotatePost` reads all key rows with a nil transaction, then performs four separate writes with **no transaction, no locking, and no constraint standing behind them**:

1. `DeleteKeyPair(nil, previousKey.Id)`
2. Guard: `if currentKey == nil || nextKey == nil` return 500
3. `currentKey.State = previous`, `UpdateKeyPair`
4. `nextKey.State = current`, `UpdateKeyPair`
5. Generate a 4096-bit RSA key, encrypt it, marshal the JWK, `CreateKeyPair` as the new next

Four things are wrong with that, and the last turns a robustness note into a real hazard.

### The guard runs after the delete

Step 2 is the prerequisite check and step 1 already ran. A deployment already in a broken state has its **remaining previous key destroyed** and is then told the rotation cannot proceed.

### Any failure between steps 3 and 5 is terminal

A failure at step 4 leaves no row in the `current` state. A failure at step 5 leaves no `next`, and the next rotation attempt is refused by the guard at step 2, having first deleted the new previous.

### No current key is a nil dereference, not a graceful failure

`GetCurrentSigningKey` follows the house convention and returns `(nil, nil)` when no row matches. `GenerateTokenResponseForAuthCode` checks only the error:

```go
keyPair, err := t.database.GetCurrentSigningKey(nil)
if err != nil {
	return nil, err
}
privKey, err := keyPair.ParsePrivateKey()
```

So the first token request after a half-completed rotation **panics**. Recovery needs direct database surgery, and the bearer-authenticated admin API depends on token validation, so the operator may not be able to fix it through the product.

### The concurrency window is seconds wide, not microseconds

Generating a 4096-bit RSA key sits **between step 4 and step 5** and takes a noticeable fraction of a second to several seconds. A second rotation arriving in that window reads: the new current (from request A), the new previous (A's old current), and **no next**, because A has not inserted it yet. It then deletes the previous, which is A's old current, **the key that signed every token still in flight**, and only afterwards hits the guard and refuses.

So request A's rotation completes successfully while request B has silently destroyed the key A had just demoted for exactly the grace period rotation exists to provide. Every token signed by it stops validating, and `/certs` stops publishing it, so third-party validators reject it too.

**A double-clicked "Rotate" button is enough.**

`key_pairs` has `CREATE INDEX idx_state ON key_pairs(state)`, a plain index, so nothing at the database level prevents two `current` rows or two `next` rows either.

## Fix

- **Generate the replacement key before touching any state**, so the slow operation is outside the critical section.
- Perform the whole transition **in one transaction** taken with a locking read (`SELECT ... FOR UPDATE`, or the engine equivalent the data layer already abstracts elsewhere).
- **Add unique constraints** so at most one row can hold `current` and at most one `next`, which turns the concurrency bug into a caught error rather than a corrupted state. This is a migration, so four engines and data-layer tests.
- **Move the prerequisite check above the delete.**
- **Add a nil check at every `GetCurrentSigningKey` call site**, because a missing current key should be a 500 with a clear message rather than a panic.

## Tests

- Two concurrent rotations: one succeeds, the other is refused, and the previous key survives.
- A failure injected at each of steps 3, 4 and 5 leaves the key set in a state the next rotation can recover from.
- A rotation attempted with no `next` row refuses **without** deleting the previous key.
- `GetCurrentSigningKey` returning nil produces a 500 with a clear message, not a panic, at every call site.
- Data-tier tests for the new unique constraints across all four engines.

## Related

- **#249** is the other missing-unique-constraint issue from the same review. Same class of defect, different tables; whoever writes one migration will be comfortable writing the other.
- **#211** (`GenerateSecurityRandomString` swallows CSPRNG failure) is the neighbouring "a failure in key material handling cannot be propagated" defect.

leodip:
Fixes #251.

## What this is for

Rotating a signing key is five statements in one handler, each with its own connection and none of
them in a transaction. Two of them are destructive and they run in the wrong order: the existing
`previous` key is **deleted before** the check that a `current` and a `next` key exist at all, so a
deployment missing its `next` row loses the key that signed every token still in flight and is then
refused. Between the promotion of `next` to `current` and the insert of a replacement sits a
4096-bit RSA key generation, seconds wide, and a failure anywhere in the sequence commits whatever
ran before it. A failure at the promotion leaves the deployment with **no row in the `current`
state**, which eight call sites dereference without a nil check, so the server panics on every token
it tries to sign or validate. The admin API is bearer-authenticated and one of those sites is on the
token-validation path, so an operator in that state cannot use the product to repair it.

OpenID Connect Core 1.0 §10.1.1 says the JWK Set at `jwks_uri` "SHOULD retain recently decommissioned
signing keys for a reasonable period of time to facilitate a smooth transition". The `previous` state
exists to satisfy that, and the defect destroys the very key it is retaining.

## The rule this establishes

**Every refusal happens before the transaction commits, so a losing rotation's delete of the previous
key rolls back with it.** The whole transition runs in one transaction, each state change written as
a conditional UPDATE that asserts the state it read and reports through `RowsAffected()` whether it
was the call that made the transition. A second concurrent rotation either succeeds after the first
or is refused having written nothing. The prerequisite check runs before any write. Key generation
moves outside the transaction, so the slow operation is no longer in the critical section.

Alongside it, `GetCurrentSigningKey` returns an error rather than `(nil, nil)` when no key is
current. All eight call sites already check `err` and return, so all eight stop panicking without
being touched, and a ninth cannot be added unguarded.

### Stage 1: the compare-and-set primitive and the current-key contract
`0dc1af61`

Two data-layer primitives, both additive. Nothing calls either one yet.

`UpdateKeyPairState(tx, keyPairId, fromState, toState) (bool, error)` moves one key from an
expected state to a new one and reports whether this call made the transition. It follows
`MarkCodeAsUsed` statement for statement: a conditional `UPDATE` whose `WHERE` carries the id
**and** the from-state, with `RowsAffected() == 1` as the answer. That predicate is the whole
point. Without it two concurrent rotations both write the snapshot they read, and the loser
deletes the previous key the winner had just demoted, retiring every token that key had signed.
A `false` return is not an error, so it leaves the caller's transaction usable, which is what
will let the losing rotation roll its own delete back in stage 2. A zero id is an error rather
than an unfiltered update. Implemented once in `commondb`, with a one-line delegating wrapper in
each of the four dialects, and the `Database` mock regenerated.

`GetCurrentSigningKey` now returns an error when no key is in the `current` state, instead of the
`(nil, nil)` this codebase returns for a lookup that may legitimately miss. All eight production
call sites dereference the result to read key material, so a nil pair was a nil-pointer panic at
each of them rather than a diagnosable failure, and a ninth site could be added unguarded. All
eight already check `err` and return, so all eight became correct without being edited; each was
re-read to confirm none dereferences first, and the 39 mock stubs across the suite were checked
to return a key pair or an error, never `(nil, nil)`.

**Tiers:** unit 3563 pass / 0 fail. Data 1227 pass / 0 fail on **all four engines** locally
(mysql, postgres, mssql, sqlite). Integration is not applicable: no handler, route or schema
moved and nothing calls the new code, so there is no endpoint through which to reach it.

**Mutations:** five, all caught. The one worth naming serialises the new concurrency case's two
goroutines, and the case fails as it should, because the loser then returns in 215µs against the
100ms it requires. Without that check, "exactly one winner" is satisfied by two calls run in
sequence, and the concurrency case would prove nothing about contention.

### Stage 2: the rotator
`05243dd8`

`oauth.SigningKeyRotator` in `src/core/oauth/signing_key_rotator.go`, holding the whole transition
in one transaction. Nothing calls it yet; stage 3 rewires the endpoint onto it.

`Rotate()` generates the replacement key, opens a transaction, reads and classifies the key set
inside it, refuses if `current` or `next` is missing, deletes the previous row, compare-and-sets
the demotion and then the promotion, inserts the new `next`, and commits. **Every refusal happens
before the commit**, so a rotation that cannot proceed leaves the key set exactly as it found it.
That is the whole defect: the delete used to be committed the moment it ran and the refusal
arrived afterwards.

The guard now runs before any write. Reversed, as it was, a deployment with no `next` key lost its
`previous` key and was then refused anyway. That key signed every token still in flight: `/certs`
publishes it and the token parser falls back to it, so deleting it retires those tokens at once,
which is what OIDC Core 10.1.1 asks a JWK Set not to do when it says recently decommissioned keys
SHOULD be retained for a smooth transition.

Two more orderings are load bearing and each is commented at the line it governs. The key material
is generated **before** the transaction opens, because a 4096-bit generation is the slow step by
three orders of magnitude and holding a transaction across it is what made the window wide enough
to hit. The `previous` row is deleted **before** the current one is demoted, since demoting first
puts two rows in the `previous` state within one statement, which stage 4's constraint will refuse
on every engine. Any error returns at once rather than continuing to the next statement: on
postgres a failed statement aborts the whole transaction and every later command in it is refused
with SQLSTATE 25P02. A compare-and-set reporting no rows is not an error, which is why the two
refusals can still roll back cleanly.

Two sentinels, `ErrRotationInProgress` and `ErrKeySetIncomplete`, are what stage 3 maps to a 409
and a 500. Anything else is a genuine engine or crypto failure.

**Tiers:** unit 3582 pass / 0 fail. Data 1655 pass / 0 fail on **all four engines** locally
(mysql, postgres, mssql, sqlite), covering one full rotation against real SQL and the issue's own
"refuses without deleting the previous key" case. Integration is not applicable: nothing routes to
the rotator yet, so no endpoint can reach it.

**Mutations:** six, all caught. The one worth naming reintroduces the defect verbatim by moving
the guard back below the delete; the guard cases fail on it because each asserts the delete never
ran, not merely that the refusal happened. Another moves key generation inside the transaction,
caught by a case that makes generation fail deterministically (`crypto/rsa` refuses a 512-bit key)
and then asserts the transaction was never opened.

**Not asserted, deliberately:** two goroutines calling `Rotate()` concurrently. Generation now runs
before the transaction and takes about 300ms with wide variance, so the two may not overlap, and
when they do not both rotations succeed, which is correct behaviour that would fail such a test.
The race is covered where it is deterministic, at stage 1's data-tier concurrency case, and both
losing branches are driven directly at the unit tier.

### Stage 3: the handler, the admin console, the documented refusal

Landed in `774b3223`.

`HandleAPISettingsKeysRotatePost` loses the five unsynchronised writes and becomes a call to
`oauth.SigningKeyRotator.Rotate()` plus a four-row mapping: `nil` audits and answers 200 as
before, `ErrRotationInProgress` is **409 `ROTATION_IN_PROGRESS`**, `ErrKeySetIncomplete` is 500
`KEY_SET_INCOMPLETE`, anything else is 500 `INTERNAL_ERROR`. 409 rather than 200 for the loser of
a race because it rotated nothing: reporting success would announce one rotation twice and invite
a retry that burns a 4096-bit key generation and rotates a second time. The signature is unchanged
and the rotator is constructed in the factory body, so `routes.go` is untouched.

The admin console makes that legible. `HandleAPIErrorJson` forwards 409 alongside 400, and the
rotate handler routes through it instead of calling `JsonError` directly, so the operator reads
"Another key rotation is in progress" rather than "An unexpected server error has occurred" with
a request id. Three lines of behaviour, no new catalog strings: the API's English description is
surfaced verbatim, which is the position #122 already settled for that helper.

Both documentation surfaces name the 409. `site/src/content/docs/integration/rest-api.mdx` gains a
line saying not to retry it; `src/authserver/web/openapi.yaml` gains a `'409'` on
`rotateSettingsKeys`, which declared only 200 and 401. That file is embedded and served at
`GET /openapi.yaml`, so it is a published contract and a generated client would not have known the
status exists.

**Tests.** The endpoint had no unit coverage at all before this, which is a large part of how the
defect survived since v0.7. `handler_api_settings_keys_test.go` adds 4 cases driving the real
rotator over a mocked database: the mapping, the audit on success and only on success, and by
omission that neither refusal reaches `DeleteKeyPair`, `CreateKeyPair` or `CommitTransaction`, the
last of which is the old defect stated as a test. `api_error_helper_test.go` adds 3 cases and is
the first behavioural test in `src/adminconsole/internal/handlers`; it pins the 400 forwarding
alongside the new 409, so widening that condition cannot silently replace #122's behaviour rather
than join it.

Locally: unit 3589 pass, data 1655 pass on all four engines, integration 1268 pass on sqlite,
0 failures. Five mutations against the new tests, all caught, including `case err == nil:` →
`case true:`, which is what proves the audit entry is written on success and nowhere else.

One deviation: `src/adminconsole/go.mod` gains `github.com/stretchr/objx // indirect`. Using
`testify/mock` in that module for the first time requires it. `go.sum` already carried it and did
not change, so nothing new entered the dependency graph.

### Stage 4: the unique constraint

Landed in `ac4b168c`. Schema only: no production Go changed, because the Go that has to respect the
constraint already did as of stage 2.

Migration **000030** on all four engines adds `idx_key_pairs_state`, a plain `UNIQUE (state)` on
`key_pairs`, preceded by a sweep that keeps the highest `id` per state. Every reader already assumed
one row per state and nothing enforced it: `GetCurrentSigningKey` is a bare `WHERE state = 'current'`
with no `ORDER BY` and no `LIMIT`, so with two `current` rows it returns whichever the driver yields
first; `/certs` assigns `previousKey` in a loop and publishes only the last duplicate it sees; the
admin console offers to revoke "the previous key", singular. This is what makes that single row
provable rather than arbitrary, which is the same defect #249 describes on two other tables.

The sweep exists because existing deployments may already carry duplicates: until stage 2, two
concurrent rotations could both insert a new `next` row. `MAX(id)` is the newest on every engine
(sqlite AUTOINCREMENT, mysql AUTO_INCREMENT, postgres sequence, mssql IDENTITY). The derived table in
the `DELETE` is required for MySQL, which otherwise refuses a subquery naming the table being deleted
from. Sweeping rather than failing on a duplicate is deliberate: a failed migration stops startup and
leaves the version dirty, and a deployment that cannot start cannot be repaired through the product.
Extra `next` rows are the only duplicate the defect can actually produce, and a `next` key has signed
nothing, so in the reachable case the sweep destroys nothing anyone holds a token from. The down
migration drops the index and says in the file that it cannot restore the swept rows. The four
`schema.sql` snapshots are updated in the same commit.

**Tests.** A new migration test runs against an isolated database per dialect: seed 2 `current`, 3
`next` and 2 `previous` at 000029, apply 000030, and assert one row survives per state and that it is
the **highest** id of that state, which is what says `MAX(id)` rather than `MIN(id)`. The index's
shape is read from each engine's catalog through the existing `describeIndex` helper (unique,
covering exactly `[state]`), with `idx_state` used as a polarity control because MySQL reports
uniqueness inverted and an assertion of "unique" would otherwise pass on an engine whose flag was
being read backwards. It then attempts a duplicate insert, because the catalog reporting a shape and
the engine enforcing it are two different claims, and exercises the down/up round trip.
`unique_constraint_test.go` gains `TestUnique_KeyPairState`, taking that file from eleven constraints
to twelve.

Three existing tests are brought under the constraint. `createTestKeyPair` stamped every row
`current` and could no longer be called twice, so it collapses into `createKeyPairInState`, which
clears the destination first; `TestGetAllSigningKeys` held two `current` rows and now holds one
`current` and one `next`, still exercising what the method is for; `TestUpdateKeyPair` moved its
fixture into `previous` without clearing it, which the index refuses whenever a `previous` row
survives in the never-dropped shared test database, and it now clears first, matching the ordering
production observes.

Locally: unit 3589 pass, data 1243 pass on all four engines, integration 957 pass on sqlite, 0
failures. Stage 1's and stage 2's data-tier cases were re-run under the index, since this is the
first moment it could falsify them, and none needed a change. Four mutations of the migration files
were all caught: `MAX` to `MIN`, `UNIQUE` dropped, the sweep neutered, and the down migration
dropping the wrong index.

### Stage 5: the spec reconciled with the handlers, and the lint that keeps it there

Landed in `97cb74df`. The follow-up this run left behind, and everything the same sweep turned up
beside it. No production Go changed: this is the published contract catching up with the code, plus
the instrument that stops it falling behind again.

`src/authserver/web/openapi.yaml` is embedded by `web/embed_fs.go` and served at `GET /openapi.yaml`,
so it is what a caller generates a client from rather than a README. Nothing held it to what the
handlers do. Reconciling its 96 operations against the 106 routes `routes.go` registers moved the
declared statuses like this:

| Status | Before | After | Why |
|---|---|---|---|
| `500` | 0 | 96 | every handler can fail on storage and none said so |
| `400` | 44 | 81 | malformed ids and validation paths |
| `404` | 59 | 72 | 15 added, 2 removed |
| `409` | 1 | 2 | the follow-up: `createUser` |
| `403` | 0 | 2 | the two account ownership checks |
| `201` | 6 | 7 | `addGroupMember` was documented as 200 |

The follow-up itself is the `409` row. `HandleAPIUserCreatePost` answers a duplicate email with 409
`EMAIL_ALREADY_EXISTS` at two sites, one for the email field and one for the derived username path,
and `createUser` declared 201, 400 and 401 only, so the single failure a caller creating users hits
routinely reached a generated client as an unmodelled error. It is now written out in place, naming
the code, alongside a line in `site/src/content/docs/integration/rest-api.mdx`.

Four things the sweep found are worth naming on their own.

**`addGroupMember` answers 201 and the spec said 200.** A client checking for 200 read a successful
add as a failure. The spec follows the handler, since the handler is what is deployed.

**Two operations declared a 404 no handler can produce.** `deleteSettingsKey` answers 400 with
`VALIDATION_ERROR` for an id matching no key, and `getResourcePermissions` never looks its resource
up at all, so an unknown id comes back as 200 with an empty list. Both now say so where it happens
rather than promising a branch that never runs. That the second is arguably a handler defect is not
this change's to settle: the spec's job is to describe what the server does.

**Three 500s are not the JSON envelope.** `getUserConsents`, `deleteUserConsent` and
`getClientLogoImage` fail through `httpHelper.InternalServerError`, which renders the shared HTML
error page. Documenting them with the JSON `ErrorResponse` the other 93 use would have been a false
statement, so those three are written out with `text/html` and say why.

**`rotateSettingsKeys` gained its own 500.** Stage 3 documented this endpoint's 409 and not the 500
beside it. It now names `KEY_SET_INCOMPLETE`, which is an operator problem a retry will not clear,
apart from a transient `INTERNAL_ERROR`, and says nothing was rotated either way.

**The lint.** `src/authserver/internal/server/openapi_contract_lint_test.go` parses `routes.go`, the
non-test handler sources and the embedded spec, and holds them to each other in both directions:
every status a handler can write is declared, every failure declared is one a handler can write, 429
appears exactly where a rate limiter does and nowhere else, and every route under `/api/v1` is either
documented or named in `knownUndocumentedRoutes` with the reason it is not. Eleven routes are on that
list, all needing schemas the document does not have yet (image bytes, multipart uploads, the audit
log query surface); the list is exact in both directions, so documenting one forces its line to go.
Two further tests hold up the ground the first three stand on: that `writeJSONError` is always passed
an `http.Status` constant, without which the lexical scan would have a blind spot it could not see,
and that every `$ref` into `components/responses` resolves and every shared response is used.

The scan is lexical, and the file says so at the top along with what that costs: it reads the
statuses a handler mentions rather than the ones it can be driven to return, so an unreachable branch
is still documented. That is the conservative direction, since a caller told about a failure that
cannot happen loses nothing.

**Tiers:** modules 1102 pass / 0 fail. Data 1247 pass / 0 fail on **all four engines** locally.
Integration 957 pass / 0 fail on sqlite. The data tier is not affected by this stage and was run
anyway, since the stage is the first to touch a file the whole suite embeds.

**Mutations:** sixteen, fourteen caught. The two that survived are accounted for rather than
excused. Dropping the `writeValidationError` to 400 rule changes no result today, because every
handler that calls it also names `http.StatusBadRequest` somewhere else; the rule is kept for the
first handler that validates and does nothing else, and the comment at the line says exactly this.
Restricting the forward check to failures also survives on its own, because the 2xx defect it would
hide is already fixed; running that mutation together with reverting `addGroupMember` to 200 lets
that defect through, which is what proves the branch is load bearing. The fourteen caught include
reintroducing the phantom 404, misspelling a shared `$ref` target, mangling the route prefix,
pointing the handler walk at a tree with no handlers, and removing each of the three extraction
rules that let the scan see a status no constant is written for.

## Tests

Four tiers, and it matters which of them ran where.

**Local, at the final tree `97cb74df`:** `--type modules` 1102 passed / 0 failed, the data tier 1247
passed / 0 failed across **sqlite, mysql, postgres and mssql**, and integration 957 passed / 0 failed
on sqlite. The module count moved from 1097 by the five tests stage 5 adds. Per-stage counts are in
each stage's section above; they were re-run at every gate rather than once at the end.

**Integration ran on sqlite locally**, at the stages that changed a handler, route or schema. Stages
1, 2 and the review's fix round did not, and each says so with its reason: nothing routed to the new
code yet, or only test files moved, so there was no endpoint through which to reach it.

**The four-engine data and integration result is the check suite's**, not a local claim. Green on
`97cb74df`, every job:
[run 32659461759](https://github.com/leodip/goiabada/actions/runs/32659461759), and green on
`aed37751` before stage 5 landed:
[run 32656041739](https://github.com/leodip/goiabada/actions/runs/32656041739). That covers
`Tests / {sqlite,mysql,postgres,mssql}`, `Unit / {core,authserver,adminconsole}`, `Build validation`,
`Lint`, `Vulnerabilities` and `Versions`, the last three of which nothing local runs. Every stage's
suite concluded green on its own commit as well.

**Thirty-six mutations were run against the new tests across the five stages, and thirty-four were
caught.** The ones worth naming: reintroducing the defect verbatim by moving the prerequisite guard
back below the delete; moving key generation inside the transaction; `MAX(id)` to `MIN(id)` in the
sweep; and `case err == nil:` to `case true:`, which is what proves the audit entry is written on a
successful rotation and nowhere else. The two survivors are both stage 5's and both are explained in
its section rather than left as a gap. A test that cannot fail transfers confidence it has not
earned, and this is the evidence these can.

### Review

Reviewed by `gpt-5.6-sol` at effort `xhigh`, 2 rounds over the whole branch, against a maximum of 4.
No stage was reviewed on its own: the whole change was read once, after the last stage landed.

|---|---|---|---|
| 1 | 2, both significant | 0 | both fixed, `aed37751` |
| 2 | 0 | 0 | **clean** |

- **significant, quality:** the admin console suite did not prove the rotate handler sends its 409
  through `HandleAPIErrorJson`; swapping it back for a direct `JsonError` left the suite green, which
  would have put `ROTATION_IN_PROGRESS` back behind a generic 500. Fixed in `aed37751`: new
  `handler_admin_settings_keys_test.go`, asserting through the real encoder over four statuses. The
  mutation survived before and is caught after.
- **significant, quality:** none of the seven compare-and-set data cases made the statement itself
  fail, so swallowing an `ExecSql` error as `false, nil` survived them all, which would report a
  storage failure to the operator as "another rotation is in progress" — a rotation that never
  happened, described as one that did. Fixed in `aed37751`:
  `TestUpdateKeyPairState_StorageFailureIsAnError`, deterministic on all four engines via
  `sql.ErrTxDone`. The mutation survived before and is caught after.

Nothing was raised on conformance or security in either round, and no finding contested any of the
eight settled decisions. Round 2 ran five mutations of its own and caught all five, which is what
separates a clean verdict from an unexamined one.

Full account, with the evidence, the `unchecked` lists and the tree hashes, in
[this comment](https://github.com/leodip/goiabada/pull/254#issuecomment-5387635844).

## Deviations

One, and it is a build requirement rather than a design change. `src/adminconsole/go.mod` gained
`github.com/stretchr/objx v0.5.3 // indirect`: using `testify/mock` in that module for the first time
requires it. `go.sum` already carried it and did not change, so nothing new entered the dependency
graph. Without the line `go vet` and `go test` in that module refuse with "updates to go.mod needed".

Stage 5 brings a second of the same kind. `src/authserver/go.mod` gained `gopkg.in/yaml.v3 v3.0.1`
as a direct dependency: the lint parses the embedded spec with it rather than with a regular
expression over the document's indentation. It was already in the file as indirect and `go.sum` did
not change, so nothing new entered the dependency graph there either.

Everything else landed as agreed. No decision was opened by the run, no stage deviated from its plan,
and no count the plan asked to be recounted disagreed with the agreement.

## For the release notes

**Migration 000030 deletes duplicate `key_pairs` rows, and cannot restore them.** Before applying the
`UNIQUE (state)` index, it sweeps `key_pairs` down to the highest `id` per state. A deployment that
was bitten by this defect may carry extras, because until this change two concurrent rotations could
both insert a new `next` row. The sweep keeps the newest of each state on every engine, and the only
duplicate the defect can actually produce is an extra `next` key, which has signed nothing, so in the
reachable case nothing anyone holds a token from is destroyed. Sweeping rather than failing is
deliberate: a failed migration stops startup and leaves the version dirty, and a deployment that
cannot start cannot be repaired through the product. The down migration drops the index and says in
the file that it cannot bring the swept rows back.

**`POST /api/v1/admin/settings/keys/rotate` can now answer 409.** It previously answered 200 or 500.
A caller that loses a race with a concurrent rotation now gets 409 `ROTATION_IN_PROGRESS`, and
**should not retry**: the rotation it conflicted with already happened, so a retry burns a 4096-bit
key generation and rotates a second time. Existing scripts and generated clients that treat any
non-2xx as a transient failure worth retrying are the ones to check.

**`POST /api/v1/admin/groups/{id}/members` has always answered 201, and `openapi.yaml` said 200.**
The server's behaviour is unchanged; the document was wrong. A generated client or a script built
from the old spec that treats anything other than 200 as a failure has been reading a successful add
as one, and is the thing to check.

**The spec now declares the failures it always had.** 500 on every operation, 400 and 404 where the
handlers answer them, 403 on the two account endpoints that check ownership, and 409
`EMAIL_ALREADY_EXISTS` on `POST /api/v1/admin/users/create`. No behaviour changed with any of them:
these are statuses the server already returned and the contract did not mention, so a regenerated
client gains branches rather than losing any.

Nothing else asks an operator to do anything. Atomicity, the transaction and the constraint are
invisible to any caller that is not racing itself.

## The closing comments

- [Decisions](https://github.com/leodip/goiabada/pull/254#issuecomment-5387629707) — all eight
  settled before the run, none deferred, none raised by the run.
- [Follow-ups](https://github.com/leodip/goiabada/pull/254#issuecomment-5387631470) — one was
  drafted, and it was solved here in stage 5 rather than filed. Two candidates were considered and
  rejected on purpose.
- [The code review](https://github.com/leodip/goiabada/pull/254#issuecomment-5387635844) — both
  rounds in full, every finding with its evidence and disposition, and all nine `unchecked` entries
  answered.

## Status

**Complete.** All five stages landed, the check suite is green on `97cb74df`, and stages 1 to 4 were
reviewed over two rounds ending clean. The defect is fixed: every refusal now happens before the
transaction commits, so a losing rotation's delete of the previous key rolls back with it, the
prerequisite check runs before any write, and the invariant the transition relies on is enforced by
the database rather than only by its one writer. `GetCurrentSigningKey` returns an error where it
used to return a nil pair into eight unguarded dereferences.

Stage 5 landed after that review and has not been through it. It changes no production Go: the
published contract now names every status the handlers answer, and a lint holds the two together.

Ready for review.

### DIFF
--- a/site/src/content/docs/integration/rest-api.mdx
+++ b/site/src/content/docs/integration/rest-api.mdx
@@ -1235,6 +1235,8 @@ POST /api/v1/admin/settings/keys/rotate
 
 Rotates to a new signing key. The old key remains available for validation.
 
+Answers `409` with error code `ROTATION_IN_PROGRESS` when another rotation won the race. Do not retry: that rotation already happened, and a retry rotates a second time.
+
 ```
 DELETE /api/v1/admin/settings/keys/{id}
 ```
--- a/site/src/content/docs/integration/rest-api.mdx
+++ b/site/src/content/docs/integration/rest-api.mdx
@@ -179,6 +179,7 @@ Common `error_code` categories:
 - `INVALID_REQUEST_BODY` - Malformed JSON in request
 - `VALIDATION_ERROR` - Input validation failed
 - `NOT_FOUND` - Resource not found
+- `FORBIDDEN` - The token is valid but the resource belongs to another user
 - `TOO_MANY_REQUESTS` - Rate limit reached; the response carries a `Retry-After` header
 - `INTERNAL_SERVER_ERROR` - Server-side error
 
@@ -274,6 +275,8 @@ POST /api/v1/admin/users/create
 | <span style="white-space: nowrap;">`familyName`</span> | string | No | Last name |
 | <span style="white-space: nowrap;">`setPasswordType`</span> | string | No | `"now"` to set password immediately, `"email"` to send setup email |
 | <span style="white-space: nowrap;">`password`</span> | string | Conditional | Required if `setPasswordType` is `"now"` |
+
+Answers `409` with error code `EMAIL_ALREADY_EXISTS` when the address is already registered. Retrying the same address will not succeed.
 
 #### Update user profile
 
--- a/src/adminconsole/go.mod
+++ b/src/adminconsole/go.mod
@@ -51,6 +51,7 @@ require (
 	github.com/pmezard/go-difflib v1.0.1-0.20181226105442-5d4384ee4fb2 // indirect
 	github.com/remyoudompheng/bigfft v0.0.0-20230129092748-24d4a6f8daec // indirect
 	github.com/shopspring/decimal v1.4.0 // indirect
+	github.com/stretchr/objx v0.5.3 // indirect
 	github.com/sym01/htmlsanitizer v1.1.1 // indirect
 	github.com/toorop/go-dkim v0.0.0-20250226130143-9025cce95817 // indirect
 	github.com/xhit/go-simple-mail/v2 v2.16.0 // indirect
--- a/src/adminconsole/internal/handlers/adminsettingshandlers/handler_admin_settings_keys.go
+++ b/src/adminconsole/internal/handlers/adminsettingshandlers/handler_admin_settings_keys.go
@@ -99,7 +99,12 @@ func HandleAdminSettingsKeysRotatePost(
 		}
 
 		if err := apiClient.RotateSettingsKeys(jwtInfo.TokenResponse.AccessToken); err != nil {
-			httpHelper.JsonError(w, r, err)
+			// Not JsonError directly: the API answers 409 when another rotation won the race,
+			// and JsonError's generic branch would show the administrator "An unexpected
+			// server error has occurred" with a request id, for something neither unexpected
+			// nor a server error. HandleAPIErrorJson forwards the API's description instead,
+			// so the modal reads "Another key rotation is in progress" (#251).
+			handlers.HandleAPIErrorJson(httpHelper, w, r, err)
 			return
 		}
 
--- a/src/adminconsole/internal/handlers/adminsettingshandlers/handler_admin_settings_keys_test.go
+++ b/src/adminconsole/internal/handlers/adminsettingshandlers/handler_admin_settings_keys_test.go
@@ -1,0 +1,147 @@
+package adminsettingshandlers
+
+import (
+	"context"
+	"encoding/json"
+	"errors"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+
+	"github.com/leodip/goiabada/adminconsole/internal/apiclient"
+	"github.com/leodip/goiabada/core/constants"
+	"github.com/leodip/goiabada/core/handlerhelpers"
+	"github.com/leodip/goiabada/core/oauth"
+)
+
+// stubApiClient embeds apiclient.ApiClient so its hundred-odd other methods come for free
+// and any of them that a test does not stub panics on a nil interface, which is the right
+// outcome for a call the test did not expect. There is no generated mock: adminconsole has
+// no .mockery.yaml. Same shape as adminclienthandlers' stub, for the same reason.
+type stubApiClient struct {
+	apiclient.ApiClient
+	rotateErr error
+}
+
+func (s *stubApiClient) RotateSettingsKeys(accessToken string) error {
+	return s.rotateErr
+}
+
+// TestHandleAdminSettingsKeysRotatePost_APIErrorReachesTheBrowser owns the wiring between
+// the rotate handler and HandleAPIErrorJson. The helper's own forwarding is pinned next
+// door in api_error_helper_test.go, but that test cannot see which of the two error paths
+// this handler calls: swapping HandleAPIErrorJson back to a direct JsonError leaves the
+// helper's tests and the whole admin console suite green while putting the 409 back behind
+// "An unexpected server error has occurred" and a request id.
+//
+// That sentence is the entire point of the change. The auth server answers 409
+// ROTATION_IN_PROGRESS when another rotation won the race, which is a fact about the
+// administrator's own request and something they can act on, and the modal that
+// sendAjaxRequest opens shows error_description verbatim (#251).
+//
+// The rows for the other statuses are what keeps the forwarding narrow: a genuine server
+// fault must stay a generic 500 with its detail in the log rather than on the screen.
+func TestHandleAdminSettingsKeysRotatePost_APIErrorReachesTheBrowser(t *testing.T) {
+
+	const refusal = "Another key rotation is in progress"
+
+	const generic = "An unexpected server error has occurred"
+
+	testCases := []struct {
+		name            string
+		apiErr          error
+		wantStatus      int
+		wantError       string
+		wantDescription string
+	}{
+		{
+			name:            "a 409 is forwarded with its status and description",
+			apiErr:          &apiclient.APIError{Code: "ROTATION_IN_PROGRESS", Message: refusal, StatusCode: http.StatusConflict},
+			wantStatus:      http.StatusConflict,
+			wantError:       "ROTATION_IN_PROGRESS",
+			wantDescription: refusal,
+		},
+		{
+			name:            "a 400 is forwarded too, which is what #122 established",
+			apiErr:          &apiclient.APIError{Code: "VALIDATION_ERROR", Message: "bad request", StatusCode: http.StatusBadRequest},
+			wantStatus:      http.StatusBadRequest,
+			wantError:       "VALIDATION_ERROR",
+			wantDescription: "bad request",
+		},
+		{
+			name:            "an incomplete key set stays generic, since 500 is not the administrator's mistake",
+			apiErr:          &apiclient.APIError{Code: "KEY_SET_INCOMPLETE", Message: "Expected current and next keys to exist", StatusCode: http.StatusInternalServerError},
+			wantStatus:      http.StatusInternalServerError,
+			wantError:       "server_error",
+			wantDescription: generic,
+		},
+		{
+			name:            "a transport error stays generic",
+			apiErr:          errors.New("connection refused"),
+			wantStatus:      http.StatusInternalServerError,
+			wantError:       "server_error",
+			wantDescription: generic,
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+
+			// templateFS is nil because JsonError renders no template. This is the real
+			// helper rather than a mock so the assertions below are on the bytes the
+			// browser receives.
+			httpHelper := handlerhelpers.NewHttpHelper(nil)
+
+			req := httptest.NewRequest(http.MethodPost, "/admin/settings/keys/rotate", nil)
+			req = req.WithContext(context.WithValue(req.Context(), constants.ContextKeyJwtInfo,
+				oauth.JwtInfo{TokenResponse: oauth.TokenResponse{AccessToken: "an-access-token"}}))
+
+			rec := httptest.NewRecorder()
+
+			handler := HandleAdminSettingsKeysRotatePost(httpHelper, &stubApiClient{rotateErr: tc.apiErr})
+			handler.ServeHTTP(rec, req)
+
+			assert.Equal(t, tc.wantStatus, rec.Code)
+
+			var response map[string]string
+			err := json.Unmarshal(rec.Body.Bytes(), &response)
+			assert.NoError(t, err)
+
+			assert.Equal(t, tc.wantError, response["error"])
+			assert.Contains(t, response["error_description"], tc.wantDescription)
+
+			if tc.wantStatus == http.StatusInternalServerError {
+				// The generic branch must not leak the API's message either: that is
+				// what sends it to the log rather than the screen.
+				assert.NotContains(t, response["error_description"], tc.apiErr.Error())
+			}
+		})
+	}
+}
+
+// TestHandleAdminSettingsKeysRotatePost_SuccessIsUnchanged is here so that routing the
+// failure path through a different helper cannot alter what a successful rotation answers.
+func TestHandleAdminSettingsKeysRotatePost_SuccessIsUnchanged(t *testing.T) {
+
+	httpHelper := handlerhelpers.NewHttpHelper(nil)
+
+	req := httptest.NewRequest(http.MethodPost, "/admin/settings/keys/rotate", nil)
+	req = req.WithContext(context.WithValue(req.Context(), constants.ContextKeyJwtInfo,
+		oauth.JwtInfo{TokenResponse: oauth.TokenResponse{AccessToken: "an-access-token"}}))
+
+	rec := httptest.NewRecorder()
+
+	handler := HandleAdminSettingsKeysRotatePost(httpHelper, &stubApiClient{rotateErr: nil})
+	handler.ServeHTTP(rec, req)
+
+	assert.Equal(t, http.StatusOK, rec.Code)
+
+	var response struct {
+		Success bool
+	}
+	err := json.Unmarshal(rec.Body.Bytes(), &response)
+	assert.NoError(t, err)
+	assert.True(t, response.Success)
+}
--- a/src/adminconsole/internal/handlers/api_error_helper.go
+++ b/src/adminconsole/internal/handlers/api_error_helper.go
@@ -40,9 +40,11 @@ func HandleAPIErrorWithCallback(httpHelper HttpHelper, w http.ResponseWriter, r 
 //
 // Routes on HTTP status the same way: 400 Bad Request from the auth server API is a
 // user-correctable validation failure, so its status and English description are forwarded
-// to the browser. Anything else, including an error that is not an *apiclient.APIError,
-// falls through to JsonError's other branch: HTTP 500, the detail in the server log, and a
-// request id on screen.
+// to the browser. 409 Conflict is forwarded for the same reason: it says another operation
+// won a race, which is a fact about the administrator's own request rather than a server
+// fault, and it names an action they should not repeat. Anything else, including an error
+// that is not an *apiclient.APIError, falls through to JsonError's other branch: HTTP 500,
+// the detail in the server log, and a request id on screen.
 //
 // This exists because JsonError preserves a status and a description only for
 // *customerrors.ErrorDetail. An *apiclient.APIError handed to it directly takes the generic
@@ -55,7 +57,8 @@ func HandleAPIErrorWithCallback(httpHelper HttpHelper, w http.ResponseWriter, r 
 // escaping is load-bearing rather than defensive, because showModalDialog assigns the
 // description to innerHTML.
 func HandleAPIErrorJson(httpHelper HttpHelper, w http.ResponseWriter, r *http.Request, err error) {
-	if apiErr, ok := err.(*apiclient.APIError); ok && apiErr.StatusCode == http.StatusBadRequest {
+	if apiErr, ok := err.(*apiclient.APIError); ok &&
+		(apiErr.StatusCode == http.StatusBadRequest || apiErr.StatusCode == http.StatusConflict) {
 		httpHelper.JsonError(w, r, customerrors.NewErrorDetailWithHttpStatusCode(
 			apiErr.Code, apiErr.Message, apiErr.StatusCode))
 		return
--- a/src/adminconsole/internal/handlers/api_error_helper_test.go
+++ b/src/adminconsole/internal/handlers/api_error_helper_test.go
@@ -1,0 +1,98 @@
+package handlers
+
+import (
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"github.com/leodip/goiabada/adminconsole/internal/apiclient"
+	"github.com/leodip/goiabada/core/customerrors"
+	mocks_handlerhelpers "github.com/leodip/goiabada/core/handlerhelpers/mocks"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/mock"
+	"github.com/stretchr/testify/require"
+)
+
+// This is the first behavioural test in this package: the three files here today are two lint tests
+// and TestMain.
+//
+// It owns what HandleAPIErrorJson forwards, which is the only observable half of #251's admin
+// console change. The modal itself is sendAjaxRequest in utils.js, unchanged, and no test in this
+// repository drives a browser, so this is where "the administrator reads the API's sentence rather
+// than a request id" is pinned. The distinction it tests is invisible at runtime unless you read the
+// screen: both branches call JsonError, and only the argument differs.
+//
+// mocks_handlerhelpers.HttpHelper is the core module's mock. Its method set is a superset of this
+// package's HttpHelper interface, so it satisfies it without a hand-written stub.
+
+// captureJsonError registers JsonError and returns a pointer to the error it was handed.
+func captureJsonError(httpHelper *mocks_handlerhelpers.HttpHelper) *error {
+	var captured error
+	httpHelper.On("JsonError", mock.Anything, mock.Anything, mock.Anything).
+		Run(func(args mock.Arguments) {
+			captured, _ = args.Get(2).(error)
+		}).Return().Once()
+	return &captured
+}
+
+// TestHandleAPIErrorJson_ForwardsConflict is #251's case. A 409 from the auth server API says
+// another rotation won the race, which is a fact about the administrator's own request, so its code,
+// description and status reach the browser instead of "An unexpected server error has occurred".
+func TestHandleAPIErrorJson_ForwardsConflict(t *testing.T) {
+	httpHelper := mocks_handlerhelpers.NewHttpHelper(t)
+	captured := captureJsonError(httpHelper)
+
+	HandleAPIErrorJson(httpHelper, httptest.NewRecorder(),
+		httptest.NewRequest(http.MethodPost, "/admin/settings/keys/rotate", nil),
+		&apiclient.APIError{
+			Code:       "ROTATION_IN_PROGRESS",
+			Message:    "Another key rotation is in progress",
+			StatusCode: http.StatusConflict,
+		})
+
+	detail, ok := (*captured).(*customerrors.ErrorDetail)
+	require.True(t, ok, "expected an *customerrors.ErrorDetail, got %T", *captured)
+	assert.Equal(t, "ROTATION_IN_PROGRESS", detail.GetCode())
+	assert.Equal(t, "Another key rotation is in progress", detail.GetDescription())
+	assert.Equal(t, http.StatusConflict, detail.GetHttpStatusCode())
+}
+
+// TestHandleAPIErrorJson_ForwardsBadRequest pins the behaviour #122 established. It is here so that
+// adding 409 to the condition cannot quietly replace 400 rather than join it.
+func TestHandleAPIErrorJson_ForwardsBadRequest(t *testing.T) {
+	httpHelper := mocks_handlerhelpers.NewHttpHelper(t)
+	captured := captureJsonError(httpHelper)
+
+	HandleAPIErrorJson(httpHelper, httptest.NewRecorder(),
+		httptest.NewRequest(http.MethodPost, "/admin/clients/1/redirect-uris", nil),
+		&apiclient.APIError{
+			Code:       "VALIDATION_ERROR",
+			Message:    "Invalid redirect URI",
+			StatusCode: http.StatusBadRequest,
+		})
+
+	detail, ok := (*captured).(*customerrors.ErrorDetail)
+	require.True(t, ok, "expected an *customerrors.ErrorDetail, got %T", *captured)
+	assert.Equal(t, "VALIDATION_ERROR", detail.GetCode())
+	assert.Equal(t, "Invalid redirect URI", detail.GetDescription())
+	assert.Equal(t, http.StatusBadRequest, detail.GetHttpStatusCode())
+}
+
+// TestHandleAPIErrorJson_GenericBranchForOtherStatuses is the other side of the boundary. A 500 from
+// the API is a server fault, so it keeps going to the log with a request id on screen rather than
+// having its English text shown to an administrator who can do nothing with it.
+func TestHandleAPIErrorJson_GenericBranchForOtherStatuses(t *testing.T) {
+	httpHelper := mocks_handlerhelpers.NewHttpHelper(t)
+	captured := captureJsonError(httpHelper)
+
+	apiErr := &apiclient.APIError{
+		Code:       "INTERNAL_ERROR",
+		Message:    "Failed to rotate signing keys",
+		StatusCode: http.StatusInternalServerError,
+	}
+	HandleAPIErrorJson(httpHelper, httptest.NewRecorder(),
+		httptest.NewRequest(http.MethodPost, "/admin/settings/keys/rotate", nil), apiErr)
+
+	assert.Same(t, apiErr, *captured,
+		"a status outside the forwarded set must reach JsonError unwrapped, for its generic branch")
+}
--- a/src/authserver/go.mod
+++ b/src/authserver/go.mod
@@ -16,6 +16,7 @@ require (
 	github.com/pkg/errors v0.9.1
 	github.com/pquerna/otp v1.5.0
 	github.com/stretchr/testify v1.11.1
+	gopkg.in/yaml.v3 v3.0.1
 )
 
 replace github.com/leodip/goiabada/core => ../core
@@ -66,7 +67,6 @@ require (
 	golang.org/x/sync v0.22.0 // indirect
 	golang.org/x/sys v0.47.0 // indirect
 	golang.org/x/text v0.41.0 // indirect
-	gopkg.in/yaml.v3 v3.0.1 // indirect
 	modernc.org/libc v1.75.3 // indirect
 	modernc.org/mathutil v1.7.1 // indirect
 	modernc.org/memory v1.12.0 // indirect
--- a/src/authserver/internal/handlers/apihandlers/handler_api_settings_keys.go
+++ b/src/authserver/internal/handlers/apihandlers/handler_api_settings_keys.go
@@ -1,24 +1,20 @@
 package apihandlers
 
 import (
-	"crypto/x509"
 	"encoding/base64"
 	"encoding/json"
-	"encoding/pem"
 	"net/http"
 	"strconv"
 	"time"
 
 	"github.com/go-chi/chi/v5"
-	"github.com/google/uuid"
 	"github.com/leodip/goiabada/authserver/internal/handlers"
 	"github.com/leodip/goiabada/core/api"
 	"github.com/leodip/goiabada/core/constants"
 	"github.com/leodip/goiabada/core/data"
-	"github.com/leodip/goiabada/core/encryption"
 	"github.com/leodip/goiabada/core/enums"
-	"github.com/leodip/goiabada/core/models"
-	"github.com/leodip/goiabada/core/rsautil"
+	"github.com/leodip/goiabada/core/oauth"
+	"github.com/pkg/errors"
 )
 
 // HandleAPISettingsKeysGet - GET /api/v1/admin/settings/keys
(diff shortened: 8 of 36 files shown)


leodip:
## Decisions: none deferred

**Every one of the eight decisions in this change was settled with the user before the run started, and the run raised none of its own.** There is no assumption to review here and nothing whose reversal is pending. This comment exists so that is stated rather than inferred from an absent one.

For the record, what was settled and what it cost:

| # | Question | Settled as |
|---|---|---|
| 1 | How the transition is serialised | One transaction around the whole transition, each state change a conditional `UPDATE` asserting its from-state and checked through `RowsAffected()`. Key generation moved ahead of the transaction. Rejected: the locking read the issue proposes (no such abstraction in the data layer, and SQLite has no `SELECT ... FOR UPDATE`), and leaning on the unique constraint alone to refuse the loser (the refusal would arrive as an engine-specific driver error). |
| 2 | Where the transition lives | A `SigningKeyRotator` type in `src/core/oauth`, beside `CodeIssuer` and `TokenIssuer`. Named for the signing key specifically, because `RotateEncryptionKeyIfNeeded` on the `Database` interface already rotates a different key. Rejected: a `RotateSigningKeys` method on the `Database` interface, because mocking the method under test proves nothing. |
| 3 | Whether the unique constraint is part of this change | Yes, same PR, its own stage so it could be dropped without unpicking the handler work. Rejected: folding it into #249, which would close #251 without doing what its own Fix section asks. |
| 4 | The constraint's shape and its sweep | Plain `UNIQUE (state)` as `idx_key_pairs_state`, preceded by a sweep keeping the highest `id` per state. One statement each, identical on all four engines. |
| 5 | What `GetCurrentSigningKey` does when there is no current key | Returns an error. All eight call sites already check `err` and return, so all eight became correct untouched. Rejected: the issue's proposal of a nil check at each of the eight sites, which leaves the ninth unguarded. |
| 6 | Whether rotation repairs a broken key set | No repair anywhere. A broken key set is only reachable from the old code, and once the transition is atomic the state cannot be produced again. |
| 7 | The status code for a refused rotation | 409 `ROTATION_IN_PROGRESS` for the loser of a race, 500 `KEY_SET_INCOMPLETE` for a missing `current` or `next`, 500 `INTERNAL_ERROR` otherwise. |
| 8 | What the administrator sees when a rotation is refused | The API's own description, forwarded through `HandleAPIErrorJson`, so the modal reads "Another key rotation is in progress" rather than a generic 500 with a request id. |

Decision 6 is the one worth a second look if you disagree with anything, since it is the only one that chooses *not* to do something: rotation refuses on a broken key set rather than repairing it. The reasoning was that a deployment with no `current` row cannot mint or validate a token and so cannot authenticate anyone to call a repair endpoint, which leaves startup as the only place a repair could run at all. Reversing it would mean promoting `next` to `current` at startup, and it would give the endpoint a third outcome for decisions 7 and 8 to express.

leodip:
## Follow-ups: one, solved here rather than filed

**Update, after stage 5.** The one follow-up below was not filed. It was solved on this branch in
`97cb74df` instead, along with the wider reconciliation its last paragraph asks for: `openapi.yaml`
now declares every status the handlers answer, and
`src/authserver/internal/server/openapi_contract_lint_test.go` holds the two together in both
directions so the drift cannot come back silently. The draft is left below unedited, since it is
what the work was scoped from. Its `gh issue create` command should not be run.

The 409 on `createUser` is now written out in place naming `EMAIL_ALREADY_EXISTS`, and the sweep it
suggested pairing found five more things: 500 undeclared on all 96 operations, 400 on 37 and 404 on
15, the two account ownership 403s, `addGroupMember` documented as 200 where the handler answers 201,
and two operations declaring a 404 no handler can produce.

---

Found during this run, verified against the code, and not this change's to fix. Duplicate search re-run just now (2026-08-23) over `gh issue list --state all --search` for "openapi 409", "openapi.yaml", "409 conflict documented" and "EMAIL_ALREADY_EXISTS": the only hits on any of them are #54 and #43, both closed and both about something else. Nothing tracks this.

---

### 1. `POST /api/v1/admin/users` returns 409 but `openapi.yaml` documents only 201, 400 and 401

**Labels:** `bug`, `documentation`

**Body:**

> `HandleAPIUserCreatePost` in `src/authserver/internal/handlers/apihandlers/handler_api_users_crud.go` answers a duplicate email with HTTP 409 and `error_code` `EMAIL_ALREADY_EXISTS`, at two call sites (one for the email field, one for the derived username path). The API's own contract does not say so.
>
> `src/authserver/web/openapi.yaml` is embedded by `src/authserver/web/embed_fs.go` and served at `GET /openapi.yaml` (`src/authserver/internal/server/routes.go`), so it is what a caller generates a client from. Its `createUser` operation declares `201`, `400` and `401` and nothing else. A generated client therefore has no branch for the one failure a caller creating users hits routinely, and will usually surface it as an unmodelled error rather than "that email is already registered".
>
> `grep -oE "^\s+'[0-9]{3}':" src/authserver/web/openapi.yaml` shows the file currently declares 200, 201, 304, 400, 401, 404 and 429 and no 409 anywhere, so this is the only 409 the product returns today and the only one missing.
>
> **What closing it takes:** a `'409'` entry on the `createUser` operation whose description names `EMAIL_ALREADY_EXISTS`, with `content` pointing at the existing `ErrorResponse` schema, matching the shape of the `400` and `401` entries around it. One block, no code change. Worth pairing with a sweep of the other operations for statuses the handlers return and the spec omits, since nothing enforces the two agreeing.

**How it was found:** while adding the rotate endpoint's new 409 to that same file in stage 3. That made this the second 409 in the product and the first to be documented, which is what surfaced the first one being absent.

**File it with:**

```bash
  --title "POST /api/v1/admin/users returns 409 but openapi.yaml documents only 201, 400 and 401" \
  --label bug --label documentation \
  --body '`HandleAPIUserCreatePost` in `src/authserver/internal/handlers/apihandlers/handler_api_users_crud.go` answers a duplicate email with HTTP 409 and `error_code` `EMAIL_ALREADY_EXISTS`, at two call sites (one for the email field, one for the derived username path). The API'"'"'s own contract does not say so.

`src/authserver/web/openapi.yaml` is embedded by `src/authserver/web/embed_fs.go` and served at `GET /openapi.yaml` (`src/authserver/internal/server/routes.go`), so it is what a caller generates a client from. Its `createUser` operation declares `201`, `400` and `401` and nothing else. A generated client therefore has no branch for the one failure a caller creating users hits routinely, and will usually surface it as an unmodelled error rather than "that email is already registered".

`grep -oE "^\s+'"'"'[0-9]{3}'"'"':" src/authserver/web/openapi.yaml` shows the file currently declares 200, 201, 304, 400, 401, 404 and 429 and no 409 anywhere, so this is the only 409 the product returns today and the only one missing.

**What closing it takes:** a `'"'"'409'"'"'` entry on the `createUser` operation whose description names `EMAIL_ALREADY_EXISTS`, with `content` pointing at the existing `ErrorResponse` schema, matching the shape of the `400` and `401` entries around it. One block, no code change. Worth pairing with a sweep of the other operations for statuses the handlers return and the spec omits, since nothing enforces the two agreeing.

Found while adding the rotate endpoint'"'"'s 409 to the same file in #254.'
```

---

**Nothing else was drafted, and two candidates were considered and rejected on purpose:**

- **`keys/get-order` loops `previous` without a `break`.** Rejected at stage 2's gate: the change's scope puts it out by name, because once duplicates are prevented there is nothing left for that loop to pick between.
- **"Give the admin console browser tests."** Raised twice by the reviewer as an unchecked item, both times as an observation rather than a defect. No JavaScript or template file is in this diff, the escaping it would cover is already pinned by `TestUtilsJS_ErrorDescriptionIsEscaped`, and no test in this repository drives a browser. That is a repo-wide appetite, not this change's debt, so filing it against #251 would misattribute it.

leodip:
## The code review

**Reviewed by `gpt-5.6-sol` at effort `xhigh`, two rounds over the whole branch, against a maximum of four.** No stage was reviewed on its own: the whole change was read once, after the last stage landed, with everything built on top of it in view. All three axes (conformance, quality, security) returned `reviewed` in both rounds.

| Round | Scope | Findings | Blocking | Outcome |
|---|---|---|---|---|
| 1 | `origin/main...HEAD` at `ac4b168c`, 33 files, +1826 / -139 | 2, both `significant` / `quality` / `confidence: high` | 0 | both fixed in `aed37751` |
| 2 | `origin/main...HEAD` at `aed37751`, 34 files | 0 | 0 | **clean** |

Round 1's own note: *"Two significant test-coverage gaps were found. The reviewed production behavior itself is conformant and currently sound."* Round 2's: *"Round 1's two quality findings are fully closed. No new conformance, quality, or security finding survived this review."*

Nothing was raised on conformance or security in either round, and **no finding contested any of the eight settled decisions**. Both round 1 findings arrived `needs_human: true` with a populated `forced_answer`, which makes them findings the code settles rather than escalations: nothing was being chosen and the answer was demonstrable. Neither was escalated, and both were verified independently before being worked.

---

### Round 1, finding 1 — significant, quality: the rotate handler's error wiring was unpinned

*Target:* `handler_admin_settings_keys.go:HandleAdminSettingsKeysRotatePost`.
*Family:* cross-layer handler-to-error-helper wiring for changed admin console AJAX behaviour.

**The claim.** The admin console suite does not prove the rotate handler sends its 409 through `HandleAPIErrorJson`. Replacing that call with the old direct `JsonError` leaves the full suite green, which would put `ROTATION_IN_PROGRESS` back behind a generic 500 and a request id: the exact outcome decision 8 exists to remove.

**Verified, not taken on trust.** Two independent checks. `grep -rn HandleAdminSettingsKeysRotatePost --include=*.go src/` returns the definition and the route registration and nothing else, and the only test file in `adminsettingshandlers` was `test_main_test.go`. Then the mutation itself, re-run rather than believed:

```
where.sh mutate src/adminconsole/.../handler_admin_settings_keys.go \
  --replace 'handlers.HandleAPIErrorJson(httpHelper, w, r, err)' \
  --with 'httpHelper.JsonError(w, r, err)' --type adminconsole
→ MUTATION SURVIVED
```

The reviewer's reasoning about why the helper's own tests cannot close this is correct and is the part worth keeping: `api_error_helper_test.go` pins what `HandleAPIErrorJson` forwards, but **both** of its branches end in a call to `JsonError` and only the argument differs, so no test of the helper can observe which of the two paths its caller takes.

**Fixed** in `aed37751`, new `handler_admin_settings_keys_test.go`. It follows `adminclienthandlers/handler_admin_client_redirect_uris_test.go`, the same shape for the helper's only other caller: a `stubApiClient` embedding `apiclient.ApiClient` so an unstubbed method panics rather than answering, the real `handlerhelpers.NewHttpHelper(nil)` rather than a mock so the assertions are on the bytes the browser receives, and a table over four statuses. The 409 forwards its status, `ROTATION_IN_PROGRESS` and its description; the 400 still forwards, so adding 409 to the helper's condition cannot be shown to have *replaced* 400 rather than joined it; the `KEY_SET_INCOMPLETE` 500 and a transport error both stay generic and are asserted **not** to leak the API's message, which is what keeps the forwarding narrow. A second test pins that a successful rotation still answers 200 with `Success: true`. Same mutation after: **MUTATION CAUGHT**.

### Round 1, finding 2 — significant, quality: a storage fault read as a lost race

*Target:* `commondb/key_pair.go:UpdateKeyPairState` and the seven `TestUpdateKeyPairState_*` cases.
*Family:* storage failures collapsed into benign compare-and-set losses.

**The claim.** None of the seven data cases makes the statement itself fail, so returning `false, nil` on an `ExecSql` error survives them all.

**The consequence is why this was worth a fix rather than a note.** `Rotate` maps a false return to `ErrRotationInProgress`, decision 7 answers that with 409 `ROTATION_IN_PROGRESS`, and decision 8 renders it in the modal as "Another key rotation is in progress". An administrator reading that sentence understands the rotation they asked for already happened, and does not retry. A storage failure reported down that path is a rotation that **never occurred**, described to the operator as one that did.

**Verified**, by re-running the mutation:

```
where.sh mutate src/core/data/commondb/key_pair.go \
  --replace 'return false, errors.Wrap(err, "unable to update keyPair state")' \
  --with 'return false, nil' --type data --db sqlite --run TestUpdateKeyPairState
→ MUTATION SURVIVED
```

**Fixed** in `aed37751`, `TestUpdateKeyPairState_StorageFailureIsAnError` in `tests/data/key_pair_test.go`. It runs the update on a transaction that has already been rolled back and requires a non-nil error with `moved` false, then reads the row back to confirm it really is untouched, so the error cannot be reporting a failure after a transition that happened anyway. `database/sql` answers `sql.ErrTxDone` from `tx.Exec` itself, before any driver is reached, so this is deterministic on all four engines rather than a per-dialect quirk, and it exercises the real implementation rather than an injected fake. Green on sqlite, mysql, postgres and mssql. Same mutation after: **MUTATION CAUGHT**.

The reviewer's `why_exhaustive` names the residual honestly and it is **accepted rather than closed**: after argument validation the method has two external error sources, and no supported driver exposes a stable way to make `RowsAffected` fail after a statement has succeeded. That boundary stays unchecked, deliberately. The code handles it correctly one line above the return, byte-for-byte the house idiom at `TryClaimCleanupRun` (`commondb/settings.go`).

---

### Round 2: clean, and what makes that mean something

A reviewer that finds nothing has either confirmed the tests or failed to test them, and only mutation evidence tells the two apart. **Round 2 ran five mutations and caught all five.**

| Mutation | Result |
|---|---|
| Replace `HandleAPIErrorJson` with a direct `JsonError` | **caught**, by the new handler test |
| Swallow the CAS `ExecSql` error as `false, nil` | **caught**, by the new storage-failure test |
| Bypass the closed transaction through the connection pool | **caught**, by the required error assertion |
| Bypass the transaction and return an error after writing | **caught**, by the unchanged-row assertion |
| Change the rotate success response to `false` | **caught**, by the success test |

The first two are round 1's own findings re-run against the fixes: both now fail as they should. The reviewer re-verified exact hashes and clean restoration after all five runs.

It also read the request, the full agreement, all four stage logs, the probes and the complete 34-file diff; traced the new admin test through the JWT context, the stub API client, `HandleAPIErrorJson`, the real JSON encoder and the response assertions for 409, 400, generic failures and success; traced `TestUpdateKeyPairState_StorageFailureIsAnError` through every dialect wrapper, `CommonDatabase` and `database/sql` closed-transaction behaviour; and reconciled rotation interleavings, migrations, callers, writers, documentation, catalogs, authorization, CSRF, audit, secret storage, error oracles and SQL parameterization.

**The tree the clean verdict is about was independently re-checked before the verdict was accepted**, not after. HEAD `aed37751` as claimed; merge base with `origin/main` equal to `origin/main` itself, so the base did not move under the review; 34 files, +2022 / -139 with two-dot and three-dot agreeing; worktree clean; all 38 code anchors resolving to exactly one line in the named function; and every category hash unmoved from the value round 1's fixes left (production `8882fd21b4de`, unit-tests `bf01e6ffc1b4`, data-tests `5b5ea49eeaa8`, integration-tests `16e1b1da7a5d`, docs `f6ca29a48df0`, aggregate `b0dfe2aba20b`, which is the value the reviewer computed for itself). The reviewer read this branch and no other.

One bookkeeping error is worth recording because a wrong number in a review request looks like evidence later: the round 2 request described the diff as "+1998 / -139" where it is +2022 / -139, 24 insertions low, carried forward from round 1's prose rather than recounted after the fixes landed. It changed nothing — the file count was right and the reviewer verified the tree by hash rather than by that sentence, reporting the same six values above.

### The `unchecked` lists, all nine entries answered

The reviewer's `unchecked` array is a work list, not a disclaimer, so every entry got an answer rather than being left for a later round to rediscover. Across both rounds: **nine entries, nine answered, none filed as a follow-up.** The four that recur and matter:

1. **No browser or live admin-console-to-auth-server HTTP hop was run.** Dismissed, with the chain verified by reading end to end rather than assumed. The auth server answers 409 `ROTATION_IN_PROGRESS` (stage 3's handler test); `HandleAPIErrorJson` forwards 400 and 409 as an `ErrorDetailWithHttpStatusCode` (pinned by `api_error_helper_test.go`); the new `handler_admin_settings_keys_test.go` asserts through the **real** encoder that the 409 emits status 409 with the refusal in `error_description`, the exact wire field the browser reads; `sendAjaxRequest` renders `escapeHtml(err.error_description)` into the modal; and the rotate button supplies `"modalId": "modal0"`. **No `.js` or `.html` file is in the diff**, so the unexecuted link is unchanged code shared with roughly twenty other call sites, and its escaping is separately pinned by `TestUtilsJS_ErrorDescriptionIsEscaped`. No test in this repository drives a browser, so this cannot be closed by adding one.
2. **Two complete `Rotate()` calls were not raced.** Dismissed, and rejected on purpose at seal time. Decision 1 moves the 4096-bit generation ahead of the transaction, generation measures ~299ms with wide variance (`probe/probe_251c_output.txt`), so the two transactions may never overlap, both rotations then succeed, and that **correct** outcome would fail such a test. The reviewer reached the same conclusion independently and checked the deterministic conditional-update race instead, plus both loser branches. That race is covered by `TestUpdateKeyPairState_Concurrent` on all four engines.
3. **A statement that succeeds and whose `RowsAffected` then errors.** Dismissed, per finding 2 above. Not inducible through any supported driver; the code handles it.
4. **Unchanged surfaces (PKCE, redirect, credential, session, token validation) not re-audited.** Dismissed, and confirmed rather than accepted: `git diff --name-only origin/main...HEAD` matches nothing under any of them. Re-auditing unchanged surfaces is what a scope exists to prevent.

### Tiers behind the fix round

Round 1's fixes are test-only, so `production` is unmoved at `8882fd21b4de` and the reviewer's reading of the production code still stands. That match is also the check that all four mutation runs in that session were fully restored.

- `--type modules`: 1097 passed, 0 failed, 8s.
- `--type data`: sqlite, mysql, postgres, mssql. 1247 passed, 0 failed, 44s. `TestUpdateKeyPairState_StorageFailureIsAnError` green on each of the four.
- Integration: not applicable to that round. Two test files moved and no handler, route or schema, so no endpoint behaviour changed. CI runs it on four engines regardless.
- `check-anchors.sh`: all 38 rows resolve to exactly one line, in the named function.

**Round 2 changed nothing**, because nothing was found: no fix commit, no edit to any tracked file, every category hash identical. Re-running the local tiers would have executed the same bytes that were already green at that exact tree, so it was not done, and this sentence says so rather than implying a run that did not happen.

**CI is green on `aed377512003edff7dba2d8056548ce321f5a0fa`, all 11 jobs**, which is the authority for the four-engine claim: `Tests / {sqlite,mysql,postgres,mssql}`, `Unit / {core,authserver,adminconsole}`, `Build validation`, `Lint`, `Vulnerabilities`, `Versions`.

No finding was refuted, none was deferred, and no follow-up came out of either round.

Issue #247, opened 2026-08-22
Ceremony pages carry no cache directives, and the OTP enrolment page renders a plaintext TOTP seed


leodip:
Found by the #239 authorization ceremony review (finding [F30](https://github.com/leodip/goiabada/issues/239#user-content-f30)).

**Sequence this after #242**, which changes how the OTP enrolment secret is generated and stored.

## No ceremony page sets cache directives

`HttpHelper.RenderTemplate` sets `Content-Type` and, optionally, a status code. It sets no `Cache-Control` and no `Pragma`. `MiddlewareSecurityHeaders` sets `X-Frame-Options`, `Content-Security-Policy: frame-ancestors 'none'`, `X-Content-Type-Options`, `Referrer-Policy` and `Strict-Transport-Security`, and no cache directives either.

Grepping the tree, `Cache-Control: no-store` is set on the token endpoint, the DCR endpoint, `EncodeJson`, the rate limiter's refusal, the two `form_post` emitters and the health check. **It is set on no rendered HTML page in the ceremony.** So the password form, the OTP form, the consent screen and the OTP enrolment page all go out with no cache directives at all.

## The enrolment page is the one that matters

`auth_otp_enrollment.html` renders the QR code as a `data:` image and the seed itself in plain text:

```html
<pre class="mt-2 text-center">{{.secretKey}}</pre>
```

That is a long-lived shared secret, the whole of the second factor, in a cacheable 200 response.

The session `Set-Cookie` on the response does not help, and RFC 9111 7.3, "Caching of Sensitive Information", says so directly:

> Implementation and deployment flaws (often led to by the misunderstanding of cache operation) might lead to the caching of sensitive information (e.g., authentication credentials) that is thought to be private, exposing it to unauthorized parties.
>
> Note that the Set-Cookie response header field does not inhibit caching; a cacheable response with a Set-Cookie header field can be (and often is) used to satisfy subsequent requests to caches. Servers that wish to control caching of these responses are encouraged to emit appropriate Cache-Control response header fields.

**Sizing it honestly.** A shared cache is unlikely to store this without explicit directives, since heuristic freshness has nothing to compute from with no `Last-Modified`. The realistic retainers are the browser's own disk cache and back/forward cache, a managed workstation's cache, and a diagnostic or inspecting proxy. That is a smaller blast radius than "any CDN", and it is still a reusable authenticator seed sitting on disk after the ceremony and after logout, which nothing else in this codebase would accept for a credential.

[...]

Issue #144, opened 2026-08-07
Security: the OTP claim binds to otp_enabled, not to the authenticator


leodip:
After #111, `TryConsumeUserOTPStep` refuses a verification claim unless `otp_enabled` is still true, so removing an authenticator invalidates a claim from a request that loaded it. A re-enrolment restores `otp_enabled`, so a verification request holding the *old* secret can still claim a code it already spent, provided the new enrolment claimed a strictly lower time step. The resulting token asserts `otp` for an authenticator that no longer exists, without the `AuditEnabledOTP` trail that enrolling would leave. No access is gained that the disable did not already grant (see #111 decision 10), so this is a false factor assertion rather than an authentication bypass.

The enrolment claim has the mirror image of the same gap, which predates #111: it passes `requireOTPEnabled` false and so binds nothing about which authenticator it establishes, while the enable is a whole-user write. Two overlapping enrolment ceremonies both succeed and the later one's write replaces the earlier one's secret.

Closing both properly needs the claim bound to the authenticator rather than to the enrolment flag: an enrolment generation column on `users`, incremented on every enable and disable, carried on the request and compared in the claim predicate. That is a migration plus a second column, which #111 deliberately did not take on. Decide whether the window is worth that before building it.

Adjacent to #140, which is the same class of defect by a different mechanism.

leodip:
The code this issue describes has moved. Nothing about the defect changed — #387 (#422, branch `issue-387-separate-application-services`) is behaviour-preserving — but the enrolment-generation column proposed here now has one place to be written and one place to be compared, instead of three enable/disable sites spread over two handler packages.

**Where it is now.** `src/authserver/internal/otpcredential/`, a package that owns the stored credential. Four operations:

| Was | Is |
|---|---|
| `handlers.EnableUserOTPTx` + the caller's `user.SetOTPSecret` and `user.OTPEnabled = true` | `otpcredential.Establish(ctx, db, user, seed) (generation, error)` |
| `apihandlers.disableUserOTP` | `otpcredential.Remove(ctx, db, user) error` |
| `user.GetOTPSecret` + `otp.MatchStep` + `TryConsumeUserOTPStep(requireOTPEnabled=true)`, inline in the browser handler | `otpcredential.VerifyStored(...)` |
| `otp.MatchStep` + `TryConsumeUserOTPStep(requireOTPEnabled=false)`, inline at two sites | `otpcredential.VerifySupplied(...)` |

**Why that matters to this issue specifically.**

[...]

PR #255, opened 2026-08-23
fix(auth): make the second factor's re-prompt survive an abandoned ceremony

### LINKED ISSUE #242: Security: the second factor's re-prompt machinery is unreliable in four ways
Found by the #239 authorization ceremony review (findings [F3](https://github.com/leodip/goiabada/issues/239#user-content-f3), [F22](https://github.com/leodip/goiabada/issues/239#user-content-f22), [F27](https://github.com/leodip/goiabada/issues/239#user-content-f27) and [F8](https://github.com/leodip/goiabada/issues/239#user-content-f8)).

Four defects on the level 2 path. They are filed together because they are reachable in one pass over the same handlers, and because the first two compound: [F22](https://github.com/leodip/goiabada/issues/239#user-content-f22) silently disables the mechanism [F3](https://github.com/leodip/goiabada/issues/239#user-content-f3) is about.

## 1. `Level2AuthConfigHasChanged` does not survive to the prompt it exists for ([F3](https://github.com/leodip/goiabada/issues/239#user-content-f3))

Three defects in one field.

1. **Spent before the factor is verified.** `HandleAuthLevel1CompletedGet` clears and commits the flag before `/auth/level2` renders anything, so an abandoned ceremony spends the re-prompt.
2. **Set on at most one session.** `handler_api_account_otp.go` flags only the caller's own session, so the user's other devices never learn that the authenticator configuration changed.
3. **Never set by the in-ceremony enrolment.** `HandleAuthOtpPost`'s enrolment branch sets it nowhere at all.

Net effect: a user who enables an authenticator can keep being issued grants with `acr: urn:goiabada:level2_optional` and `amr: ["pwd"]` for up to the session maximum lifetime, on the very sessions they were trying to protect.

**Fix:** clear the flag where the factor is verified (`HandleAuthOtpPost`), not where the redirect is decided. Set it on every session of the user, at both enrolment sites. Consider whether it wants to be a column on `users` rather than on `user_sessions`, which would make "every session" free and would make the clearing question moot.

## 2. The account OTP endpoint swallows both errors while flagging ([F22](https://github.com/leodip/goiabada/issues/239#user-content-f22))

```go
userSession, err := database.GetUserSessionBySessionIdentifier(nil, sid)
if err == nil && userSession != nil {
	userSession.Level2AuthConfigHasChanged = true
	_ = database.UpdateUserSession(nil, userSession)
}
```

Both errors are discarded. A database fault silently skips the flag, and the endpoint returns 200 saying the authenticator configuration changed.

Combined with the above, the flag is the only thing that re-prompts an existing session, so a swallowed error is a silently weakened session with a success response in front of it.

## 3. Reloading the OTP enrolment page regenerates the secret ([F27](https://github.com/leodip/goiabada/issues/239#user-content-f27))

`HandleAuthOtpGet` is a GET that writes. For a user who has not enrolled, it calls `otpSecretGenerator.GenerateOTPSecret` on **every** request and overwrites `sess.Values[constants.SessionKeyOTPSecret]` and `SessionKeyOTPImage`. The state stays `level2_otp`, so nothing blocks the reload.

The sequence that fails: the user reaches the enrolment page, scans the QR code into their authenticator, then reloads (a slow render, a fat-fingered [F5](https://github.com/leodip/goiabada/issues/239#user-content-f5), a mobile browser restoring the tab). The server issues a new secret. The authenticator is still showing codes for the old one. Every code they type is refused with "incorrect code", **each refusal spends a credential-failure budget through `RecordCredentialFailure`**, and nothing on the page suggests re-scanning. Enough refusals and the rate limiter locks them out of a first login they have not completed.

`SessionKeyOTPSecret` is a single global session slot rather than a per-ceremony one, which is the same root cause: it has no identity to compare against, so the handler cannot tell "this is the ceremony that generated you" from "this is a fresh visit".

**Fix:** generate once per ceremony and reuse. If the session already holds a secret and the ceremony has not changed, render that one instead of a new one. The ceremony id is already on the auth context and already carried by the template, so binding the stored secret to it is the natural shape, and it also fixes the neighbouring case where a second `/auth/authorize` overwrites the slot of a ceremony still on screen.

## 4. `HandleAuthLevel2Get` dereferences a nil user ([F8](https://github.com/leodip/goiabada/issues/239#user-content-f8))

```go
user, err := database.GetUserById(nil, authContext.UserId)
if err != nil { ... }

targetAcrLevel := authContext.GetTargetAcrLevel(client.DefaultAcrLevel)
switch targetAcrLevel {
case enums.AcrLevel2Optional:
	if user.OTPEnabled {
```

`CommonDatabase.getUserCommon` returns `(nil, nil)` when no row matches. Every other ceremony handler that calls `GetUserById` checks for nil: `HandleAuthOtpGet`, `HandleAuthOtpPost`, `HandleConsentGet`, `HandleConsentPost`, `HandleAuthCompletedGet` and `handleImplicitFlow` all do. `HandleAuthLevel2Get` is the only one that does not.

Reachable by deleting the user between `/auth/level1completed` and `/auth/level2`, which an administrator can do. The result is a nil pointer dereference caught by `middleware.Recoverer`, so the user sees a 500 and the log carries a panic.

One line. Worth filing because "every handler nil-checks except this one" is the kind of gap that regresses.

## Tests

- Enabling an authenticator flags every session of that user, not just the caller's.
- A database fault during the flag write is surfaced, not swallowed behind a 200.
- The re-prompt survives an abandoned ceremony: the flag is still set on the next attempt.
- Reloading `/auth/otp` during enrolment returns the same secret, and a code from the first render still verifies.
- `/auth/level2` with a deleted user answers rather than panics.

## Related

- #144 (the OTP claim binds to `otp_enabled`, not to the authenticator) and #142 (TOTP skew window) are neighbours in the same subsystem. Not folded in: they are about what OTP *means* and how it is *verified*, while this issue is about when the user is *asked*.
- #203 records that a panicking request is logged with status 0 rather than 500, so the nil dereference above also lands in the log wrong.
- The cache-directives issue that follows this one touches the same enrolment template and the same secret. **Sequence it after this**, because how the secret is stored changes there.

leodip:
Fixes #242.

## What this is for

The machinery that decides when a user must prove a second factor again is unreliable in four ways.
All four were found by the spike in #239 and all four are still present.

1. **The re-prompt can be spent without being served.** `Level2AuthConfigHasChanged` is a boolean on
   `user_sessions`. The only reader that acts on it, in `HandleAuthLevel1CompletedGet`, clears it and
   commits the session *before* redirecting to `/auth/level2`. A browser closed at the OTP prompt has
   already spent the obligation, and the next sign-in is a password alone. The same field also only
   ever reaches one session, and the browser enrolment path sets it on none, so a user who enrols an
   authenticator on one device keeps a token stamped `acr: urn:goiabada:level2_optional` with
   `amr: ["pwd"]` on every other. Goiabada defines `level2_optional` as password plus OTP if the user
   has OTP enabled, so that token asserts an authentication class the authentication performed did
   not satisfy (OpenID Connect Core 1.0 section 2: `acr` identifies the class "the authentication
   performed satisfied").

2. **The write that raises the obligation is best effort.** In `HandleAPIAccountOTPPut` the session
   read swallows its error and the write is `_ =`, so an authenticator can be enabled with nothing
   recording that any session now owes a second factor.

3. **The enrolment secret has no owner.** It lives in two global slots in the browser session,
   rewritten on every `GET /auth/otp`, so reloading the enrolment page silently replaces the secret
   behind a QR code the user may already have scanned.

4. **`/auth/level2` panics on a deleted user.** `GetUserById` returns `(nil, nil)` for a missing row
   and the handler dereferences it without a check, where six sibling ceremony handlers all check.

## The rule this establishes

**Deciding that a second factor is owed must not be able to discharge it.**

`users.otp_config_generation` becomes the authoritative counter, advanced by one inside the same
transaction as every write that establishes or removes an authenticator.
`user_sessions.otp_config_generation` records the value that session last satisfied. A session owes a
level 2 re-prompt when the two differ.

Both readers then compare two integers and write nothing, so nothing a visitor can abandon, close,
replay or race discharges an obligation. Discharging happens once, at `/auth/completed`, and only to
a value captured before the ceremony's level 2 question was answered. Because the counter hangs off
the user rather than the session, "every session of that user" is one statement rather than an
enumeration, which is what closes the one-session and no-session halves of part 1.

`Level2AuthConfigHasChanged` is retired entirely: the column, the model field, both API response
fields, and `PUT /api/v1/admin/user-sessions/{sessionIdentifier}`, whose request body contained that
field and nothing else. `GET` and `DELETE` on that path stay.

## Status

**Complete.** All three stages landed, the check suite is green on the final commit `cb599d10`
([run 32670209981](https://github.com/leodip/goiabada/actions/runs/32670209981)), and the whole
branch was reviewed and came back clean. Ready for review.

The counter is in place, `Level2AuthConfigHasChanged` is gone from the tree entirely, and the
enrolment secret now rides the auth context, which closes the last of the issue's four parts.

### Stage 1: retire the `level2AuthConfigHasChanged` API surface

`2de75465`. The published field and the endpoint built around it are gone: `Level2AuthConfigHasChanged`
off `UserSessionResponse` and `EnhancedUserSessionResponse` and off the four places that mapped it,
`UpdateUserSessionRequest` deleted, `HandleAPIUserSessionPut` and its route deleted, and the `put:`
operation, the schema property and the request schema removed from `openapi.yaml`. `GET` and `DELETE`
on that path stay, so a `PUT` to it now answers 405 rather than 404.

The admin console's only caller went with it, and it was wrong in its own right: the block read the
session identifier out of the **operator's** cookie, so an administrator who disabled somebody else's
authenticator re-prompted themselves and left the target user's sessions untouched. It was only ever
correct when an administrator edited their own account. `AuthServerClient.UpdateUserSession`, its
interface entry and the `hasDisabledOTP` bookkeeping that existed to reach it are deleted, which left
`HandleAdminUserAuthenticationPost`'s fourth parameter unused, so it is dropped at the handler and at
the route.

Nothing here touches the counter, the `level2_auth_config_has_changed` column or
`models.UserSession`: the re-prompt behaves exactly as before, minus the operator's manual lever. The
column and the model field are dropped together in stage 2, because neither order works split.

Two cases pin the deletion, since it is otherwise invisible to the suite:
`TestUserSessionResponses_OmitLevel2AuthConfigHasChanged` asserts the JSON of both session structs
carries no such key, and `TestAPIUserSessionPut_MethodNotAllowed` asserts the 405.
`TestOpenAPI_DescribesEveryAPIRoute` passes when the route and the spec block are restored together
and checks no response bodies at all, which is why those two were needed.

Tiers: unit 3603 pass / 0 fail across the three modules, integration on sqlite 1266 pass / 0 fail.
Three mutations, all caught. Data is not applicable: no query, no schema change, no model field
moved.

The `Lint` job then went red on `2de75465`, and `3eba17cb` settles it. Removing the
`Level2AuthConfigHasChanged:` field deleted the widest key from a set of struct literals, and gofmt
aligns a literal to its widest key, so seven files were left indented to the width of a name that is
no longer there. Whitespace only: `git diff -w` on the fix is empty. It is worth knowing because
`gofmt -l` runs first in the Lint job's per-module loop and the loop exits the module on failure, so
that commit also lost its `vet`, `unparam` and `golangci-lint` run on `authserver`; all three are
clean now. `TestModuleSourcesAreGofmted` was added with the fix so the unit tier catches the next
one, which matters because four of the seven files live under `./tests/`, a directory that tier
never compiles. Re-run against the fix: unit 3604 pass / 0 fail, the one addition being that guard,
data 1667 pass / 0 fail on all four engines, integration on sqlite 1266 pass / 0 fail.

### Stage 2: `otp_config_generation` replaces the flag

`88c22dfb`. The mechanism is replaced rather than repaired. `users.otp_config_generation` is the
authoritative per-user counter, advanced by one at every site that establishes or removes an
authenticator; `user_sessions.otp_config_generation` records the value that session last satisfied,
and a session owes a level 2 re-prompt whenever the two differ.

**Reading is not writing.** Both readers, `/auth/level1completed` and `handlePromptNone`, now compare
two integers they already hold and write nothing at all. That is what closes part 1.1: the handler
that decided to send a visitor to the OTP form used to clear the boolean and commit it at that
moment, so closing the browser at the prompt spent the re-prompt and the next attempt let the
visitor through on a password alone.

**The counter is per user, so "every session" is one statement.** That closes parts 1.2 and 1.3,
where the old writer flagged only the row named by the caller's own `sid` claim, reached no session
at all for a token without one, and discarded both its read error and its write error. The admin API
that disables another user's authenticator flagged nothing anywhere, and now goes through the same
transaction.

**The advance commits with the write that changed the authenticator**, at all three code paths, so
there is no state in which the authenticator has changed and no session knows. `disableUserOTP`
gained a third statement inside its existing transaction; the two enable sites share the new
`handlers.EnableUserOTPTx`.

The snapshot is captured onto `AuthContext` at password verification and on every arm of
`/auth/level2`, the legitimate skip for a user with no authenticator included, and promoted once, at
`/auth/completed`. Captured rather than read live, so an authenticator change landing mid-ceremony
is not discharged by a ceremony that never saw it. The browser enrolment overwrites its capture with
the value the increment returned, because it observed generation N and then moved the user to N+1;
promoting N would leave a session that had just enrolled and verified owing another prompt at once.

Migration 000031 adds both columns, seeds `-1` over the sessions the old flag marked **plus** every
session of a user who currently has an authenticator, then drops the old column. That second clause
is decision 9, answered on this thread: without it, a session that authenticated at
`level2_optional` with `amr: ["pwd"]` while its user had no authenticator keeps that false assertion
for its whole remaining life after the upgrade, because the old per-`sid` writer never flagged it.

Part 4 lands here too: `/auth/level2` answers 500 "user not found" rather than dereferencing a nil
user, matching the six sibling ceremony handlers that already check.

Also updated: `concepts/acr-amr.mdx`'s "OTP configuration changes" section, which described the flag
and is now false, plus the equivalent sections of `CLAUDE.md` and `AGENTS.md`.

**Tiers.** Unit 3614 pass. Data 1271 pass on mysql, postgres, mssql and sqlite. Integration 958 pass
on sqlite; the other three engines are the check suite's to report on this commit. Sixteen mutations
run against the new seams, all caught. Two survived first and were fixed before the gate: a capture
in the password handler that no end-to-end path could distinguish, now asserted at the unit tier
where it is observable, and a `prompt=none` case whose fixture was refused by the ACR check before
the counter was ever consulted, now built at `level2_optional` so it reaches the reader it names.

**One deviation worth knowing.** The shared enable helper is `handlers.EnableUserOTPTx`, exported in
the `handlers` package rather than sitting beside `disableUserOTP` in `apihandlers`. The browser
enrolment handler cannot reach an unexported function in `apihandlers`, and `TerminateUserSessionTx`
already lives in `handlers` and is called from `apihandlers` for the same reason.

### Stage 3: the enrolment secret rides the auth context

Commit `cb599d10`.

The TOTP enrolment seed and its QR code lived in two slots on the browser session, written on every
GET of `/auth/otp`. Reloading the enrolment page therefore replaced the secret behind a QR code the
user had already scanned: their next code was checked against a secret they had never been shown and
refused, with the page still showing that code's own QR code. Nothing about that is visible to the
user, and the only way out is to scan again.

The pair moves onto `AuthContext` as `OTPSecret` and `OTPImage`, and the enrolment render generates
only when the ceremony has no seed yet, which makes the GET idempotent. That is the fix. The binding
to the ceremony is then **structural rather than compared**: a second `/auth/authorize` mints a new
auth context and takes the old seed with it, so there is no shared slot for one ceremony's seed to be
read out of by another and no ceremony id to check it against.

The already-enrolled arm clears the pair where it used to delete the slots, because
`HandleAuthOtpPost` picks the template for its error rerender by whether a seed is present and a
stale one would redraw the enrolment page for a user who is already enrolled.

`ClearAuthContext` at `/auth/issue` now removes the seed as a consequence, so it no longer outlives
the ceremony in the cookie. Deleting it at the successful enrolment save is #247's half and is
deliberately not taken here.

With their last readers gone, `SessionKeyOTPSecret` and `SessionKeyOTPImage` are deleted and both OTP
handlers lose the session store parameter those slots were their only use for.

Tiers: unit 3615 pass; data on sqlite, mysql, postgres and mssql 1691 pass; integration on sqlite
1271 pass. Two new integration cases at the browser ceremony seam, both reading the secret out of the
rendered page rather than out of storage: one reloads `/auth/otp` and then submits a code generated
from the **first** render, one runs two authorization requests in a **single cookie jar** and
requires different secrets, which is the only place a shared slot and a per-ceremony field disagree.
Three mutations against the new behaviour, all caught.

**One thing worth knowing.** `startOtpCeremony` in the integration suite was split so a caller can
supply its own cookie jar, which is what makes the second-authorize case expressible in one browser.
Every existing caller keeps the fresh-jar signature unchanged.

## How it was tested

|---|---|---|
| Unit, three modules | local | 3615 pass, 0 fail |
| Data, sqlite + mysql + postgres + mssql | local | 1691 pass, 0 fail |
| Integration, sqlite | local | 1271 pass, 0 fail |
| Data + integration, **all four engines** | check suite, `cb599d10` | pass |
| Lint, Vulnerabilities, Build validation, Versions | check suite, `cb599d10` | pass |

The counts are the final stage's; each stage's own numbers are under its heading above. The
four-engine integration claim is the check suite's, not this machine's: locally the integration tier
runs on sqlite only, and mysql, postgres and mssql were never run here.

**22 mutations** were run against the new seams across the three stages, one per behaviour the change
claims to have made observable, and all 22 were caught. Two survived on first attempt in stage 2 and
were fixed before that gate closed: a capture in the password handler that no end-to-end path could
distinguish, now asserted at the unit tier where it is observable, and a `prompt=none` case whose
fixture was refused by the ACR check before the counter was ever consulted, now built at
`level2_optional` so it reaches the reader it names.

New coverage worth naming: `TestMigration000031_OTPConfigGeneration` exercises the backfill on all
four engines, seeding both old flag states before the migration runs and asserting values and column
shapes, then down and up again for SQL Server's default constraints. Two integration cases pin the
enrolment seed at the browser ceremony seam, both reading the secret out of the rendered page rather
than out of storage.

## Review

Reviewed by `gpt-5.6-sol` at effort `xhigh`. Two gates: the plan before any code landed, and the
whole branch once all three stages had. **No stage was reviewed on its own** — this run reviewed the
complete change once, at the end.

|---|---|---|---|
| Plan, 1 | 4 | 3 | 1 escalated and answered, 3 fixed in the plan before any code |
| Final, 1 | 0 | 0 | clean, `static-review-exhausted`, all three axes reviewed |

- **blocking, security:** migration 000031 would have treated sessions the old writer never flagged
  as already current, leaving a live `acr`/`amr` bypass alive for up to a session lifetime after
  upgrade. Escalated as decision 9 and answered `C`: the seed is now the union of the old flag's rows
  and every session of a user who has an authenticator.
- **blocking, conformance:** browser enrolment advanced the counter *after* the value completion
  promotes was captured, so a user who had just enrolled and verified immediately owed another
  prompt. Fixed in the plan: `IncrementUserOtpConfigGeneration` returns the value that landed and the
  enrolment writes it to the auth context inside the same transaction.
- **blocking, quality:** the migration backfill was left untested on the false premise that no
  harness can seed rows before a migration runs. Fixed: `newIsolatedDB` does exactly that, and the
  four-engine migration test above was added.
- **significant, quality:** stage 1 as planned could not tell the retirement from restoring the whole
  API surface. Fixed: a 405 case at the endpoint and a JSON-absence case per response struct.

Nothing was refuted at either gate. The final round's three `unchecked` entries were each answered
and dismissed with evidence, including a measured rebuttal of the cookie-inflation risk.

Full account, with the evidence and what the reviewer executed, in
[this comment](https://github.com/leodip/goiabada/pull/255#issuecomment-5388912922).

## Decisions and follow-ups

- [Decisions](https://github.com/leodip/goiabada/pull/255#issuecomment-5388912799) — nine settled,
  **none deferred**. One came back to you on this thread (decision 9) and you answered `C`.
- [Follow-ups](https://github.com/leodip/goiabada/pull/255#issuecomment-5388912853) — one issue
  drafted ready to file, one fix folded into this change, one observation that belongs on an existing
  issue rather than a new one.

## For the release notes

**The admin API loses one endpoint and one field.** `PUT /api/v1/admin/user-sessions/{sessionIdentifier}`
is removed; its request body contained `level2AuthConfigHasChanged` and nothing else, and the
mechanism behind that field no longer exists. A `PUT` to that path now answers **405**, not 404, since
`GET` and `DELETE` on it stay. The `level2AuthConfigHasChanged` property is also gone from the
`UserSessionResponse` and `EnhancedUserSessionResponse` bodies, so anything reading it will find it
absent. `openapi.yaml` is updated to match.

**Users with an authenticator are asked for their second factor once more after the upgrade.**
Migration 000031 cannot tell which pre-upgrade sessions had genuinely satisfied a second factor,
because the old per-session flag was never written for most of the sessions that owed one. It
therefore marks every session of a user who currently has an authenticator as owing one re-prompt.
That prompt happens once, at the user's next sign-in whose target ACR is above level 1, after which
the session behaves normally. Users with no authenticator are unaffected. This is deliberate: without
it, a session that had asserted `acr: urn:goiabada:level2_optional` with `amr: ["pwd"]` would keep
making that false assertion for the rest of its life.

**The admin console's manual "re-prompt for second factor" lever is gone.** It was removed rather
than repaired because it never did what it appeared to: the block read the session identifier out of
the *operator's* own cookie, so an administrator who disabled somebody else's authenticator
re-prompted themselves and left the target user's sessions untouched. Disabling a user's
authenticator now advances that user's counter directly, in the same transaction as the change, which
re-prompts every one of their sessions.

**Migration 000031 runs automatically** on all four supported engines. It adds
`users.otp_config_generation` and `user_sessions.otp_config_generation`, seeds the re-prompt
population above, and drops `user_sessions.level2_auth_config_has_changed`.


leodip:
## Decision 9: which pre-upgrade sessions does migration 000031 seed as owing a second factor re-prompt?

The plan review found a case the sealed agreement's decision 8 does not cover, and it is a live
bypass rather than a missed prompt. It needs your answer before stage 2 writes the migration.

### The case

Today, when a user enables OTP through `PUT /api/v1/account/otp`, the server flags at most **one**
session: the one named by the `sid` claim of the caller's own token. That is part 1.2 of this issue.
So consider a session that authenticated at `acr: urn:goiabada:level2_optional` with
`amr: ["pwd"]` back when the user had no authenticator, and the user later enables OTP from a
different browser. That session is never flagged.

This change replaces the flag with a counter on `users` and a snapshot on `user_sessions`. A session
owes a re-prompt when the two differ. Decision 8 seeds the snapshot to `-1` only where the old
boolean was set, and everything else starts at 0 matching a counter of 0.

That session therefore comes out of the upgrade looking current. Both readers pass it, and it keeps
obtaining tokens stamped `acr: urn:goiabada:level2_optional` with `amr: ["pwd"]` for a user who now
has an authenticator. OpenID Connect Core 1.0 section 2 defines `acr` as identifying the class
"the authentication performed satisfied", and Goiabada defines `level2_optional` as password plus
OTP **if the user has OTP enabled**. So the token asserts a class the authentication did not satisfy.
The bypass this issue exists to close survives the fix, for the rest of that session's life: by
default at most 24 hours, longer wherever you have raised `UserSessionMaxLifetimeInSeconds`.

Decision 8's closing sentence says the partial preservation's error "can only ever be one extra
prompt rather than one fewer". This case is one fewer.

### The options

**A. As decision 8 has it.** Seed `-1` where `level2_auth_config_has_changed = 1`, nothing else.
Nobody is prompted at upgrade who was not already going to be. Every session the old writers missed
keeps its false assertion until it expires.

**B. Seed `-1` on every session.** No legacy session is assumed current. Cost: every user with a
live session answers the level 2 question once more at their next authorization whose target ACR is
above level 1. For a user with an authenticator that is one extra OTP prompt. For a `level2_optional`
user without one it is invisible, an internal redirect through `/auth/level2` that skips and
promotes.

**C. Seed `-1` on every session of a user who currently has `otp_enabled` set.** Exactly the users
for whom a `level2_optional` session can assert a class it did not satisfy. Users with no
authenticator are untouched. Costs one extra `WHERE` clause in the seed, on four engines.

### Recommendation: C

It closes the case above, costs nothing for the population that cannot be affected by it, and keeps
A's seeded rows as a subset rather than replacing them. The only difference between B and C is users
with no authenticator, for whom no false level 2 assertion is possible in either direction, so B
buys strictly more prompts and no more safety.

### What it costs to be wrong

A leaves the bypass alive for up to one session lifetime after the upgrade. B or C cost one extra
second factor prompt per affected session, once, and cannot cost a bypass. Whichever you pick gets
pinned by a new four-engine migration test the same review added, so changing your mind later is one
line of seed SQL and one assertion.

Reply with **A**, **B** or **C**.

### Three other findings from the same review, already handled, for context

- Browser enrolment would have left a session owing another OTP prompt the instant it finished
  enrolling, because the plan captured the counter before the enrolment incremented it. Fixed in the
  plan: the increment now returns the value that landed, as `IncrementUserAuthStateGeneration`
  already does, and the enrolment records it inside its own transaction.
- The migration's backfill was going to ship untested on the belief that no harness can seed rows
  before a migration runs. It can: `newIsolatedDB` does, and eight migrations already use it. Stage 2
  now carries `TestMigration000031_OTPConfigGeneration` on all four engines.
- Stage 1 had no assertion that could tell the API retirement from restoring the whole surface, since
  the OpenAPI lint checks no response bodies and its route check passes when route and spec are
  restored together. Stage 1 now adds a 405 case and a JSON-absence case per response struct.

leodip:
C

leodip:
## Decision 9 applied: the migration seeds the union, flagged rows plus every session of a user who has an authenticator

Answered **C**. Recorded, and here is the one thing your answer left to derive rather than choose,
stated plainly because it changes the SQL.

### C is a union with decision 8, not a replacement

Decision 8 seeds `-1` wherever `level2_auth_config_has_changed = 1`, and it stays in force. C's
recommendation said it "keeps A's seeded rows as a subset rather than replacing them", and that
sentence is only true if the seed carries both clauses.

C's clause taken alone would drop one population: a session flagged by its user's own **disable**.
The user removed their authenticator, the old writer flagged the caller's session, and `otp_enabled`
is now false, so a seed reading only `WHERE otp_enabled` would discharge exactly the obligation
decision 8 exists to hand over. So:

```sql
UPDATE user_sessions SET otp_config_generation = -1
 WHERE level2_auth_config_has_changed = <true>
    OR user_id IN (SELECT id FROM users WHERE otp_enabled = <true>)
```

`<true>` is `1` on sqlite, mysql and sql server and `true` on postgres, which declares both columns
`boolean` where the others use `numeric`, `tinyint(1)` and `BIT`. `= 1` is a type error on postgres
and `= true` is one on sql server, so the four migration files cannot share the literal. SQL Server
wraps the whole statement in `EXEC`, as migration 000029 does and for the reason it records.

The counter starts at 0 and only rises, so `-1` never matches for either population: each seeded
session is re-prompted once, then promoted like any other.

### What the migration test now pins

`TestMigration000031_OTPConfigGeneration` seeds **all four** combinations of the old flag against
`otp_enabled` rather than two flag states, and asserts that the three rows the union covers came out
at `-1` and the fourth, neither flagged nor OTP-enabled, came out at `0`. That fourth row is the
only case that distinguishes your answer from option B, so without it the test would pass against
either.

The mutation list gained the seed's two `WHERE` clauses **separately**. Dropping the first regresses
the seed to option C-alone and discharges what decision 8 preserves; dropping the second regresses
it to option A and leaves the bypass you just closed. Each has to fail the test on its own or the
seed is only half-pinned.

### State

No code was touched and nothing was committed, since no stage has started. Decision 9 is `Decided`,
decision 8 carries a `Widened by decision 9` line, and stage 2's plan no longer reads
`Blocked on decision 9`. The run is unblocked and stage 1, retiring the
`level2AuthConfigHasChanged` API surface, is next.

leodip:
## Decisions: nine settled, none deferred

**Nothing in this change rests on an assumption the run took to keep moving.** All nine decisions in
the agreement closed `Decided`. There is no deferred list, so there is nothing here you have to
ratify after the fact and nothing that would need revisiting if an assumption turns out wrong.

One decision could not be settled by the run and came to you on this thread.

### Decision 9: which pre-upgrade sessions does migration 000031 seed as owing a re-prompt?

Raised by the plan reviewer, answered by you on 2026-08-23 with `C`.

The plan had the migration seed only the sessions the old boolean had flagged (decision 8). The
reviewer showed that population is smaller than "the sessions that owe a re-prompt", and that the
gap is a **live bypass rather than a missed prompt**: a session authenticated at
`urn:goiabada:level2_optional` with `amr: ["pwd"]` while its user had no authenticator is never
flagged when that user later enables OTP somewhere else, because the old writer only touched the row
named by the caller's own `sid`. After the upgrade its counter and snapshot would match, both readers
would pass it, and it would keep obtaining tokens asserting a class it did not satisfy for the rest of
its life: up to 24 hours on the seeded defaults, longer wherever an operator raised them.

**What `C` settled.** Seed `-1` on every session of a user who currently has an authenticator, as
well as on the sessions the old flag marked. The two clauses are a union rather than a replacement,
which matters: a session flagged by its user's own *disable* belongs to a user whose `otp_enabled` is
now false, so C's clause taken alone would have discharged exactly the obligations decision 8 exists
to hand over.

```sql
UPDATE user_sessions SET otp_config_generation = -1
 WHERE level2_auth_config_has_changed = <true>
    OR user_id IN (SELECT id FROM users WHERE otp_enabled = <true>)
```

`<true>` is `1` on sqlite, mysql and SQL Server and `true` on postgres, because postgres declares both
columns `boolean` while the others use `numeric`, `tinyint(1)` and `BIT`. It is not a literal the four
migration files can share.

**What it costs, and what reversing costs.** Every user with an authenticator and a live session
answers the second factor once more at their next ceremony targeting above level 1. Once, then the
session is promoted like any other; the counter starts at 0 and only rises, so `-1` never matches
again. Users with no authenticator are untouched. The alternative left the bypass alive for up to one
session lifetime after the upgrade. Reversing it is a one-line change to the seed plus one assertion,
because `TestMigration000031_OTPConfigGeneration` pins the chosen behaviour on all four engines.

The other eight decisions were settled in the interview before the seal and are recorded in the
agreement with their reasoning.

leodip:
## Follow-ups: one to file, one folded in, one already tracked

Everything the run found next to the work, including the part it fixed rather than filed.

---

### 1. To file: only `authserver` has a local gofmt guard

Found while settling the red check suite on stage 1's commit `2de75465`. Filed rather than fixed
because no module other than `authserver` was implicated in that failure, and a repo-wide guard is
not this issue's to add unasked.

Duplicate search re-run at the finish: `gh issue list --state all --search "gofmt formatting lint
test"`, `--search "unformatted gofmt"` and `--search "gofmt guard module"`, all three empty over
every state.

**Title:** `Only authserver has a local gofmt guard; core, adminconsole and goiabada-setup can still fail CI on formatting`

**Labels:** `go`, `enhancement`

**Body:**

> `src/authserver/internal/server/gofmt_lint_test.go` holds every `.go` file in the `authserver`
> module to gofmt's canonical formatting, so an unformatted file fails the local unit tier. The
> other three modules the `Lint` job checks, `core`, `adminconsole` and `cmd/goiabada-setup`, have
> no equivalent, so for those an unformatted file is invisible until CI.
>
> That matters more than a formatting nit, because of how the job is shaped. In
> `.github/workflows/check.yml`, the `gofmt, vet, golangci-lint, unparam` step loops over the four
> modules and runs `gofmt -l .` first:
>
> ```
> unformatted=$(gofmt -l .)
> if [ -n "$unformatted" ]; then
>   echo "::error::$m has unformatted files:"
>   echo "$unformatted"
>   exit 1
> fi
> go vet ./...            || exit 1
> ```
>
> The `exit 1` leaves the subshell for that module, so an unformatted file also costs that module
> its `go vet`, `unparam` and `golangci-lint` run. A whitespace slip therefore silently reduces lint
> coverage rather than merely reporting itself, and the job's summary reads as one formatting
> complaint.
>
> Nothing local notices, for two compounding reasons: formatting is not a compile error, so no tier
> can fail on it; and the unit tier does not compile every directory in a module anyway. In
> `authserver` it runs `./internal/...` and `./web` (`run-tests.sh`), so `./tests/...` is never built
> by that tier at all.
>
> This is not theoretical. It happened on `2de75465`, where deleting the widest key from a set of
> struct literals left seven files over-indented, the `Lint` job went red, and `authserver` lost its
> vet, unparam and golangci-lint run on that commit. `authserver` now has a guard because of it; the
> other three modules are in exactly the position `authserver` was in before.
>
> **What closing it takes:** the same test, once per module. About thirty lines, no new dependency:
> walk the module's `.go` files, compare each to `go/format.Source`, which is the library `gofmt`
> itself is built on, and fail on a walk that finds no files so a wrong root cannot pass for the
> wrong reason. `src/authserver/internal/server/gofmt_lint_test.go` is the working copy to follow.
> The one judgement per module is where the file lives such that the module's unit tier actually
> reaches it, which differs: `core` and `adminconsole` legs run `./...`, so anywhere works there,
> while `cmd/goiabada-setup` needs checking.
>
> Alternatively, one guard could walk all four modules from a single home, which avoids four near
> copies at the cost of a test in one module reaching across module boundaries. Per-module
> duplication is the established pattern here (`credential_read_lint_test.go` exists once per module
> for the same kind of reason), so the four-copy version is the lower-surprise choice.

**File it with:**

```bash
gh issue create \
  --title "Only authserver has a local gofmt guard; core, adminconsole and goiabada-setup can still fail CI on formatting" \
  --label go --label enhancement \
  --body-file /path/to/body.md
```

---

### 2. Folded in: the `authserver` gofmt guard itself

`3eba17cb`. Stage 1 removed `Level2AuthConfigHasChanged:`, the widest key in a set of struct
literals, and gofmt aligns a literal to its widest key, so seven files were left indented to the
width of a name that no longer existed. The `Lint` job went red on `2de75465`.

`TestModuleSourcesAreGofmted` (`src/authserver/internal/server/gofmt_lint_test.go`) was added with
the whitespace fix so the local unit tier catches the next one. It fit inside the stage by the
fold-in test: it rides on a failure that stage caused, there was nothing to settle, and it is the
whole of the fix for the module that broke. Four of the seven affected files live under `./tests/`,
which the unit tier never compiled, which is why a guard was needed rather than just the reformat.

The three modules it does not cover are item 1 above.

---

### 3. Already tracked, not drafted: #136's root cause at a second call site

Stage 2's first data-tier run failed `TestGetUserByUsername` on mysql only. `createTestUser` names
users with `gofakeit.Username()`, and the never-reset shared `goiabada_data` already held a row from
2026-08-11 carrying the same value, so `GetUserByUsername` returned the older row and every field
comparison in `compareUsers` reported a mismatch. It passed on the immediate re-run and on every run
since.

That is exactly the condition **#136** describes ("Flaky: TestSearchUsersPaginated collides with
accumulated users in the never-reset shared databases"), at a different call site. Not drafted as its
own issue, because a second issue for one cause is a duplicate. The part worth adding to #136 if you
touch it: the collision is not specific to `TestSearchUsersPaginated`. Any lookup keyed on a
`gofakeit` value can hit it, and it presents as dozens of unrelated field mismatches rather than as a
duplicate.

Stage 2 adds six users to that database through `tests/data/otp_config_generation_test.go`. That
raises the collision odds marginally; it does not create the condition.

leodip:
## Review account

Two gates were reviewed by Codex: the plan before any code landed, and the whole branch once all
three stages had. This run set `LEO_REVIEW_MODE=end`, so **no stage was reviewed on its own** and the
final round is the only reading the code itself got. The verdicts live in a gitignored directory that
the next run overwrites, so this comment is the durable copy.

| Gate | Reviewer | Findings | Blocking | Outcome |
|---|---|---|---|---|
| Plan, round 1 | `gpt-5.6-sol` @ xhigh | 4 | 3 | 1 escalated to you (decision 9), 3 fixed in the plan |
| Final, round 1 | `gpt-5.6-sol` @ xhigh | 0 | 0 | clean, `static-review-exhausted` |

Nothing was refuted at either gate.

---

## Plan gate, round 1

Request `242-20260823T202029Z`, fresh Codex session. All three axes `reviewed`. The reviewer wrote and
ran its own scratch state model of the generation design rather than reading it, which is where
findings 1 and 2 came from, and ran four focused suites from the tree to establish that the current
code passes the assertions the plan proposed to rely on.

**1. Migration 000031 treats sessions the old writers never flagged as already current.**
*Blocking, security.* → **Escalated to you as decision 9**, answered `C`, see the decisions comment.

A pre-upgrade session at `acr: urn:goiabada:level2_optional` with `amr: ["pwd"]`, whose user later
enabled OTP from a different session, is never flagged, because the old writer touches only the row
named by the caller's own `sid` claim. Both new counters would then migrate to 0, they match, and both
readers pass the session. It keeps obtaining tokens asserting a class it did not satisfy for the rest
of its life: up to 24 hours on the seeded defaults, longer where an operator raised them. This is
evidence against decision 8's claim that its partial preservation "can only ever be one extra prompt
rather than one fewer" — here it is one fewer. Not resolvable by the run: the safer options cost a
user-visible extra prompt at upgrade, so it was a choice rather than a forced repair.

**2. Browser enrolment advances the counter after the value completion promotes was captured.**
*Blocking, conformance.* → **Fixed in the plan**, stage 2.

Capture N at `/auth/level2`, increment to N+1 at the enrolment, promote N at `/auth/completed`: a
session that has just established and verified an authenticator immediately owes another second
factor prompt, and `prompt=none` answers `interaction_required`. The repair was forced from two
sides. The reviewer's own live-read alternative failed its control by laundering an unrelated
concurrent change. And `IncrementUserAuthStateGeneration` in `commondb/user.go` is the same shape
already in the tree: it returns the value that landed, requires a transaction because "outside a
transaction another increment can land between them, and this caller would then return the OTHER
caller's generation", and its comment refuses computing the successor in Go. So
`IncrementUserOtpConfigGeneration` returns its new value and the enrolment writes it to the auth
context inside the same transaction.

**3. The migration backfill was left untested on a false premise.** *Blocking, quality.* → **Fixed in
the plan**, stage 2.

The agreement held that no harness can seed rows before a migration runs. `newIsolatedDB` at
`tests/data/migration_testdb_helper.go` does exactly that, and eight migrations from 000021 onward
already have such a test. Stage 2 gained `TestMigration000031_OTPConfigGeneration` on four engines:
migrate to 000030, seed both old flag states, apply, assert values and column shapes, then down and up
again for SQL Server's default constraints.

**4. Stage 1 could not tell the retirement from restoring the whole API surface.** *Significant,
quality.* → **Fixed in the plan**, stage 1.

`openapi_contract_lint_test.go` says in its own header that it checks no response bodies or schemas,
and its route test is two-way, so restoring the route and the spec operation together passes. Stage 1
as planned deleted the only PUT tests and the only field assertion and added nothing that failed on
the pre-stage tree. It now adds a 405 case at the endpoint and a JSON-absence case per response
struct, with three mutations.

The verdict's `unchecked` array held three entries, all answered in the plan log: two dismissed to the
stage gates that own them, one dismissed as untouched surface.

---

## Final gate, round 1: the whole branch

Request `242-20260823T222559Z`, written 2026-08-23T22:25:59Z, verdict 2026-08-23T22:40:10Z. Scope was
the whole branch, `0c558525...cb599d10`, 79 files, +3187 −1031. The request named no area as worth
attacking: the reading is worth paying for precisely because it arrives without the author's picture
of which part is interesting.

**Verdict: `static-review-exhausted`, zero findings.** Conformance, quality and security all reported
`reviewed`. No follow-ups and no bookkeeping items raised.

### What the reviewer executed

Recorded because a clean verdict is worth exactly what the work behind it was worth.

1. Validated all six recorded tree hashes and 22 of 22 contract anchors.
2. Confirmed the exact final commit's four-engine data and integration tiers, the three module unit
   suites, lint, vulnerabilities, build validation and version checks are green.
3. Ran targeted unit, sqlite data and sqlite integration tests, including both admin OTP disable paths.
4. Ran four counter and snapshot mutations: both `dont-update` guards, the increment changed from
   `+ 1` to `+ 0`, and a neutralised promoted value. Every owning test failed as expected, and each
   mutation was restored.
5. Ran an encrypted `ChunkedCookieStore` round-trip probe for the OTP seed and QR image, confirming
   neither plaintext appears in `Set-Cookie`.
6. Ran a scratch ceremony-carryover mutation copying an old ceremony's seed into a new auth context.
   `TestAuthOtp_SecondAuthorizeInTheSameBrowserGetsANewSecret` caught it.
7. Checked protocol behaviour against OpenID Connect Core 1.0 sections 2 and 3.1.2.1, and the touched
   OAuth surface against RFC 9700. No conflict found.

### The `unchecked` array, answered

Three entries, three answers. An unanswered entry is one the next round rediscovers at the price of a
full round.

**1. No scheduler-controlled two-writer database interleaving test was added. → Dismissed.**
The property such a test would pin is guaranteed by the statement rather than by scheduling.
`IncrementUserOtpConfigGeneration` in `src/core/data/commondb/user.go` issues one
`UPDATE users SET otp_config_generation = otp_config_generation + 1 WHERE id = ?`, a database-side
read-modify-write serialised by the row lock on all four engines, so two concurrent enrolments cannot
lose an increment. It refuses a nil `tx` outright, errors unless exactly one row matched, and reads
the landed value back inside the same transaction rather than computing `N+1` in Go, which is the case
a race would otherwise corrupt. The direction of any residual skew is safe: the level 1 reader compares
with `!=` and not `<`, so a snapshot somehow ahead of the counter re-prompts rather than passing. The
data tier already proves transaction enlistment against the real implementation on four engines by
rolling one back and reading the column unmoved. A goroutine-scheduled test would assert Go's
scheduler, not the engine's lock, and would be the flakiest file in the tree.

**2. No manual visual browser pass was performed. → Dismissed.**
The one browser-only risk this change could plausibly introduce is cookie inflation, since the
enrolment seed and its QR image moved onto `AuthContext`, and a Go cookie jar does not enforce browser
cookie limits where a real browser does. It is not a new risk: `SessionKeyOTPSecret` and
`SessionKeyOTPImage` were slots in the **same browser session** as the auth context, so those bytes
already rode in the same cookie and this change relocates them inside the same payload rather than
adding to it. Measured rather than assumed: the real `OTPSecretGenerator` produces a 32-byte secret
and a 2844-byte base64 QR image for a long email at a long app name. `ChunkedCookieStore` splits at
`DefaultChunkSize` 3800, under the 4096-byte per-cookie limit, with `DefaultMaxChunks` 50, giving a
190KB ceiling this payload uses two or three chunks of, and it **errors** at
`chunked_cookie_store.go:158` rather than truncating when a session would exceed it. The remaining
visual question, that the QR still renders, is held by the strict `len(bind) != 10` assertion in the
OTP GET test plus the integration cases that read the secret out of the rendered page. The one
genuinely browser-only behaviour left, a reload served from the browser's own cache rather than the
server, is #247's subject by name and out of scope.

*(Measurement method: a scratch `TestQRSizeProbe242` in `src/core/otp/`, run and then deleted. The
tree was re-hashed afterwards and `production 2291e7d75e24`, `unit-tests e59b2b6f9eb9`,
`data-tests 25c05d785330`, `integration-tests f38edeba5df9`, `docs 2f2efaf2d127`, `all d1fc3e458311`
all still equal the values `cb599d10` was reviewed at, with `git status` clean. Nothing scratch
survived.)*

**3. Settled neighbouring issues were not reopened. → Dismissed**, and intended. #144, #142, #247, the
stale `amr` a disabled authenticator leaves on a reused session, and #203 are each out of scope with a
reason, and the request declared them settled ground. The reviewer states no new evidence connected
any of them to this branch, and the ingest found none either.

### Disposition

Nothing to fix, defer or refute. No production code changed at this gate, so the six tree hashes above
stand unmoved from the ones the round was reviewed at, and the four-engine claim remains the check
suite's on `cb599d10`
([run 32670209981](https://github.com/leodip/goiabada/actions/runs/32670209981)).

Issue #205, opened 2026-08-11
The server accepts unbounded input before anything bounds it: no MaxHeaderBytes, no DCR body limit, and a debug middleware that buffers before authentication


leodip:
Both servers construct `http.Server` without setting `MaxHeaderBytes`, so Go's 1 MB default applies: an unauthenticated client can send about a megabyte of request line and headers on every request, and `net/http` allocates several times that parsing it before any handler or middleware runs.

Four `&http.Server{}` literals leave it unset:

- `src/authserver/internal/server/server.go:135` and `:151`
- `src/adminconsole/internal/server/server.go:116` and `:130`

Each sets `Handler`, `Addr` and the TLS config and nothing else. `MaxHeaderBytes` appears nowhere in `src/`.

That default is the input bound behind every cost measured at #159's final review: `url.ParseRequestURI` alone allocates 5.41 MB for a 900000-byte high-byte path, inside `net/http`, before any Goiabada code runs.

#159 bounded what its own request logger does with such a request and deliberately did not change what the server accepts, because that is a behaviour change with its own compatibility question: a long `redirect_uri` alongside a large `Cookie` header is a legitimate request, and the bound is a number somebody has to choose.

## This was already identified in #39, and the header cap is only half of it

#39 was closed as not planned, but its closing analysis reached this exact code and left it deliberately unfiled:

> `http.MaxBytesReader` appears only in the three image upload handlers, and the `http.Server` instances at `server.go:135,151` set no `MaxHeaderBytes`. So the GET path is bounded around 1MB by Go's default header limit, while the POST path is bounded only by net/http's form defaults. That is the actual resource-exhaustion surface, it covers the whole body rather than two parameters, and it would be addressed by middleware rather than by length validation. Left unfiled for now; worth its own issue if request-size limits become a priority.

So there are two halves and `MaxHeaderBytes` closes only the first:

1. **Request line and headers.** Set `MaxHeaderBytes` on all four `http.Server` literals.
[...]

Issue #190, opened 2026-08-09
Security: the API debug middleware buffers the whole request body before authentication and before the handler's size limit


leodip:
When `GOIABADA_AUTHSERVER_DEBUG_API_REQUESTS` is enabled, `APIDebugMiddleware` reads the entire request body into memory with `io.ReadAll` before any authentication middleware runs and before any handler installs `http.MaxBytesReader`. An unauthenticated request to any route under `/api/v1/admin` or `/api/v1/account` is therefore buffered in full, whatever its size, and only then rejected with a 401. The response side buffers without a bound too, through `responseWriter.Write` teeing into a `bytes.Buffer`.

`APIDebugMiddleware` is the first `r.Use` on both mounts in `src/authserver/internal/server/routes.go`, ahead of `RequireBearerTokenScope`, `RequireUserBoundToken` and `RequireValidSession`. The three handlers that do limit a body install `http.MaxBytesReader` inside themselves (`grep -rn "MaxBytesReader" --include=*.go src/ | grep -v _test.go` returns 3 hits, all in profile-picture and client-logo handlers), which is strictly after the middleware has already read it.

This is behind a flag that is off by default, which is what kept it out of #145. #145 bounds what is formatted into the log at 256 KB, but the read has already happened by then.

Fix by bounding the read in the middleware itself, with `http.MaxBytesReader` or an `io.LimitReader`, and logging a placeholder when the limit is hit. A limit in the middleware also has to leave the handler's own `MaxBytesReader` working, since the handler replaces `r.Body` afterwards.

Related: #93 is the same class on `/connect/register`, a different endpoint and a different code path.

leodip:
Merged into #205, which now covers all three places the server reads unbounded input before anything bounds it: the missing `MaxHeaderBytes`, the DCR body, and this middleware. The detail is carried over in full, including that it is behind a flag that is off by default and that #145 bounded what is formatted rather than what is read.

Closed during a backlog consolidation pass.

PR #256, opened 2026-08-24
fix(security): refuse caching on credential-bearing responses, and bind OTP enrolment

### LINKED ISSUE #247: Ceremony pages carry no cache directives, and the OTP enrolment page renders a plaintext TOTP seed

leodip:
Fixes #247.

## What this is for

Two defects around the same credential, the TOTP enrolment seed.

**Nothing Goiabada renders tells a cache not to keep it.** `RenderTemplate` writes `Content-Type`
and nothing else, and the security-headers middleware sets five headers, none of them a cache
directive. So the password form, the OTP form, the consent screen and the OTP enrolment page, which
displays the authenticator seed in plain text and its QR code as an inline PNG, all go out as a 200
that a cache is permitted to store. RFC 9111 section 3 allows storing a response with no directives
at all, since 200 is heuristically cacheable, and section 4.2.2 says origin servers should send an
explicit directive if they wish to prevent it. RFC 6749 section 5.1 is stronger than that and
unqualified as to endpoint: the server MUST send `Cache-Control: no-store` and `Pragma: no-cache` on
"any response containing tokens, credentials, or other sensitive information". The seed is served by
three surfaces, not one, and the Admin API's `GET /api/v1/admin/clients/{id}` returns a **decrypted**
client secret with no directive either.

**The account API enrols whatever seed the caller sends.** `PUT /api/v1/account/otp` matches the
submitted code against `req.SecretKey`, the caller's own value, and then stores that value as the
user's authenticator. The server never checks it issued it. `GET /api/v1/account/otp/enrollment`
also mints a fresh secret on every call, so reloading an enrolment page silently invalidates a QR
code the user has already scanned, which is the defect #242 fixed for the `/auth/otp` ceremony and
left live on this surface.

## The rule this establishes

A response carrying a credential is never storable, and the directive is written in one place per
mechanism so a page or a route added later inherits it rather than shipping uncovered:
`RenderTemplate` for every HTML page either module renders, and a middleware mounted first on both
`/api/v1` groups for every API response, refusals included. A test fails if a new route escapes it.

And the server stops trusting a caller-supplied TOTP seed: the enrolment GET records what it issued,
the PUT enrols that seed and no other, and the GET is idempotent while an enrolment is pending.

## Breaking change

`PUT /api/v1/account/otp` no longer accepts `secretKey` and refuses a request carrying one with 400
`SECRET_KEY_NOT_ACCEPTED`. Every existing integration needs updating on the day this ships. The
refusal is loud by choice: silently ignoring the field would leave an integrator wondering which
secret was enrolled. The admin console is updated in this change. Upgrade guidance is under
`## For the release notes` at the end of this body.

## Status

**Complete.** All five stages have landed, the whole branch has been reviewed and the check suite is
green on the final commit. Every goal in the agreement is met:

- every rendered HTML page in both modules, and every response from an endpoint under
  `/api/v1/admin` and `/api/v1/account`, carries `Cache-Control: no-store` and `Pragma: no-cache`,
  written in one place per mechanism so a page or route added later inherits it;
- a test fails if a new route escapes it;
- the enrolment QR no longer occupies the browser's cookie, and a spent enrolment key no longer
  travels through the rest of the ceremony;
- the account API records the TOTP enrolment it issued and enrols that seed and no other,
  idempotently while one is pending.

**This change is breaking.** `PUT /api/v1/account/otp` no longer accepts a `secretKey`. See
"For the release notes" below.

### The closing record

Three comments carry what the diff cannot say for itself:

- **[Deferred decisions](https://github.com/leodip/goiabada/pull/256#issuecomment-5395131830)**:
  none. All sixteen decisions are settled; the one escalated to you, decision 16, is recorded there
  with the cost it accepts and what reversing it would take.
- **[Follow-ups](https://github.com/leodip/goiabada/pull/256#issuecomment-5395143016)**: one issue
  to file, ready to paste; one finding already tracked as #205; four things folded in rather than
  filed, each with its commit; and one dropped by your decision.
- **[The review account](https://github.com/leodip/goiabada/pull/256#issuecomment-5395157742)**:
  every round, every finding, and what answered it, with the evidence.

### Stage 1: rendered HTML carries the pair, and the 500 page stops committing its status

Landed in `35209225`.

`RenderTemplate` sets `Cache-Control: no-store` and `Pragma: no-cache` beside the `Content-Type` it
already wrote, after the template has rendered successfully into its buffer. That covers all 136
render sites across both modules at once, including the two password forms outside `/auth` and the
admin console's own OTP page, and it structurally cannot reach static assets, images, JWKS or
discovery, which stay cacheable. Bare `no-store` is the value both RFC 6749 5.1 and OAuth 2.1 3.2.3
name; `Pragma` goes with it because RFC 6749 5.1 makes both header fields a MUST, and because the
nine places in the tree that already send a cache directive send exactly this pair.

Folded in: `InternalServerError` was calling `w.WriteHeader` before rendering, which freezes the
header map, so every header `RenderTemplate` set afterwards was dropped. The 500 page has been
shipping without its `Content-Type` in both modules, and would have shipped without the new
directives. The status now travels in the bind map as `_httpStatus`, which `RenderTemplate` already
understood, and the `http.Error` fallback stops writing a header that has already been sent.

Every header assertion in the helper's tests moved from `httptest.ResponseRecorder.Header()` to
`w.Result().Header`. The live map and the snapshot disagree in exactly the `WriteHeader`-first case
above, so the existing `Content-Type` assertion on the 500 page was green for a header that never
shipped.

Tiers: unit (3618 pass) and integration on sqlite (1273 pass), the latter including two new
end-to-end cases that require both header fields on the password form and on the OTP enrolment page
as a client receives them. Four mutations, all caught, including one that reintroduces the
`WriteHeader`-first defect and one that hoists the header write above the error return.

Docs: one bullet in `site/src/content/docs/reference/security.mdx`, scoped to rendered pages, saying
what is sent and honestly what it binds. `no-store` binds HTTP caches; the browser's history and
back/forward cache are left free to redisplay a page, so it is not a promise that nothing is
retained.

### Stage 2: every response under `/api/v1` carries it

Landed in `e3350c1e`.

A new `MiddlewareNoStore` in `src/core/middleware/`, mounted **first** in the `/api/v1/admin` and
`/api/v1/account` groups, so every response under both, refusals included, carries
`Cache-Control: no-store` and `Pragma: no-cache`. Two of those responses hand out a credential
outright: `GET /api/v1/account/otp/enrollment` serves a TOTP enrolment seed, and
`GET /api/v1/admin/clients/{id}` returns a client secret decrypted. Both went out with
`Content-Type` and nothing else.

**Why a middleware rather than a header inside the JSON encoder.** The API writes success bodies
two ways, and 47 of its `EncodeJson` call sites commit the status with `w.WriteHeader` before
encoding. A header set after the status is committed is silently dropped, so an encoder-level write
would have been lost in those 47 places with every test still green. A middleware runs ahead of the
handler and is immune to that.

**Why first in the group is a requirement and not a preference.** The 401 and 403 refusals are
written by `emitAuthError`, which sets its own `Content-Type` and calls `WriteHeader`. Anything
mounted behind the guards would never write the pair on a refusal.

**The guard against a route added later is structural, not a list.**
`TestInitRoutes_EveryApiRouteRefusesToBeStored` builds the real router, enumerates it with
`chi.Walk` and calls all **105** registered `/api/v1` routes unauthenticated, so a route or a whole
group added under that prefix is covered without the test being edited. Worth knowing: a mutation
moving the mount later *within* a group passed that sweep, because no unauthenticated request
reaches the middleware in between and `RequireValidSession` writes refusals of its own from there.
The test now also asserts `MiddlewareNoStore` is first in each route's chain, which closes it.

**What is deliberately not covered.** Four responses produced *before* routing still ship with
neither header field: a CORS preflight, a settings-lookup 500, a cookie-reset redirect and a
session-identifier error. None carries a credential, so RFC 6749 section 5.1's MUST does not reach
them. That was put to the user as a decision and answered; the sweep states the limit in its own
header comment rather than implying a coverage it does not have.

Tiers: unit 3728 pass across three modules, data 1691 pass on all four engines, integration 1275
pass on sqlite. Seven mutations, all caught, two of them only after the positional assertion above
was added.

Docs: the security reference bullet widens from rendered pages to the API responses too, the REST
API integration page gains a `### Caching` section, and the OpenAPI `info.description` gains a
`## Caching` block. No per-operation `headers:` blocks: all 95 operations send the pair, and
declaring it on the two credential-bearing ones would read as though only those two did.

### Stage 3: the QR leaves the cookie, and the spent seed leaves the ceremony

Landed in `5bfd9c28`.

The enrolment ceremony carried the TOTP seed and a 2.4 KB base64 PNG of its QR code as two fields on
the auth context, which is an encrypted cookie. That cost an entire extra cookie chunk and about
5 KB on every request to the origin for the duration of the enrolment. Neither field was ever
cleared, so a spent credential then travelled through `/auth/completed`, `/auth/consent` and
`/auth/issue`, and sat in the cookie of a ceremony abandoned after enrolling until a later
`/auth/authorize` replaced the context.

One field now holds the key's `otpauth://` URL, about 150 bytes, and both the secret and the image
are derived from it: `otp.SecretFromKeyURL` for the verifier and `otp.RenderQRCodeImage` for the
page. Two fields could come to disagree about which key a user scanned; one cannot. Measured with
the real `ChunkedCookieStore` at its production chunk size, the context falls from 3693 bytes over
3 cookies and a 7058-byte `Cookie` request header to 1016 bytes over 2 and 2274.
`HandleAuthOtpPost` then blanks that field before the `SaveAuthContext` that advances the state,
the same discipline as `User.SetOTPSecret`, which blanks the plaintext seed once it has encrypted
it (#82).

**The library validates nothing, so this code does.** `otp.NewKeyFromURL` runs `url.Parse` and keeps
whatever comes back: it accepts `""`, answers `Secret()` with `""`, and will happily render a
scannable QR code of the empty string. Trusting it would mean verifying passcodes against an empty
secret and showing a user a QR code that enrols their authenticator in nothing. The new
`parseKeyURL` requires a non-empty argument, a `totp` key type and a non-empty secret, and each
refusal case asserts that the library **accepts** the same input, so the tests record why the guard
exists rather than only that it fires.

**A present-but-unusable value regenerates rather than failing.** #242's sentinel was "empty means
this ceremony has generated nothing". A structured field has a second failure two flat strings did
not, and refusing it would wedge the ceremony on a 500 that every reload repeats. The enrolment GET
therefore regenerates when the stored URL will not parse as well as when it is absent. #242's
idempotence is untouched: a parseable URL is never replaced, and a mutation forcing regeneration on
every GET is caught.

Tiers: unit green on all three modules, 3738 ok 0 fail; data green on sqlite, mysql, postgres and
mssql, 1691 ok 0 fail; integration green on sqlite, 1275 ok 0 fail. The load-bearing one is that
`auth_otp_enrolment_secret_test.go` passes **unchanged**: it reads the secret off the rendered page,
generates a code from it and submits it, so it proves the seed a user scans is still the seed the
server verifies now that both come from a URL rather than sitting side by side. Six mutations, all
caught.

One existing handler test changed what it pins, and it is worth knowing why. It set a secret with no
image, a state no GET ever wrote, which the old two-field template condition read as "not
enrolling". With one field that inconsistency cannot be expressed, so the case now pins the branch
that does remain: a ceremony carrying no key falls back to the verification page and binds neither
the image nor the secret.

### Stage 4: a pending enrolment on the user row

Commit `6915ce5b`. The storage the account API's new contract needs. Nothing reads it yet; the
endpoints change in stage 5.

Migration `000032` adds two nullable columns to `users`, `otp_enrollment_secret_encrypted` and
`otp_enrollment_issued_at`, in the shape of the three pending-credential pairs already on that
table: an encrypted blob beside an `issued_at`, dormant as `NULL`, cleared on success. Four engine
directories, plus the four `schema.sql` snapshots.

**The ciphertext is the whole `otpauth://` URL, not the bare base32 seed**, and that was settled by
running it rather than by argument. A repeat enrolment GET has to return the same QR image;
`otp.RenderQRCodeImage` renders one only from an `otpauth://` URL and refuses a bare secret with
`otp key url is not a totp key`; and nothing in `core/otp` rebuilds a URL from a seed. Adding
something that did is the hazard weighed and rejected earlier in this change, because a rebuilt URL
that does not reproduce `totp.Generate`'s defaults exactly yields a QR scanning to a different
configuration than the server verifies against. The column keeps the name it was given; the model
field's doc comment says which of the two values it carries.

Both model fields are tagged `dont-update`, for the reason `AuthStateGeneration`, `LastOTPStep` and
`OtpConfigGeneration` are. `UpdateUser` writes every field not tagged `pk` or `dont-update`, and
fourteen production sites load a user and later write it back, so in the ordinary update set a model
loaded before issuance would erase the seed installed in between, and two concurrent enrolment
requests would each invalidate the other's QR code.

They move only through two narrow writes on the `Database` interface.
`TryInstallPendingOTPEnrollment` is one conditional `UPDATE` whose `WHERE` requires the row, requires
`otp_enabled` false so a seed cannot be parked on an already-enrolled account, and requires the
existing pending value to be absent or issued before a caller-supplied cutoff. It reports whether it
won through `RowsAffected`, which is what lets concurrent callers all answer with one seed rather
than several. `ClearPendingOTPEnrollment` returns the pair to `NULL` with a literal `NULL` rather
than a bound nil, because sqlbuilder sends an untyped Go nil that SQL Server types `nvarchar` and
then refuses to convert to `varbinary(max)`.

`EnableUserOTPTx` calls the clear inside the transaction it already owns, so no committed state has
an authenticator enabled with a live pending seed still behind it, and that covers the browser
ceremony as well as the account API.

The encrypted column joins `commondb.aesProtectedColumns`, without which its ciphertext would
survive a data-key rotation decryptable only under the retired key, and both fields join
`api.sensitiveUserFields` so neither can reach a `UserResponse`.

Tiers: unit 3928 pass; data 1723 pass on mysql, postgres, mssql and sqlite; integration 1275 pass on
sqlite. Integration covers no new case here, since nothing writes these columns until stage 5; it
runs because `EnableUserOTPTx` gained a statement and that deserves a real engine. Eight mutations
against the new tests, all caught, including dropping either `dont-update` tag, dropping the
`otp_enabled` term, widening the expiry cutoff, and removing the column from the rotation registry.

### Stage 5: the account API enrols only the seed it issued

`71fd5e98`. The endpoints, the admin console that calls them, and the published contract. Stage 4
built the storage and nothing read it; this reads it.

`GET /api/v1/account/otp/enrollment` became the issuing step rather than a secret generator. It
answers with the enrolment already pending and mints one only when there is none, which is the
idempotence: reloading an enrolment page no longer replaces the seed behind a QR code the user has
already scanned. The install is a compare-and-set, so two concurrent calls agree on one QR code, and
the one that loses reads the row back and answers with the winner's seed rather than handing out a
second the PUT could never accept. A pending enrolment lasts 15 minutes.

`PUT /api/v1/account/otp` loads that seed, checks the code against it and enrols it. With no live
pending enrolment it refuses with `OTP_ENROLLMENT_NOT_PENDING` rather than falling back to anything,
so a partially migrated or rolled back deployment fails closed. `OTP_CODE_AND_SECRET_REQUIRED`
became `OTP_CODE_REQUIRED` and `INVALID_OTP_SECRET` retired, unreachable now that a server-issued
seed is well formed by construction. The clear rides in the transaction `EnableUserOTPTx` already
owns, so no committed state has OTP on with a live pending seed behind it.

A request still carrying `secretKey` is refused rather than ignored. The check reads the raw body
rather than the struct, so `"secretKey": null` and a non-string value refuse exactly as a string
does, and it runs before the user is loaded and applies to a disable request too: it is a statement
about the shape of the request, and an upgrading integrator needs the reason rather than an
authentication failure. One consequence worth naming: such a request no longer spends the shared
credential-failure budget, which is correct, since it never reaches a password comparison.

The admin console deletes both hidden inputs and stops sending the seed. Those two form reads were
the enable form's only source, so every rerender now calls the enrolment GET again, which is
idempotent and therefore redraws the QR code the user has already scanned. Without that the page
would still have rendered, with an empty image and an empty `<pre>`, and the user could not have
finished enrolling. An enrolment that succeeded elsewhere reloads the page instead of redrawing a
form the enrolment endpoint would refuse to feed.

Folded in: a 64 KB bound on the PUT's body, which is now read whole because it is parsed twice.

Two comments justifying the unbounded wrong-code path claimed the code was "verified against the
secret the caller supplied in the same request". That premise is false after this change; the
conclusion survives on a true one and both were restated rather than left standing.

Unit 3765, data 1723 on four engines, integration 1281 on sqlite. Ten mutations against the new
tests, all caught: the GET ignoring its pending value, a pending value never expiring, a lost
compare-and-set ignored, the `secretKey` refusal skipped, the no-pending refusal skipped, a seed
other than the issued one stored, the blank-code refusal skipped, the console rerender binding an
empty QR, the already-enabled redirect removed, and the hidden `secretKey` input put back.

New coverage: the console's `accounthandlers` package had no handler test at all before this, and
`rendertest` gained the only seam that can see whether a form carries a hidden input.

## How this was verified

**Locally, on the final tree:**

| Tier | Engines | Result |
|---|---|---|
| unit, three modules | n/a | 3956 pass, 0 fail |
| data | sqlite, mysql, postgres, mssql | 1723 pass, 0 fail |
| integration | sqlite | 1281 pass, 0 fail |

**On all four engines, by the check suite, which is where that claim comes from.** Local integration
runs on sqlite only, by policy, so nothing in this run asserted a four-engine integration result from
a session. CI is green on the final commit `4a87e383` across all eleven jobs, `Tests / sqlite`,
`Tests / mysql`, `Tests / postgres`, `Tests / mssql`, `Unit / authserver`, `Unit / adminconsole`,
`Unit / core`, `Lint`, `Vulnerabilities`, `Versions` and `Build validation`:
[run 32723360876](https://github.com/leodip/goiabada/actions/runs/32723360876). Every stage commit
was green in its own right too, so no stage landed on a red suite and was repaired afterwards.

**Thirty-eight mutations were run against this change's new tests** and every one is caught: four in
stage 1, seven in stage 2, six in stage 3, eight in stage 4, ten in stage 5, and three in the review.
Two of those three were the reviewer's own and had **survived** when it ran them, which is what
turned two of its findings from opinions into demonstrations. Each mutation ran through a harness
that restores from a trap and compares the file by hash afterwards, so no mutation residue reached a
commit.

### Review

Reviewed by `gpt-5.6-sol` at effort `xhigh`, three rounds: one over the plan before any code was
written, and two over **the whole branch against `origin/main`**. No stage was reviewed on its own.
Nine findings in total, all nine verified against the code, none refuted.

|---|---|---|---|---|
| plan | 1 | 4 | 2 | 3 fixed in the plan before code existed, 1 escalated as decision 16 |
| final | 1 | 4 | 0 | all 4 fixed, `4a87e383` |
| final | 2 | 1 | 1 | mechanism confirmed, attribution refuted, already tracked as #205 |

- **blocking, security:** the pending-enrolment columns as plain model fields were a read-then-write
  race, and fourteen whole-row user writes would have clobbered them. Fixed before stage 4 was
  built: both fields `dont-update`, a compare-and-set install, the loser reading back the winner's
  seed.
- **blocking, conformance:** deleting the console's hidden inputs would have left every enable-error
  rerender with a broken QR image and an empty seed, with no test in that package to catch it. Fixed
  before stage 5 was built: the rerender calls the idempotent enrolment GET, and the package gained
  its first handler tests.
- **significant, security:** the encrypted pending seed was missing from the key-rotation registry;
  the reviewer's probe showed a live enrolment would have survived rotation unreadable. Fixed in
  stage 4 with a rotation fixture that must decrypt only under the new key.
- **significant, security:** the unauthenticated route sweep cannot see a handler that sets its own
  `Cache-Control`, and the credential-bearing client-detail response had no cache assertion. Fixed in
  `4a87e383`; the reviewer's surviving mutation is now caught.
- **significant, security:** the folded-in 64 KB body bound had no boundary test, and deleting the
  cap left the module suite green. Fixed in `4a87e383` with a case whose body is valid JSON, which is
  the only shape that separates the two outcomes.
- **significant, conformance:** three tests proved storage held the right ciphertext rather than that
  a caller can authenticate with the installed authenticator. Fixed in `4a87e383`: the success case
  now drives a real level 2 ceremony with a code from the issued seed.
- **blocking, security (round 2):** `APIDebugMiddleware` buffers the whole request body before
  authentication. Real, and **not this change's**: that file is byte-identical to `origin/main`, its
  pre-authentication mount order pre-dates this branch, and `MiddlewareNoStore` never touches
  `r.Body`. Already tracked as #205, which #190 was closed into.

[this comment](https://github.com/leodip/goiabada/pull/256#issuecomment-5395157742).

## Deviations worth knowing

- **The scope was widened before the run started.** Two follow-ups drafted at the first seal, the
  account API enrolling a caller-supplied seed and the QR living in the cookie, were folded in on
  your decision, taking the full contract fix over the contained subset with the breakage stated.
  That is what makes goals 6 to 8, stages 3 to 5, and the breaking change.
- **Goal 4 was narrowed after it was sealed**, from every response under `/api/v1` to every route the
  router registers. Four responses produced *before* routing carry neither header field, none of them
  carries a credential, and the route sweep now states that limit in its own header comment rather
  than implying a coverage it does not have. That was decision 16 and you answered it.
- **Three of five stages were rewritten by the plan review before any code was written**, which is
  the point of reviewing a plan: the concurrency tags and compare-and-set of stage 4, the console
  rerender of stage 5, and the key-rotation registration. None of it changed what the agreement set
  out to do.
- **One test rewrite in the review changed what is pinned, not just how.** The account API's
  success case used to decrypt the stored column; it now drives a browser ceremony and submits a code
  from the issued seed. The column assertions were not lost, they already lived in the tier that runs
  on all four engines.
- **Four fixes were folded in rather than filed**, and each is named with its commit in the
  [follow-ups comment](https://github.com/leodip/goiabada/pull/256#issuecomment-5395143016), because
  the goal does not mention them: the two scope additions above, the 500 page committing its status
  before rendering, and the 64 KB bound on the OTP PUT's body.

## For the release notes

**`PUT /api/v1/account/otp` no longer accepts a `secretKey`, and refuses a request carrying one.**

The server now issues the TOTP enrolment, records it, and enrols that seed and no other. A caller
could previously enrol a secret the server had never generated, because the code was verified
against the secret the same request carried and that value was then stored.

What to do: call `GET /api/v1/account/otp/enrollment` first, show the user the QR code and secret it
returns, and send only `enabled`, `password` and `otpCode` to the PUT.

What breaks if you do not: the request is refused with `400 SECRET_KEY_NOT_ACCEPTED`. It is refused
rather than ignored deliberately, so an integrator learns at once instead of wondering which secret
was enrolled.

Also changed on that endpoint:
- `OTP_CODE_AND_SECRET_REQUIRED` is now `OTP_CODE_REQUIRED`.
- `INVALID_OTP_SECRET` no longer occurs.
- `OTP_ENROLLMENT_NOT_PENDING` is new: no enrolment was started, or the one started has expired.

`GET /api/v1/account/otp/enrollment` is now idempotent while an enrolment is pending, returning the
same secret and QR code for 15 minutes. If your client assumed each call minted a fresh secret, it
no longer does, and that is the fix: repeat calls used to invalidate a QR code the user had already
scanned.

The admin console is updated in this same change and needs nothing from an operator.


leodip:
## Decision 16: does the API's cache directive run ahead of the global middleware chain, or only inside the route groups?

The plan review for #247 found a gap between what the agreement's goal 4 says and what the mechanism
it picked can reach. Answering **A** or **B** below unblocks the run; stage 1 is unaffected either way
but the guard stops the whole run on an open decision, so nothing proceeds until this is settled.

### What the change was going to do

Goal 4 of this issue: *"Every response under `/api/v1/admin` and `/api/v1/account` carries it too, so
the decrypted client secret on `GET /api/v1/admin/clients/{id}` stops going out storable alongside the
seed."* The mechanism chosen for it was a new `MiddlewareNoStore` writing `Cache-Control: no-store`
and `Pragma: no-cache`, mounted **first** inside each of the two `/api/v1` route groups in
`routes.go`, ahead of the auth guards, so refusals carry the pair as well as successes.

### What the review found

Four middleware are mounted **globally**, in `Server.initMiddleware`, and each can answer a request
without ever calling `next`. A group mount sits inside `initRoutes`, which those responses never
reach, so none of them gets the pair:

| Path | How it answers early |
|---|---|
| CORS preflight to any `/api/v1` path | `MiddlewareCors` is mounted **first**, before everything. Its `AllowOriginFunc` allows only `/.well-known/openid-configuration`, `/certs`, `/auth/token`, `/auth/logout` and `/userinfo`, and go-chi's CORS handler ends a preflight itself unless `OptionsPassthrough` is set. Result: **200, no CORS headers, no cache headers.** |
| `MiddlewareSettings` | returns a 500 on a settings lookup failure |
| `MiddlewareCookieReset` | redirects on a cookie it cannot decode |
| `MiddlewareSessionIdentifier` | returns on a session store or database error |

The reviewer composed the real `MiddlewareCors` in production order around the planned group mount and
got `preflight=true status=200 cache="" pragma=""`. I verified the registration order and each early
return in the source independently.

There is a second consequence. The planned router sweep, which walks the real `chi` route tree and
asserts every `/api/v1` route carries the pair, builds only `initRoutes`. It therefore cannot see any
of the four, so describing it as exhaustive over the response surface was too broad however this is
answered. That part is being corrected regardless.

### What the specification actually says

RFC 6749 section 5.1:

> The authorization server MUST include the HTTP "Cache-Control" response header field [RFC2616] with
> a value of "no-store" in any response containing tokens, credentials, or other sensitive
> information, as well as the "Pragma" response header field [RFC2616] with a value of "no-cache".

OAuth 2.1 draft -13 section 3.2.3 keeps that scope word for word and drops `Pragma`.

**The MUST is scoped to responses that contain a credential.** A CORS preflight has an empty body. So
do the settings 500, the cookie-reset redirect and the session error. **The spec requires neither
option and permits both.** This is a genuine choice about how literally to read goal 4's own sentence,
not a conformance obligation, and I am not treating it as one.

### Option A: keep the group mounts, and say honestly what is covered

Leave the mechanism exactly as agreed. Restate goal 4 and the router sweep to own **every route the
router registers under `/api/v1`, and nothing answered before routing**.

- The four uncovered responses carry no credential, so goal 4's stated purpose, the client secret and
  the TOTP seed, is fully met.
- It matches the codebase as it already is: `MiddlewareSecurityHeaders` is mounted **third**, behind
  CORS, so the five existing security headers have the identical hole on a preflight today. One shape
  for response-policy middleware rather than two.
- Cost if wrong: four reachable `/api/v1` responses ship storable. All are empty or an error page.

### Option B: move the policy ahead of the global chain

Mount `MiddlewareNoStore` first on the root router, ahead of CORS, writing the pair only when
`r.URL.Path` is under `/api/v1/`, and drop the two group mounts.

- Goal 4 becomes true as written. One mount instead of two. The router sweep keeps its value: a new
  API group whose responses lack the pair still fails it.
- Cost: the covered surface becomes a **path prefix string inside the middleware** rather than a
  structural property of the route tree. Deciding 7 rejected exactly that kind of maintained list, and
  a future group mounted outside `/api/v1/` would silently miss.
- About four lines of change either way.

### Recommendation: A

No uncovered path carries a credential, and matching where `MiddlewareSecurityHeaders` already sits is
worth more than the letter of a sentence whose purpose is already satisfied. Adding the pre-routing
mount later, if a credential ever does get emitted before routing, is a small change.

**Stated against my own recommendation, because it is the strongest argument for B:** when the same
narrow-or-wide question was put during the interview, about whether to cover just the enrolment
endpoint or the whole `/api/v1` surface, you chose wide. If that preference holds here, B is the
consistent answer and it costs almost nothing.

Reversal in either direction is roughly a mount line and one test assertion. Nothing persisted,
published or user-visible depends on it.

### Also fixed from this review, no answer needed

Three other findings arrived with the fix forced by the codebase's own established patterns, so they
were applied to the plan rather than asked about:

1. **The pending enrolment columns would have been corrupted by ordinary user updates** and issuance
   was a read-then-write race, losing the idempotence goal 7 promises. Now tagged `dont-update` with a
   narrow compare-and-set install and a clear joined to the enable transaction, following
   `TryConsumeForgotPasswordCode` and the three fields already tagged that way on `users`.
2. **Deleting the admin console's hidden seed and QR inputs** would have left every enable-error
   rerender showing a broken image and an empty secret, with no way to finish enrolling. The rerender
   now reloads from the server, which is the same mechanism the interview already chose.
3. **The new encrypted column was missing from the key-rotation registry**, so a live enrolment would
   have survived a data-key rotation unreadable. Now registered, with rotation test coverage.

leodip:
A

leodip:
## Decision 16 applied: **A**, the group mounts stand and the two over-broad claims are narrowed

Answered above with `A`. Recorded, and the run is unblocked.

### What is now fixed

`MiddlewareNoStore` is mounted **first inside** each of the `/api/v1/admin` and `/api/v1/account`
route groups, ahead of the debug middleware and every guard, exactly as originally planned. Nothing
about the mechanism changes. The pre-routing root mount filtered on `r.URL.Path` is rejected: it
would have made the goal true as written for about four lines, but it turns the covered surface into
a path-prefix string inside the middleware rather than a structural property of the route tree, and
a future API group mounted outside `/api/v1/` would then silently miss.

### Two claims narrowed to what is actually true

Both were written unqualified and neither was; the second was owed whichever option won.

1. **The goal** now reads *every **route the router registers** under `/api/v1/admin` and
   `/api/v1/account`*, with responses produced before routing named as outside it.
2. **The router sweep** (the `chi.Walk` test that calls every registered `/api/v1` route
   unauthenticated and asserts both header fields) now states what it does not own. It builds
   `initRoutes` alone, so it is exhaustive over registered routes and **not** over the API's whole
   response surface. It will not be described as the latter.

### The accepted cost, written down so it is not rediscovered as a bug

Four responses under `/api/v1` will ship with neither `Cache-Control: no-store` nor
`Pragma: no-cache`, because each is produced by a globally mounted middleware that answers without
calling `next`:

| Response | Why it never reaches the route group |
|---|---|
| CORS preflight to any `/api/v1` path | `MiddlewareCors` is mounted first; `OptionsPassthrough` is unset, so go-chi's handler ends the preflight itself |
| settings-lookup failure | `MiddlewareSettings` returns 500 |
| undecodable cookie | `MiddlewareCookieReset` redirects |
| session-store or database error | `MiddlewareSessionIdentifier` returns |

Every one of them is empty or an error page. RFC 6749 section 5.1 scopes its MUST to *"any response
containing tokens, credentials, or other sensitive information"*, so it does not reach them, and the
purpose the goal exists for, the decrypted client secret on `GET /api/v1/admin/clients/{id}` and the
TOTP enrolment seed, is met in full. This also matches where `MiddlewareSecurityHeaders` already
sits, third in the chain and behind CORS, so the five existing security headers have the identical
hole today and the codebase keeps one shape for response-policy middleware instead of two.

Reversing this later is a mount line and one test assertion. The trigger worth watching for is a
credential emitted before routing, which no current middleware does.

### Where the run is

Nothing has been built on the open decision: no stage has landed and the branch carries only the
empty commit that opened this pull request. Stage 1, the rendered-HTML pair and the 500 page, is
next and was never affected by this question. Stage 2 is unblocked and gets expanded when stage 1
lands.

leodip:
## Deferred decisions: none

Nothing in this change rests on an assumption the run took because it could not ask. All sixteen
decisions in the agreement are `Decided`, and none was resolved by deferral, so there is no list
here of "we assumed X, and reversing it costs Y".

Fifteen were settled with the user before the agreement was sealed, during the interview. The
sixteenth was raised by the run itself, escalated to this thread, and answered by the user.

### The one escalation, and what it settled

**Decision 16: whether the API's cache directive runs ahead of the global middleware chain.** Raised
during the plan review, before stage 1 started, when the reviewer showed that a group-mounted
middleware structurally cannot cover a response produced *before* the router runs.

The options put to you were **A**, keep `MiddlewareNoStore` mounted first inside the `/api/v1/admin`
and `/api/v1/account` groups and narrow the two claims that overstated its reach, or **B**, mount it
on the root router ahead of CORS and write the pair only when `r.URL.Path` is under `/api/v1/`.

The escalation is [this comment](https://github.com/leodip/goiabada/pull/256#issuecomment-5389598691),
your answer is [this one](https://github.com/leodip/goiabada/pull/256#issuecomment-5393046336), and
what it changed is recorded in
[this one](https://github.com/leodip/goiabada/pull/256#issuecomment-5393097026).

**The accepted cost, restated so it is not rediscovered later as a defect.** Four responses under
`/api/v1` ship with neither header field, because four globally mounted middleware can answer a
request without the router ever running:

1. a CORS preflight, which `MiddlewareCors` answers `200` itself, since `OptionsPassthrough` is
   unset and its `AllowOriginFunc` allows only discovery, `/certs`, token, logout and userinfo;
2. `MiddlewareSettings`'s `500` when the settings lookup fails;
3. `MiddlewareCookieReset`'s redirect on a cookie it cannot decode;
4. `MiddlewareSessionIdentifier`'s return on a session-store or database error.

Each is empty or an error page. RFC 6749 section 5.1's MUST is scoped to "any response containing
tokens, credentials, or other sensitive information", so it does not reach any of the four, and the
purpose of the goal, the decrypted client secret on `GET /api/v1/admin/clients/{id}` and the TOTP
seed on `GET /api/v1/account/otp/enrollment`, is met in full. `MiddlewareSecurityHeaders` is mounted
third, behind CORS, so the five existing security headers have this identical hole today; A keeps
one shape for response-policy middleware rather than introducing a second.

**What reversing it would take**, since that is the one thing this list exists to tell you: a mount
line and one scope assertion. Move `MiddlewareNoStore` to `s.router` ahead of CORS, gate it on the
path prefix, and rewrite seam C's claim. The trigger to watch for is a credential emitted before
routing, which no current middleware does.

**Where the goal and the tests were corrected to match.** Goal 4 now reads "every route the router
registers", `routes_no_store_test.go` names the four exclusions in its own header comment rather
than implying a coverage it does not have, and the three published texts that overstated it,
`rest-api.mdx`, `security.mdx` and `openapi.yaml`, were narrowed in round 1 of the review to attach
the guarantee to a response from an **endpoint** in either group.

leodip:
## Follow-ups: one to file, two folded in, one dropped, one already tracked

The run's account of everything it found next to the work, including the things it fixed here rather
than leaving for you.

---

## 1. To file: recovery codes for accounts with OTP enabled

The only follow-up this run leaves owed. The duplicate search was re-run just now over every state,
`gh issue list --state all --search "recovery code"` and `--search "backup code"`: nothing open
tracks it, and the closed hits (#40 recovery *email*, #20 a general verification epic) are different
things.

This came out of deciding what shape the pending-enrolment columns should take, and it is written
down because it would otherwise be re-derived by whoever picks it up.

**Title:** `Feature: recovery codes for accounts with OTP enabled`
**Labels:** `enhancement`, `go`, `security`

**Body:**

> Nothing in the tree implements recovery codes. `grep -rniE "recovery.?code|backup.?code"` over
> `src/` and `site/` returns nothing, and no open issue tracks it. A user whose authenticator is
> lost or whose phone is wiped has no self-service route back into their account today: the only
> way back in is an administrator disabling OTP on the user row.
>
> **Shape.** Recovery codes are many per user, stored **hashed** rather than encrypted, since unlike
> the TOTP seed they are never displayed again after generation, and single-use with history. That
> makes them their own table, `user_recovery_codes(user_id, code_hash, used_at)`, rather than
> columns on `users`. Worth stating explicitly: the two pending-enrolment columns added by #247,
> `otp_enrollment_secret_encrypted` and `otp_enrollment_issued_at`, are **not** a step toward this
> and do not block it.
>
> **The open design question is when they are generated.** Showing them after a successful
> enrolment needs nothing new. Showing them *during* enrolment, so the user must confirm they have
> saved them before OTP is switched on, means staging them beside the pending seed, which #247's
> columns accommodate with a third column.
>
> **Also owed whenever it is built:** a way to consume one at the OTP prompt, an audit event for
> that consumption, and rate limiting on it, since a recovery code is a credential check and this
> codebase counts failures on every other one (#219).

**Files it:**

```bash
gh issue create \
  --title "Feature: recovery codes for accounts with OTP enabled" \
  --label enhancement --label go --label security \
  --body 'Nothing in the tree implements recovery codes. `grep -rniE "recovery.?code|backup.?code"` over `src/` and `site/` returns nothing, and no open issue tracks it. A user whose authenticator is lost or whose phone is wiped has no self-service route back into their account today: the only way back in is an administrator disabling OTP on the user row.

**Shape.** Recovery codes are many per user, stored **hashed** rather than encrypted, since unlike the TOTP seed they are never displayed again after generation, and single-use with history. That makes them their own table, `user_recovery_codes(user_id, code_hash, used_at)`, rather than columns on `users`. Worth stating explicitly: the two pending-enrolment columns added by #247, `otp_enrollment_secret_encrypted` and `otp_enrollment_issued_at`, are **not** a step toward this and do not block it.

**The open design question is when they are generated.** Showing them after a successful enrolment needs nothing new. Showing them *during* enrolment, so the user must confirm they have saved them before OTP is switched on, means staging them beside the pending seed, which #247'"'"'s columns accommodate with a third column.

**Also owed whenever it is built:** a way to consume one at the OTP prompt, an audit event for that consumption, and rate limiting on it, since a recovery code is a credential check and this codebase counts failures on every other one (#219).'
```

---

## 2. Already tracked, so not filed: the unbounded pre-authentication body read

Round 2 of the review raised exactly one finding, marked blocking: `APIDebugMiddleware` reads
`io.ReadAll(r.Body)` before calling `next`, so with `GOIABADA_AUTHSERVER_DEBUG_API_REQUESTS=true`
the whole request body is resident before any guard runs, and the response side tees into an
unbounded buffer the same way.

**The mechanism is real and was confirmed in the source. The attribution is not.** That file is
byte-identical to `origin/main` on this branch, `APIDebugMiddleware` was already the first `r.Use`
in both API groups on `origin/main`, and `MiddlewareNoStore` sets two header fields and calls
`next`, never touching `r.Body`. This change neither introduced nor worsened it. What it did add to
that surface is the opposite of a regression: a 64 KB bound on an endpoint that had none, effective
end to end in the default configuration, and the debug flag is off by default.

**It would have become a follow-up here, and it does not, because the duplicate search found it
twice:**

- **#190** (closed, `NOT_PLANNED`) is this finding verbatim, closed by you with "Merged into #205,
  which now covers all three places the server reads unbounded input before anything bounds it".
- **#205** (open, `bug, go, security`) carries it now, as one of three halves alongside the missing
  `MaxHeaderBytes` on all four `http.Server` literals and the DCR body limit, and its "Choosing the
  numbers" section is already the sizing discussion the fix needs.

Filing again would put a third issue on one defect. It is recorded here so that a reader who meets
the reviewer's finding can see it was verified rather than missed, and can see where it lives.

---

## 3. Folded in rather than filed

Four things were found beside the work and closed inside it. Each rides on a surface the change
already touched, and each is named here because the goal does not mention it, so without this the
hunk reads as scope creep.

- **The account API enrols a caller-supplied TOTP seed** (drafted at the first seal as follow-up
  9.1). You widened the scope and took the full contract fix over the contained subset, with the
  breakage stated. It became goals 6 and 7 and stages 4 and 5, `6915ce5b` and `71fd5e98`. This is
  the breaking half of the change.
- **The enrolment QR was carried in the cookie rather than regenerated** (drafted as 9.2, deferred
  on a hazard that turned out not to apply: the concern assumed rebuilding the `otpauth://` URL
  from its parts, while `otp.Key.URL()` and `otp.NewKeyFromURL` are the library's own serialization
  and its inverse). Folded in as goal 8 and stage 3, `5bfd9c28`. The auth context falls from 3693
  bytes over 3 cookies to 1016 over 2.
- **`InternalServerError` committed its status before rendering**, in stage 1, `35209225`. It called
  `w.WriteHeader` first, which freezes the header map, so every header `RenderTemplate` set
  afterwards was silently dropped: the 500 page has been shipping without its `Content-Type` in both
  modules, and would have shipped without the new cache directives too. It fit because it is the
  same function the stage was already editing and it is the whole of the fix. Found because the
  existing test asserted the header through `httptest.ResponseRecorder.Header()`, the live map,
  which shows a header that never ships.
- **A 64 KB bound on `PUT /api/v1/account/otp`'s request body**, in stage 5, `71fd5e98`. That
  endpoint is the one API JSON endpoint that reads its body whole with `io.ReadAll`, because stage 5
  made it parse twice: once as a raw object to detect a `secretKey` field at all, once into the
  request struct. The bound landed with the read that needed it. Its boundary test came a round
  later, in `4a87e383`, after the reviewer's mutation showed the cap could be deleted with the
  module suite still green.

---

## 4. Dropped, by your decision, so nothing tracks it

**Retiring `Pragma: no-cache` across the tree.** Drafted as follow-up 9.3 and dropped by you on
2026-08-23 rather than deferred: there is to be no tracked task for it.

Kept here as the record of what was weighed, because the decision to emit the pair turns on it. RFC
9111 section 5.4 deprecates `Pragma` and records that its meaning in a *response* was never
specified, and the IANA field-name registry marks it `deprecated`. OAuth 2.1 draft -13 section 3.2.3
drops it, keeping only `Cache-Control: no-store`. RFC 6749 section 5.1 still makes both a MUST,
which is why this change emits the pair, and the pair is what the nine existing sites in the tree
already write. If OAuth 2.1 is published and this is revisited, it is a sweep of its own.

leodip:
## The review account

Every round this change went through, what was found, and what happened to it. The verdicts live in
`.review/`, which is gitignored and overwritten by the next round, so this comment is the only
durable record of them. Pasted rather than pointed at, for that reason.

**Reviewer throughout:** `codex`, model `gpt-5.6-sol`, effort `xhigh`, read from `.review/reviewer.env`
rather than inferred.

**Three rounds: one at the plan gate, two over the finished branch.** This run executed under
`LEO_REVIEW_MODE=end`, so **no stage was reviewed on its own**: the whole branch was read once, as a
whole, after all five stages had landed. Nine findings in total, all nine verified against the code
before being worked, **none refuted**.

|---|---|---|---|---|
| plan | 1 | 4 | 2 | 3 fixed in `plan.md` before any code was written, 1 escalated to you as decision 16 |
| final | 1 | 4 | 0 | all 4 fixed in `4a87e383` |
| final | 2 | 1 | 1 | mechanism confirmed, attribution refuted with evidence; already tracked as #205 |

---

# Plan gate, round 1 (2026-08-24)

Scope: `plan.md` against the sealed agreement, all three axes, before a line of code existed. The
reviewer ran `check-anchors.sh` (29 pass), matched the tree hash, checked the cited RFC 6749, RFC
9111 and RFC 9700 text against the publisher copies, and wrote four probes of its own. Production was
byte-identical before and after, and `git status --porcelain` was empty.

### 1. Blocking, security. Pending OTP enrolment issuance is a read-then-write race, and whole-row user writes can overwrite it. **Fixed in the plan, before stage 4 was built.**

The plan added `otp_enrollment_secret_encrypted` and `otp_enrollment_issued_at` as ordinary
`models.User` fields. `commondb.UpdateUser` writes every field not tagged `pk` or `dont-update`, so
all **fourteen** production sites that load a user and later save it would rewrite the pair,
including `EnableUserOTPTx`, which the browser ceremony uses. Two concurrent
`GET /api/v1/account/otp/enrollment` calls would both see no pending value, return different seeds
and leave only the last usable, losing the idempotence goal 7 asks for.

The reviewer's interleaving probe printed `returned=(seed-A,seed-B), stored=seed-B,
first-still-live=false`, then a stale full-row update restoring an empty pending value, then browser
enrolment leaving `otp=seed-B pending=seed-A both-live=true`. Verified independently: the tree
already answers this class three times on the same table (`AuthStateGeneration`, `LastOTPStep` and
`OtpConfigGeneration` are each `fieldtag:"dont-update"`), and `TryConsumeForgotPasswordCode` is the
compare-and-set precedent.

**What changed:** stage 4 tags both fields `dont-update`, adds a narrow conditional install whose
`WHERE` requires OTP disabled and the pending value absent or expired, with the loser re-reading the
winner's seed, and joins a narrow clear to `EnableUserOTPTx`'s transaction. Seam F gained seven
cases. Not escalated: the verdict carried a populated forced answer and the remedy restores two
things already decided.

### 2. Blocking, conformance. Deleting the admin console's hidden inputs leaves every enable-error rerender with no QR or seed. **Fixed in the plan, before stage 5 was built.**

`HandleAccountOtpPost` read `secretKey` and `base64Image` from the submitted form and they were its
**only** source, feeding the blank-code path and the bind map on every handled API 400.
`account_otp.html` renders the QR unconditionally in the enable branch, so after the planned deletion
a blank code, wrong password or wrong OTP code would rerender a broken image and an empty `<pre>`
with no way to finish enrolling. Its hard-coded error switch still named two error codes stage 5
retires, so a new code would fall through to a generic 500, and the package carried only
`test_main_test.go`, so nothing would have caught it.

**What changed:** the mechanism was right and stands; the factual claim that the console needed
nothing else was wrong. The rerender now calls the idempotent enrolment GET and binds what it
returns, the error switch is updated for all four code changes, an already-enabled race reloads
state instead of the form, and a new `handler_account_otp_test.go` covers five paths.

### 3. Significant, security. The encrypted pending seed was absent from the key-rotation and API secret-guard registries. **Fixed in the plan.**

`commondb.aesProtectedColumns` enumerates the seven AES-GCM columns re-keyed on rotation and drives
both legacy-key migration and env-to-env rotation; `api.sensitiveUserFields` lists both halves of the
three existing pending pairs. Both are hand-maintained and the plan named neither. The reviewer's
SQLite probe added the proposed column, encrypted a pending URL under the old key, ran
`RotateEncryptionKeyIfNeeded` and printed `rotated=true pending-decrypts-old=true
pending-decrypts-new=false`: **a live enrolment would have survived rotation unreadable.**

**What changed:** stage 4 registers the column with a rotation fixture required to decrypt only under
the new key, and both model fields with their sentinel case.

### 4. Significant, conformance. Group-mounted middleware cannot cover responses short-circuited earlier in the chain. **Escalated to you as decision 16, answered A.**

`initMiddleware` mounts `MiddlewareCors` first, ahead of `initRoutes`, and go-chi ends a preflight
itself unless `OptionsPassthrough` is set, so a preflight to an `/api/v1` path is answered 200 with
neither header field; the reviewer's probe over the real middleware in production order printed
`preflight=true status=200 cache="" pragma=""`. `MiddlewareSettings`, `MiddlewareCookieReset` and
`MiddlewareSessionIdentifier` each return before calling `next` on their error paths.

This carried `needs_human: true` with an **empty** forced answer, which is the definition of a
choice, and the choice was between narrowing goal 4's unqualified sentence and contradicting the
named mechanism of an earlier decision. Neither is the run's to make. It was escalated as the class
rather than the CORS instance, so one answer settles all four paths. You answered **A**; the detail
is in the deferred-decisions comment above.

**Round productivity:** productive. It raised a decision and amended three of five stages. Production
code was byte-identical throughout, which at a plan gate is expected rather than a stall.

---

# Final gate, round 1 (2026-08-24, 11:12 to 11:27 UTC)

Scope: **the whole branch against `origin/main`**, 58 files, +3205 / -322, all three axes.
`axis_coverage` reads `reviewed` on all three. The reviewer's own note: **"No current production
vulnerability was found."** All four findings are about test machinery or published prose.

**What it ran, which is why that absence has a stated scope:** the complete agreement, plan, five
stage logs, closing record, probes and the entire 58-file diff; targeted unit coverage for render,
middleware, routes, OTP cleanup, API enrolment, pending storage, migrations, key rotation and the
console; targeted data tests on all four engines; targeted sqlite integration tests; and **two
mutations of its own, both of which SURVIVED**, which is what turned findings 1 and 3 from opinions
into demonstrations. It re-verified HEAD, all 29 anchors, the tree hash and the recorded green CI.

### 1. Significant, security. The route sweep cannot see a handler cache override, and the second credential-bearing API response had no cache assertion. **Fixed.**

`MiddlewareNoStore` deliberately lets a handler replace `Cache-Control`, since `Header().Set`
replaces. `TestInitRoutes_EveryApiRouteRefusesToBeStored` calls all 105 registered routes **without
credentials**, so it stops in a guard and never reaches a handler, yet the comment on
`TestMiddlewareNoStore_HandlerCanOverrideCacheControl` claimed that sweep was what would catch a
handler setting its own policy. **That sentence was false.** And
`TestAPIClientGet_ConfidentialIncludesSecretInDetailButNotList`, which drives the one authenticated
success response carrying the decrypted client secret, asserted only the body.

Production was correct: no handler under `/api/v1` sets a `Cache-Control` today. This was a
regression gap, not an exposure.

**Fix:** `assertNotStorable` now runs on the admin client detail response, immediately after the
assertion that the body really carries the decrypted secret. One line was added beyond the forced
answer: the same assertion on the client **list** response, which carries no credential but is an
authenticated **success**, the half of the API surface the unauthenticated sweep structurally cannot
reach. The false comment was replaced with what actually owns the property.

**Evidence, the reviewer's own mutation re-run:** `Cache-Control: public, max-age=300` added to
`HandleAPIClientGet`'s success path. It SURVIVED for the reviewer. It is now **MUTATION CAUGHT**.

### 2. Significant, conformance. Three tests observed endpoint effects through the user row, which the seam forbids by name. **Fixed, and the replacements are stronger than what they replaced.**

Seam G says, verbatim: "Observed as a caller sees it, through status codes, bodies and a second call,
never by reading the user row." Three tests read `OTPEnabled`, decrypted the stored secret and read
the pending columns. The success case therefore proved that **storage held the right ciphertext**,
and never that a caller can authenticate with the authenticator that was installed.

**Fix:**
- *Caller-supplied secret refused:* the enrolment GET is called again and must answer with the
  **same seed**. That subsumes the old assertion and adds what the row read never checked, that the
  pending enrolment was not consumed either.
- *No pending enrolment:* the enrolment GET must still issue. A 200 there is the observable form of
  "the stranger's seed did not become this account's authenticator".
- *Success:* a fresh level 2 authorization ceremony is driven to the OTP form with a
  `level2_mandatory` client and a fresh cookie jar, and a code from the **issued** seed is accepted,
  redirecting to `/auth/completed`.

`nextStepCode` is used rather than the code the PUT spent, because the enable branch claims its time
step through `TryConsumeUserOTPStep` (#111), so resubmitting would be refused *correctly* and would
read as a broken enrolment. The direct column assertions were not lost: they were already in the tier
that runs on all four engines, as `TestTryInstallPendingOTPEnrollment_RoundTrip` and
`TestEnableUserOTPTx_ClearsThePendingEnrollment`.

**Evidence:** `user.SetOTPSecret(pendingSecret)` replaced with a hard-coded foreign seed in the enable
branch. **MUTATION CAUGHT**: the enrolment succeeds and the ceremony then refuses the code.

### 3. Significant, security. The folded-in 64 KB body bound had no boundary test. **Fixed.**

`HandleAPIAccountOTPPut` is the one API JSON endpoint that reads its body whole with `io.ReadAll`,
because it parses twice. Stage 5 folded in the bound and wrote no test for it. **The reviewer's
mutation, replacing the capped read with `io.ReadAll(r.Body)`, left the whole module suite green.**

**Fix:** `TestHandleAPIAccountOTPPut_OversizedBodyIsRefused`, with expectation-free database and
audit mocks and a recording limiter stub. The body is **syntactically valid JSON** with an ignored
`padding` field, which is the whole design of the case: a *malformed* oversized body answers 400
`INVALID_REQUEST_BODY` from the `json.Unmarshal` below the read whether the cap exists or not, so it
would pass with `MaxBytesReader` deleted and pin nothing. With well-formed JSON the two outcomes
separate. It also asserts the shared credential budget was not spent, so an oversized body cannot be
used to exhaust the account limiter.

**Evidence, the reviewer's own mutation re-run:** capped read to uncapped read. It SURVIVED for the
reviewer. It is now **MUTATION CAUGHT**.

### 4. Minor, conformance. Three published descriptions overstated the boundary decision 16 settled. **Fixed.**

`rest-api.mdx`, `security.mdx` and `openapi.yaml` each said, in their own words, that **every**
response from both API groups carries the pair, refusals included. Decision 16 accepted, on your
answer A, that four responses under those paths are produced before routing and carry neither field.
Goal 4 was restated for exactly that reason and `routes_no_store_test.go` documents the four
exclusions accurately; the three published texts did not follow.

**Fix:** in all three, in each page's own voice, the guarantee now attaches to a response from an
**endpoint** in either group, refusals from the token and scope checks included, followed by a short
sentence naming the pre-routing responses and saying none of them carries a credential.

### The `unchecked` list, all three answered

1. **"No interactive browser pass over the admin console OTP error rerenders."** **Dismissed with a
   reason.** The properties a browser pass would establish are pinned by four handler tests in
   `handler_account_otp_test.go`, and the admin console has **no integration tier in this repository
   at all**, so a browser pass has no seam to live in; building one is a change of its own.
2. **"Authenticated success branches were not driven for every non-credential API route."**
   **Covered in part, remainder dismissed.** This round added the assertion to the admin client list
   response. Full exhaustiveness over all 105 routes' success branches is dismissed: the middleware
   writes both fields **before** the handler runs, so the only way a success can lack them is a
   handler calling `Header().Set` itself, a source sweep found no such call, and that exact override
   is now caught by a mutation.
3. **"No simultaneous real enrolment GET stress test under the race detector."** **Dismissed.** The
   atomicity claim is a property of `TryInstallPendingOTPEnrollment`'s `WHERE` clause, which the data
   tier executes on all four engines, and the loser branch has two controlled unit interleavings. A
   concurrent HTTP stress probe would be non-deterministic evidence for something already established
   deterministically.

### Tiers after the round

| Tier | Engines | Result |
|---|---|---|
| unit | n/a | 3956 pass, 0 fail |
| data | sqlite, mysql, postgres, mssql | 1723 pass, 0 fail |
| integration | sqlite | 1281 pass, 0 fail |

All 29 anchors resolve. All three mutations ran through the harness that restores from a trap and
compares by hash, and `git status --porcelain` afterwards listed exactly the seven intended files and
**no production Go file**. The fixes are `4a87e383`, and CI is green on all eleven jobs there.

*One bookkeeping note, recorded because it explains the round numbering.* Round 1's verdict was left
sitting in the review mailbox because the session that ingested it failed to return, which would have
made round 2 impossible to open and, if cleared naively, would have made the reviewer believe it was
re-reading round 1. It was repaired using the driver's own names and arithmetic before round 2 ran.
Nothing was re-ingested and no finding was re-worked.

---

# Final gate, round 2 (2026-08-24, 11:55 to 12:06 UTC)

Scope: **the whole branch against `origin/main`**, 59 files, the complete 5,051-line diff, all three
axes, `reviewed` on all three.

**What it ran:** the complete agreement, plan, five stage records, closing record, request, probes
and RFC excerpt, and the whole diff. It matched all six split tree hashes exactly, resolved all 29
anchors exactly once, confirmed the worktree clean, confirmed all 11 CI jobs green at `4a87e383`,
ran the complete module tier, ran focused sqlite integration cases for caller-secret refusal,
no-pending refusal, issued-seed browser verification, pending clear, client secret cache directives
and enrolment JSON cache directives, and ran an overlay probe through `APIDebugMiddleware` with a
1,048,576-byte counting body against a downstream 65,536-byte `MaxBytesReader`.

It confirmed the round 1 repairs are genuine, **including that the two mutations which had survived
now fail**. Its own summary of everything else: *"Apart from the finding above, the branch implements
the sealed contract."*

### 1. Blocking, security. `APIDebugMiddleware` buffers the whole request body before authentication and before the endpoint's own limit. **Mechanism confirmed. Attribution refuted with evidence. Already tracked as #205.**

The mechanism, verified by reading the source rather than by trusting the probe:
`api_debug_middleware.go` reads `reqBody, _ = io.ReadAll(r.Body)` and only then calls
`next.ServeHTTP`, so with the debug flag on the whole body is resident before any guard runs. The
256 KB `maxLoggedBody` bound is applied *after* the allocation and governs only what is formatted
into the log. The response side tees into an unbounded buffer the same way. All accurate as reported.

**Why it is not this change's defect, established four ways:**

1. `api_debug_middleware.go` is **byte-identical to `origin/main`** on this branch;
   `git diff --stat origin/main...HEAD` does not name it.
2. On `origin/main`, `APIDebugMiddleware` is **already** the first `r.Use` in both API groups, ahead
   of `authHeaderToContext`. The pre-authentication ordering the finding turns on is pre-existing.
3. `io.ReadAll(r.Body)` sits at the same line on `origin/main`, and the three upload handlers'
   `MaxBytesReader` calls pre-date this branch, so the "upload limits are bypassed too" half of the
   family is entirely pre-existing.
4. This branch's only change to `routes.go` is 9 inserted lines: the two `MiddlewareNoStore` mounts
   and their comments. That middleware sets two header fields and calls `next`. It never reads, wraps
   or replaces `r.Body`.

**What this change actually did to that surface is the opposite of a regression:** it added a 64 KB
bound to an endpoint that had none, effective end to end in the default configuration. The debug flag
is off by default, is written `false` in every shipped compose file and in the setup generator, and
the security reference already says it "is off by default and is meant for development".

**Nothing this change published overstates the bound**, which is the one part that could have been
contained here, since it is the shape of round 1's finding 4. Checked: the diff to `openapi.yaml` and
to `site/` makes no claim about body size at all.

**Not fixed here, and not on scope alone.** A partial remedy was considered and rejected on its
merits: replacing `r.Body` with a bounded prefix truncates the body for downstream handlers and would
break the three upload endpoints outright; preserving the stream with `io.MultiReader` over a
`LimitReader` prefix would bound the middleware's own buffer but changes the streaming semantics of a
middleware shared by every Admin and Account route, needs its own tests across that surface, and
still leaves the number to choose. The reviewer says as much itself: *"The design needs a human
choice because the OTP JSON endpoint, ordinary JSON endpoints, and upload endpoints have different
legitimate sizes."* That is #205's work, at #205's altitude. Details in the follow-ups comment above.

### `unchecked`, both entries answered

1. *"Did not rerun the full four-engine data tier or complete integration tier."* **Covered.** CI is
   green on all 11 jobs at `4a87e383`, including `Tests / sqlite`, `Tests / mysql`, `Tests / postgres`
   and `Tests / mssql`, which are the four-engine data and integration tiers. That is stronger
   evidence than a local run, since it names a commit and anyone can reproduce it.
2. *"Did not attempt actual process exhaustion."* **Dismissed as a method, filed in substance.**
   Declining to exhaust the workspace was the right call and the counting-reader probe established
   the read without it. The substance belongs to #205.

**Tiers.** No code changed this round, so none was re-run. The tree is byte-identical to what the
reviewer read, every one of the six split hashes equal to the value recorded in the request and
independently matched by the reviewer, which is itself the check that no mutation residue was left
behind. Working tree clean at `4a87e383`.

**Round closed, and the review with it.** Round 2 changed no production code, so no further round is
funded, and the tree a round 3 would read is the one round 2 has just read in full. Nothing is left
to raise.

Issue #225, opened 2026-08-16
Admin console proxy handlers turn downstream 400 responses into generic 500s


leodip:
Every admin console handler that proxies a call to the auth server API discards the API's status and message when the call fails. `HttpHelper.JsonError` (`src/core/handlerhelpers/http_helper.go`) preserves a status code and description only for `*customerrors.ErrorDetail`. An `*apiclient.APIError`, which is what `src/adminconsole/internal/apiclient/auth_server_client.go` returns for every non-2xx API response, takes the other branch: HTTP 500, `slog.Error("%+v")` of the whole error, and the fixed description "An unexpected server error has occurred. For additional information, refer to the server logs. Request Id: `<id>`".

`HandleAdminClientRedirectURIsPost` was repaired in #224, and it is the only one: every other AJAX proxy handler still takes the generic branch.

Two consequences.

1. **A user-correctable validation failure is presented as a server fault.** The administrator is told the server broke, with no indication of which input was rejected or why, even though the API returned a precise sentence one hop away. `HandleAdminClientWebOriginsPost` is the sharpest remaining case, because its endpoint validates web origins the same way the redirect URIs endpoint validates URIs, so the same first-try mistake produces the same unreadable 500.
2. **The rejected value is written to the admin console log as unbounded text.** The `%+v` carries the API's whole message, which for these endpoints embeds the submitted value verbatim, for example `Invalid redirect URI: <the submitted string>`. No request body size limit applies on these paths, so a principal holding `manage-clients` or `manage` can put an arbitrarily large, arbitrary-content string in the log by pressing Save.

This codebase already has the intended pattern, in `src/adminconsole/internal/handlers/api_error_helper.go`:

> `HandleAPIErrorWithCallback` ... Routes on HTTP status: 400 Bad Request is treated as a user-correctable validation failure and surfaced back to the form via renderErrorFunc; anything else escalates to InternalServerError.

The form-post handlers use it. Until #224 the AJAX proxy handlers had no equivalent at all.

**Most of the work is already done, and this issue is the adoption.** #224 added `HandleAPIErrorJson` to that same file, the AJAX counterpart routing on status the same way, and `sendAjaxRequest` now escapes `error_description` before `showModalDialog` assigns it to `innerHTML`, so the sink is closed for every caller rather than only for the one handler that adopted the helper.

[...]

Issue #260, opened 2026-08-25
Three defects in the Admin and Account API: Go field names as JSON keys, a response shape no handler writes, and client saves that discard concurrent updates


leodip:
Three defects found by the #245 OpenAPI census, in what the Admin and Account API sends and how it
writes. They are grouped because all three live in `src/core/api` and
`src/adminconsole/internal/apiclient`, and all three change something a caller or the admin console
already depends on, so each needs a compatibility decision rather than a patch.

They are independent of each other. Defects 1 and 2 are wire format, defect 3 is write semantics.

---

## 1. Nested group, permission, redirect URI and web origin objects use Go field names as JSON keys

`api.UserResponse` carries `Groups []models.Group` and `Permissions []models.Permission`, and
`api.ClientResponse` carries `RedirectURIs []models.RedirectURI` and `WebOrigins []models.WebOrigin`.
None of those `core/models` types declares a `json:` tag, only `db:`, so `encoding/json` falls back
to the exported Go field name. Inside responses whose every other key is lowerCamelCase, the wire
carries capitalised keys.

Marshalling each type directly:

```
models.Group       -> {"Id":0,"CreatedAt":{"Time":"0001-01-01T00:00:00Z","Valid":false},"UpdatedAt":{...},"GroupIdentifier":"admins","Description":"d","Attributes":null,"Permissions":null,"IncludeInIdToken":false,"IncludeInAccessToken":false,"MemberCount":0}
models.Permission  -> {"Id":0,"CreatedAt":{...},"UpdatedAt":{...},"PermissionIdentifier":"read","Description":"d","ResourceId":0,"Resource":{"Id":0,"CreatedAt":{...},"UpdatedAt":{...},"ResourceIdentifier":"","Description":""}}
models.RedirectURI -> {"Id":0,"CreatedAt":{...},"URI":"https://x/cb","ClientId":3}
models.WebOrigin   -> {"Id":0,"CreatedAt":{...},"Origin":"https://x","ClientId":3}
```

`models.Resource` is in the same position, reached through `Permission.Resource`, so it is five
types rather than four.

Three consequences, all live:

1. A caller reading `user.groups[0].groupIdentifier` or `client.redirectURIs[0].uri` gets nothing.
   Until #245 those were the names `openapi.yaml` declared, so a generated client was guaranteed to
   be wrong. The spec now declares the real capitalised names, which makes the document honest and
   the API no less odd.
[...]

Issue #245, opened 2026-08-22
Security: a public client can run without PKCE, and flipping a client to public deletes its secret without forcing PKCE or invalidating its grants


leodip:
Found by the #239 authorization ceremony review (finding [F36](https://github.com/leodip/goiabada/issues/239#user-content-f36)).

Two halves. The second is the sharper one and it is a plain bug rather than a policy question.

## 1. The configuration permits a public client with PKCE off

`HandleAuthorizeGet` derives the requirement from `client.IsPKCERequired(settings.PKCERequired)`, which returns the per-client override whenever it is set, so an explicit `false` beats a globally enabled setting. The OAuth2-flows update endpoint assigns `client.PKCERequired = req.PKCERequired` with no guard; its only public-client rule is `ClientCredentialsEnabled = false`.

At redemption, `ValidateTokenRequest` requires a verifier only when the stored code carries a challenge, and requires a secret only when `!client.IsPublic`. **So a public client whose code carries no challenge presents nothing at all and is issued tokens.** `TestValidateTokenRequest_PKCE_NoPKCEUsed_NoVerifierProvided_Success` pins the redemption half.

### Framing, so this is not overstated

RFC 9700 2.1.1's *"Public clients MUST use PKCE"* is addressed to **clients**, and every MUST that section and 4.8.2 place on the authorization server is already met here. The accurate statement is not that this server violates RFC 9700. It is that this server ships, and exposes through its admin API, a switch whose only effect is to let a public client violate a MUST written for it, with no warning at the point of configuration and no enforcement at the point it matters.

That is a security-defaults defect, and RFC 9700 4.8.1 spells out the attack it re-enables: an attacker strips `code_challenge` from the request, *"If the authorization server allows for flows without PKCE, it will create a code that is not bound to any PKCE code challenge"*, and the injected code is redeemed under the victim's client.

## 2. Flipping a client to public removes its only authentication

`UpdateClientAuthentication` handles the switch like this:

```go
if req.IsPublic {
	// Switching to public: remove secret and disable client credentials
	client.IsPublic = true
	client.ClientSecretEncrypted = nil
	client.ClientCredentialsEnabled = false
}
```

It deletes the secret. It does not set `PKCERequired`, and it does not touch any grant the client already holds. So at the instant of the flip:

[...]

PR #263, opened 2026-08-25
fix(core,api,adminconsole)!: each flow is gated by its own switch, and web origins become manageable and canonical

### LINKED ISSUE #250: Outside the ceremony: AuthorizationCodeEnabled is used as a proxy for "this client does OAuth at all"
Found by the #239 authorization ceremony review (findings [F39](https://github.com/leodip/goiabada/issues/239#user-content-f39) and [F38](https://github.com/leodip/goiabada/issues/239#user-content-f38)). **Outside the ceremony proper**, filed separately so it can be scheduled on its own terms.

One mistake in two places: `AuthorizationCodeEnabled` is used as a proxy for "this client does OAuth at all".

## 1. Every refresh is gated on the authorization code flow, before the token's origin is known ([F39](https://github.com/leodip/goiabada/issues/239#user-content-f39))

`ValidateTokenRequest`'s `refresh_token` arm opens with:

```go
case "refresh_token":
	if !client.AuthorizationCodeEnabled {
		return nil, customerrors.NewErrorDetailWithHttpStatusCode("unauthorized_client",
			"The client associated with the provided client_id does not support authorization code flow.",
			http.StatusBadRequest)
	}
```

`isROPCToken := !refreshToken.CodeId.Valid` is computed **seventy-three lines later**. The code learns how to tell the two kinds of token apart well after it has already refused on the wrong flag, and the arm contains no ROPC check at all.

| `AuthorizationCodeEnabled` | ROPC enabled | Redeeming an ROPC refresh token |
|---|---|---|
| on | on | works |
| **off** | **on** | ROPC issues the token, redemption is refused with *"does not support authorization code flow"*. Dead on arrival. |
| on | **off** | **still works.** Turning ROPC off stops nothing already issued. |

The documentation promises the second row works: *"ROPC always returns a refresh token. Use it to obtain new tokens without re-prompting for credentials."*

The error is also wrong. RFC 6749 5.2 defines `unauthorized_client` as *"The authenticated client is not authorized to use this authorization grant type."* The grant type here is `refresh_token`, which the client **is** authorized for, and there is no `RefreshTokenEnabled` flag for the message to be talking about.

### Decision, already taken

**A refresh is governed by the switch of the flow that issued the token.**

RFC 6749 6's MUST list is exactly three items and none is a grant-type flag: *"require client authentication for confidential clients or for any client that was issued client credentials"*, *"authenticate the client if client authentication is included and ensure that the refresh token was issued to the authenticated client"*, and *"validate the refresh token."* Flow switches are local policy the specification neither requires nor forbids, so gating on the issuing flow and gating on nothing both conform.

The deciding argument is what the ROPC switch is for. ROPC is deprecated in OAuth 2.1, the per-client switch exists to kill it, and the documentation records that ROPC refresh tokens are *"always offline type"* and use the offline idle timeout and maximum lifetime. **A kill switch that leaves long-lived offline tokens alive is not a kill switch.**

```go
isROPCToken := !refreshToken.CodeId.Valid
if isROPCToken {
    if !client.IsResourceOwnerPasswordCredentialsEnabled(settings.ResourceOwnerPasswordCredentialsEnabled) {
        return unauthorized_client   // message naming ROPC
    }
} else if !client.AuthorizationCodeEnabled {
    return unauthorized_client       // message naming the authorization code flow
}
```

**Breaking change, needs a release note.** A deployment with the authorization code flow on and ROPC off has live ROPC refresh tokens today that will stop working. That is the intended direction and it must be announced. The opposite row is a pure fix.

**Not a concern:** `client.Enabled` is checked once before the grant-type switch, so a total kill switch exists either way.

### The fixtures are part of the fix

`createROPCClient` sets `AuthorizationCodeEnabled: true` on every fixture, which is why this was invisible. `TestROPC_RefreshToken_OpenIdOnly`'s own comment records that this same fixture family already let one bug ship:

> The four ROPC tests above assert only that a refresh token comes back and never redeem one, which is why this shipped.

So the same blind spot has now produced two defects. The fixtures should stop enabling the authorization code flow, and at least one ROPC test should actually redeem the refresh token it is handed.

## 2. An implicit-only client's redirect URIs cannot be managed ([F38](https://github.com/leodip/goiabada/issues/239#user-content-f38))

`HandleAPIClientRedirectURIsPut` opens with:

```go
if !client.AuthorizationCodeEnabled {
	writeJSONError(w, "Authorization code flow is disabled for this client.", "VALIDATION_ERROR", http.StatusBadRequest)
	return
}
```

Implicit is also a redirect-based flow and also requires registered redirect URIs, so a client with `ImplicitGrantEnabled` and `AuthorizationCodeEnabled = false` is refused management of the only setting that makes it work. An administrator cannot add, rotate, or **urgently remove** such a client's callback without temporarily enabling the authorization code flow or editing the database.

The urgent-removal case is exactly the one #241 is about, so this blocks that issue's remediation for implicit-only clients.

**The test suite hides it from both sides.** `TestAPIClientRedirectURIsPut_AuthCodeDisabledRejected` pins the rejection using a fixture that is `IsPublic: true, AuthorizationCodeEnabled: false`, and `createImplicitFlowClient` in the implicit integration tests builds its client and redirect URI straight through the models, never through the API. So nothing exercises the combination that fails.

**Fix:** gate on either redirect-based flow being enabled rather than on the authorization code flow alone, and reverse the direction of the existing test so it refuses a client with neither.

## Tests

- An ROPC-only client (authorization code flow off) can redeem the refresh token ROPC gave it.
- Disabling ROPC stops an outstanding ROPC refresh token; disabling the authorization code flow does not.
- The error message names the grant that was actually refused.
- ROPC fixtures no longer enable the authorization code flow, and at least one ROPC test redeems its refresh token end to end.
- An implicit-only client can have its redirect URIs managed through the admin API; a client with neither redirect-based flow cannot.

## Related

- **#241** needs part 2 fixed before an implicit-only client's callback can be urgently removed.
- **#125** covers the other ROPC token-claim defect (`auth_time` reset on refresh, plus the `acr` value).

leodip:
Closes #250.

`AuthorizationCodeEnabled` was being read as "this client does OAuth at all". It is not: it is one
flow's switch, and three places treated it as the gate for something else. This fixes all three and,
while it is in the web origins page, everything else wrong there.

**The rule this establishes:** a setting is gated by the switch that actually governs it. A refresh
is governed by the switch of the flow that issued the token. A redirect URI is governed by whether
the client has any redirect-based flow. A web origin is governed by nothing, because "this client's
app is JavaScript in a browser" is not a question any flow flag answers.

## What is wrong today

**A refresh token issued by ROPC cannot be redeemed by an ROPC-only client.** The `refresh_token`
arm opens by refusing any client with the authorization code flow off, saying it "does not support
authorization code flow", and it does that 81 lines before it works out whether the token came from
an authorization code or from ROPC. So a client that only does ROPC is handed a refresh token it can
never use, against a documented promise that it can.

The same gate has a second half, and it is the one that matters more: **turning ROPC off stops
nothing already issued.** ROPC refresh tokens are always offline type, so what survives the switch
being turned off is a long-lived offline grant. The switch governs issuance only.

**An implicit-only client's redirect URIs cannot be managed.** `PUT
/api/v1/admin/clients/{id}/redirect-uris` refuses any client with the authorization code flow off,
and so does the admin console page behind it. A redirect URI is exactly as load-bearing for implicit
as for authorization code, and rather more dangerous, so an administrator who needs to remove a
compromised callback from an implicit-only client has to turn on a flow the client does not use, or
edit the database. The authorization endpoint already knows the distinction and exempts implicit;
only the management path does not.

**The same gate is on the web origins endpoint and page**, which the issue does not mention, plus
five defects of its own found while grounding: values are accepted that CORS can never match (six of
ten realistic inputs, including the trailing slash a browser's address bar produces), only the
console's JavaScript normalizes, the per-client scoping the URL and the UI promise is not what the
server enforces and cannot be, rows already stored may be permanently dead, and every CORS-checked
request reads the whole table with no preflight caching.

### Stage 1: the refresh arm gates on the flow that issued the token
`32ba53b5`. The `!AuthorizationCodeEnabled` refusal is gone from `ValidateTokenRequest`'s refresh
arm, which now holds no flow rule at all, and the rule lives in `HandleTokenPost` instead. There it
asks which flow minted the presented token: an ROPC token is judged by the ROPC switch, resolved
through the client's override and then the global exactly as the issuing arm resolves it, and an
authorization code token keeps the refusal it has today, word for word. Both stay
`unauthorized_client`, because logging in again cannot help when the same switch blocks a fresh
login. The ROPC sentence is now one exported constant that both the issuing arm and the refresh arm
emit, so they cannot drift.

It sits **below** the replay-containment block and **above** `MarkRefreshTokenAsRevoked`. Below,
because whether a theft is detected must not depend on which switches are on: the old placement
refused a replayed token before containment ran, so a stolen token whose flow was off left its
rotation family live and nothing audited. Above, so a token the gate refuses is not spent and is
still live if the operator turns the switch back on.

The ROPC integration fixture no longer enables the authorization code flow, so all 16 ROPC clients
now run as the ROPC-only shape this issue is about, and `TestROPC_RefreshToken_OpenIdOnly` stops
passing vacuously.

Tiers: unit (3807 pass, 0 fail) and integration on sqlite (1303 pass, 0 fail) locally; the
four-engine data and integration tiers are the check suite's on this commit. Data tier not
applicable, no query, migration or schema change. Five mutations against the new gate and against
replay containment, all caught.

Two notes worth knowing. The sealed plan said six existing handler subtests needed repairing for the
new gate; five did, because one of them stops inside containment and never reaches it. And
`concepts/tokens.mdx` said refresh tokens are used in the authorization code flow with PKCE, full
stop, which is false about precisely what this stage changes; the sentence was corrected here rather
than filed.

### Stage 2: an implicit-only client's redirect URIs can be managed
Landed in `3e8e48ca`.

A redirect URI belongs to whichever redirect-based flow a client uses, and implicit is one of them.
`PUT /api/v1/admin/clients/{id}/redirect-uris` now refuses only a client with **neither**
redirect-based flow, resolving the per-client implicit override against the global setting. Those
two flags are the complete enumeration rather than a guess: `oauth.ResponseTypeInfo.IsImplicitFlow`
treats any response type containing `code` as not implicit, so nothing outside the pair can produce
a redirect.

The admin console reached the same wall independently, by clicking, so it moved too.
`HandleAdminClientRedirectURIsGet` fetches the general settings, resolves the three-state implicit
setting itself, and binds one boolean, which keeps the template to a single condition and the
inheritance rule in one place. The three `redirect_uris.requires_auth_code_*` catalog keys are
replaced in both locales by one `requires_redirect_flow` key naming both flows: a key called
`requires_auth_code` whose text no longer says that is this bug in miniature.

The OAuth2 flows endpoint was deliberately left alone. An implicit-only client was always a legal
shape; it was management of one that was broken. The identical gate on the **web origins** endpoint
is stage 5's, where it is deleted rather than corrected, because "this client's app is JavaScript in
a browser" is not a question any flow flag answers.

Widening who may manage a redirect URI cannot widen what one permits: a registered URI is inert
until `/auth/authorize` reads it, and that endpoint checks the requested flow itself.

Tiers: unit 4003 pass / 0 fail across all three modules, integration on sqlite 1304 pass / 0 fail.
Data not applicable, since the stage adds no query, interface method, migration or schema change.
Five mutations were run against the new cases, all caught, including one that reduces the API gate
to its pre-fix form and is caught by the implicit-only acceptance case.

Docs: `concepts/clients.mdx`'s `## Redirect URIs` opening sentence now names both redirect-based
flows (18 words to 21), and `oauth2-flows/implicit.mdx`'s per-client setup list gains one step
pointing at the client's Redirect URIs page (11 words, beside a 12-word step 1).

### Stage 3: `urlutil.CanonicalOrigin`

`6dc488c0`. The pure half of the web-origins repair: `CanonicalOrigin(raw string) (string, bool)` in
`src/core/urlutil`, beside the redirect-URI helpers that already do this kind of string work.

It returns the exact string a browser sends in an `Origin` header, `scheme "://" host [ ":" port ]`,
and false when the input cannot be one. It **converts** only case, surrounding whitespace, a
trailing slash, a path, a query, a fragment, and a default port (`:80` on http, `:443` on https). It
**refuses** every other shape a browser would serialize differently or reject outright: a non-ASCII
host, userinfo, any IPv6 literal, a zone identifier, an empty, leading-zero or out-of-range port,
and a host that ends in an IPv4 number without already being a strict dotted quad, which is what
`http://127.1`, `http://2130706433` and `http://0x7f000001` all are.

The asymmetry is deliberate and load-bearing. Rows stored before this change are repaired by a SQL
migration on four engines, and SQL has no punycode, no IPv6 compression and no IPv4 expansion, so
converting a shape the migration cannot reach would create a row that is permanently dead. The set
the write path accepts is therefore exactly the set the repair can reach.

The function is textual rather than parse-and-render, on the precedent of `stripPortRaw` in the
sibling file, because which component moved is the whole contract and `url.URL.String()` moves
components on its own initiative. It has no caller yet: stage 5's handler and stage 4's migration
test are the first two. A length cap is not here, being a fact about a column rather than about
origins.

Tiers: unit 3928 pass / 0 fail, of which 111 leaves are this function; data 1761 pass / 0 fail on
sqlite, MySQL, PostgreSQL and SQL Server, run because it is free rather than because the stage
claims it. Integration is not applicable, since nothing reaches an uncalled function through a
route. Ten mutations were run against the new table, one per behaviour the stage claims, and all ten
were caught.

**The check suite then went red on `6dc488c0`, on `Lint` alone**, and `df59534b` settles it. The
four database jobs were still running and every other job passed; staticcheck reported three
`QF1001: could apply De Morgan's law` findings, all in the new `origin.go`, at the default-port
test, the empty-label test and the hexadecimal-digit test. Each negated a parenthesised
conjunction. `golangci-lint` is the single definition of lint-clean here (`.golangci.yml`) and it
runs only in CI, and the QF family is absent from the standalone `staticcheck` binary `make check`
uses, so no local path could have shown it.

Each site now names what it tests, `isDefaultPort`, a local `isTrailingDot`, and `isLowerHexDigit`,
rather than inverting the expression in place, so the negation is over a call or an identifier and
each predicate reads as the comment beside it already claimed. No behaviour changed, and four
mutations prove it: `isDefaultPort` forced false, the trailing-dot test widened to admit any
interior empty label, and `isLowerHexDigit` forced true and separately extended to `'g'`.

**One of those four survived at first, and the fix is folded into the same commit.** No case
exercised a last label opening with `0x` whose remaining digits are not all hexadecimal, for example
`example.0xzz`, which is an ordinary domain rather than the IPv4 number the prefix suggests and must
be accepted verbatim. `TestCanonicalOrigin_Converts` gains `http://example.0xzz` and
`http://example.0x1f2g`. The gap predates the refactor: the same predicate was there inline.

Re-run after the fix, counting leaf `--- PASS` lines: unit 4120 pass / 0 fail, data 1761 pass / 0
fail on all four engines, and CI's own lint loop clean across all four modules with the pinned
`golangci-lint` 2.13.1. Integration remains not applicable while the function has no caller.

### Stage 4: migration 000034 and `WebOriginExists`

`b64d2421`. The storage the canonicalizer constrains: migration 000034 on all four engines, and a
new `Database.WebOriginExists`.

**The migration** repairs the rows already stored and then states an invariant the table has never
expressed. In order, and the order is load-bearing: trim, lowercase, cut everything from the first
`/`, `?` or `#` after the authority, drop `:80` on http and `:443` on https, delete what no
conversion can reach, dedupe on `(origin, client_id)` keeping the earliest row, then
`CREATE UNIQUE INDEX idx_web_origins_origin_client ON web_origins(origin, client_id)`.
Canonicalizing is what creates the duplicates, so the dedupe cannot come first, and the cut must
precede the port strip or `:443/x?y` never looks like a default port. The down migration drops the
index and stops, deliberately: a repaired value is one that already worked, and the original text is
recorded nowhere.

A row that is still not a canonical origin after the conversion is deleted. It has never matched an
`Origin` header and never could, so removing it removes nothing that ever worked, and leaving it
leaves exactly the silent dead row this change exists to remove.

**The delete predicate lists what is definitely wrong rather than whitelisting host characters**,
and that is the shape of the whole file. The two directions of error are not symmetric: deleting a
row the endpoint would accept destroys a live CORS origin, while leaving one it would refuse leaves
a row that was already dead. A whitelist of `[a-z0-9.-]` would be a quarter of the size and would
delete `http://foo_bar`, which is unusual, legal and works today. All four engines run the same
ordered conjunction over a derived table naming `authority`, `host` and `port`, and differ only in
the character-class idiom: `GLOB` on SQLite, `LIKE` classes on SQL Server, `REGEXP` on MySQL, `~` on
PostgreSQL.

**`WebOriginExists(tx, origin) (bool, error)`** joins the `Database` interface with one
implementation in `commondb` and four forwarders, and the regenerated `database_mock.go`. It
replaces `GetAllWebOrigins` in the CORS middleware from stage 5 onward, and the index leads with
`origin` because that is the only order this lookup can use: a browser's preflight carries no client
identity, so the question is about an origin with no client in hand. It checks `rows.Err()` rather
than falling out of a bare `rows.Next()` into a benign `false`, because this method gates
cross-origin access and a failure collapsed into `false` fails closed silently while one collapsed
into `true` fails open, with every mock-backed test green either way.

Tiers: unit 1164 pass / 0 fail; data 4/4 packages `ok` on SQLite, MySQL, PostgreSQL and SQL Server
in 61s, with the new migration test passing on each. Integration is not applicable, since nothing
above the data layer calls either piece until stage 5. The new data test seeds 38 shapes the old
validator permitted and asserts each one's exact surviving value or its absence, with every
expectation cross-checked against `CanonicalOrigin` so the write path and the repair cannot drift
apart. Six mutations, all caught, including the two that would otherwise pass on a plausible
mistake: the index built as `(client_id, origin)`, which enforces the identical invariant and is
useless to the lookup, and a dedupe on `origin` alone, which matches what CORS enforces while
deleting a second client's legitimate row.

Four deviations from the plan, all in `log/stage-4.md`: the cut takes `#` as well as `/` and `?` so
the migration converts everything `CanonicalOrigin` converts; the 256-character cap is its own
`DELETE` naming the column, since the predicate's derived table never carries `origin` itself; only
spaces are trimmed, because no writer the server has ever had could store a control character; and
the dedupe keeps `MIN(id)` rather than 000030's `MAX(id)`, because the survivors are byte-identical
and the earliest registration is the truer record.

One test defect was found and fixed inside the stage. The transaction-enlistment mutation **was**
caught, but as a ten-minute test-binary timeout: on SQLite a pool read blocks behind the caller's
own uncommitted write rather than answering wrongly. That read now carries a 15-second deadline and
a message naming what went wrong, and the same mutation is caught in 20 seconds.

Docs: none, internals only. Nothing here is observable until stage 5 calls it.

### Stage 5: web origins become manageable, canonical and honestly scoped

Landed in `f45241ce`.

The flow gate on `PUT /api/v1/admin/clients/{id}/web-origins` is deleted rather than corrected. A
web origin is needed when a client's app is JavaScript running in a browser, which no flow flag
expresses: an ROPC client whose single-page app calls `/auth/token` needs one and enables no
redirect-based flow at all. The Redirect URIs page next door keeps a gate on purpose, because a
redirect URI does belong to a redirect-based flow, and a comment at each site records why the two
differ so that restoring symmetry does not restore this bug.

The endpoint now stores only what CORS could match. It called `url.ParseRequestURI`, checked the
scheme, lowercased the value and stored it verbatim, so six of ten realistic inputs were accepted
and then permanently dead. It now stores `urlutil.CanonicalOrigin`'s answer, or refuses the value
with a message naming it and saying what an origin should look like, deduplicates on the canonical
form, and refuses a canonical origin over 256 characters, which is the column's width on MySQL,
PostgreSQL and SQL Server and a 500 on all three today.

The replacement became one transaction. Every insert and delete ran on its own connection, so a
failure part way through committed half a list under a 500 and two concurrent saves produced the
union of both lists. `BeginTransaction`, `AcquireClientRow`, the read, the writes,
`CommitTransaction`, with a deferred rollback.

`MiddlewareCors` calls `WebOriginExists` instead of reading every row on every checked request, and
sets `MaxAge: 600`. The cached preflight is safe because the real request still runs the origin
check, so a removed origin bites immediately.

And the console stops implying a scoping the server cannot honour. The Web Origins page says that an
origin registered against any client is permitted for every client, and renders the effective
server-wide list beside the client's own rows, so an administrator can answer "what may call these
endpoints from a browser today" without opening every client. `concepts/clients.mdx` says the same,
and that the value is a bare origin rather than a URL.

Folded in: `HandleAdminClientWebOriginsPost` handed the API's error to `JsonError`, which keeps a
status and a description only for a `*customerrors.ErrorDetail`, so every refusal above would have
reached the administrator as "An unexpected server error has occurred" with the useful sentence in
the log. One line. It is an instance of #225, whose body records that only
`HandleAdminClientRedirectURIsPost` had been repaired; two are repaired now and #225 stays open for
the rest.

Deviation worth knowing: section 4 said `GetAllWebOrigins` "stays: the API handler still needs the
full list". It does not, and nothing in production calls it now. It stays on the interface anyway,
with its data-tier coverage, because the five web-origin methods are one family.

Tiers: unit 3941 pass on three modules; data 1773 pass on all four engines; integration 1309 pass on
sqlite. Eleven mutations run against the new seams, all caught.

Docs: `site/src/content/docs/concepts/clients.mdx`, the `## Web origins` section, 52 words to 115.

## How it was verified

**The check suite is green on the final commit**, `6c867218`:
[run 32955982605](https://github.com/leodip/goiabada/actions/runs/32955982605). All eleven jobs
pass, which is the four-engine data and integration claim for this branch, plus the `Lint`,
`Vulnerabilities` and `Versions` jobs no local tier runs.

| Tier | Where it ran | Engines | Result |
|---|---|---|---|
| unit | local, and CI per module | n/a | 3941 cases, 0 failures, all three modules |
| data | local **and** CI | mysql, postgres, mssql, sqlite | 1773 cases, 0 failures, 4/4 engines |
| integration | local on sqlite only; **four engines are the check suite's** | sqlite locally, all four in CI | 1309 cases pass on sqlite; CI green on all four |

Eleven mutations were run against the new seams across the five stages, and every one was caught.
Two more were run against migration 000034's new double-port guard at the final review, both caught.
A mutation that survives is a test that cannot fail, so this is the evidence the new tests are worth
their green rather than a second pass count.

**What no job covers.** `.github/workflows/docs.yml` builds the Astro site only on a push to `main`,
a published release or a manual dispatch, so nothing on this pull request compiles `site/`. The
documentation changes here add no MDX construct the surrounding lines did not already carry.

### Review

Reviewed by `gpt-5.6-sol` at effort `xhigh`. The plan was reviewed once before any code existed, and
the whole branch was reviewed twice after every stage had landed. This run was under
`LEO_REVIEW_MODE=end`, so **no stage was reviewed on its own.**

|---|---|---|---|---|
| plan | 1 | 4 | 2 | 2 resolved into the plan, 3 escalated as decisions 13/14/15, 1 bookkeeping fix |
| final | 1 | 1 | 1 | fixed, `d97c5e6a` |
| final | 2 | 1 | 0 | fixed, `6c867218` |

- **blocking, security (plan):** the flow gate as planned sat above the replay-containment block, so
  a refusal would suppress `RevokeRefreshTokenFamily` and `AuditRefreshTokenReplayDetected`. A
  pre-existing family with nine members on `main`. Escalated as decision 13 and answered by moving
  the gate into `HandleTokenPost` below containment, which repairs one of the nine rather than
  adding a tenth.
- **blocking, conformance (plan):** the origin model was incomplete for the migration and for three
  engines, including a 267-character origin accepted against 256-character columns. Escalated as
  decision 14.
- **significant, quality (plan):** the template tests bypassed the two handlers deriving the new UI
  state. Resolved in the plan: a tenth seam on the GET handlers, with cases in stages 2 and 5.
- **significant, conformance (plan):** web-origin replacement was partial and unserialized.
  Escalated as decision 15, answered B.
- **blocking, security (final round 1):** migration 000034 stripped a scheme-default port *before*
  the delete that refuses a malformed authority, so `https://double.example:80:443` survived as
  `https://double.example:80` and an upgrade would have widened the CORS allowlist to a host and
  port nobody registered. Reproduced on all four engines against the unmodified migration. Fixed in
  `d97c5e6a`: the strip now requires the origin to hold exactly two colons. Round 2 independently
  re-derived the fix on all four dialects and mutation-tested the guard.
- **minor, conformance (final round 2):** the Web origins documentation told JavaScript clients to
  register an origin before calling `/token` and `/logout`, which the server does not mount. Fixed
  in `6c867218`.

Full account of every round, with the evidence, the family boundaries and the answered `unchecked`
lists, in [this comment](https://github.com/leodip/goiabada/pull/263#issuecomment-5423808136).

### Deviations worth knowing

- **Stage 3's check suite went red on `Lint` alone**, with three staticcheck `QF1001` findings no
  local tier can produce: `golangci-lint` is the only definition of lint-clean here and it runs only
  in CI. Settled by `df59534b` without changing behaviour. Proving that rewrite was
  behaviour-preserving is what found a real gap in the tests, folded in with it.
- **The plan review restructured the work before stage 1 was written.** Three of its four findings
  changed §2 through §5 and the plan itself, which is why the stage list here is not the one the
  agreement was sealed with.
- **The integration tier ran locally on sqlite only, by policy.** Four engines run in CI, in
  parallel, and are reported above. Two stages did not re-run it at all, each saying why: their
  change was one SQL file per engine plus a data test, and two words of prose.

### The run's record

- [Decisions](https://github.com/leodip/goiabada/pull/263#issuecomment-5423786525): sixteen taken,
  none deferred, three escalated to and answered by the repository owner.
- [Follow-ups and fold-ins](https://github.com/leodip/goiabada/pull/263#issuecomment-5423795818):
  one issue drafted ready to file, three fixes folded into stage commits, one gap recorded and
  deliberately not filed.
- [The code review](https://github.com/leodip/goiabada/pull/263#issuecomment-5423808136): all three
  rounds in full.

## For the release notes

**Turning the ROPC flow off now stops refresh tokens that flow already issued, not only new
logins.** Before this, the switch governed issuance alone, so outstanding ROPC refresh tokens, which
are always offline type, kept refreshing indefinitely after an operator had turned ROPC off.

Affected deployments are those with the authorization code flow on, ROPC off (on the client or
globally, including clients left inheriting a global that is off) and live ROPC refresh tokens.
Those tokens stop working at upgrade and the clients holding them must log in again. Turning ROPC
back on for the client restores them, since the gate reads current configuration rather than a
snapshot taken when the token was minted.

Nothing changes for the authorization code flow: a refresh token whose client has that flow off is
refused exactly as it was, with the same message.

**A web origin that is not already in canonical form is refused by the Admin API rather than
stored.** `PUT /api/v1/admin/clients/{id}/web-origins` accepted anything that parsed as an http or
https URI and stored it lowercased and verbatim, so a value with a trailing slash, a path, an
explicit `:80` or `:443`, a non-ASCII host, an IPv6 literal or an abbreviated IPv4 address was
accepted and then never matched a browser's `Origin` header. Callers that scripted their setup with
one of those shapes now receive a 400 naming the value and saying what an origin should look like.
Nothing that worked stops working, because none of those values ever worked. A canonical origin over
256 characters is refused too, where before it was a 500 on MySQL, PostgreSQL and SQL Server.

Migration 000034 repairs the rows already stored, in the same release: it canonicalizes what it can,
deletes what it cannot turn into a valid origin, deduplicates, and adds `UNIQUE (origin, client_id)`.
A deleted row is one that had never matched an `Origin` header and never could.

**A client with no redirect-based flow gets a new refusal message from
`PUT /api/v1/admin/clients/{id}/redirect-uris`**, and the three
`adminconsole.admin_clients.redirect_uris.requires_auth_code_*` message-catalog keys no longer
exist, replaced by one `requires_redirect_flow` key in both locales. Anything asserting on the old
sentence or reading those keys needs updating. No client that could manage its redirect URIs before
loses that ability.

## Status

**Complete.** All five stages have landed, the review is closed and the check suite is green on
`6c867218`.

The five stages, in order: the refresh arm's gate (`32ba53b5`), the redirect-URIs gate (`3e8e48ca`),
the origin canonicalizer (`6dc488c0`, with `df59534b` settling `Lint`), migration 000034 with
`WebOriginExists` (`b64d2421`), and the web origins surface itself (`f45241ce`). Two review fixes sit
on top: `d97c5e6a` for the migration's double-port defect and `6c867218` for the documentation's
endpoint names.

**This is a breaking change**, in two ways, both described under `## For the release notes` above:
turning ROPC off now stops refresh tokens that flow already issued, and a web origin that is not
already in canonical form is refused rather than stored, with migration 000034 repairing or deleting
the rows already there.

Ready for review.


leodip:
## Decision 13: when a stolen refresh token is replayed while its flow is switched off, should the server still detect the theft?

Three questions came out of the plan review, all verified against the code. This is the first.

**What a refresh token is here.** It lets an app get a fresh access token without asking the user to
sign in again. Every time one is used, Goiabada issues a replacement and retires the old one. That
retirement is also the theft detector: if a refresh token leaks, sooner or later the thief and the
real app both try to use the same one, whichever goes second presents a token that was already
retired, and that is the server's only signal the token leaked. When it sees that, it revokes every
token in that chain and writes a "refresh token replay detected" entry to the audit log.

**Where the detection sits.** In the token endpoint handler, `handler_token.go`, in the
`refresh_token` case. It runs only after the request has passed validation, so any refusal during
validation happens first and the detection never runs at all.

**What this change puts in front of it.** Part 1 of this issue makes a refresh honour the switch of
the flow that issued the token: an ROPC refresh token stops working when ROPC is switched off, an
authorization-code one stops when the authorization code flow is switched off. The plan puts that
refusal in validation, which is where the current one already lives. So while a flow is switched
off, replaying a stolen token from that flow gets a plain "this client is not authorized to use that
grant type" and nothing is revoked and nothing is logged.

**Who is affected and what it costs.** Nobody is locked out, because while the switch is off nothing
refreshes anyway. What is lost is that the stolen chain survives. Turn the switch back on and the
thief's token works again, and there is no audit entry anywhere saying a theft was ever presented.

**The part that makes this less clear-cut than it sounds.** The server already behaves this way, for
the flow almost everybody uses. Today's gate is the very first thing the refresh path does, and it
applies to every refresh token regardless of which flow issued it. So today, with the authorization
code flow switched off, a replay is already uncontained and unlogged. This change moves one row into
the good column and one row out of it: an ROPC token replayed with the **authorization code** flow
off is detected after this change and is not today, and an ROPC token replayed with **ROPC** off is
detected today and would not be after this change. There are nine other refusals in validation
already sitting in front of the detector: disabled user, superseded credentials, wrong client,
terminated session, the PKCE rule, three session and lifetime checks, and scope.

**Worked example.** An app has been using ROPC. Its refresh token leaks in a log export. You notice
something odd and switch ROPC off for that client while you investigate. During that window the
thief's client retries with the token it holds, which is the one your app already rotated away from.

- Under **A**, the retry is refused, and that is all. Nothing in the audit log. When you switch ROPC
  back on, the token the thief holds is dead but the rest of that chain is alive.
- Under **B** or **C**, the retry is refused **and** the whole chain is revoked and a replay entry is
  written. You come back to the audit log and see the theft. Switching ROPC back on restores nothing
  to the thief.

**The one axis.** Which replays the server detects, and where the flow refusal has to live to get
that.

| | A. leave it in validation | B. move it into the token handler | C. restructure validation |
|---|---|---|---|
| Replays detected when a flow is off | no, for either flow | yes, for both flows, including the authorization-code case that is broken today | yes, and also for the other nine refusals |
| Where the refusal is written | validation, as planned | the token handler, just after the theft check and before the token is spent | validation, with a new result type so the handler can act first |
| Tests that own it | the validator's unit table | the token handler's unit tests, which already own the theft-detection case | both, plus every existing refusal's case |
| Size | as planned | one block moves between two files | a change to what validation returns, and every refusal reworked |
| Extra escalation risk | none | none: everything the moved gate needs is already in hand there | this is the token path's core contract |

What does not differ: the refusal the client sees is identical under all three, `unauthorized_client`
with the same sentence, and under all three it still happens only after the server has established
the token belongs to the client presenting it, so a stranger holding somebody else's token still
learns nothing about which flow issued it. Stages 2 to 5 are unaffected either way.

**Recommendation: B.** Whether a theft gets detected should not depend on which switches happen to
be on. B is the only option that makes that true without touching what validation returns, it fixes
the authorization-code case that is already broken rather than just declining to make it worse, and
the pieces are all there: the handler already has the client, already knows whether the token came
from ROPC, and already has a unit test covering the theft-detection block that the new refusal would
sit next to. C is the thorough answer and I am not recommending it, because it reworks the contract
between validation and the token handler on the token path, which is a much bigger change than this
issue asked for and would be better as its own.

**Why I am asking rather than doing it.** B contradicts a decision already sealed in this agreement,
which put the refusal in validation and gave reasons for that placement. Those reasons still hold
under B, but reversing a sealed placement on the token path is yours to make. If you pick A, say so
and the run will record the gap and file it; that is a legitimate answer and it keeps this change to
exactly what the issue asked for.

---

## Decision 14: what should count as a valid web origin, and what should happen to the ones already stored that cannot be fixed?

**What a web origin is here.** It is the address of a JavaScript app in a browser, registered so that
app is allowed to call `/auth/token`, `/auth/logout` and `/userinfo` from the browser. Browsers send
that address in a header, and Goiabada compares it to the stored value character for character. One
character different and the call is blocked.

**Where you meet it.** Admin console, Clients, a client, Web Origins. Behind it,
`PUT /api/v1/admin/clients/{id}/web-origins`.

**What happens today.** The endpoint accepts almost anything that looks like a web address, lowers
the case, and stores it exactly as typed. This issue already found that six out of ten realistic
values are accepted and can then never match anything, including `https://app.example.com/`, which
is what you get by copying out of the browser's address bar. The plan review found a second group,
and I re-ran all of them against the live endpoint's validator to confirm: an address with accented
or non-Latin letters (`https://bücher.example`), one with a username in front of it
(`https://user@example.com`), an IPv6 address written out in full, a port left empty
(`https://example.com:`) or written with a leading zero (`:0443`), an IP written short
(`http://127.1`, `http://2130706433`), a port number that does not exist (`:99999`), and an IPv6
address with a network-interface suffix. All ten are accepted and stored. A browser sends something
different for the first seven and refuses to load the page at all for the last two.

**Who is stuck.** The administrator saves, gets a green result and a row in the list, and the app is
still blocked, with nothing anywhere saying why. The console's own page happens to clean up the
common cases in JavaScript before sending, so this bites whoever sets clients up through the API.

Two things have to be decided, and they are connected.

**14a. Should the server convert what you type, or insist you type the exact thing?**

The plan repairs the rows already in the database with a SQL migration. SQL can lowercase, cut a
trailing slash or a path, and drop a default port. SQL cannot convert `bücher` into the form a
browser actually sends, cannot compress an IPv6 address, and cannot expand `127.1` into `127.0.0.1`.
So whatever the save endpoint accepts and converts, the migration has to be able to convert too, or
old rows in those shapes stay dead forever.

| | A. convert only what the migration can | B. convert everything a browser would | C. accept only exact values |
|---|---|---|---|
| `https://APP.example.com/` | converted and saved | converted and saved | refused, with the correct form shown |
| `https://bücher.example` | refused, with the correct form shown | converted and saved | refused, with the correct form shown |
| `https://example.com:99999` | refused | refused | refused |
| Old rows in the exotic shapes | can be handled, because the endpoint can never create one again | cannot be fixed in SQL, so the repair has to move into the app's startup instead | can be handled |
| Effect on the console | none, it already cleans up before sending | none | the page's cleanup becomes load-bearing |

**14b. There is no length limit, and the column has one.** A legitimate origin can be 267 characters.
The column holding it is 256 characters on MySQL, PostgreSQL and SQL Server, and unlimited on SQLite.
So a valid long address is a server error on three databases out of four, and our SQLite test run
cannot see it. Either widen the three columns in the same migration, or refuse anything over 256
characters at the endpoint with a message saying so.

**Recommendation: 14a option A, and for 14b, refuse over 256 characters.** Option A has a property
the other two do not: the set of values the endpoint accepts becomes exactly the set the migration
can repair, so this class of dead row can never be created again and the repair is complete by
construction. It also keeps the fix where the plan put it. Alongside it I would have the migration
**delete** the rows it cannot turn into a valid origin rather than leave them sitting there. Such a
row has never matched anything and never could, so deleting it removes nothing that ever worked, and
the alternative is leaving exactly the silent dead rows this issue exists to remove. On 14b, refusing
is strictly better than today on three databases out of four, where the value currently produces a
server error, and it avoids adding a column change to a migration that is already doing string
surgery four different ways.

**What it costs to get this wrong.** Choosing B means the row repair moves out of a one-time
migration and into code that runs at every startup, which you rejected earlier in this issue for good
reasons. Choosing C means an administrator who pastes a URL out of a browser gets a refusal instead
of a save, on an endpoint that has never refused one. Widening the columns instead of capping means
three more engine-specific statements in the riskiest file of this change.

---

## Decision 15: should saving a client's web origins be one all-or-nothing operation?

**What happens when you save.** The Web Origins page replaces the whole list. The server reads what
is stored now, inserts the ones you added, deletes the ones you removed, and each of those is its own
separate write that takes effect immediately and on its own.

**Where you meet it.** The same page as decision 14, and the same endpoint behind it.

**What goes wrong.** If the server fails part way through, you get an error, and the list has already
half-changed. And if two administrators save different lists at the same time, both are told it
worked, and what ends up stored is the two lists merged together, which is neither of the things
either of them asked for. The unique index this change adds stops the same origin appearing twice for
one client and does nothing about either of these.

**Worked example.** You have three origins registered. You remove two and add one, and a colleague at
the same moment removes the one you are adding. Both saves report success. The list afterwards can
contain your addition that your colleague deleted, or their state plus your deletions, depending
purely on ordering. Neither of you can tell from the page that anything went wrong.

**Why this is a question at all.** When we widened this change to cover web origins, we listed eight
specific defects and said that list was the boundary. This is a ninth, found after that, and it was
not on the list. Your instruction at the time was "all fixes related to web origins, I want it
perfect and solid in the end", which does cover it. So this is a scope call, not a technical one.

| | A. leave it | B. web origins only | C. web origins and redirect URIs |
|---|---|---|---|
| Partial save on failure | still possible | no | no |
| Two administrators at once | still merges | one wins cleanly | one wins cleanly |
| Redirect URIs page, same defect | untouched | untouched, filed as a follow-up | fixed |
| New machinery needed | none | none, this is already used by the client settings page next door | none |
| Size | nothing | one handler, plus a test that a failure rolls back | two handlers |

What does not differ: no new database method, no schema change, no migration, and nothing a user
sees when the save succeeds. All three are the same amount of work to reverse.

**Recommendation: B.** The pieces already exist and are already used a few hundred lines up in the
same file, including a unit test asserting the read happens inside the transaction, so this is small
and low-risk. C is tempting because the Redirect URIs page has exactly the same defect, but redirect
URIs were never in scope for anything except the flow gate, and widening this change twice is how it
stops landing. Under B, the redirect-URIs version gets written up as a ready-to-file follow-up
instead. If you would rather this whole thing were a follow-up too, A is a perfectly good answer and
the write-up happens either way.

---

**Answering.** Three letters is enough, for example "13 B, 14 A + refuse, 15 B", with any reasoning
you want recorded. The run stops until then and none of the five stages has started, so nothing has
to be unwound whichever way these go.

leodip:
All three your recommendations

leodip:
## Decisions: sixteen taken, none deferred

Every decision this change rested on is **Decided**. Nothing was deferred, so there is no assumption
here for you to accept or reverse, and no decision is left open.

That is worth stating explicitly rather than leaving as an absent comment: a run that defers posts
its assumptions here for you to overrule, and this one has none to post.

**Thirteen were settled during the interview, with you.** They are what the change is: where the
flow gate sits in the refresh arm and which error it answers (1, 2), that the ROPC gate resolves the
global setting rather than reading only the per-client override (3), that the redirect-URIs gate
becomes "any redirect-based flow" rather than disappearing (4), that web origins lose their flow
gate entirely (5), that every web-origins defect found while grounding came into scope (5b), what
happens to the admin console's blocked panels and their catalog keys (6), how the ROPC fixture
changes across 16 call sites (7), where the origin canonicalizer lives and which tier proves each
behaviour (8), and the four web-origins repairs: the per-client assertion (9), the migration that
repairs stored rows (10), the index that ends the CORS table scan (11), and the unique index (12).

**Three were raised by the run at the plan gate and escalated to you**, because each was a scope or
security call the run did not own. You answered all three on 2026-08-26, in
[this comment](https://github.com/leodip/goiabada/pull/263#issuecomment-5418893813), and they were
applied to §2, §3, §4, §5 and the plan before stage 1 was written. Restated here so this comment
stands on its own:

**13. Whether a replayed refresh token still triggers theft containment when its flow's switch is
off.** Answered: it does. The flow refusal moved out of `ValidateTokenRequest` and into
`HandleTokenPost`, below the replay-containment block. The finding behind it was that every
validator refusal suppresses `RevokeRefreshTokenFamily` and `AuditRefreshTokenReplayDetected`,
because the handler runs only after the validator returns and the validator never reads `Revoked`.
This was a pre-existing family with nine members on `main`; the placement chosen repairs one of
them rather than adding a tenth.

**14. What counts as a web origin, and what happens to rows that cannot be repaired.** Answered in
two parts: the save endpoint converts exactly what the repair migration can convert (case, trailing
slash, path, query, fragment, default port) and refuses the shapes a browser would send differently
or reject outright; and a canonical origin over 256 characters is refused rather than the column
widened, which today is a 500 on MySQL, PostgreSQL and SQL Server. The migration deletes the stored
rows it cannot repair, all of which had never matched an `Origin` header and never could.

**15. Whether replacing a client's web origins becomes one transaction.** Answered B: yes for web
origins, and the identical defect on the redirect-URIs endpoint is deliberately left out rather
than widening this change a second time. That exclusion is the drafted follow-up in the next
comment, and it is the one thing here you may want to reverse: filing it costs an issue, and
leaving it means `PUT /api/v1/admin/clients/{id}/redirect-uris` keeps the partial-save and
concurrent-merge behaviour this change removed from its neighbour.

**What each of these costs to reverse.** Decisions 3 and 5 are the breaking ones and both are
described under `## For the release notes` in the pull request body: turning ROPC off now stops
refresh tokens that flow already issued, and non-canonical web origins are refused rather than
stored. Reversing either is a code change plus a release note retraction, not a configuration
switch. Decision 10's migration is the only one with a data consequence that an operator cannot
undo by editing settings: it deletes stored web-origin rows it cannot repair, and its down
migration cannot restore them, by design, since the rows it deletes are ones CORS could never have
matched.

leodip:
## What was found next to the work: one follow-up to file, three fixes folded in

This is the run's complete account of everything it found beside the change itself. The half that
was left undone is first; the half that was fixed in passing is below it, because a fold-in you
were never shown is one you cannot object to.

---

## To file: one issue

**Duplicate search re-run at the finish, 2026-08-26**, because the tracker moves while a run works.
`gh issue list --state all --search "redirect URIs transaction"` and `"partial save atomic redirect"`
return only **#239**, the closed spike this change came from. `--search "AcquireClientRow"` returns
**#260**, which is new since the draft was written and is worth naming: its defect 3 is *also* about
concurrent client saves, but it is a different endpoint and a different mechanism. #260 defect 3 is
`UpdateClient` projecting every column, so two administrators saving two different **sections** of a
client lose one section. This draft is the sub-resource endpoint performing N autocommitted writes
with no transaction at all. Same family, neither covers the other. **#249** is the same family again
on `user_session_clients` and `user_consents`. No duplicate. Filing it as it stands is correct.

**Title:** `PUT /api/v1/admin/clients/{id}/redirect-uris` can commit half a redirect URI list under a 500

**Labels:** `bug`, `go`

**Body:**

> `HandleAPIClientRedirectURIsPut` in
> `src/authserver/internal/handlers/apihandlers/handler_api_clients.go` replaces a client's whole
> redirect URI set, and it performs every write outside a transaction. It calls
> `database.ClientLoadRedirectURIs(nil, client)` to read the current list, then
> `database.CreateRedirectURI(nil, ...)` once per added URI and `database.DeleteRedirectURI(nil, ...)`
> once per removed one, each on its own connection.
>
> Two things follow.
>
> **A failure part way through commits part of the save.** A database error on the second of three
> writes leaves the first one persisted, answers `500 INTERNAL_ERROR`, and leaves the client's list
> as neither what the administrator sent nor what it was before. The case where that matters is the
> case an administrator is most likely to be in a hurry for: removing a compromised callback and
> adding its replacement in one save, which can end up having done only the removal, or only the
> addition.
>
> **Two concurrent saves produce the union of both lists.** Each request reads the same current list
> and writes its own diff of it, so an administrator removing a URI while another adds one can find
> the removed URI still registered. The read is not serialized against the writes by anything.
>
> `HandleAPIClientWebOriginsPut`, the endpoint next door, was repaired for exactly this in #250:
> `BeginTransaction`, a deferred `RollbackTransaction`, `AcquireClientRow` to serialize the read
> against a concurrent save, the current-list read, the inserts and deletes, and
> `CommitTransaction`. `updateClientNotOwningAuthenticationMode` in the same file has run that
> sequence since #245. Nothing new is needed on the `Database` interface: all four methods are
> already on it and already covered by the data tier.
>
> Redirect URIs were deliberately left out of #250 to keep that change from widening twice; it
> fixed the flow gate on this endpoint and nothing else about it.
>
> **What closing it takes:** the same wrapping, plus a unit case on `handler_api_clients_test.go`'s
> strict-mock harness asserting that the current list is read under `AcquireClientRow` inside the
> writing transaction, and one asserting that a failed write reaches `RollbackTransaction` and never
> `CommitTransaction`. Both shapes exist for the web-origins endpoint and can be copied. No data
> tier and no migration.
>
> Related but distinct: #260 defect 3 (whole-client column projection), #249 (the same family on
> `user_session_clients` and `user_consents`).

Files as it stands:

```bash
gh issue create \
  --title 'PUT /api/v1/admin/clients/{id}/redirect-uris can commit half a redirect URI list under a 500' \
  --label bug --label go \
  --body '`HandleAPIClientRedirectURIsPut` in `src/authserver/internal/handlers/apihandlers/handler_api_clients.go` replaces a client'"'"'s whole redirect URI set, and it performs every write outside a transaction. It calls `database.ClientLoadRedirectURIs(nil, client)` to read the current list, then `database.CreateRedirectURI(nil, ...)` once per added URI and `database.DeleteRedirectURI(nil, ...)` once per removed one, each on its own connection.

Two things follow.

**A failure part way through commits part of the save.** A database error on the second of three writes leaves the first one persisted, answers `500 INTERNAL_ERROR`, and leaves the client'"'"'s list as neither what the administrator sent nor what it was before. The case where that matters is the case an administrator is most likely to be in a hurry for: removing a compromised callback and adding its replacement in one save, which can end up having done only the removal, or only the addition.

**Two concurrent saves produce the union of both lists.** Each request reads the same current list and writes its own diff of it, so an administrator removing a URI while another adds one can find the removed URI still registered. The read is not serialized against the writes by anything.

`HandleAPIClientWebOriginsPut`, the endpoint next door, was repaired for exactly this in #250: `BeginTransaction`, a deferred `RollbackTransaction`, `AcquireClientRow` to serialize the read against a concurrent save, the current-list read, the inserts and deletes, and `CommitTransaction`. `updateClientNotOwningAuthenticationMode` in the same file has run that sequence since #245. Nothing new is needed on the `Database` interface: all four methods are already on it and already covered by the data tier.

Redirect URIs were deliberately left out of #250 to keep that change from widening twice; it fixed the flow gate on this endpoint and nothing else about it.

**What closing it takes:** the same wrapping, plus a unit case on `handler_api_clients_test.go`'"'"'s strict-mock harness asserting that the current list is read under `AcquireClientRow` inside the writing transaction, and one asserting that a failed write reaches `RollbackTransaction` and never `CommitTransaction`. Both shapes exist for the web-origins endpoint and can be copied. No data tier and no migration.

Related but distinct: #260 defect 3 (whole-client column projection), #249 (the same family on `user_session_clients` and `user_consents`).'
```

---

## Folded in: three fixes, each in a stage commit

Each passed the four tests for folding in rather than filing: it rides on a surface the stage
already touched, there was nothing to settle, the fix named is the whole of the fix, and the
change's goals stay true as written.

**1. `concepts/tokens.mdx` said refresh tokens are an authorization-code feature.** Commit
`32ba53b5`, stage 1, as its `docs(site)` bullet. The `## Refresh tokens` section opened "Refresh
tokens are used in the authorization code flow with PKCE (the client credentials flow doesn't use
refresh tokens)". That is false about exactly what stage 1 changes: ROPC issues a refresh token on
every login, the site's own ROPC page says so, and after stage 1 an ROPC-only client can finally
redeem it. A reader running ROPC would have concluded the whole section, including the offline idle
timeout and maximum lifetime it goes on to describe, was not about them. One sentence rewritten to
name both flows, 19 words to 27.

**2. The hexadecimal-label predicate had no acceptance case.** Commit `df59534b`, alongside the lint
fix stage 3's red check suite forced. `endsInIPv4Number` in `src/core/urlutil/origin.go` decides
whether a host's last label is an IPv4 number, and a label opening `0x` is one only while every
digit after the prefix is hexadecimal. The tests had the positive side only, so forcing the hex test
to accept every byte still left the suite green. Found by mutation while proving the lint rewrite
was behaviour-preserving, not introduced by it: the same predicate was there inline before.
`TestCanonicalOrigin_Converts` gains `http://example.0xzz` and `http://example.0x1f2g`, and both
mutations are then caught.

**3. `HandleAdminClientWebOriginsPost` swallowed the API's 400 into a generic 500.** Commit
`f45241ce`, stage 5, one line: `handlers.HandleAPIErrorJson` in place of `httpHelper.JsonError`.
Goal 8 asks for a refusal "with a message saying what the value should look like", and stage 5 is
what makes that message load-bearing, because the page's own `new URL().origin` converts a non-ASCII
host, an IPv6 literal and an abbreviated IPv4 address that `CanonicalOrigin` refuses. A message the
administrator never reads does not discharge the goal.

This third one is an instance of open issue **#225**, "Admin console proxy handlers turn downstream
400 responses into generic 500s", whose body records that `HandleAdminClientRedirectURIsPost` "was
repaired in #224, and it is the only one". Two are repaired now. **#225 stays open for the rest**,
and no duplicate was filed. Pinned by
`TestHandleAdminClientWebOriginsPost_APIRefusalReachesTheBrowser`, two cases, mutation caught.

---

## Recorded, deliberately not filed: no local seam gates lint

Stage 3's check suite went red on `Lint` alone, with three staticcheck `QF1001` findings no local
tier could have produced. `golangci-lint` is the single definition of lint-clean here
(`.golangci.yml`, `docs/ci-v2-design.md` 5.8) and it runs only in CI; the `QF` family is also absent
from the standalone `staticcheck` binary `make check` uses, so that path passes too.

Nothing is filed against this repository for it, on purpose. A test asserting the linter's opinion
would have to run the pinned 2.13.1 from inside the unit tier and would then fail on any machine
without it, the host included, which is a worse gate than the one CI already provides. The gap is in
the run's own habits, which live outside this repository. Recorded here so the decision not to file
is visible rather than silent.

leodip:
## The code review, in full

This run was under `LEO_REVIEW_MODE=end`, so **no stage was reviewed on its own**. The plan was
reviewed once before any code existed, and the whole branch was reviewed twice at the end. The
verdicts themselves live in a gitignored directory that the next round overwrites, so this comment
is the durable copy: every round, every finding, the evidence, and what happened to it.

Reviewer throughout: `gpt-5.6-sol` at effort `xhigh`, taken from `.review/reviewer.env` rather than
inferred.

|---|---|---|---|---|
| plan | 1 | 4 | 2 | 2 resolved into the plan, 3 escalated as decisions 13/14/15, 1 bookkeeping fix |
| final | 1 | 1 | 1 | fixed in `d97c5e6a` |
| final | 2 | 1 | 0 | fixed in `6c867218` |

---

# Plan gate, round 1 (2026-08-26)

Fresh Codex session, round 1 of a budget of 2. Scope: `plan.md` in full against the sealed contract,
all three axes. No code existed; the branch was one empty commit above `origin/main`. All three axes
returned `reviewed`. Four findings, **all four verified against the code by the run before any was
acted on. None was refuted.**

**1. blocking, security. The flow gate hides refresh-token replay from containment.** Confirmed.
`handler_token.go`'s `refresh_token` case is where `refreshToken.Revoked` is read and where
`RevokeRefreshTokenFamily` and `AuditRefreshTokenReplayDetected` run, and it executes only after
`ValidateTokenRequest` returns, so every validator refusal suppresses it. The validator never reads
`Revoked` at all. The run extended the finding: this is a **pre-existing family with nine members on
`main`**, and the change's net effect as planned was one row gained and one row lost rather than a
new defect. Escalated as **decision 13** with three options and a recommendation. The half needing
no answer, that stage 1's integration test could pass vacuously by presenting the retired parent
instead of the live rotated successor, was fixed in the plan at step 8.

**2. blocking, conformance. The origin model is incomplete for the migration and for three
engines.** Confirmed, and re-executed independently: a standalone mirror of the current validator
accepted all ten shapes the reviewer named, and the same URL implementation the console's JavaScript
uses serializes seven of them differently and refuses two. The 267-character origin was confirmed
accepted at 267 characters, against `origin` columns of 256 on MySQL, PostgreSQL and SQL Server and
`TEXT` on sqlite, read from all four `schema.sql` files. Escalated as **decision 14**.

**3. significant, quality. Template tests bypass the two handlers that derive the new UI state.**
Confirmed. `rendertest.render` takes a caller-supplied bind map and calls `RenderTemplateToBuffer`
directly, so no handler runs, and both new console facts originate in the GET handlers.
`needs_human: false`, so resolved in the plan: §5 gained a tenth seam for
`HandleAdminClientRedirectURIsGet` and `HandleAdminClientWebOriginsGet` with the reason the
rendertest seam cannot carry it, and stages 2 and 5 gained the cases, extending the `stubApiClient`
harness already in `handler_admin_client_redirect_uris_test.go` rather than inventing one.

**4. significant, conformance. Web-origin replacement is partial and unserialized.** Confirmed by
reading the handler: `ClientLoadWebOrigins(nil)`, then `CreateWebOrigin(nil)` and
`DeleteWebOrigin(nil)`, all autocommitted. Escalated as **decision 15**, because it sat outside the
scope boundary decision 5b had set, so whether it landed here was the user's call.

**Bookkeeping, fixed in passing.** The reviewer noted that §5's rejected-candidate list and decision
8's reasoning both asserted no `handler_api_clients_test.go` exists. It does: 363 lines with a
request and strict-mock harness, landed by #245 shortly before the agreement was written. Both
passages corrected; decision 8's outcome does not move, but the false premise is gone.

**The `unchecked` list, all five entries answered.** Three dismissed as true-by-construction at a
plan gate (no production code existed; nothing had changed; unrelated surfaces out of scope). One
dismissed for this gate and returned as real coverage at the final review (no interactive
admin-console browser pass, since the pages had not changed). One **filed**, as the substance of
decision 14: the complete URL Standard input boundary, which 14a decides and stage 3 then owns
exhaustively at seam 5.

**Escalation budget, named for the next `/leo-spec`.** Three blocking decisions at the plan gate on
top of thirteen settled during the interview is at the edge of normal, and all three were classes
the interview could have reached: 13 by asking where replay containment sits relative to every
refusal in the refresh arm, 14 by asking what `CanonicalOrigin`'s stated contract implies against
the column widths, 15 by noticing the W-list was built from a read of the CORS path and the
validator and never from the write path's transaction handling.

---

# Final gate, round 1: the whole branch, `origin/main`..`f45241ce`

Fresh session, 2026-08-26 02:50 to 03:10 UTC. All three axes reviewed. **1 blocking, 0 significant,
0 minor.** This was the first complete reading of the code.

## F1, blocking, security. Migration 000034 activated malformed double-port origins instead of deleting them

**The claim.** Each engine's migration strips a trailing scheme-default port *before* the delete
that refuses a malformed authority. A stored `https://double.example:80:443` therefore became
`https://double.example:80`, which the delete then judged valid and kept, **so an upgrade would
widen the CORS allowlist to a host and port nobody registered.** Reproduced on all four engines.

**Verified, and the finding stands.** Both halves checked against the code rather than taken on the
reviewer's word:

- `urlutil.CanonicalOrigin` refuses that input. It splits the port at the **last** colon
  (`strings.LastIndex`), so `isCanonicalHost` sees `double.example:80`, and `:` is in the forbidden
  set.
- The SQL disagreed. `UPDATE ... WHERE origin LIKE 'https://%:443'` ran at step 3, before step 4's
  `length(authority) - length(replace(authority, ':', '')) > 1`. Step 4's own comment rested on the
  assumption step 3 broke: "Splitting the port at the FIRST colon rather than the last, as
  CanonicalOrigin does, is equivalent here because a second colon is refused outright two clauses
  below." Step 3 removed that second colon first.

**Reproduced before fixing.** Four double-port shapes added to the migration's data test and run
against the *unmodified* migration on sqlite. Three failed, retaining
`https://dport443.example.com:80`, `http://dport80.example.com:8080` and
`https://dport2.example.com:443`. The triple-port case passed, which is the boundary: two strips are
needed to get it down to one colon and only one runs.

**The family, boundary established independently.** The reviewer named the family "lossy migration
normalization that erases evidence of an invalid authority before validation" and reported
`mutations_run`, so its `why_exhaustive` was checked rather than accepted. The members are the four
transformations that mutate `origin` before the delete:

| Step | Agrees with `CanonicalOrigin`? |
|---|---|
| 1, `trim` then `lower` | Yes on case. On whitespace the SQL trims only spaces where the Go trims all of `<= ' '`, so a tab-padded row is deleted rather than repaired: the safe direction, it cannot reach the table, already recorded in the file |
| 2, cut at the first `/`, `?` or `#` | Yes. Same rule and position as `strings.IndexAny(authority, "/?#")` |
| 3, drop the default port | **No.** This was F1 |
| 4 onward | Validation and dedupe, not normalization |

Step 3 is the only step that can reduce the authority's colon count, which is what bounds the
family: no other step can turn a value the delete would refuse into one it accepts.

**Fixed** in `d97c5e6a`, on all four engines. The strip now requires the whole origin to hold exactly
two colons, the one in `://` and one in the authority; step 2 has already cut everything after the
authority, so that count is exactly "the authority has one port separator". The forced answer's
second option was taken (guard the rewrite) rather than its first (delete before stripping), because
the over-length delete has to stay after canonicalization to measure the canonical form, per
decision 14b.

**Regression coverage,** permanent and on all four engines, in the migration's existing data test:
`https://dport443.example.com:80:443`, `http://dport80.example.com:8080:80`,
`https://dport2.example.com:443:443` and `https://tport.example.com:80:443:443`. The third is the
retained-but-still-dead variant rather than the activated one; the fourth pins the boundary. The
file's existing tying assertion covers them for free: a case marked deleted must be one
`CanonicalOrigin` refuses, so the two sets cannot drift apart again.

### Tiers after the fix

| Tier | Engines | Result |
|---|---|---|
| data | sqlite, mysql, postgres, mssql | pass, 4/4 engines, ~53s |
| unit | n/a | pass, 0 failures, all three modules |
| integration | not re-run | one SQL file per engine plus a data test; no handler, route or schema shape moved |

### Mutations

| Replacement | Result |
|---|---|
| the https guard `AND length(origin) - length(replace(origin, ':', '')) = 2` deleted | **CAUGHT** |
| the http guard deleted | **CAUGHT** |

Both restored by the mutation tool's trap and verified by hash.

### The reviewer's `unchecked`, all four answered

1. **No interactive browser pass over the two admin-console pages.** *Dismissed.* The console diff
   is markup only: the diff of `admin_clients_web_origins.html` contains no change to a function,
   listener, `new URL` call or handler. The two new console facts are assigned to the GET handlers'
   bind and to `rendertest`, deliberately, and both ran.
2. **The 1,746-input differential corpus ran only on sqlite.** *Dismissed, with the gap named.* The
   four-engine obligation is the migration's data test, which runs about forty shapes including this
   family on all four engines. The corpus is a scratch probe with no home in the repository.
3. **Unchanged signing-key custody, issuer and audience validation, session-cookie attributes, PKCE
   internals, unrelated rate-limiters.** *Dismissed.* None is touched by this branch.
4. **The full green tiers were not re-run locally.** *Covered, and superseded.* Sound for
   `f45241ce`; `d97c5e6a` changes code, so the data tier was re-run on all four engines and the unit
   tier on all three modules.

**The reviewer's two follow-ups were discharged rather than filed:** correcting the migration on all
four engines and retaining the all-engine regression test is `d97c5e6a`; rerunning the data tier and
the check suite is the table above.

**Round outcome.** Productive: one blocking finding, real, fixed in production code. Round 2 was
requested because the fix was SQL written four ways that nothing had read but its author, sitting on
the security boundary the finding was about.

---

# Final gate, round 2: the whole branch, `origin/main`..`d97c5e6a`

Resumed session, 2026-08-26 09:41 UTC. All three axes reviewed. **0 blocking, 0 significant, 1
minor.** `unchecked` was empty.

**What the reviewer actually ran**, from the verdict's `ran` list: the recorded split tree hashes;
all 26 agreement anchors; the complete agreement, plan, five stage logs, closing record and all 48
changed files; the 1,746-case migration differential on sqlite and then again through MySQL,
PostgreSQL, SQL Server and SQLite, **zero mismatches on each**; the same differential with the new
double-port guard removed from a scratch copy of the migration, which its probe caught as **25
mismatches**; the 1,705-case WHATWG URL differential, **zero false accepts**; CI green at
`d97c5e6a`; `git diff --check` and a clean tree.

**Round 1's blocking finding is independently closed.** The reviewer re-derived F1's fix on all four
dialects and mutation-tested the guard itself. That is a second, independent demonstration of what
round 1's own two mutations showed, produced by a party that did not write the fix.

## F1 (round 2), minor, conformance. The Web origins paragraph named endpoints the server does not mount

**The claim.** `site/src/content/docs/concepts/clients.mdx`, the "Web origins" section, told
JavaScript clients to register an origin before calling `/token`, `/logout` or `/userinfo`. The
server mounts and CORS-enables `/auth/token`, `/auth/logout` and `/userinfo`. Only the third
spelling was right.

`needs_human: true` arrived with a populated `forced_answer`, which is a finding the code settles
rather than a choice to escalate. It was settled from the code.

**Verified, and the finding stands.** Both halves read rather than taken on trust:

- `src/authserver/internal/server/routes.go` registers `POST /token` and `GET`/`POST /logout` inside
  `s.router.Route("/auth", ...)`, so their request paths are `/auth/token` and `/auth/logout`.
  `/userinfo` is registered at the root and is the one spelling the page had right.
- `src/core/middleware/middleware_cors.go`'s `AllowOriginFunc` switches on `r.URL.Path` and its
  origin-checking arm reads `case "/auth/token", "/auth/logout", "/userinfo":`. A request to a bare
  `/token` would fall through to `return false` even if it resolved, which it does not.

So the page sent a reader to configure CORS for three paths, two of which 404, **and the two it got
wrong are the two this branch's CORS work is about.**

**The family, boundary established independently.** The reviewer reported `mutations_run: []`, so its
`why_exhaustive` was treated as an opinion and the boundary established here. Every user-facing
surface naming a CORS-relevant endpoint:

| Surface | Names | Sound? |
|---|---|---|
| `concepts/clients.mdx` "Web origins" | `/token`, `/logout`, `/userinfo` | **No.** This was the finding |
| `reference/security.mdx` CORS bullet | no endpoint at all | Yes, nothing to drift |
| `active.en.toml`, `web_origins.intro_em1`/`em2` | `/auth/token`, `/userinfo` | Yes |
| `active.pt-BR.toml`, the same two keys | `/auth/token`, `/userinfo` | Yes, paths are not translated |
| `integration/endpoints.mdx`, `index.mdx`, `rest-api.mdx`, `concepts/tokens.mdx`, `pkce.mdx`, `user-sessions.mdx`, `concepts/index.mdx` | `/auth/token`, `/auth/logout` | Yes |

Enumerated by grepping the docs for `` `/token` ``/`` `/logout` `` against `/auth/token`/`/auth/logout`,
and by reading both catalogs. The first search returns exactly one line, `clients.mdx:191`, and it is
the finding. Nothing else in the repository spells either endpoint without its `/auth` prefix.

**Fixed** in `6c867218`: the line now reads "the `/auth/token`, `/auth/logout` or `/userinfo`
endpoints". Two prefixes added, nothing else touched. The sentence was 45 words before and is 45
after, so no documentation size budget moved.

**No mutation was run for this fix, and none can be.** The change is prose. There is no assertion to
falsify and no test reads that sentence; the demonstration that it was wrong is the router and the
CORS switch quoted above, and the demonstration that it is right is the same two files.

**The site build does not run on this branch.** `.github/workflows/docs.yml` triggers on push to
`main` with `paths: ['site/**']`, on a published release and on manual dispatch, so no check-suite
job builds the Astro site for a pull request. Stated rather than implying CI covered it. The edit
introduces no MDX construct the line did not already carry: both new spellings sit inside inline-code
spans that were already there.

### Tiers after the fix

| Tier | Engines | Result |
|---|---|---|
| unit | n/a | pass, 3941 cases, 0 failures, all three modules |
| data | mysql, postgres, mssql, sqlite | pass, 1353 cases, 0 failures, 4/4 engines, ~60s |
| integration | not re-run | two words of prose in one `.mdx` file; no handler, route, schema or Go source moved |

### Round outcome, and why there is no round 3

Productive: one real finding, verified, fixed. But the fix moved documentation only, and the
`production` tree hash is `487881215cd6` before and after, exactly as at `d97c5e6a`.

Rounds 1 and 2 at the final gate are funded whatever they changed; past that a round is funded only
if the previous one moved production code. This one did not. The rule exists for precisely this
shape: issue 240 spent all four of its final rounds on documentation, each finding a true minor edge
in the previous round's rewrite, while the production tree stayed byte-identical from round 1 on, and
the code was read once while the fourth repair shipped unread. Two rounds have now read this
branch's code, the second independently reproducing and mutation-testing the first's fix, and the
remaining delta was two path prefixes in a sentence.

Issue #197, opened 2026-08-10
Security: an implicit ceremony receives tokens for a session that was already ended


leodip:
#129 added a gate at `/auth/issue` refusing to mint a grant when the session backing the ceremony has been ended, and #133 moved it above the `response_type` dispatch so it covers the implicit branches too. Two cases still get through it, both ending with an implicit ceremony holding tokens for a session that is gone.

**1. The ceremony arrives with no session identifier.** The gate exempts `isImplicitFlow && sessionIdentifier == ""` deliberately, and that exemption covers the terminated-session case: `MiddlewareSessionIdentifier` only publishes an identifier when the row exists, so an ended session leaves an empty one. The issuer omits the `sid` claim for an empty identifier (`if len(input.SessionIdentifier) > 0` at both the access-token and ID-token claim sites), and `RequireValidSession` treats a user token with no `sid` as an offline or ROPC grant, validating it against the token's generation claim instead of a session. A termination does not change that claim, so the access token and ID token stay valid until they expire. Implicit issues no refresh token, so the exposure ends there.

**2. The session is ended after the gate's lookup.** `HandleIssueGet` performs exactly one `GetUserSessionBySessionIdentifier`; `handleImplicitFlow` receives only the identifier string and never loads the row again, so nothing between the gate and `GenerateTokenResponseForImplicit` re-asks. A termination committing in that window leaves the response carrying freshly signed tokens with the ended `sid`. The code flow closes the identical window with `RevokeCodeIfSessionGone` right after `CreateAuthCode` (#129 decision 12), which works because a code is a row that can be marked after issuance. A signed token cannot be. `RequireValidSession` does reject these on Goiabada's own endpoints, since the `sid` no longer resolves, so what is left is a third-party resource server validating the signature against `/certs`: it cannot consult the session table and accepts the token until expiry.

Fix case 1 by applying the same liveness rule to the implicit branch when the identifier is empty, with #129 decision 16's answer for `prompt=none` (`login_required`, which `redirToClientWithError` already returns in a fragment for implicit response types) and a level 1 restart otherwise. Decide first whether an implicit ceremony must have a session at all: the gate as written would refuse every implicit request reaching `/auth/issue` without one, which is a wider question than the terminated case and the reason #133 left it alone. The unit cases live in `TestHandleIssueGet_ForeignAmbientSession`, and every existing implicit subtest in `handler_auth_issue_test.go` builds its request without an identifier, so those are the ones that would move.

[...]

PR #265, opened 2026-08-26
fix(authserver,core): /auth/issue re-establishes the facts /auth/completed established

### LINKED ISSUE #241: Security: the consent screen is a check window, and /auth/issue re-establishes nothing /auth/completed established
Found by the #239 authorization ceremony review (findings [F5](https://github.com/leodip/goiabada/issues/239#user-content-f5), [F33](https://github.com/leodip/goiabada/issues/239#user-content-f33) and [F37](https://github.com/leodip/goiabada/issues/239#user-content-f37)).

Three defects that look separate in a bug list and are one defect in the code. **`/auth/completed` establishes several facts, the consent screen sits between with no bound on how long, and `/auth/issue` re-establishes none of them.**

Everything below is a time-of-check-to-time-of-use gap rather than a conformance failure, and the issue should be read that way. RFC 6749 3.1.2.3 and OIDC Core 3.1.2.2 require validation when the authorization request is processed, and that happens. Neither specification prescribes re-validating at issuance. What is defeated is the operator's expectation that a revocation takes effect, which is the entire reason those controls exist.

## 1. Session validity is not rechecked ([F5](https://github.com/leodip/goiabada/issues/239#user-content-f5))

`HandleIssueGet` asks whether the session row resolves and whose it is, and leaves idle timeout and `max_age` to `/auth/completed` on the grounds that it has already applied them.

Cost of not doing it: a code minted against an idle-expired session, an access token that works, and a refresh token that fails on first use with `invalid_grant`, which is a confusing failure to debug from the relying party's side.

**Fix:** recheck `userSessionManager.HasValidUserSession(r.Context(), ambientSession, authContext.ParseRequestedMaxAge())` at `/auth/issue` and take the existing `!mayBind` branch when it fails. That branch already does the right thing for both the interactive case (restart at level 1) and the silent case (`login_required`), so this is a widened predicate rather than a new path.

## 2. Scope permissions are filtered once and never rechecked ([F33](https://github.com/leodip/goiabada/issues/239#user-content-f33))

`HandleAuthCompletedGet` calls `permissionChecker.FilterOutScopesWhereUserIsNotAuthorized(authContext.Scope, user)` and writes the result back with `authContext.SetScope(effectiveScope)`. That is the only live permission check the ceremony performs. Everything after it trusts the stored list:

- `HandleConsentGet` builds its checkboxes from `authContext.Scope`.
- `HandleConsentPost` builds `scopeInfoArr` from the same field, saves the ticked subset as the `user_consents` row, and sets `ConsentedScope`.
- `CreateAuthCode` writes `ConsentedScope` (or `Scope`) onto the code.
- `handleImplicitFlow` signs it directly.

If an administrator removes a direct permission, removes a group permission, or removes the user from a group while the consent screen is open, the ceremony issues a grant naming a permission the user no longer holds.

**Nothing downstream catches it.** Three candidates, all miss:

- **The first code exchange does not recheck.** `ValidateTokenRequest`'s `authorization_code` arm checks the client's flow, the code hash, the redirect URI, the client match, `User.Enabled`, the generation, the 60 second expiry, PKCE, the client secret, revocation and session ownership. It never consults `permissionChecker`.
- **The generation boundary does not cover it.** The only callers of `RevokeUserAuthStateTx` are the password reset handler, the account password handler and the admin user handler (password set and account disable). No permission or group management handler advances the generation, and it should not: #106's whole point is that revoking authentication state is a credential-level event, not an authorisation-level one.
- **Only a later refresh checks.** `ValidateTokenRequest`'s refresh arm calls `permissionChecker.UserHasScopePermission(user.Id, inputScopeStr)` for every non-OIDC, non-`offline_access`, non-injected scope, and refuses `invalid_grant` when the user no longer holds it.

So the asymmetry is exact and it is backwards: **a refresh of an old grant is checked against live permissions, while the very first token minted from a brand-new grant is not.** The access token is usable for its full lifetime with a permission that was removed before the token existed.

**Fix:** re-filter immediately before issuance, in `HandleIssueGet` for the code path so `CreateAuthCode` stores a scope that was live when the code was minted, and in `handleImplicitFlow` before signing. An empty result after re-filtering takes the existing `access_denied` "not authorized to access any of the requested scopes" answer that `/auth/completed` already uses, so there is a house form.

Additionally re-filter in `HandleConsentPost` before writing the `user_consents` row, so a consent record cannot durably grant a permission removed while the screen was open. That is a separate benefit and **not a substitute**: consent and entitlement are different questions, and only the issuance-time check answers the one that matters here.

## 3. Removing a redirect URI does not stop a ceremony already using it ([F37](https://github.com/leodip/goiabada/issues/239#user-content-f37))

The redirect URI is matched against the client's current registrations exactly once, in `ValidateClientAndRedirectURI` at `/auth/authorize`. From that point the literal string is a snapshot carried on the auth context, and nothing revisits the registration:

- `CodeIssuer.CreateAuthCode` writes `input.RedirectURI` onto the code. It loads the client, but only for `client.Id`, and never calls `ClientLoadRedirectURIs`.
- `handleImplicitFlow` hands the stored URI to `issueImplicitTokens`, which delivers access and ID tokens to it directly.
- `checkRedirectURIEmittable`, the last-resort gate in front of every emission, asks only `urlutil.IsAbsoluteRedirectURI`. Its own comment is explicit that it exists to assert that every emitter sits downstream of a gate, not to re-run one.
- At redemption, `ValidateTokenRequest` compares the submitted `redirect_uri` against `codeEntity.RedirectURI`, that is, against the code, never against the client's current list.

So an administrator who removes a callback because it has been compromised, or because the host has been decommissioned or lost to a subdomain takeover, does not stop a ceremony on the consent screen from delivering a code to it, does not stop an implicit ceremony from delivering the tokens themselves, and does not stop an already-issued code from being redeemed within its 60 seconds.

**Fix:** reload the client's redirect URIs at `/auth/issue` and re-run the original match (`==`, plus `urlutil.RedirectURIMatches` under the same single-`code` response-type gate that `ValidateClientAndRedirectURI` applies, so loopback port flexibility behaves identically). On failure, **render locally rather than redirecting**: the whole point is that this URI must not receive anything, so `renderRedirectBlocked` or the refusal page is the answer and an error redirect is not.

Re-checking at redemption as well would cover removal in the window after issuance, at the cost of one query on the token path.

## Shape of the fix

One revalidation point at the top of `HandleIssueGet`, above the `response_type` dispatch, covering all three. Failing any check answers with the error that check implies rather than issuing.

Doing them together is the point of filing them together: three scattered checks bolted on separately will drift, and the next fact that needs rechecking will be added in a fourth place.

## Tests

- Integration: terminate the session while the consent screen is open, then approve. No code is issued.
- Integration: revoke a permission while the consent screen is open, then approve. The issued token does not carry the scope.
- Integration: delete the redirect URI while the consent screen is open, then approve. Nothing is delivered to it and the refusal renders locally.
- Unit: the `!mayBind` branch is reached from the widened predicate.

## Related

- #197 is this same gap on the implicit branch, for session validity specifically. **Sequence it after this issue** so its fix can reuse the revalidation helper built here.
- Combines badly with #240's public-client sibling: a public client without PKCE plus a still-live removed callback means the attacker who receives the code also needs nothing to redeem it.
- #106 established the generation boundary, which deliberately does not cover permission changes. That is correct and is why the check has to live in the ceremony.

leodip:
Closes #241.

`/auth/completed` establishes three facts about an authorization ceremony and then hands the
browser to the consent screen, which has no bound on how long it sits there. `/auth/issue`
re-establishes none of them before it mints a code or signs an implicit token:

- **Session validity.** `/auth/completed` applies idle timeout and max lifetime. `/auth/issue`
  reads the session row again and asks only whether it resolves and whose it is, so a session that
  idled out while the consent screen was open still yields a code. The access token then works and
  the first refresh fails with `invalid_grant`, which is a confusing thing to debug from the
  client's side.
- **Scope permissions.** The permission filter runs once, before consent. Everything after it reads
  the stored scope, so a permission revoked while the screen was open is still granted, and nothing
  at the token endpoint catches it: the `authorization_code` arm never consults the permission
  checker. A refresh of an *old* grant is validated against live permissions while the first token
  minted from a *brand-new* grant is not.
- **Redirect URI registration.** Matched once, at `/auth/authorize`. The matched value is copied
  onto the code and never rematched, so a callback deleted mid-ceremony still receives the
  response, and an already-minted code stays redeemable for its full 60 seconds afterwards.

None of the three is a conformance failure. RFC 6749, OIDC Core and RFC 9700 all place redirect URI
matching and parameter validation at the authorization *request*, and no source requires
re-validating at issuance or says a grant must be revoked when a client's registration changes.
What they are is time-of-check-to-time-of-use gaps against the operator's expectation that a
revocation takes effect.

## The rule this establishes

`/auth/issue` re-establishes all three facts immediately before it mints anything, above the
`response_type` dispatch so the implicit branch is covered by the same gate. Failing any one
answers with the error that check implies rather than issuing.

The property that makes it safe: **the gate can only ever refuse.** Nothing below a failing check
is reached, no row is written, no route is added, and each refusal either renders locally or takes
a refusal path that already exists. A ceremony where all three checks pass reaches the issuers with
exactly the values it reaches them with today.

It reaches further than issuance for one of the three. A sign-in already in progress also answers
its client from six other places, every one of them an error response and every one of them a
redirect: the user clicks Cancel on the consent screen, or approves with nothing ticked, or the
account is disabled mid sign-in, or the sign-in turns out to hold none of the requested permissions,
or a bad authorization request is delivered after the password is verified. All six navigate the
browser to the address stored at the start, and the shared function they pass through asks who
registered the client and whether the stored string is an absolute URI, but never re-reads the
registered list. So the registration check goes there too, as a third gate beside the two that
function already applies. **Removing a callback then stops it receiving anything from a sign-in
already in progress: a code, a token and an error redirect alike.** A withheld redirect renders the
page this server already shows when it refuses to forward a browser somewhere, which names the
destination and stops.

Two things ride with it. The consent submission filters the ticked selection against live
permissions before saving the `user_consents` row, so a saved consent never names a permission the
user did not hold when they ticked it. And code redemption re-checks the code's redirect URI
against the client's current registrations, so removing a callback has no 60 second tail.

`max_age` is deliberately **not** part of the session recheck. It bounds the age of the
authentication, not the dwell time on a consent screen, and it is measured from the session's start,
so re-applying it would make every `max_age=0` ceremony fail, restart at level 1, mint a fresh
session and fail again.

### Stage 1: the shared redirect URI matcher

`519a9e72`. The registration match, the one rule that decides whether a browser may be sent
somewhere, lived in exactly one place: `ValidateClientAndRedirectURI`. Three more places are about
to ask it. Rather than copy the loop, it moved to
`urlutil.RedirectURIIsRegistered(registered, requested, allowLoopbackPortFlexibility)`, beside the
`RedirectURIMatches` pair matcher it delegates to, and the validator became its first caller.

Exact string equality per entry runs whether or not the flag is set, so a client that registered a
URI `url.Parse` rejects can still redeem that exact string. Only beyond exact equality does the flag
admit RFC 8252 loopback port flexibility, and the flag is a caller's parameter rather than something
`urlutil` infers, because that flexibility is scoped to the authorization code flow and `urlutil`
deliberately knows nothing about OAuth2 flows (#41). The validator kept its own
`len(responseTypes) == 1 && responseTypes[0] == "code"` gate and passes the result in.

**Behaviour is identical. No check is added here and none is removed.** The rest of the branch is
what adds checks; this stage only makes one available in more than one place.

Tests: `TestRedirectURIIsRegistered`, 15 table rows. Deliberately thin over the pair matcher's own
behaviour, one accept and one reject through it, since the exhaustive table already lives directly
above. What it owns is the list: entry ordering, an empty list, and that flexibility does not leak
from a loopback entry to a non-loopback sibling. Three mutations, all caught, including one that
ignores the flag entirely, which is the mutation that survives if anyone moves the flow decision
inside `urlutil`.

Tiers: unit 1174 pass, 0 fail. Integration on sqlite 988 pass, 0 fail, run as a regression check
because this rewrites a function on the path of every authorization request. Data not applicable:
no query, no schema, no migration.

### Stage 2: the gate at /auth/issue

`12d6ba10`. `HandleIssueGet` now re-establishes all three facts, above the `response_type`
dispatch so the implicit branch is covered by the same gate. Every branch it adds can only refuse:
nothing below a failing check is reached, no row is written and no route is added, so a ceremony
that passes reaches the issuers with exactly the values it reached them with before.

**The redirect URI check goes first, and that position is the design.** Every refusal below it
answers the client by redirect, so delivering one to a callback the operator has just pulled would
navigate the browser to that host on a request the server refused, which is RFC 9700 section
4.11.2's harm arriving through this server's own error response. Placed first, the handler has one
property worth stating: nothing in it can send a browser to an unregistered callback. Its refusal
renders `auth_redirect_blocked.html` locally at 200 and is never redirected, because that URI must
receive nothing, an error included. It is also the one refusal in the file that renders anyway when
clearing the auth context fails: the usual fallback is to answer the client with `server_error`,
and answering this client is precisely what the gate exists to prevent. A client deleted
mid-ceremony refuses through the same page rather than answering a 500, since a client with no
registrations has answered the question rather than errored.

**The session check** widens `mayBind` with `HasValidUserSession`, conjoined above the implicit
exemption because that call answers false for a nil session and folding it in afterwards would
refuse every implicit ceremony arriving without one. It passes `nil` for the requested max age
deliberately: `max_age` bounds the age of the authentication, not the dwell time on a consent
screen, and `UserSession.IsValid` measures it from `Started`, so re-applying it would refuse every
`max_age=0` ceremony a microsecond after it began. The refusal now distinguishes foreign, expired
and gone in its log line, and only the expired shape writes an audit row: the other two are #133's
and #129's refusals, which write none today.

**The scope re-filter** reads whichever of `ConsentedScope` or `Scope` the issuer will read,
through a pointer so the field read and the field written cannot drift apart, and writes back only
on a non-empty result. An emptied `ConsentedScope` is refused rather than written back, because
both issuers fall back to `Scope` when it is empty, so writing it would fall through to the full
unfiltered request. A narrowed set is issued rather than refused, which is RFC 6749 section 3.3 and
what `/auth/completed` already does with the same removal seconds earlier; the client is told
through the response's `scope`.

Three audit constants join `constants.go`, one per check, so an operator can ask whether a removal
stopped anything through the admin console and `GET /api/v1/admin/audit-logs`. `auditEvent` is a
free-form string in `openapi.yaml`, so the API contract does not change.

`handleImplicitFlow` now takes the client and the user rather than loading them: the gate resolves
both above the dispatch and refuses a nil for each, so re-reading them was two queries for values
already in hand. The two `clientProvenance` calls in this handler went with them.

Docs: one sentence in `concepts/user-sessions.mdx`, extended so the last-step check covers a
session that timed out as well as one somebody ended. 61 words to 72, still one sentence. Nothing
else on any page was falsified.

Tests: three unit tables covering which refusal fires and what it does, plus the ordering case that
pins the registration check above the `id_token_hint` refusal, and three integration cases that
park a browser on the consent screen and mutate one thing before the last hop. Seven mutations, all
caught; the two that matter most are loopback port flexibility granted to an implicit response, and
the refusal redirecting instead of rendering, which the two existing gates would both have let
through.

Tiers: unit 1178 pass, 0 fail. Integration on sqlite 991 pass, 0 fail. Data not applicable: no new
`Database` method, query, migration or schema change.

### Stage 3: the gate inside the emitter

`d4a97e68`. The six error responses a ceremony in progress can send stop being an exception.

`redirectWillBeEmitted` is the one definition of "would an error response to this client actually
leave the server", asked by the emitter and by the authorization endpoint's decision about whether
to authenticate before redirecting. It gains a third gate below the provenance test (#108) and the
absoluteness test (#122): it reads the client's current registrations with
`GetRedirectURIsByClientId` and asks `urlutil.RedirectURIIsRegistered`. Seven copies of a security
gate is seven chances for one of them to be forgotten, which is why it lives in the predicate all
seven sites share rather than at each of them.

- **The loopback flag is false**, deliberately. That arm exists so a native app's ephemeral port can
  match a registered portless URI on a code request, and an error redirect carries no code, so this
  gate is stricter than the one the stored value passed at `/auth/authorize`. It can therefore admit
  nothing that was refused there.
- **A failed load answers no**, matching `clientProvenance` directly above it: every caller is
  already on its way to an error response, so a lookup that fails must not turn a working refusal
  into a 500, and an unresolved registration is the untrusted case.
- **The refusal renders `auth_redirect_blocked.html`**, the page this emitter already withholds a
  redirect through, so one condition keeps one page wherever it fires.
- **The database is threaded, not injected.** `redirToClientWithError` and `answerClientWithError`
  take one, and their nineteen call sites pass the database their handler already holds. No exported
  signature moved, `routes.go` did not change, no interface changed and no mock was regenerated.

Two comments became false and correcting them is part of the change rather than tidying. The
authorization endpoint justified evaluating two clauses before a third on the predicate having no
side effects; it now issues a query, so the argument is restated on what it actually rests on, which
is that the query is a read and the short-circuit only ever skips it. And the refusal page's file
comment said it replaces a redirect "to a client that registered itself", which is now one of three
reasons rather than the reason.

Tiers: unit and data locally (data is not applicable by the agreement's standing reason and was run
anyway, since it costs half a minute), integration on sqlite. Five mutations, all caught, including
the two that matter: the emittability gate never refusing, which must make the unemittable-URI row
reach a mock that has no expectation for the registration read, and the registration gate never
refusing, which must make the browser Cancel case emit a `Location` again.

Worth knowing: the test fanout was 67 sites across five files, not the 25 direct callers of the
emitter the plan named. What needs the new fixture is every test whose handler answers a client by
redirect. A shared `stubRegisteredRedirectURI` supplies it, in the shape `stubAuthenticatedBrowser`
was written in for #213's session requirement; the table-driven tests whose rows split between
answering and not answering stub inside the answering branch rather than being made permissive.

### Stage 4: the consent record filter

`509c1c3e`. The stored consent record stops naming permissions the user has already lost.

`HandleConsentPost` weighed the ticked selection against nothing, so a user parked on a consent
screen whose permission was revoked while they read it could store a `user_consents` row naming a
scope they no longer held. The handler gains a `PermissionChecker` and filters the selection against
the permissions the user holds now, above the consent row load so a selection the filter empties
refuses without reading or writing anything.

The row grants nothing on its own: every reader pairs it with a live permission check, and stage 2's
gate already narrows the token. What the row does is **suppress the consent screen on later
ceremonies**, so an unfiltered one lets a permission granted back months later ride in on a tick the
user made at a moment when it was gone, and the user's own consents page lists a scope the
application can never use.

- **The empty case now has two shapes and two answers.** A user who ticked nothing still gets "The
  user did not provide consent", which is a choice they made. A selection an administrator emptied
  gets `/auth/completed`'s "The user is not authorized to access any of the requested scopes", which
  is not. Both are `access_denied`, and both descriptions are pinned by tests, so neither case can
  be edited into the other without failing.
- **The new refusal needed nothing of its own to be safe.** It redirects, which makes it the seventh
  member of the family stage 3's registration gate covers. That is the whole reason stage 3 was
  ordered first, and it held.

Tiers: unit and data locally (data is not applicable by the agreement's standing reason and was run
anyway), integration on sqlite. Four mutations, all caught, including the one that restores exactly
the line this stage replaced, and one against the browser case so it is not vacuously green.

The browser case is deliberately the mirror of stage 2's: the revocation lands **before** the consent
submission rather than after it. After it, the user still held the permission when they ticked the
box and the record is a true account of what they agreed to; before it, the checkbox is on a page the
revocation cannot reach and this filter is the only thing between the tick and a stored grant.

### Stage 5: the redemption re-check

`ab399662`. Removing a callback now reaches a code that is already outstanding.

`ValidateTokenRequest`'s `authorization_code` arm compares the submitted `redirect_uri` against the
one stored on the code, and that stored value is a copy taken at minting which nothing ever
rematched against the client. So a code delivered one second before an administrator removed a
callback stayed redeemable for the rest of its sixty-second life, and the tokens it yielded outlived
that by their whole lifetime. The arm now loads the client's current registrations and asks stage
1's `urlutil.RedirectURIIsRegistered` before it returns.

- **The flexibility flag is unconditionally `true` here, and `false` in stage 3's emitter gate.**
  Not an inconsistency: a code exists only for `response_type=code`, and a native app's code stores
  an ephemeral loopback port that its registered portless form would never exact-match (RFC 8252).
  It can therefore only be more permissive than the gate that same stored value already passed at
  `/auth/authorize`, so it admits nothing that was refused there. An error redirect carries no code,
  which is why the emitter's answer is the opposite.
- **Placed at the very end of the arm**, below the revoked and ownership checks. Late, so an
  unauthenticated presenter of a stolen code learns nothing about the grant's state (#137). *Last*
  rather than merely late, because this is the one refusal in the group that says what happened: a
  revoked or cross-bound code keeps the flat "Code is invalid." it has today, and would not if this
  ran first.
- **The refusal is legible on purpose.** A new sentinel beside `ErrUserDisabled`, matched by value
  so the audit row and the wire message cannot drift apart. Reusing the generic "Invalid
  redirect_uri." would have been wrong for the person who most needs to read it: that message also
  means "you sent one that differs from the code's", so an administrator who had just rotated a
  callback could not tell the two apart. A fourth audit event, `redemption_refused_redirect_uri`,
  makes the refusal answerable from the admin console rather than only from a server log file.
- **A failed registration load propagates as itself.** An unreachable database says nothing about
  whether the URI is registered, so turning the failure into a refusal would report an outage as an
  administrative action.

Tiers: unit and data locally (data is not applicable by the agreement's standing reason and was run
anyway), integration on sqlite, run here as a regression check because every existing
authorization-code exchange now passes through the new read. Four mutations, all caught, including
one that flips the flexibility flag, which is the row that fails if somebody makes it conditional to
match stage 3's.

The handler's audit hop had **no seam in the plan** and gained one rather than shipping untested. It
asserts both halves: the sentinel produces the event, and an ordinary `invalid_grant` produces none.
That second row is the one that fails if anybody keys the predicate on the error code, which 22
unrelated failures also carry.

## For the release notes

Four things an operator will observe, all of them behaviour changes rather than configuration. There
is **no migration, no schema change, no configuration setting, no new route and no API contract
change**: the new audit event names are free-form strings in `openapi.yaml`, which does not move.

**1. Removing a redirect URI now reaches sign-ins that are already running.** It used to apply only
to ceremonies that started after it. A browser parked on the consent screen still received its code,
an already-minted code stayed redeemable for the rest of its sixty seconds, and every error response
the ceremony sends still navigated to the removed address. Now the ceremony refuses: where a redirect
would have gone to the removed callback the server renders its redirect-blocked page, which names the
destination and stops, rather than navigating there, and that holds for errors as well as successes.
At the token endpoint a code whose redirect URI is no longer registered is refused with
`invalid_grant` and a message that says specifically that, so it cannot be mistaken for the
"you sent one that differs from the code's" case. *What to do about it:* nothing, if the removal was
intended. If a callback was removed by mistake, re-register it and users retry the sign-in.

**2. A session that expires while the consent screen is open now refuses at the last step.** It used
to issue the code anyway: the access token worked and the *first refresh* failed with
`invalid_grant`, which is a confusing thing to debug from the client's side. Now the user is asked to
sign in again, at the point where the problem actually is. `max_age` is deliberately not applied
here: it bounds the age of the authentication rather than the dwell time on a consent screen.

**3. A permission revoked mid sign-in no longer rides through.** It is removed from the scope of the
token being minted, and from the `user_consents` row being saved. If every requested scope has been
revoked the ceremony refuses with `access_denied` and a description distinct from the one a user gets
for ticking nothing. Previously the token carried the revoked scope and the saved consent record
named it indefinitely, which suppressed the consent screen on later sign-ins.

**4. Four new audit events.** Anything consuming the audit log by event name should expect
`issuance_refused_session_invalid`, `issuance_refused_scope_denied`,
`issuance_refused_redirect_uri` and `redemption_refused_redirect_uri`. They exist so that a
refusal caused by an administrative change is answerable from the admin console and
`GET /api/v1/admin/audit-logs` rather than only from a server log file. Note that only the expired
session writes a row among the session refusals: a foreign or missing session is #133's and #129's
refusal, and neither writes one today.

## Tiers

27 files, +3824 -220. What ran locally and what only the check suite can say:

| Tier | Where | Engines | Result |
|---|---|---|---|
| unit | local, at every stage and every review round | n/a | **4016 passed, 0 failed** at `a52f7731` |
| integration | local, at every stage | **sqlite only** | **1314 passed, 0 failed** at `eef795c5`, and still binding at `a52f7731`: the `integration-tests` tree hash did not move, so nothing that tier executes changed |
| data | local, at stages 3, 4 and 5 | all four | not applicable by the agreement's standing reason, no new `Database` method, query, migration or schema change, and run anyway because it costs half a minute |
| data and integration | **the check suite**, four parallel jobs | sqlite, mysql, postgres, mssql | green |
| build, lint, vulnerabilities, versions | **the check suite** | n/a | green |

**The four-engine result is the check suite's, and it has concluded.** Green on all eleven jobs at
`a52f7731`, [run 33019858384](https://github.com/leodip/goiabada/actions/runs/33019858384), and green
on all eleven at `eef795c5` before it. No local run on this branch ever covered mysql, postgres or
mssql for the integration tier, and nothing here claims otherwise.

**Mutation testing, which is the evidence the new tests are worth their green.** 23 mutations across
the five stages, every one caught. 6 more in review round 1, including the loopback flag mutated in
*both* directions, since an unconditional `true` would otherwise have passed. 13 in round 2, nine of
them enumerating one family, of which four survived. 8 in round 3, all eight of which survived. Every
survivor is closed by a named test case. Each mutation ran through `where.sh mutate`, which restores
from a trap and compares by hash, and the production tree hash was checked before each commit so no
mutation residue could be committed.

`check-anchors.sh`: 41 of 41 rows resolve.

## Deviations worth knowing

- **Stage 3 was not in the sealed plan.** It exists because the plan review found the six error
  responses and you answered **B**. That is the one place this change's scope moved after the
  agreement was sealed, and it is set out in full in the decisions comment below.
- **Stage 3's test fanout was 67 sites across five files, not the 25 the plan named.** The plan
  counted direct callers of the emitter; what actually needed the new fixture was every test whose
  handler answers a client by redirect.
- **`handleImplicitFlow` takes the client and the user rather than loading them.** The gate resolves
  both above the `response_type` dispatch and refuses a nil for each, so re-reading them was two
  queries for values already in hand.
- **Stage 4's filter went above the consent row load, not beside the scope join.** Section 4 allowed
  either; the higher placement was chosen so a selection the filter empties refuses without reading
  or writing anything.
- **Stage 5 gained a test seam the plan did not have.** The handler's audit hop had no seam in
  section 5, so its production code would have shipped untested. It asserts both halves: the sentinel
  produces the event, and an ordinary `invalid_grant` produces none.
- **Section 0's anchors were re-swept four times** as stages moved the code they cite, once for the
  emitter's gate call in two different rounds. That is anchors following code, not citations that
  were wrong when written.
- **`plan.md` was damaged during stage 3's expansion** by a splice script that read a stale temp file
  left by an unrelated run. Caught immediately on reading the file back and rewritten. No committed
  file was affected and no other stage was touched.

## Review

The code was not reviewed stage by stage: this run used `LEO_REVIEW_MODE=end`, so the whole branch
got one reading at the end. Reviewed by `gpt-5.6-sol` at effort `xhigh`, three rounds over
`main...HEAD`. The gate is closed: round 3 changed no production code, so the budget funds no
fourth round.

|---|---|---|---|
| 1 | 4 | 1 | all fixed, `0d637b97` |
| 2 | 3 | 0 | all fixed, `eef795c5` |
| 3 | 1 | 0 | fixed, `a52f7731` |

Nothing was contested, nothing was deferred, and no round left an `unchecked` entry or filed a
follow-up.

**Round 1**, all fixed in `0d637b97`:

- **blocking, security:** `redirectWillBeEmitted` is asked twice in one authorization request, and
  once it started reading the registration table those two answers could disagree. A failing first
  read is what routes a logged-out browser to an immediate answer, on the reasoning that no redirect
  leaves the server; a second read that recovered then emitted it anyway, which is RFC 9700 4.11.2
  attack 1 and the exact harm #213 closed. Fixed: the first refusal travels to the emitter as a
  floor, so a later read may narrow a permission to a refusal and never the reverse.
- **significant, conformance:** the same gate refused loopback port flexibility unconditionally, so a
  native app that started a code flow on an ephemeral loopback port received successes but had its
  protocol errors withheld, silently. RFC 8252 section 7.3 makes that port part of what "registered"
  means for a code request and RFC 6749 section 4.1.2.1 then requires the error to be delivered
  there. Fixed: the flag comes from the response type, and implicit responses stay strict.
- **significant, quality:** the three new fail-closed storage-error branches had no test, so widening
  any of them to use the requested authority left the whole tier green. Fixed with three strict
  error-path cases; all three mutations that survived the review are now caught.
- **minor, conformance:** four concept pages named self-registration as the only exception to a
  client receiving error redirects. There are two now. Fixed in the seven sentences that were false.

**Round 2**, all fixed in `eef795c5`:

- **significant, quality:** the refusal floor round 1 introduced was untested through
  `answerClientWithError`'s clear-failure fallback, and so was the response type the loopback gate
  reads, at two composed seams. Both are the same defect, a field of `redirectErrorInput` failing to
  reach the gate, and they were closed as one family rather than as two findings. All nine members
  were enumerated and mutated, and **four survived rather than the three reported**: `state` across
  the fallback survived as well, which would have emitted a `server_error` carrying no `state` echo,
  against RFC 6749 section 4.1.2.1. Two new tests close all four.
- **minor, conformance:** a comment in `token_validator.go` still described the pre-round-1 loopback
  rule. Rewritten; no behaviour changed.

**Round 3**, fixed in `a52f7731`:

- **significant, quality:** the three refusals this change adds each clear the auth context before
  answering, and none observed what happens when that clear fails; two of them can also fail to
  produce the answer itself. Every refusal that predates this issue pins both, in a triple the
  neighbouring blocks already run. The review reported three members; enumerating the house form
  found **eight**, and all eight survived as mutations against the full internal tier. Eight cases
  close them. The deregistration refusal gets a pair rather than a triple, because it renders
  locally instead of answering the client, so its answer has one way to fail rather than two, and a
  failure to render must stay a 500 rather than falling back to a redirect toward the very host the
  gate just refused. No production code changed: the production tree hash is identical before and
  after.

**The full account of every round**, with the reviewer's exact findings, the spec quotes, the
mutation tables and the per-page documentation word counts, is in
[this comment](https://github.com/leodip/goiabada/pull/265#issuecomment-5431924628). It is worth
opening: the verdicts themselves are gitignored and were overwritten by each following round, so that
comment is the only surviving record of what was found and what answered it.

## The rest of the record

Three comments carry what does not belong in a diff:

- [**Decisions**](https://github.com/leodip/goiabada/pull/265#issuecomment-5431919657). Eleven
  settled, none deferred. The one you answered, decision 11, is set out with what B decided and what
  reversing it would cost.
- [**Follow-ups and fold-ins**](https://github.com/leodip/goiabada/pull/265#issuecomment-5431921927).
  None. Nothing was drafted for you to file and nothing was folded in, which is stated explicitly
  because an empty list and an unwritten one look identical from outside. It also notes that #197 is
  now unblocked.
- [**The code review**](https://github.com/leodip/goiabada/pull/265#issuecomment-5431924628). Four
  rounds, one on the plan and three on the whole branch.

## Status

**Complete.** All five stages landed, the whole branch was reviewed over three rounds and the gate is
closed, and the check suite is green on all eleven jobs at `a52f7731`.

The shared matcher exists, `/auth/issue` re-establishes the session, the scope and the redirect URI
before it mints anything, the error-redirect emitter weighs every response a ceremony in progress
sends against the client's live registrations, the stored consent record no longer names permissions
the user has already lost, and a code whose callback was deregistered is no longer redeemable. All
four facts are closed, at every point a ceremony can reach a client: **a sign-in already running says
nothing at all to a callback that client no longer has, not a code, not a token, and not an error
redirect sixty seconds later.**

Ready for review.


leodip:
## Decision 11: when you delete a client's redirect URI, must a sign-in already in progress stop sending the browser there at all, or only stop giving it a code?

**What a redirect URI is.** It is the callback address an application is allowed to send a user
back to when a sign-in finishes. Goiabada matches it exactly against the registered list every time
an application starts a sign-in, which makes it the most security-sensitive setting on a client
after its secret. Deleting one is how an administrator cuts an application off from receiving
users, and the usual reason to do it in a hurry is that the address has been compromised.

**Where you meet it.** Admin console, Clients, pick a client, Redirect URIs. Behind that page is
`PUT /api/v1/admin/clients/{id}/redirect-uris`.

**What happens today.** The deletion takes effect on the *next* sign-in that starts. A sign-in
already underway keeps the address it was handed at the beginning, because nothing looks at the
registered list again. That is exactly what this issue is about closing, and the plan I wrote
closes it for the moment a code or a token is handed over. It does not close it for the moments
where a sign-in ends in an *error* instead, and there are six of those. Every one of them
navigates the browser to the address stored at the start.

The plan review caught this and reproduced it: park an authenticated browser on the consent
screen, delete the registered callback, click Cancel. Goiabada answers
`302 Location: https://removed.example/callback?error=access_denied&state=...`, to the address
deleted moments earlier. I confirmed it by reading the code rather than taking the reproduction on
trust. All six go through one function, `redirToClientWithError`, and its only gate,
`redirectWillBeEmitted` (`src/authserver/internal/handlers/handler_authorize.go:936`), asks two
questions: did the client register itself rather than being created by an administrator (#108), and
is the stored string an absolute URI naming a host (#122). It never loads the registered list. An
administrator-created client whose callback was deleted a minute ago passes both.

| When the sign-in ends in an error | Where |
|---|---|
| the user clicks Cancel on the consent screen | `handler_consent.go:380` |
| the user approves but ticks nothing | `handler_consent.go:274` |
| a bad authorization request, answered after the password is verified | `handler_auth_level1.go:102` |
| the account was disabled mid sign-in | `handler_auth_completed.go:334` |
| the user turns out to hold none of the requested permissions | `handler_auth_completed.go:376` |
| *(new, from this change)* the permission filter empties the ticked selection | stage 3 |

**Who is stuck.** The administrator who just deleted a compromised callback. Whoever controls that
address still receives browser navigation from every sign-in that was already open, along with the
`state` value the application sent. They do not get a code and they do not get a token: the sealed
plan already blocks those, and this question does not touch that. What they get is traffic and a
protocol response from your server, on a request you refused. None of this is new, incidentally,
it is all true on `main` today and no code has landed on this branch yet.

### The worked example

Say you delete `https://app.example/cb` at 10:00 because the host was taken over. Two people have
that application's sign-in open in a tab, one sitting on the consent screen and one on the password
page. At 10:05 the first clicks Cancel.

- **Under A**, their browser is navigated to `https://app.example/cb?error=access_denied&state=...`,
  the host you deleted five minutes ago.
- **Under B**, they get a local page reading "You have not been sent anywhere. This application
  asked to send you to https://app.example/cb. The request stops here", and nothing leaves.

Neither one ever receives a code. Both block that.

### The choice

The one thing it turns on: **which responses have to pass a fresh registration check.**

| | **A. Only the ones the final step sends** (what is sealed) | **B. Every response a sign-in in progress sends** |
|---|---|---|
| Deleted callback can still receive a code or token | no | no |
| Deleted callback can still receive an error redirect | yes, from the six above plus the new one | no |
| Where the check lives | the final issuance step only | the final issuance step, plus the one shared function every error response already passes through |
| Work left in this run | none, the plan stands | roughly half a stage: one function gains a database parameter, its two callers change, six unit cases and one browser test |
| Extra database read | on issuance | on issuance, and on each error response, which is not a hot path |
| Reopens a sealed decision | no | yes, one: which page a refusal shows |

**What does not differ.** The other two halves of this change, the session re-check and the
permission re-check, are untouched either way, as are the shared matcher and the code-redemption
check. The one-sentence documentation change is the same. Both are equally easy to reverse later:
it is one gate in one function, not a schema or a wire format.

### What I would do, and why I am asking rather than doing it

**B.** The reason is that A ships a promise it does not keep. The whole change exists so that
removing a callback reaches a sign-in already running, and the release note for A would have to
read "removing a callback stops it receiving a code, but it can still receive an error redirect
from six places" — which is a strange thing to ask an administrator to hold in their head, and the
harm the guard is aimed at (RFC 9700 4.11.2, navigating a browser to a destination someone else
controls on a request the server refused) is the same harm at both. This codebase already built
that guard twice, in #108 and #122, in that same function. This would be the same reasoning applied
to a third input, in the same place, and the "You have not been sent anywhere" page it refuses
through already exists and is already what those two gates use.

I am not just doing it because it widens what this issue agreed to build, and because it reverses
part of a decision we settled during the interview: we chose the ordinary error page for this
refusal specifically so that one condition would not show two different pages, and putting the gate
in the shared function means it refuses through the other one. Under B I would take the withheld
page for all of them, one condition and one page, unless you say otherwise. That part is cheap to
change afterwards.

And the honest case for A: none of this is a regression, it is all reachable on `main` today, so
"not this issue's job, file it" is a real answer and I will file it as a well-evidenced issue with
the reproduction if that is what you want.

Nothing is committed. The branch is still identical to `main` apart from the empty commit that
opened this pull request, and no stage will start until this is answered.

leodip:
B

leodip:
## Decision 11 applied: every response a sign-in in progress sends now has to pass a fresh registration check

Answer taken as **B**. Nothing is committed against it yet: the branch is still just the empty
commit that opened this pull request. The run is unblocked and stage 1 is next.

### What B decided

Deleting a client's redirect URI has to reach a sign-in that is already underway, not just the next
one that starts. Under the sealed plan that held for the moment a code or a token is handed over.
It now also holds for the moments a sign-in ends in an error instead, which is six places today and
a seventh that this change was going to add: Cancel on the consent screen, approving with nothing
ticked, a bad authorization request delivered after the password is verified, the account disabled
mid sign-in, the user holding none of the requested permissions, and the new refusal when the
permission filter empties a ticked selection.

Every one of those redirects the browser to the address stored when the sign-in began. They all go
through one function, and it asks two questions, who registered the client and whether the stored
string is a real absolute address. It never re-reads the registered list. B adds that as a third
question in the same place, so there is still one function deciding whether an error redirect leaves
this server, and it now has one more reason to say no.

The worked example from the question, resolved: you delete `https://app.example/cb` at 10:00, and
at 10:05 someone with that sign-in open clicks Cancel. They now see a local page naming the
destination and saying the request stops there. Nothing is sent to `app.example`.

### The page a refusal shows, which the answer reopened

The sealed agreement had this refusal render the ordinary authorization error page, chosen so that
one condition would not show two different pages. Putting the gate inside the shared emitter
reverses which page that principle picks: the emitter can only refuse through the "You have not been
sent anywhere" page, so keeping the other one at the final step is what would give the condition two
pages. Both refusals now use the withheld page, which is what the question offered and what the
answer did not say otherwise about. It already exists, it already names the destination, and it is
already in both language catalogs, so no new text is written.

The front door is untouched. An authorization request that *names* an unregistered address is a bad
request and still gets the ordinary error page; this is about a registration withdrawn while a
sign-in is running, which is a different condition.

### What it cost the plan

One new stage, and no new dependency anywhere. The shared function gains a database handle and its
callers pass one down; all five handlers involved already hold one, so no endpoint signature
changes, no route moves, no interface changes and no mocks are regenerated. The work is a third gate,
its six unit cases, one browser test that parks on the consent screen, deletes the callback and
clicks Cancel, and threading an argument through call sites that already have it in scope.

There are now five stages instead of four. The new one is stage 3, and the two that follow it kept
their work and moved up a number:

1. the shared redirect URI matcher
2. the gate at `/auth/issue`
3. **the gate inside the emitter** (new)
4. the consent record filter
5. the redemption re-check

Stage 3 sits before the consent-record filter deliberately. That stage adds the seventh error
redirect, so gating first means no commit on this branch ever carries a refusal that could send a
browser to a deleted callback.

### One thing worth knowing

Nothing on the documentation site is made wrong by this, so nothing there changes. No page claims
that removing a callback reaches a sign-in already in progress, and the page describing redirect URI
rules is about their shape rather than their registration.

What does change is the line an operator would read on a release: **removing a callback now stops it
receiving anything from a sign-in already in progress, a code, a token and an error redirect alike**,
with no caveat attached. That was the reason given for preferring B, and it is now true.

leodip:
## Decisions

The agreement settled **11 decisions, none of them deferred**, so there is no assumption on this
branch waiting to be confirmed and nothing here that a later reader has to go back and check. Ten
were settled during the interview, before the seal. One was raised by the run itself, escalated to
you here, and answered.

This comment exists so that the one decision that moved the change's scope after it was sealed is
readable without the agreement, which is gitignored and does not survive the run.

### Decision 11: how far does the re-check reach? Answered **B**

**What raised it.** The plan review found, with an end-to-end reproduction, that `/auth/issue` is
not the only place a sign-in in progress answers its client. Six other places do, every one of them
an error response and every one of them a redirect: the user clicks Cancel on the consent screen, or
approves with nothing ticked, or the account is disabled mid sign-in, or the sign-in holds none of
the requested permissions, or a bad authorization request is delivered after the password is
verified. All six navigate the browser to the address stored at the start of the ceremony, and all
six pass through one predicate, `redirectWillBeEmitted`, which weighs who registered the client
(#108) and whether the stored string is an absolute URI (#122) but never re-reads the registered
list. So a client whose callback an administrator deleted a minute earlier passes both gates and
still receives the error redirect. That was confirmed by reading the call graph, not by trusting the
reproduction, and **all six predate this branch: they behave that way on `main` today.**

**Why it could not be resolved inside the run.** Closing it meant a signature change on a pure
predicate, a security property, and widening the sealed goal from `/auth/issue` to every response a
ceremony in progress sends. The goal is frozen at the seal, so widening it is yours to authorise.

**The options put to you, and your answer.**

- **A:** only the responses the final step sends, which was the sealed scope.
- **B:** every response a sign-in in progress sends.

You answered **B**, at 2026-08-26T12:03:25Z on this thread, with no instruction beyond the choice,
so it was applied as written.

**What B decided.** The run grew a stage it was not sealed with, stage 3, `d4a97e68`, which puts the
registration check inside `redirectWillBeEmitted` where all seven sites share it rather than at each
of them. It was ordered before the consent filter, because that stage adds the seventh site, so no
commit on this branch ever carries a ceremony refusal that could navigate a browser to a
deregistered callback. It also amended decision 4, which your answer explicitly offered: the refusal
renders `auth_redirect_blocked.html` at both sites rather than the generic error page, because a gate
inside an emitter can only refuse through that interstitial, and keeping the other page at
`/auth/issue` would have given one condition two pages.

The operator-facing sentence B bought is in the PR body under **For the release notes**: removing a
callback stops it receiving anything from a sign-in already in progress, a code, a token and an
error redirect alike. Under A that sentence would have needed a caveat naming six exceptions.

**What reversing costs.** Reverting to A means reverting stage 3, one commit, `d4a97e68`, plus the
test fixture it threaded through 67 sites in five test files. Stages 1, 2, 4 and 5 stand without it,
but stage 4's new refusal then joins the six that can still reach a deregistered callback, so the
guarantee above loses its "and an error redirect alike" half and has to be restated with the
exceptions listed.

### The other ten

Settled in the interview and unchanged by the run, except decision 4 as amended above and decision 9,
which was re-tested against the widening and stands: no documentation page promised that removing a
callback reaches a sign-in already in progress, so nothing on the site was made false by making it
true.

leodip:
## Follow-ups and fold-ins

**The run drafted no follow-up issues, and folded nothing in.** There is nothing here for you to
file, and nothing moved in this diff that the goal does not account for.

That is worth stating rather than leaving as an absence, because an empty follow-up list and an
unwritten one look identical from outside. Each of the five stage gates and each of the three review
rounds recorded the same finding: no defect was found beside the work, and no reviewer finding was
real but out of scope. Every finding the reviewer raised over three rounds was inside this change and
was fixed in it. Nothing was deferred and nothing was refuted.

The one thing found beside the work was a coverage gap rather than a defect: stage 5's handler audit
hop had no seam in the sealed test plan, so its production code would have shipped untested. That was
covered in the stage that created it, `TestHandleTokenPost_RedemptionRegistrationRefusalAudit`,
matching `TestHandleTokenPost_ScopeDenialAudit` next door. Coverage the plan missed belongs in the
stage that needs it, so it is not a follow-up and is recorded here only so the list is complete.

### One existing issue this unblocks

**#197**, open, *"an implicit ceremony receives tokens for a session that was already ended"*. Its
owner comment says to land this change first and reuse the helper, and that is now possible:
`urlutil.RedirectURIIsRegistered` exists as of stage 1, and `/auth/issue`'s gate sits above the
`response_type` dispatch, so the implicit branch already passes through the session, scope and
registration re-checks this change adds. #197 is not closed by this branch and the run did not touch
it. Judging what it still needs is yours, not the run's.

The duplicate search was re-run at the finish over all states. Nothing else in the tracker overlaps.

leodip:
# The code review, in full

This is the run's complete review account, posted here because it lives nowhere else: the verdicts
were written into `.review/`, which is gitignored and overwritten by the next round, so without this
comment a reader of this pull request would see the fix commits with no way to learn what prompted
them, what was refuted, or what was deliberately left.

**No stage was reviewed on its own.** This run used `LEO_REVIEW_MODE=end`, so the whole branch got
one reading after the last stage landed, by `gpt-5.6-sol` at effort `xhigh`, in a fresh session each
round, scoped to `main...HEAD`. There was one earlier reading, of the plan rather than the code,
before any stage was built; its single finding is decision 11 and is answered in the decisions
comment above.

Summary, then every round verbatim as it was recorded at the gate:

|---|---|---|---|---|
| plan | 1 | 1 | 1 | escalated as decision 11, answered **B**, applied |
| final | 1 | 4 | 1 | all fixed, `0d637b97` |
| final | 2 | 3 | 0 | all fixed, `eef795c5` |
| final | 3 | 1 | 0 | fixed, `a52f7731` |

Nothing was contested, nothing was deferred, `unchecked` was empty at every final round, and the
reviewer filed no follow-ups. Round 3 changed no production code, which is what closed the gate: a
round that only hardens tests has had its answer, and the budget funds no fourth.

Two of the four rounds found more than they reported. In round 2 the two coverage findings were one
nine-member family, and enumerating it turned up a **fourth** surviving mutation the review had not
named: `state` failing to cross the clear-failure fallback, which would have emitted a `server_error`
carrying no `state` echo, against RFC 6749 section 4.1.2.1. In round 3 the reported three-member
family enumerated to **eight**, and all eight survived as mutations. Resolving the family rather than
the instance is why those two rounds cost what they did and why there was no fourth.

---

## 10. Review rounds

### Plan gate, round 1, 2026-08-26

Reviewer `gpt-5.6-sol`, effort `xhigh`, fresh session. Scope: `plan.md`'s four stages, their
ordering, stage 1's full specification, and the three sketches. All three axes reviewed. Verdict
`findings`: one finding.

The reviewer independently re-ran `check-anchors.sh` (38 anchors, each resolving once),
`tree-hash.sh --split` (unchanged before and after), both probes (`registered-table` 15 rows /
0 mismatches, `maxage-at-issue` reproducing the `max_age=0` loop), and four scratch mutations of
a model of the stage 1 matcher, all four killed by named table rows. It recounted the fanout
numbers and confirmed CI green on the base commit `078e820b`.

**Finding 1, blocking, security. Accepted; escalated as decision 11, still open.**

*The claim.* The plan re-checks current redirect URI registration only in `HandleIssueGet`, but a
ceremony can end before that handler and send an error straight to the stored callback. The
reviewer parked an authenticated browser on the consent screen of an administrator-created client,
deleted the registered redirect URI, and submitted Cancel through the suite's own `postConsent`
helper. It answered `302` with `Location: https://removed.example/callback?error=access_denied...`
where a local `200` was expected, twice.

*Verified independently, by call graph rather than by re-running the reproduction.* All six sites
answer through `redirToClientWithError` (`src/authserver/internal/handlers/handler_authorize.go:948`),
whose only gate is `redirectWillBeEmitted` (`:936`). That function returns false for a nil or
DCR-created client (gate 3, #108) and for a URI failing `checkRedirectURIEmittable` (gate 4, #122),
and consults no registrations at all. An administrator-registered client holding an absolute URI
deleted a minute earlier passes both gates. The six: consent cancelled (`handler_consent.go:380`),
nothing ticked (`:274`), the deferred authorization error after password verification
(`handler_auth_level1.go:102`), the disabled user (`handler_auth_completed.go:334`) and the empty
effective scope (`:376`) refusals. Stage 3's new filter-emptied refusal would be a seventh.

*Why it was escalated rather than folded in.* The remedy needs a third gate inside
`redirectWillBeEmitted`, which is pure today, so it takes a `data.Database` parameter: a signature
change, a security property, and a widening of §2's goal from `/auth/issue` to every response a
ceremony sends a client. It also reopens decision 4, which weighed `blocked/render` for this exact
condition and rejected it for giving one condition two pages depending on when it fires, since a
gate inside the emitter refuses through exactly that page. The run may neither overturn a `Decided`
item nor narrow its own scope, so both available answers belong to the user. The defect predates
the branch, which is byte-identical to `origin/main` at `078e820b` apart from an empty commit, so
declining to widen is a legitimate answer rather than a failure.

*Disposition.* No production code changed. Decision 11 posted to PR #265 with the family
tabulated, two options compared, and the recommendation to widen at the shared emission boundary.

**Bookkeeping, fixed in passing.** `plan.md` justified its "unparseable registered value, exact
equality still wins" row with the claim that `RedirectURIMatches` returns false for a value
`url.Parse` rejects. It does not: it opens with `if registered == requested { return true }`
(`src/core/urlutil/redirect_uri.go:234`). The row's expected value and the planned implementation
are unchanged; the note now states what the row actually pins. No round spent.

**Coverage the reviewer declined, and what answered it.** Four `unchecked` entries, each answered
in `log/plan.md`: implementation-level mutations and audit assertions dismissed as inherent to a
plan gate and owned by the stages that write them; the unchanged tiers covered by the matching
`--split` hashes and green CI on `078e820b`; the absent interactive browser pass dismissed because
the finding turns on a status code and a `Location` header, which the harness observes directly;
the unmodified surfaces dismissed as outside §2, matching the reviewer's own `why_exhaustive`,
which scopes its family to the ceremony handlers and excludes logout.


## Final gate, round 1 request written (2026-08-26)

Every stage had landed and nothing had reviewed the code, so `.review/final-review.md` was the
mailbox. This invocation wrote `.review/request.md`, `type: code-review`, `gate: final`, `round: 1`,
`status: awaiting-review`, scoped to the whole branch against its base (24 files, +2613 -188, five
stage commits plus the empty commit that opened PR #265). No stage was reviewed on its own: the run
used `LEO_REVIEW_MODE=end`, so this is the first and only reading the code gets.

Two facts recorded here because the mailbox is throwaway and does not survive the next round.

**`tree-hash.sh --verify` at HEAD `ab399662`.** Current tree: production `e05521a00b03`, unit-tests
`eeca5eb8965e`, data-tests `497298e9c253`, integration-tests `4d912ef9b082`, docs `66191b384928`,
all `b1aae2a973fe`.

| Stage | unit | data | integration | docs | moved since |
|---|---|---|---|---|---|
| 1 | stale | not recorded | stale | - | production, unit-tests, integration-tests |
| 2 | stale | not recorded | stale | live | production, unit-tests, integration-tests |
| 3 | stale | stale | stale | live | production, unit-tests, integration-tests |
| 4 | stale | stale | stale | live | production, unit-tests |
| 5 | live | live | live | live | nothing |

Four of the five entries record a result against a tree that has since moved, which is what stages
landing on top of each other does and not a fault in any entry. It means stage 5's tiers are the
only ones still free to believe, and the mutations stages 1 to 4 ran are recorded against trees that
no longer exist. Stages 1 and 2 recorded no data-tests hash, so those claims cannot be settled
either way from the record. The block was pasted into the request so the reviewer can tell which
claims still bind rather than re-deriving all of them or believing all of them.

**CI green on HEAD `ab3996629ffd1677ec57460f1f64b6d8e8d48832`**, all eleven jobs: Build validation,
Lint, Vulnerabilities, Versions, Unit on core, authserver and adminconsole, and Tests on sqlite,
mysql, postgres and mssql, each running the data then the integration tier.
<https://github.com/leodip/goiabada/actions/runs/32976596221>. That is current where the per-stage
entries are not, and it is the only four-engine claim in the record that names the tree it ran on.
The working tree is clean and identical to that commit.

The verdict is the driver's to fetch. Nothing else was done this invocation.

## Section 10, final gate round 1: verdict ingested (2026-08-26)

Reviewer `gpt-5.6-sol` at effort `xhigh`, fresh session, `gate: final`, round 1 of 4, over the whole
branch against `main`. Verdict `static-review-exhausted`, all three axes reviewed. Four findings:
one blocking, two significant, one minor. **All four were fixed this round**; none was deferred and
none was refuted.

Every one arrived with a populated `forced_answer`, so none escalated. Per `decisions.md` section 1
a `needs_human: true` finding whose `forced_answer` is populated is a finding the code settles
rather than a choice, and findings 1 to 3 each mechanically restore an invariant the agreement had
already decided. Two independent demonstrations exist for each: the reviewer's, and this session's
mutation below.

### Finding 1, blocking, security. A refused registration read could be widened into a redirect

**Claim.** `redirectWillBeEmitted` is asked twice in one authorization request: by
`authorize/answer-now` before routing, and by `authorize/redir-emitter` at the last moment. Since
decision 11 gave it a `GetRedirectURIsByClientId` read it is no longer deterministic, so the two
answers need not agree. A failing or empty first read makes `answerClientNow` true, on the
reasoning that no redirect leaves the server so RFC 9700 4.11.2 has nothing to govern. If the
second read then recovers, or sees the callback re-added, the redirect is emitted to a browser that
never authenticated. The refusal that made answering-at-once safe has been discarded while its
conclusion was kept.

**Verified independently.** Structural: the two call sites are the only production readers, and both
now perform the same live read. The reviewer reproduced it with a scratch overlay returning an error
then the registered URI, getting `302 Location: https://legit.example/cb?error=invalid_scope`. This
session reproduced the same shape as a committed test rather than an overlay.

**This is the branch's own defect, not a pre-existing one.** Before decision 11 both gates were pure,
so the two evaluations could not disagree in the unsafe direction. Making the predicate impure
introduced the window.

**Fixed.** `redirectErrorInput` gains `redirectAlreadyWithheld`, a floor and never a ceiling:
`redirToClientWithError` reads it first and short-circuits above the live gates, so a refusal already
given cannot be overturned, while a later read may still narrow a yes to a no. `HandleAuthorizeGet`
memoises the predicate in the shape `hasValidUserSession` already uses and sets the field from it.
Both directions are pinned by `TestHandleAuthorizeGet_RegistrationReadDisagreesWithItself`, whose two
rows are separated by `AssertNumberOfCalls` on the read: one read means the floor short-circuited,
two means the emitter asked and took the stricter answer. The dangerous second answer is deliberately
left available to the mock rather than withheld, so the row proves the floor by making the wrong
answer reachable and asserting the code never went back for it. Caching the first answer instead of
flooring it would make both rows read once and fail the second.

### Finding 2, significant, conformance. Code-flow errors to native loopback clients were withheld

**Claim.** The emitter passed `false` for loopback port flexibility unconditionally, so a native
client registered at `http://127.0.0.1/cb` starting a code flow at `http://127.0.0.1:49152/cb` could
receive a success but not a protocol error.

**The agreement's own reasoning was wrong here, and this is the one finding that overturns a written
design statement.** Section 4's decision-11 text says the flag is false because "an error redirect
carries no code, so this gate is deliberately stricter". That asks the wrong question. The gate
weighs whether the destination is one the client *still has registered*, and both specs were fetched
before deciding rather than recalled:

- RFC 8252 section 7.3: "The authorization server MUST allow any port to be specified at the time of
  the request for loopback IP redirect URIs, to accommodate clients that obtain an available
  ephemeral port from the operating system at the time of the request."
- RFC 6749 section 4.1.2.1 confines the refusal to a request that "fails due to a missing, invalid,
  or mismatching redirection URI"; for every other failure "the authorization server informs the
  client by adding the following parameters to the query component of the redirection URI".

So the ported URI *is* registered for a code request, and the errors are owed there. Passing `false`
made a native app's consent cancellation, deferred validation error and scope refusal vanish into
this server's interstitial with nothing on the wire, which is a hang rather than an error it can
report. It was also a regression against `main`, where the emitter performed no registration check
at all and those errors were delivered.

**And it is not a widening of what the ceremony already admits.** The same flexible rule ran over the
same URI at `/auth/authorize` for this same ceremony, so a destination this arm now accepts is one
the front door accepted; and a loopback IP literal names the browser's own machine, not a host
somebody else controls. Strictness is not free on a gate that can only refuse: an over-strict answer
here is a delivery failure with no security to show for it.

**Fixed.** `redirectWillBeEmitted` takes the response type and computes the flag with the same
token-sequence test `validator/loopback-gate` and `/auth/issue` use, so exactly one token equal to
`code` buys flexibility and every implicit response stays strict. Both readers pass the same string:
`authContext.ResponseType` and `redirectErrorFromRequest`'s `r.FormValue("response_type")` are the
same raw value, assigned at `handler_authorize.go:113`. Decision 7's shape is untouched: the gate
still lives with the caller, and the flag is computed once inside the one predicate its two readers
share rather than copied to a fourth site. `TestRedirectWillBeEmitted` replaces its single
"flexibility is NOT granted here" row with three: granted on `code`, denied on `token`, denied on
`code code`.

### Finding 3, significant, quality. Three new fail-closed branches had no test

**Claim.** The issuance registration load, the issuance permission filter and the consent permission
filter all fail closed, and no test made any of those dependencies error, so widening each to use the
requested or submitted authority left the whole authserver internal tier green. Three mutations
survived in the reviewer's run.

**Verified.** All three branches read, all three answer `InternalServerError`. The positive tables
above each prove what the check refuses when an answer arrives; none made the answer fail to arrive.
The gap matters most on the permission seams, where the empty string is the refusal signal, so a
filter that errors and a filter that removed everything return the same value and only `err` keeps
them apart.

**Fixed.** `TestHandleIssueGet_TheLiveChecksFailClosedOnAStorageError` (two rows) and
`TestHandleConsentPost/The filter failing closed records nothing`. Each registers no expectation for
anything downstream, so a call reaching the issuer, the audit logger, the consent read or write, the
context save or the renderer fails the case: the absence is the assertion. All three of the
reviewer's surviving mutations are now caught, below.

### Finding 4, minor, conformance. Four pages promised redirects the gate now withholds

**Claim.** Decision 9 concluded nothing on any surface was falsified, and decision 11 re-tested it and
let it stand. Both were wrong about the same thing: four concept pages state categorically that an
administrator-created client receives ordinary error redirects and that every `prompt=none` failure
returns to the client, naming self-registration as *the* exception. There is now a second exception.

**Verified, page by page.** Eight sites, seven false: `authorization-lifecycle` at the validation
Aside, the deferred-error table row, the `prompt=none` paragraph and the Declining table;
`prompt-parameter`'s "Clients an administrator registered are unaffected"; `user-sessions`'s
silent-renewal paragraph, which this run wrote itself in stage 5; and `clients`'s "Clients an
administrator created keep the ordinary redirect".

**Fixed, as the smallest edit that makes each page true**, per `docs.md` section 3: no new section,
no new page, each correction made in the sentence that is wrong. The eighth site is a precision edit
rather than a correction: `clients.mdx`'s unbadged-client paragraph says "the refusal behaviour above
does not apply to it", which was true of the self-registered rule and would have become false once a
second rule joined the paragraph above it, so it now names the self-registered refusal specifically.

**Word counts, per site, before and after.** `docs.md` section 3 asks for these and asks that they be
compared rather than merely recorded.

| Page | Site | Before | After | Delta |
|---|---|---|---|---|
| authorization-lifecycle | validation Aside | 17 | 30 | +13 |
| authorization-lifecycle | deferred-error table row | 19 | 24 | +5 |
| authorization-lifecycle | `prompt=none` paragraph | 24 | 42 | +18 |
| authorization-lifecycle | Declining table row | 25 | 41 | +16 |
| prompt-parameter | silent-request Aside, last sentence | 6 | 21 | +15 |
| user-sessions | silent-renewal paragraph | 33 | 45 | +12 |
| clients | security considerations, refusal paragraph | 8 | 40 | +32 |
| clients | unbadged-client paragraph | 12 | 12 | +0 |

**+111 words over four pages, and it is a deviation from decision 9**, which said one sentence in
`docs/session-consent-check` and nothing else. Recorded as one rather than absorbed. The deviation is
forced: decision 9's premise was that no page was falsified, and seven sentences were. Each is
counted against its own siblings, per `docs.md`: the table cells sit beside cells of 12 to 25 words,
and `clients.mdx`'s 40 words sit in a section whose paragraphs run 40 to 90. That paragraph is the
largest single addition and the only one carrying an operator action, which is the test `docs.md`
sets: an administrator who deletes a redirect URI now learns it takes effect on sign-ins already
under way. The other seven are boundary corrections to sentences that had become false, which is the
one thing `docs.md` says a page is always owed.

### Mutations run this round

Every one against the tier named, through `where.sh mutate`, which restores from a trap and compares
by hash. Five of them are the reviewer's own surviving mutations, re-run against the new tests.

|---|---|---|---|
| `handler_authorize.go` | `if input.redirectAlreadyWithheld \|\|` to `if false \|\|` | modules | **CAUGHT** |
| `handler_authorize.go` | loopback flag to always-false (`responseTypes[0] == "zzz"`) | modules | **CAUGHT** |
| `handler_authorize.go` | loopback flag to always-true (`len >= 1 && [0] != ""`) | modules | **CAUGHT** |
| `handler_auth_issue.go` | registration-load error treated as registered | modules | **CAUGHT** (survived for the reviewer) |
| `handler_auth_issue.go` | permission-filter error treated as the original scope | modules | **CAUGHT** (survived for the reviewer) |
| `handler_consent.go` | permission-filter error treated as the submitted selection | modules | **CAUGHT** (survived for the reviewer) |

The loopback flag is mutated in *both* directions on purpose. Always-false is the defect finding 2
reported; always-true is the property the implicit and `code code` rows exist to pin, and without it
the fix could have been written as an unconditional `true` and no row would have objected.

**Mutation residue checked by hash before committing.** `git status` shows `handler_auth_issue.go`
and `handler_consent.go` unmodified, so both were restored byte-identically; `git diff --check` is
clean.

### Answers to the verdict's `unchecked` array

All four entries answered, per `reviewer.md`.

1. *No real native loopback app or interactive browser was driven.* **Covered, differently.** The
   claim it supports is finding 2, and the fix is now pinned by three `TestRedirectWillBeEmitted`
   rows plus the existing `/auth/issue` recheck table, which already carries the same code-flow and
   implicit loopback pair. The decision is at the HTTP boundary and both specs were fetched.
2. *No real database outage was timed between the two registration reads.* **Covered.** The strict
   sequential mock is now a committed test rather than an overlay, and the call-count assertion is
   what makes the timing irrelevant: the fix removes the second read entirely on the refusal path.
3. *Concurrent admin writes after a successful final recheck but before mint or emission were not
   measured.* **Dismissed.** Section 2 scopes this change to point-in-time rechecks, and the residual
   window between a passing check and the write it guards is the one every such check has. Closing it
   needs a locking or generation contract, which is a different change with a different design.
4. *Full tiers were not rerun locally.* **Covered.** This session ran them: unit 3998 passed 0 failed,
   integration on sqlite 1314 passed 0 failed, both against the post-fix tree.

### Tiers, this round

| Tier | Command | Engines | Result | Time |
|---|---|---|---|---|
| unit | `where.sh test --type modules` | n/a | 3998 passed, 0 failed | ~1m |
| integration | `where.sh test --type integration --db sqlite` | sqlite | 1314 passed, 0 failed | 57s |
| data | not run | - | **not applicable**, unchanged | - |

**Data is not applicable and the hash proves it rather than asserting it.** `data-tests` is
`497298e9c253` and `integration-tests` is `4d912ef9b082`, both identical to the values stage 5
recorded: no data-tier file moved this round. No new `Database` method, query, migration or schema
change, exactly as section 5 said.

**Tree after the fixes:** production `6e4a48005970`, unit-tests `a3509bfed790`, data-tests
`497298e9c253`, integration-tests `4d912ef9b082`, docs `943e6e7f5fea`, all `fea1cf7c8661`.
Production, unit-tests and docs moved; the two test-data categories did not.

`check-anchors.sh`: 41 of 41 rows resolve. One row was re-swept, `authorize/redir-emitter`, whose
locate string was the emitter's gate call and is now `if input.redirectAlreadyWithheld ||`. That is
the only section 0 edit.

### What this round did not do

No finding was deferred and none was refuted, so nothing goes to section 9 from this round. The
round changed production code, so the budget is not the constraint on whether a round 2 happens.

## Section 10: final gate, round 2 (2026-08-26)

Reviewer `gpt-5.6-sol` at effort `xhigh`, resumed Codex session, round 2 of a maximum 4, over the
whole branch against `main` at HEAD `0d637b97`. Request `241-20260826T172148Z`.

All three axes `reviewed`, `unchecked` empty. The reviewer verified the split tree hashes and all 41
anchors itself, confirmed CI green on all eleven jobs for HEAD, ran focused handler and validator
tests, and ran five independent mutations of its own, two caught and three surviving. That is a
review that wrote its own programs rather than re-running ours, which is the kind worth paying for.

**Three findings, all three fixed, none refuted, none deferred.** No finding was blocking. Every one
carried `needs_human: true` with a populated `forced_answer`, which per `reviewer.md` section 3 is a
finding the code settles rather than an escalation, so none went to the user.

**Finding 1, `significant`, `quality`.** The monotonic redirect-refusal floor is not tested through
the `answerClientWithError` clear-failure fallback: resetting `fallback.redirectAlreadyWithheld` to
`false` left the whole authserver internal tier green. Reproduced before acting.

**Finding 2, `significant`, `quality`.** Response-type propagation into the loopback registration
gate is untested at two composed seams: passing an empty response type from `redirToClientWithError`
into `redirectWillBeEmitted`, and erasing `fallback.responseType`, both left the tier green. Either
regression would withhold a native client's code-flow errors from the ephemeral loopback port that
RFC 8252 section 7.3 requires be accepted. Reproduced before acting.

Findings 1 and 2 were resolved as **one family**, not two, because both are the same shape: a field
of `redirectErrorInput` failing to reach the gate. The reviewer's two `why_exhaustive` statements
each scoped themselves to the single field their finding was about, so neither established the
family's boundary. Enumerated all nine members and mutated every one:

| # | Member | Before | After |
|---|---|---|---|
| 1 | fallback preserves `client` | caught | caught |
| 2 | fallback preserves `responseMode` | caught | caught |
| 3 | fallback preserves `redirectURI` | caught | caught |
| 4 | fallback preserves `state` | **SURVIVED** | caught |
| 5 | fallback preserves `responseType` | **SURVIVED** | caught |
| 6 | fallback preserves `redirectAlreadyWithheld` | **SURVIVED** | caught |
| 7 | gate argument `input.client` | caught | caught |
| 8 | gate argument `input.redirectURI` | caught | caught |
| 9 | gate argument `input.responseType` | **SURVIVED** | caught |

**Four survived, not the three reported.** Member 4 is the one the review missed: dropping `state`
across the fallback emits a `server_error` with no `state` echo, contrary to RFC 6749 section
4.1.2.1, and no test failed. It is a conformance gap the reviewer's own enumeration could not have
found, since it never asked what else the fallback preserves.

Fixed by two new tests in `handler_authorize_test.go`:
`TestRedirToClientWithError_PassesTheResponseTypeIntoTheGate`, three rows proving the ephemeral
loopback port is reached for exactly one `code` token and refused for `token` and `code code`; and
`TestAnswerClientWithError_TheFallbackPreservesWhatItDoesNotSwap`, three cases pinning the floor,
the response type and `state` across the code-and-description swap. The floor case gives the
emitter a database mock with no expectations at all, so a second registration read fails the case
outright, which is stronger than stubbing a recovered read and asserting it was ignored.

**Finding 3, `minor`, `conformance`.** `token_validator.go`'s registration-boundary comment still
justified its unconditional `true` by contrast with the emitter, "unlike the emitter's gate, which
passes false", a rule round 1 finding 2 had reversed one commit earlier. Rewritten to give the
reason positively: holding an authorization code is itself the proof of the code flow, so redemption
needs no response type, while the emitter derives the flag because it runs where no code exists yet.
Swept the branch for the family (`passes false`, `unconditionally`, `port flexibility`, `portless`
across all 27 changed files) and confirmed independently that this was the only remaining false
statement. No behaviour changed.

**Tiers, at `production 8f1f052c1cb0`, `unit-tests f6e16175ddfd`:**

| Tier | Command | Engines | Result | Time |
|---|---|---|---|---|
| unit | `where.sh test --type modules` | n/a | 4006 passed, 0 failed | ~1m |
| integration | `where.sh test --type integration --db sqlite` | sqlite | 1314 passed, 0 failed | 67s |
| data | not run | - | not applicable, per section 5: no new `Database` method, query, migration or schema change | - |

`data-tests 497298e9c253` and `integration-tests 4d912ef9b082` are byte-identical to round 1's, so
no data-tier or integration-tier file moved this round.

`check-anchors.sh`: 41 of 41 rows resolve, no re-sweep needed.

Thirteen mutations were run this round, nine enumerating the family and four re-run after the fix.
The tree was verified byte-identical by split hash after every batch, and `git status` was clean of
everything but the two intended files.

## Section 10, final gate, round 3

**Reviewer:** `gpt-5.6-sol` at effort `xhigh`, session resumed, gate `final`, round 3 of 4.
Request `241-20260826T174801Z`, verdict written 2026-08-26T22:12:06Z against HEAD `eef795c5`.
Ingested 2026-08-26. Fixed in `a52f7731`.

**Axis coverage:** conformance `reviewed`, quality `reviewed`, security `reviewed`. `unchecked` is
empty. `## Follow-ups` is None. One finding, `significant`, `quality`, confidence `high`,
`needs_human: true` with a populated `forced_answer`, which per `reviewer.md` section 3 is a finding
the code settles rather than a choice that escalates.

**What it cleared, with what it ran.** It reproduced the six split hashes exactly, confirmed all 41
anchors resolve, and confirmed the Check workflow concluded green on `eef795c5` itself across lint,
vulnerabilities, all three module jobs, and the sqlite, mysql, postgres and mssql data and
integration jobs. It ran the two round 2 repair tests directly and then thirteen independent
mutations, of which ten were caught: erasing fallback state, fallback response type and the fallback
refusal floor; dropping the emitter response type; erasing response type at each of the two input
constructors; suppressing the issuance scope write-back; removing session validity from the binding
predicate; persisting the unfiltered consent selection; and disabling the code-redemption
registration refusal. It traced all eleven decided requirements through the whole `main...HEAD`
production diff and every production reader of `redirectWillBeEmitted`, and found no product or
documentation conformance defect and no security defect.

### Finding 1, significant, quality: the new refusals do not test what happens when they themselves fail

**Family as reported:** "ClearAuthContext failure behavior on refusals newly introduced by issue
241". Three members, one per new refusal, each demonstrated by a surviving mutation against the full
authserver internal tier.

**Verified, and the enumeration confirmed.** `handler_auth_completed.go` and `handler_authorize.go`
carry the same number of `ClearAuthContext` calls on `main` and at HEAD; `handler_auth_issue.go`
goes from 4 to 6 and `handler_consent.go` from 2 to 3. So the branch adds exactly three, and the
reviewer's boundary for its own family is right.

**The family is bigger than the family it named, and reading the test files is what shows it.** The
established house form for a refusal in these two files is a **triple**, not a single case:

- `handler_auth_issue_test.go`, the no-session `prompt=none` refusal: "failing clear - server_error
  to the client", "failing clear and an unusable form_post template - last-resort 500", "unusable
  form_post template - 500 when the refusal itself cannot be sent".
- The same three for the `id_token_hint` mismatch refusal in the same file.
- The same three in `handler_consent_test.go` for "User cancels consent" and for "No consent given".

Two of the three refusals this issue added answer the client by redirect, so each owes that whole
triple and had none of it. The third renders locally, so it owes a pair rather than a triple: it has
one way for its answer to fail rather than two. **Eight members, against the three reported.**

Every one was demonstrated by a mutation against the full `--type internal` tier, and **all eight
survived**:

| # | Arm | Mutation | Before | After |
|---|---|---|---|---|
| A1 | issuance, deregistered URI | `return` inserted after the failed-clear log, so the interstitial is never rendered | SURVIVED | CAUGHT |
| A2 | issuance, deregistered URI | the 500 after a failed `renderRedirectBlocked` deleted | SURVIVED | CAUGHT |
| B1 | issuance, empty scope | `server_error` swapped for `access_denied` in the clear-failure fallback | SURVIVED | CAUGHT |
| B2 | issuance, empty scope | the fallback's last-resort 500 deleted | SURVIVED | CAUGHT |
| B3 | issuance, empty scope | the 500 on the ordinary `access_denied` path deleted | SURVIVED | CAUGHT |
| C1 | consent, filter emptied the selection | `server_error` swapped for `access_denied` in the clear-failure fallback | SURVIVED | CAUGHT |
| C2 | consent, filter emptied the selection | the fallback's last-resort 500 deleted | SURVIVED | CAUGHT |
| C3 | consent, filter emptied the selection | the 500 on the ordinary `access_denied` path deleted | SURVIVED | CAUGHT |

A1, B1 and C1 are the three the review reported. A2, B2, B3, C2 and C3 are the siblings that share
the house form, and they were found by enumerating it rather than by taking the finding as handed.

**Fixed in `a52f7731`,** eight cases in the shape the neighbouring refusals already use:

- `TestHandleIssueGet_RedirectURIRefusalSurvivesItsOwnFailures`, a pair. The failing-clear case
  arms `RenderTemplate` as an expectation rather than an assertion, so a handler that returned early
  fails the mock for an unmet expectation, and it carries no `InternalServerError` expectation so a
  bare 500 fails too. The unrenderable-page case asserts the 500 and asserts `Location` stays empty,
  which is the property that matters: a failure to render the refusal must not fall back to a
  redirect toward the host the gate just refused.
- `TestHandleIssueGet_ScopeRefusalSurvivesItsOwnFailures`, the full triple, `form_post` throughout
  for the two 500 cases because it is the only response mode whose arm can fail after the redirect
  URI has been validated.
- Three subtests in `TestHandleConsentPost`, named for the refusal they belong to, sitting directly
  above the "No consent given" triple they copy.

**No production code changed.** `production` is `8f1f052c1cb0` before the round and after it, which
is also the value the round was reviewed at, so every one of the sixteen mutation runs is proven
restored by hash.

### Answering the earlier round that called this ground closed

Round 1 finding 3 was the same shape, and its `why_exhaustive` read: "Every new error-returning
storage and permission-check call was matched against explicit mock-error fixtures; only these three
lacked one." `ClearAuthContext` is an error-returning call added at three new sites by this branch,
and none of the three was matched. So round 1's exhaustiveness claim was wrong, and this round is
the second consecutive one to land in the coverage table for the branch's new refusal arms.

That is the third round in a row where enumerating the family found more than the review reported:
round 2's nine-member enumeration turned up a fourth surviving mutation on `state`, and this one
turned up five. The lesson for the next `/leo-spec` is that a repository with a strong house form
for a code shape should have that form named in section 5 as a seam obligation, so a new refusal
arrives owing its triple rather than acquiring it three review rounds later.

### Tiers, run against `a52f7731`

| Tier | Command | Engines | Result | Time |
|---|---|---|---|---|
| unit | `where.sh test --type modules` | n/a | 4016 passed, 0 failed | ~11s |
| integration | not re-run | - | round 3's sqlite result still binds | - |
| data | not run | - | not applicable, per section 5 | - |

Integration was not re-run and the hashes are why. `production` is unchanged at `8f1f052c1cb0` and
`integration-tests` is unchanged at `4d912ef9b082`, so nothing the integration tier executes moved
and the 1314-passed sqlite run recorded at `eef795c5` binds unchanged. Only `unit-tests` moved,
`f6e16175ddfd` to `b272d6d9ba21`. The unit count rose by exactly ten, 4006 to 4016: eight new
subtests plus the two new top-level functions that hold five of them.

`check-anchors.sh`: 41 of 41 rows resolve, no re-sweep needed.

**The four-engine result is the check suite's, on `a52f7731`.** It was green on all eleven jobs at
`eef795c5`, the reviewer confirmed that independently, and this commit adds test files only.

### Why this gate closes here

Nothing is contested, nothing is deferred, `unchecked` is empty and the reviewer filed no
follow-ups. Round 3 changed no production code, so under `reviewer.md` section 5 the budget does not
fund a round 4: a gate whose last round only hardened tests has had its answer. `.review/request.md`
is marked `ingested` and `.review/final-review.md` is deleted, which lets the finish run.

---

Issue #268, opened 2026-08-27
golang-migrate links a second Postgres driver, a second JWT library and five Azure modules into both binaries, to run numbered .sql files


leodip:
`golang-migrate/migrate/v4` is used for four symbols. In exchange it links **eight modules** into both binaries, including a second Postgres driver and a second JWT library that nothing in Goiabada calls.

## What we actually use

Every one of the four `NewMigrator()` implementations (`core/data/{sqlitedb,mysqldb,postgresdb,mssqldb}/db.go`) is the same shape, and the whole API surface across the repository is:

| Symbol | Call sites |
|---|---|
| `migrate.NewWithInstance` | 4 |
| `migrate.Migrate` (step to a version, used by tests) | 6 |
| `migrate.ErrNoChange` | 11 |
| `Up()` | 4 |

That is it. No `Down()`, no source drivers beyond `iofs`, no CLI, no locking beyond what the database drivers give us.

## What it costs

`go mod why -m` from `src/core`:

```
github.com/lib/pq
  core/data/postgresdb -> migrate/v4/database/postgres -> github.com/lib/pq

github.com/Azure/go-autorest/autorest/adal
  core/data/mssqldb -> migrate/v4/database/sqlserver -> Azure/go-autorest/autorest/adal

github.com/golang-jwt/jwt/v4
  core/data/mssqldb -> migrate/v4/database/sqlserver -> Azure/go-autorest/autorest/adal -> golang-jwt/jwt/v4
```

So the migration runner drags in:

| Module | Why it is absurd |
|---|---|
| `github.com/lib/pq` | **A complete second PostgreSQL driver**, linked alongside `jackc/pgx/v5`, which is the one we actually talk to Postgres with |
| `github.com/Azure/go-autorest` + `/autorest/adal` + `/autorest/date` + `/logger` + `/tracing` | Five modules of Azure AD token machinery, for a SQL Server driver we never authenticate to Azure with |
| `github.com/golang-jwt/jwt/v4` | **A second JWT library**, linked alongside `golang-jwt/jwt/v5`, reached only through the Azure module above |

Eight modules of the roughly 41 linked into each binary, to read numbered files and run them in order.

## What replaces it

A `core/data/commondb` migration runner. The contract it has to honour is small and already visible in the schema snapshots:

```sql
CREATE TABLE schema_migrations (version uint64, dirty bool);
CREATE UNIQUE INDEX version_unique ON schema_migrations (version);
```

Behaviour to reproduce exactly:

[...]

Issue #259, opened 2026-08-25
Security: a client-scoped revocation cannot reach ROPC refresh grants, and PKCE-optional is broken on SQLite


leodip:
Two defects that the #245 review found and deliberately did not close. They are independent and
can land in either order or as separate pull requests. They are filed together because both need a
new migration and four-engine data tests, and both live in `src/core/data` and the token endpoint.

**The first is a security defect and does not depend on the second.** If only one gets done, do
that one.

---

## 1. A client-scoped revocation cannot reach an ROPC refresh grant

Switching a client from confidential to public revokes its outstanding grants in one transaction
(#245). For authorization-code grants that revocation is **durable**: the flip marks the client's
authorization codes revoked, a rotated refresh token inherits its parent's `code_id`, and the
refresh arm refuses any token whose code is marked, so a replacement inserted after the flip commits
is born already rejected.

**ROPC grants have no such handle.** An ROPC refresh token carries a NULL `code_id`, there is no
code row to mark, and every code-borne check in `ValidateTokenRequest` is guarded by `isROPCToken`,
which is defined as `!refreshToken.CodeId.Valid`. The revoked-code check and the PKCE boundary #245
added are both behind it. So `RevokeClientGrants` reaches only the ROPC tokens that exist at the
instant its transaction reads `refresh_tokens` through `GetRefreshTokensByClientId`.

Rotation writes in **two separate autocommit operations**. `HandleTokenPost` claims the presented
token with `MarkRefreshTokenAsRevoked(nil, ...)`, a conditional `UPDATE ... WHERE revoked = false`
that commits on its own, and `TokenIssuer` then inserts the replacement with
`CreateRefreshToken(nil, ...)` in `generateRefreshTokenForROPC`. Both pass `nil` for the
transaction. So this ordering is reachable:

1. A holder of an ROPC refresh token starts a refresh. It validates while the client is confidential.
2. The presented token is marked revoked. Committed.
3. The flip runs, finds that token already revoked and no replacement yet, and commits.
4. The refresh inserts the replacement as a **live** token.

The administrator is told the client's grants were revoked. One survives, and it rotates
indefinitely with nothing presented at all, because a public client's ROPC arm skips authentication
entirely: the whole secret check in `ValidateTokenRequest`'s `password` arm sits under
`if !client.IsPublic`.

**Why it is real rather than theoretical.** The window is a fixed fraction of every rotation cycle
[...]

Issue #284, opened 2026-08-27
Nothing checks that the four database engines build the same schema


leodip:
The four migration sets are written by hand, in parallel, and no test compares the result. Every divergence in #282 reached main the same way: a migration was written four times and one of the four said something slightly different.

`000007_make_pkce_nullable` is the clearest case. Three engines drop a NOT NULL constraint; the SQLite file is a `SELECT 1;` no-op justified by a comment claiming SQLite does not enforce NOT NULL, which it plainly does. That comment has been in the tree since the migration landed and nothing has ever contradicted it.

The `schema.sql` snapshots do not help here. They are documentation, never loaded by Go, and they have been out of sync with the migrations before.

## Shape

Reading the catalogs is straightforward on all four engines: columns with type, nullability and default; indexes with uniqueness and key columns; foreign keys with their referential actions; check constraints; collation. I have a working dumper that does exactly this and produced the findings in #282, so the query half is a known quantity.

The awkward part is that the data tier runs one engine per invocation, so no single test process ever sees all four schemas. Two steps solve it:

1. **Per engine, in the data tier.** Dump the schema of a freshly migrated database into a normalised structure and assert it matches a golden file committed alongside that engine's migrations. This also turns the `schema.sql` snapshots into something checked rather than believed, since the golden file is generated from the migrations that actually run.
2. **Once, as a plain unit test with no database.** Read the four golden files and assert they agree, modulo an explicit allowlist of engine idioms.

The allowlist is the valuable half. It is where we write down, once, that `bigint unsigned` on MySQL against signed `bigint` elsewhere is deliberate, that MySQL's automatic foreign key indexes are not a divergence, that SQLite's `TEXT` stands in for every `varchar(N)`, and that `refresh_tokens` cannot cascade on SQL Server because the engine refuses multiple cascade paths. Anything not on that list fails the build at the moment the migration is written, which is the only time it is cheap to fix.

Every divergence found so far would have been caught by this, including the collation split, provided the dump includes collation, which it should.

---

### Schema standardization

**#282** the five structural divergences, one migration each.

**#283** string comparison semantics, the one that changes behaviour rather than just shape.

**#284** the parity test that stops all of it drifting apart again.

PR #286, opened 2026-08-27
fix(core): the four database engines build the same schema

### LINKED ISSUE #282: The four database engines do not build the same schema: five structural divergences
The four engines are built from four separate migration sets maintained by hand, and nothing checks that they agree. I applied all 35 migrations to a fresh database on each engine and diffed the catalogs: columns with type, nullability and default, indexes with uniqueness and key columns, foreign keys with their referential actions, check constraints, collations. Not from the `schema.sql` snapshots, which are documentation-only and have drifted before.

Most of it holds up. All four build the same 25 tables, with the same column names and nothing missing anywhere, the same primary keys, the same unique constraints, and no type mismatch at all between MySQL, PostgreSQL and SQL Server: no `varchar(64)` on one engine against `varchar(128)` on another. No engine has a CHECK constraint.

Five places disagree.

## 1. `refresh_tokens(client_id)` has no index on SQLite

`000011_ropc_refresh_token` creates `idx_refresh_tokens_client_id` for MySQL, PostgreSQL and SQL Server. The SQLite file never creates it, and never did: SQLite picks up `idx_refresh_tokens_user_id` later, in `000024_add_auth_state_generation`, but the client one is simply absent.

`commondb.DeleteClient` opens with `deleteRefreshTokensByColumn(tx, "client_id", clientId)`, which is a full table scan on SQLite and an index seek everywhere else.

Fix: one migration, SQLite only, purely additive.

```sql
CREATE INDEX idx_refresh_tokens_client_id ON refresh_tokens(client_id);
```

## 2. `audit_logs.details` has no default on MySQL

`000018_add_audit_logs` declares `details TEXT NOT NULL DEFAULT '{}'` on SQLite, PostgreSQL and SQL Server, and `details TEXT NOT NULL,` on MySQL.

MySQL rejects a literal default on a TEXT column outright, which is presumably why it was dropped rather than translated. The expression form does work, but only inside `CREATE TABLE` or `ALTER TABLE ... MODIFY`, never through `ALTER COLUMN ... SET DEFAULT`. Verified against the MySQL 9.5 in the dev container:

```sql
ALTER TABLE audit_logs ALTER COLUMN details SET DEFAULT '{}';       -- ERROR 1101
ALTER TABLE audit_logs ALTER COLUMN details SET DEFAULT ('{}');     -- ERROR 1101
ALTER TABLE audit_logs MODIFY details TEXT NOT NULL DEFAULT ('{}'); -- works
```

Fix: one migration, MySQL only, using the `MODIFY` form. It needs MySQL 8.0.13 or newer, which is already implied: `NewMySQLDatabase` creates the database with `COLLATE utf8mb4_0900_ai_ci`, a collation that does not exist before 8.0.

## 3. `codes.code_challenge` and `code_challenge_method` are still NOT NULL on SQLite

`000007_make_pkce_nullable` drops the constraint on MySQL, PostgreSQL and SQL Server. The SQLite file is a `SELECT 1;` no-op, on these stated grounds:

```
-- The columns are already effectively nullable in SQLite (it doesn't enforce NOT NULL strictly for VARCHAR)
```

That is not true. SQLite enforces NOT NULL like any other engine, whether the value is written explicitly or omitted:

```
sqlite> CREATE TABLE t (code_challenge TEXT NOT NULL);
sqlite> INSERT INTO t (code_challenge) VALUES (NULL);
Error: NOT NULL constraint failed: t.code_challenge
```

So an insert that omits either column succeeds on three engines and fails on SQLite. Nothing reaches it today, because `sqlbuilder.NewStruct` writes every non-pk column and `models.Code` holds both as plain `string`, so the value is always `''` and never NULL.

Two ways to close it, and the cheaper one is also the more honest:

- **NOT NULL on all four** (recommended). The Go model has no pointer and so can never represent NULL. Backfill `NULL` to `''` and add the constraint back on the three engines. `codes` rows are short-lived, so the backfill costs nothing.
- Nullable on all four, which needs a full table rebuild on SQLite in exchange for a state the application cannot produce.

## 4. Five columns are non-Unicode `varchar` on SQL Server

Every other string column on SQL Server is `nvarchar`. These five are not:

| column | declared in |
|---|---|
| `codes.code_challenge` | `000007_make_pkce_nullable` |
| `codes.code_challenge_method` | `000007_make_pkce_nullable` |
| `user_profile_pictures.content_type` | `000008_add_profile_picture` |
| `clients.include_open_id_connect_claims_in_id_token` | `000013_add_oidc_claims_in_id_token_setting` |
| `client_logos.content_type` | `000014_add_client_logo` |

The first two are a regression rather than a choice: `000001_initial_create` declares both `NVARCHAR`, and `000007` re-declares them `VARCHAR` on its way to making them nullable.

All five hold ASCII by nature, so nothing is broken today. What it costs is that those five columns silently transcode non-ASCII input to `?`, on SQL Server and only on SQL Server.

Fix: one migration, SQL Server only. Two traps, both verified in the dev container:

- `ALTER COLUMN` fails when an **index** references the column (`ALTER TABLE ALTER COLUMN <c> failed because one or more objects access this column`). None of these five is indexed, so it does not bite here, but it will if the recipe is reused.
- `ALTER COLUMN` also fails when a **default constraint** depends on the column, which is the same trap that makes the `000004` and `000017` down migrations unrunnable. `clients.include_open_id_connect_claims_in_id_token` carries `DEFAULT 'default'` under an auto-generated name, so it needs the drop, alter, recreate sequence:

```sql
DECLARE @n sysname = (SELECT dc.name FROM sys.default_constraints dc
  JOIN sys.columns c ON c.object_id = dc.parent_object_id AND c.column_id = dc.parent_column_id
  WHERE dc.parent_object_id = OBJECT_ID('clients')
    AND c.name = 'include_open_id_connect_claims_in_id_token');
EXEC('ALTER TABLE clients DROP CONSTRAINT [' + @n + ']');
ALTER TABLE clients ALTER COLUMN include_open_id_connect_claims_in_id_token NVARCHAR(10) NOT NULL;
ALTER TABLE clients ADD CONSTRAINT DF_clients_include_oidc_claims_in_id_token
  DEFAULT 'default' FOR include_open_id_connect_claims_in_id_token;
```

Naming the constraint on the way back also takes it out of that trap next time.

## 5. `refresh_tokens` does not cascade on SQL Server

`refresh_tokens.user_id -> users(id)` and `refresh_tokens.client_id -> clients(id)` are `ON DELETE CASCADE` on SQLite, MySQL and PostgreSQL, and `ON DELETE NO ACTION` on SQL Server (`000011_ropc_refresh_token`).

This one is not an oversight. SQL Server refuses multiple cascade paths, and deleting a user reaches `refresh_tokens` two ways: directly, and through `codes.user_id`. The engine will not accept the CASCADE, so SQL Server cannot be brought up to meet the other three.

It works today because `commondb.DeleteUser` and `commondb.DeleteClient` both clear tokens with `deleteRefreshTokensByColumn` before deleting the row. The cost is that the invariant lives in the schema on three engines and in Go on the fourth, so any future path that deletes a user or client by another route works on three and errors on one.

Fix, if the four should agree: bring the other three **down** to `NO ACTION`, so all four depend on the application clearing tokens first, and all four fail identically when something does not. Dropping and recreating a foreign key is routine on MySQL and PostgreSQL. SQLite cannot alter a constraint, so it needs a table rebuild, which is the only real work in this issue. There is precedent: `000034_canonicalize_web_origins` already rebuilds a SQLite table.

The alternative is to leave the schema alone and pin the behaviour with a cascade test that runs on all four engines. Cheaper, and it keeps the trap.

## Related, and a decision rather than a fix

InnoDB creates an index for every foreign key automatically, so MySQL carries 18 indexes the other three lack. Read the other way round: **foreign key columns are unindexed on SQLite, PostgreSQL and SQL Server**, which is where `ON DELETE CASCADE` and every join across those columns pays for it. Adding them is additive and risk-free. It is a performance question rather than a correctness one, so it wants a measurement before a migration.

## Looked at and deliberately left alone

For the record, the diff also turned these up, and none is worth touching:

- SQLite declares `TEXT` where the others declare `varchar(N)`, so length limits are enforced by the database on three engines and only by the validators on SQLite. SQLite would not enforce the length even if it were declared, so the only real fix would be CHECK constraints on 90 columns.
- SQLite declares booleans three different ways (`numeric`, `BOOLEAN`, bare `INTEGER`). All three land on the same affinity, so behaviour is identical.
- Every id and foreign key column is `bigint unsigned` on MySQL and signed `bigint` elsewhere. Changing 55 columns means dropping and recreating every foreign key for no functional gain.
- Auto-increment is spelled three ways, and SQLite reports its rowid `id` as nullable.
- Column order differs on `refresh_tokens` and `users`. MySQL, PostgreSQL and SQL Server agree exactly with each other and SQLite places a few columns elsewhere. Every query names its columns, so nothing reads positionally.

String comparison semantics are the largest difference of all and do not show up in a column diff at all. They are in #283.

---

### Schema standardization

**#282** the five structural divergences, one migration each.

**#283** string comparison semantics, the one that changes behaviour rather than just shape.

**#284** the parity test that stops all of it drifting apart again.

leodip:
Closes #282. Also strikes section 2 of #259, which is the same defect and is folded in here.

Goiabada's four migration sets are written by hand, four times, and nothing compares the result.
Diffing four fully migrated catalogs turns up five structural divergences. This change closes all
five, so the four engines build the same schema with two named and reasoned exceptions, and it
removes a live 500 that one of them causes on SQLite.

**The rule this establishes:** a migration is not finished until all four engines end up in the same
shape, and where an engine physically cannot reach that shape, the exception is named in the
migration file that creates it rather than discovered later by a diff.

## The five divergences

| | What diverges | Reaching parity |
|---|---|---|
| D1 | `refresh_tokens(client_id)` has no index on SQLite | additive `CREATE INDEX`, SQLite only |
| D2 | `audit_logs.details` has no `'{}'` default on MySQL | `MODIFY`, MySQL only. Parity on the default's effect, not on its recorded text: MySQL can only store a TEXT default as the expression `(_utf8mb4'{}')` where the others store `'{}'`. First named exception |
| D3 | `codes.code_challenge` and `code_challenge_method` are still `NOT NULL` on SQLite | relax SQLite, via a table rebuild |
| D4 | five columns are non-Unicode `varchar` on SQL Server | `ALTER COLUMN`, SQL Server only |
| D5 | `refresh_tokens` does not cascade on SQL Server | only downward. SQL Server refuses the CASCADE (`Msg 1785`, multiple cascade paths through `users <- codes <- refresh_tokens`), so the other three come down to `NO ACTION`. Second named exception: the invariant lives in Go on all four rather than in the schema on three |

**D3 is a live defect, not a latent one, and the issue is wrong about it.** #282 says nothing ever
writes NULL there and recommends restoring `NOT NULL` on all four. `models.Code` declares both
columns as `sql.NullString` and `oauth.CreateAuthCode` leaves them at `Valid:false` when the request
carried no challenge, so the insert does send NULL. On SQLite, the default engine, a confidential
client configured for PKCE-optional runs the whole ceremony and gets a 500 at `/auth/issue`. Six
integration tests skip on SQLite because of it. RFC 9700 section 2.1.1 makes PKCE RECOMMENDED rather
than required for confidential clients, so PKCE-optional is a configuration worth keeping, and
relaxing SQLite is the only direction that reaches parity without breaking it.

**D5 levels down, and that is unobservable.** The only two statements in the tree that delete a user
or a client row live inside `DeleteUser` and `DeleteClient`, and both clear that row's refresh tokens
first. That is why SQL Server has run without the cascade since migration 000011 without anyone
noticing.

### Stage 1: the shape dumper, and proving it reads each engine's catalog
`9a641d46`

`dumpTable(t, h, table)` in `src/authserver/tests/data/migration_testdb_helper.go` reads a table's
columns, indexes and foreign keys out of the configured dialect's catalog into a comparable
structure, so stage 3's hand-written SQLite rebuild can be held to a before-and-after comparison
instead of to an expectation somebody typed. That matters because a dropped column, a lost index or
a changed foreign key action are all silent: `PRAGMA foreign_key_check` sees none of them.

It compares a table against itself on one engine, never across engines, so it records each engine's
own spelling verbatim rather than mapping `TEXT`, `varchar(256)` and `nvarchar(256)` onto a common
vocabulary. Foreign keys are keyed by (local column, referenced table, referenced column, on-delete)
and not by constraint name, because SQLite's `pragma_foreign_key_list` omits the name even when the
table declares `CONSTRAINT fk_named`.

`TestDumpTable_ReadsTheCatalog` is the control: a helper that silently returns nothing would make a
before-and-after comparison read as "nothing changed" and pass. It runs against an isolated database
pinned at 000035 rather than at head, so it keeps asserting today's shape after this change alters
it, and it pins nullability and uniqueness in both polarities, the per-dialect type and default of
`auth_state_generation`, and the two divergences themselves: `codes.code_challenge` NOT NULL on
sqlite alone, `refresh_tokens.user_id` NO ACTION on mssql alone.

Seven mutations against the sqlite branches, all caught. Tiers: unit 4144 pass; data 1421 pass on
mysql, postgres, mssql and sqlite. Integration not applicable, no production code and no schema
change.

One departure from the plan, found by running it: the non-unique index control is
`idx_refresh_tokens_user_id` and not `idx_refresh_tokens_code_id`, because InnoDB indexes every
foreign key under the constraint's own name, so MySQL has no index by the latter name.

### Stage 2: the three single-engine additive fixes (D1, D2, D4)
`5b9a4e92`

Three migrations, one per engine that is out of step, no shared file between them.

**000036, sqlite:** `CREATE INDEX idx_refresh_tokens_client_id ON refresh_tokens(client_id)`. D1.

**000037, mysql:** ``ALTER TABLE `audit_logs` MODIFY `details` TEXT NOT NULL DEFAULT ('{}')``. D2.
`SET DEFAULT` fails `ERROR 1101` on a TEXT column in both spellings, so `MODIFY` with a
parenthesised expression is the only form MySQL accepts, and the result is metadata-only:
accepted with both `ALGORITHM=INSTANT` and `ALGORITHM=INPLACE`, so what is typically the largest
table in a deployment is not rebuilt.

**000038, mssql:** the five `VARCHAR` columns become `NVARCHAR`. D4. Every `ALTER COLUMN` restates
its nullability, because SQL Server makes a column nullable when the keyword is omitted whatever it
was before, and three of the five are `NOT NULL`; written without it this would have traded one
divergence for another that nothing in the tree would notice.
`clients.include_open_id_connect_claims_in_id_token` carries a default constraint whose
auto-generated name differs per database, so the name is read from `sys.default_constraints` at run
time and the replacement is named `df_clients_include_open_id_connect_claims_in_id_token`, so a
later migration can drop it without repeating the lookup.

D4's severity is narrower than #282 states. `NewMsSQLDatabase` pins
`COLLATE Latin1_General_100_CI_AI_SC_UTF8`, and under a UTF-8 collation `VARCHAR` stores Unicode
losslessly, so a database Goiabada created was never losing anything. The creation is
`IF NOT EXISTS`, so this protects a database an operator pre-created under the server's default
collation, and it repairs nothing already transcoded: a value stored as `café-é?` stays that way.

One migration test each, all four engines. Each takes two shapes, because each file exists on one
engine and `Migrate` refuses a target version the engine's source does not carry: the engine with
the file does the full 000035 → N → 000035 → N round trip, and the other three assert the state
they already have at 000035. 000036's checks the index's key columns and non-uniqueness rather than
its name, with `idx_refresh_token_jti` as the uniqueness-polarity control. 000037's asserts the
default's **effect**, inserting a row that omits `details` and reading back `{}`, and never asserts
its recorded text, which would encode a driver's charset setting rather than a schema fact. 000038's
compares before-and-after catalog dumps of all four affected tables and requires the only difference
anywhere to be the five columns' type, which catches a column dropped in passing even when nobody
thought to look for it.

Three `schema.sql` snapshots updated; `postgresdb/schema.sql` is untouched, since no divergence here
is PostgreSQL's. Seven mutations against the three migrations, all caught. Tiers: unit 4334 pass;
data 1886 pass on mysql, postgres, mssql and sqlite; integration 1325 pass on sqlite, which is what
proves 000036 applies on the real startup path rather than only under the test harness.

One departure from the plan, presentational: MySQL's snapshot carries its note about the expression
default above the `CREATE TABLE` rather than on the column line, because every other comment in that
file is a top-level one.

### Stage 3: the shared rebuild (D3 and D5)
`6c6f0416`

One migration, `000039`, on sqlite, mysql and postgres. D3 and D5 land together because both need
the same SQLite table rebuild, and doing them in one pass costs less than doing either alone. SQL
Server gets no file: it has been `ON DELETE NO ACTION` since its own 000011 and nullable on both
PKCE columns since its own 000007, so it is already where the other three are going.

**D3 is a live defect, and #282 has it backwards.** The issue says nothing reaches the column
because `models.Code` holds the challenge as a plain `string`; it holds `sql.NullString`, and
`CodeIssuer.CreateAuthCode` leaves it at `Valid:false` when the authorization request carried no
challenge, so the insert sends NULL. On SQLite, the default engine, a confidential client
configured for PKCE-optional therefore ran the whole ceremony and got a 500 at `/auth/issue`.
Restoring `NOT NULL` on all four, which the issue recommends, would have taken a one-engine failure
to four. RFC 9700 2.1.1 makes PKCE for confidential clients RECOMMENDED rather than required, so
PKCE-optional is a configuration worth keeping, and SQLite is relaxed to meet the other three.

**D5 can only be levelled down.** SQL Server refuses the cascade outright: `users <- codes <-
refresh_tokens(code_id)` alongside `users <- refresh_tokens(user_id)` is a multiple cascade path
(`Msg 1785`), reproduced on the real table shape. So `refresh_tokens.user_id` and `client_id` come
down to `NO ACTION` on the other three. It is unobservable: `DeleteUser` and `DeleteClient` are the
only statements in the codebase that delete a user or a client row, and both clear that row's
refresh tokens first, which is why SQL Server has run this way since 000011 with nobody noticing.
The invariant now lives in Go on all four engines rather than in the schema on three, which is a
named exception in this change's goal rather than an oversight.

**The SQLite ordering is not the obvious one.** golang-migrate runs each migration inside a
transaction and `PRAGMA foreign_keys` cannot be changed inside one, so `DROP TABLE codes` fires an
implicit `DELETE` that cascades into `refresh_tokens(code_id)` and destroys every refresh token
issued through the authorization code flow. Nothing reports it: `PRAGMA foreign_key_check` is clean
afterwards, so in a deployment that is every `offline_access` grant revoked with no error.
`PRAGMA defer_foreign_keys`, `PRAGMA legacy_alter_table` and `PRAGMA foreign_keys = OFF` were each
measured and none prevents it. What works is parking `refresh_tokens` in a constraint-free
`CREATE TABLE AS` copy and dropping it *before* `codes` is touched, then rebuilding it from the
parked copy with the foreign key actions wanted. Both the up and the down do that, and the down
converts NULL to `''` before restoring `NOT NULL`, which is lossless because every reader in
`ValidateTokenRequest`'s PKCE downgrade mitigation already treats the two alike.

On mysql and postgres the swap is two `DROP CONSTRAINT` / `ADD CONSTRAINT` pairs. Measured
beforehand that both explicit indexes survive it, which matters because InnoDB indexes every
foreign key and losing `idx_refresh_tokens_client_id` here would have undone stage 2 on MySQL
inside the same change.

Four migration tests, all four engines. The first is the one the rebuild's risk actually needs: it
reads both tables out of the catalog before the migration, builds the expected after-shape by
applying **only** the intended edits to it, and requires the real after-shape to equal that. Stated
as a difference rather than a checklist, it fails on a dropped column, a lost index, a changed type
spelling or a third foreign key action moving, including ones nobody thought to check. The second
covers what a shape dump cannot see, two preserved columns of the same type whose values were
swapped by the hand-written refill: it seeds through the repository's own `CreateCode` and
`CreateRefreshToken` and compares whole structs across the apply, the down and the re-apply. The
third is the challenge-less insert that used to fail, at the same boundary the diagnosis used. The
fourth is `NO ACTION` in effect rather than in the catalog: a raw `DELETE FROM users` is refused
with a ROPC-shaped token present, while `DeleteUser` still succeeds, which is the invariant above
asserted directly.

Also here: the false comment in `000007_make_pkce_nullable.up.sql`, which claimed SQLite does not
enforce `NOT NULL` (it does, on every column type), and three `schema.sql` snapshots. Six mutations
of the migration SQL, five caught by the case meant to catch them. The sixth is the value swap run
against the shape test alone, which survives on purpose: that is the argument for keeping both
tests, recorded so the row-value case is not later read as redundant.

### Stage 4: the three integration tests stop skipping on SQLite

Commit `48b4fb80`. Deletes `requireChallengelessCodesAreStorable` from
`src/authserver/tests/integration/utils_test.go`, along with its doc comment describing the SQLite
defect, and its three call sites. No production file moves.

The guard existed only because `codes.code_challenge` was NOT NULL on SQLite; stage 3's 000039
rebuild removed that, so it had nothing left to guard.

**Three was the number of call sites, not tests.** Two of the three sit inside fixture helpers
(`challengelessRefreshToken`, `challengelessCode`) rather than inside a test, so **six** tests were
skipping on SQLite and now run there:

| Was gated by | Tests |
|---|---|
| `challengelessRefreshToken` | `TestToken_Refresh_ChallengelessGrant_RefusedOnceTheClientIsPublic`, `TestToken_Refresh_ChallengelessGrant_ConfidentialClientStillRefreshes` |
| `challengelessCode` | `TestToken_AuthCode_ChallengelessCode_RefusedAfterClientBecomesPublic`, `TestToken_AuthCode_ChallengelessCode_ConfidentialClientStillSucceeds`, `TestToken_AuthCode_ChallengelessCode_SurvivesTurningPKCEOnForAConfidentialClient` |
| itself | `TestAuthorize_ConfidentialClient_PKCEOff_WithoutCodeChallenge_Succeeds` |

Tiers: unit 1246 pass; data 1446 pass on all four engines; **integration on sqlite 1010 pass with
zero skips**, where six rows reported `--- SKIP` before. mysql, postgres and mssql are the check
suite's for `48b4fb80`.

The stage adds no test, so the mutation targets what the six cover instead: setting
`authorize_validator.go`'s `if input.PKCERequired` to `if true` makes PKCE unconditional, and **all
six fail**. They really do drive a challenge-less authorization to completion on SQLite rather than
passing vacuously on a newly available engine.

## Tests

Every stage ran unit and data locally on all four engines, and integration locally on sqlite only.
Those per-stage numbers are in each stage's entry above. **The four-engine integration result exists
nowhere but the check suite**, so no local run in this branch ever produced it and none is reported
as having done so.

**The check suite is green on the final commit `48b4fb80`**, run
[33130052885](https://github.com/leodip/goiabada/actions/runs/33130052885): `Tests / sqlite`,
`Tests / mysql`, `Tests / postgres` and `Tests / mssql`, each running the data and integration tiers
on that engine, plus the three unit jobs and `Build validation`, `Lint`, `Vulnerabilities` and
`Versions`, none of which any local tier runs.

Last local run, at stage 4: unit 1246 pass; data 1446 pass on mysql, postgres, mssql and sqlite;
integration on sqlite 1010 pass **with zero skips**, where six rows reported `--- SKIP` before this
branch.

**26 mutations, 25 caught.** 21 across the four stages and 5 more run independently by the reviewer
at the final gate. The one survivor is deliberate: stage 3's value swap, run against the shape test
alone, survives on purpose to show why the row-value test is not redundant, and the same swap is
caught by the row-value test. The reviewer's five closed the last gap in that argument, so all four
of the SQLite rebuild's copy blocks are now exercised by a test that fails when their column
mappings are wrong.

### Review

Reviewed by `gpt-5.6-sol` at effort `xhigh`, 1 round over the whole branch. No stage was reviewed on
its own; the plan was reviewed separately before any code existed.

|---|---|---|---|---|
| plan | 1 | 4 | 0 | all fixed in `plan.md`, no code existed yet |
| final, whole branch | 1 | 0 | 0 | clean, `static-review-exhausted`, zero `unchecked` |

- **significant, quality:** stage 1's control required 21 `codes` columns; the table has 22. Fixed in the plan: `auth_state_generation` and `revoked` are the two it omitted.
- **significant, conformance:** stage 1 required exact foreign-key constraint names on all four engines, and SQLite cannot report one. Fixed in the plan: foreign keys are keyed by local column, referenced table, referenced column and delete action instead.
- **significant, quality:** the control asserted no column type or default and had no nullable-polarity arm on SQLite, so a stub returning nothing would have passed. Fixed in the plan: both polarities on every engine, per-dialect type and default, and five mutations instead of one.
- **significant, quality:** the tests were anchored to migration versions that do not exist on every engine, and `golang-migrate` fails outright on an absent target. Fixed in the plan: pinned at 000035, the last version all four share.

The final gate found nothing, and one `unchecked` entry at the plan gate was a real gap that changed
the plan: a shape dump cannot see two preserved columns whose *values* were swapped by the
hand-written refill. The row-value test that answered it went on to catch three of the reviewer's
five independent mutations.

Full account, with the evidence and the coverage a clean verdict rests on, in
[this comment](https://github.com/leodip/goiabada/pull/286#issuecomment-5447058635).

## For the release notes

**Deleting a `users` or `clients` row by hand-written SQL now fails if refresh tokens reference it,
on SQLite, MySQL and PostgreSQL.** Those three cascaded before; SQL Server has behaved this way
since its own migration 000011. Goiabada itself is unaffected, because `DeleteUser` and
`DeleteClient` are the only code paths that delete such a row and both clear the row's refresh
tokens first. If you have your own cleanup or GDPR-erasure SQL that deletes these rows directly,
delete from `refresh_tokens` first or it will now stop with a foreign key violation instead of
silently taking the tokens with it.

**A confidential client configured for PKCE-optional now works on SQLite.** It previously ran the
whole authorization ceremony and returned a 500 at `/auth/issue`. No action; the fix is the
migration.

**SQL Server: five columns become `nvarchar`, and this does not repair data already damaged.** A
database Goiabada created was never at risk, because it pins a UTF-8 collation. If you pre-created
your database under the server's default collation, this migration stops further loss but cannot
recover a value that was already transcoded: a name stored as `caf?-?` stays that way. Check those
five columns for replacement characters if that describes your deployment.

**Take a backup before upgrading.** On SQLite, migration 000039 rebuilds `codes` and
`refresh_tokens` rather than altering them in place, which is the only way SQLite can relax a `NOT
NULL`. The down migration is written and tested in both directions, and it converts `NULL` PKCE
values back to `''`, which every existing reader already treats alike.

## The run's record

Three comments on this PR carry what the diff cannot:

- [Deferred decisions](https://github.com/leodip/goiabada/pull/286#issuecomment-5447050962): there are none, and all six decisions are listed with the two named exceptions worth revisiting.
- [Follow-ups](https://github.com/leodip/goiabada/pull/286#issuecomment-5447053565): none filed, and **#259 needs your edit** to strike its section 2, which this change closes with the opposite fix to the one it recommends.
- [The review](https://github.com/leodip/goiabada/pull/286#issuecomment-5447058635): both gates in full, with the mutation evidence.

**Deviations from the plan, all four minor and all recorded above** with the stage that hit them:
stage 1's non-unique index control had to change name, because InnoDB indexes every foreign key
under the constraint's own name; stage 2 put MySQL's snapshot comment above the `CREATE TABLE`
rather than on the column line, to match every other comment in that file; and the sealed agreement
carried two counting errors, `codes` at 21 columns rather than 22 and a dump anchored at a migration
version SQL Server does not have, both caught by the plan review before any code was written and
both fixed in the plan rather than by editing sealed text.

One correction the record should carry: this change was written against "three integration tests
skip on SQLite". Three was the number of *call sites*, and two of them sit inside fixture helpers,
so **six** tests were skipping and six now run. The goal is met more widely than it was written.

## Status

**Ready for review.** All four stages landed, the whole branch was reviewed once at the end and the
reviewer found nothing, and the check suite is green on the final commit `48b4fb80` across all four
engines. All five divergences reported in #282 are closed, and the live 500 that divergence 3 caused
on the default engine is gone.

Migrations 000036 to 000039 are reserved by this branch; #249, #259 and #268 are all open and all
want a migration number, so whichever lands second renumbers. Note that the long-lived
`goiabada_data` and `goiabada_integration` test databases make a reused migration number fail
silently, so the run that renumbers has to drop them.

### DIFF
--- a/src/authserver/tests/data/migration_000036_refresh_tokens_client_id_test.go
+++ b/src/authserver/tests/data/migration_000036_refresh_tokens_client_id_test.go
@@ -1,0 +1,91 @@
+package datatests
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// refreshTokensClientIdIndex000036 is the index SQLite was missing, and the name the
+// other three engines have carried since their own 000011.
+const refreshTokensClientIdIndex000036 = "idx_refresh_tokens_client_id"
+
+// TestMigration000036_RefreshTokensClientIdIndex is goal 1 of #282 stated as a test:
+// refresh_tokens(client_id) is indexed on all four engines. It runs against an ISOLATED
+// database of the configured dialect (see migration_testdb_helper.go).
+//
+// The migration file exists on SQLite alone, because the other three already have the
+// index, so this test takes two shapes. On SQLite it is the usual migration shape: absent
+// at 000035, present at 000036, gone again after the down, back after a re-apply. On the
+// other three it asserts the state at 000035 and never calls Migrate past it, since
+// Migrate refuses a target version the engine's source does not carry.
+//
+// Two assertions are worth explaining:
+//
+//  1. The control. idx_refresh_token_jti is UNIQUE on all four engines, so reading it as
+//     unique here is what makes "the client_id index is non-unique" mean something: each
+//     catalog reports uniqueness differently and MySQL reports it inverted, so a
+//     non-unique reading would otherwise pass on an engine whose flag was being read
+//     backwards. This is the polarity guard migration_000030 established, run in the
+//     opposite direction.
+//  2. The key columns, not just the name. An index built on the wrong column would pass a
+//     name check while doing nothing for the scan this exists to remove.
+//
+// Stage 3's rebuild of refresh_tokens has to recreate this index; that is what makes this
+// test worth keeping rather than a one-off check of the migration.
+//
+// Run per dialect via: ./run-tests.sh --type data --db <sqlite|mysql|postgres|mssql>
+//
+//	--run TestMigration000036_RefreshTokensClientIdIndex
+func TestMigration000036_RefreshTokensClientIdIndex(t *testing.T) {
+	h := newIsolatedDB(t)
+
+	require.NoError(t, h.Migrator.Migrate(35), "migrate to 000035")
+
+	// 1. Control.
+	control := describeIndex(t, h, "refresh_tokens", "idx_refresh_token_jti")
+	require.True(t, control.Exists, "control index idx_refresh_token_jti must exist at 000035")
+	require.True(t, control.Unique,
+		"control index idx_refresh_token_jti is UNIQUE on every engine; reading it as non-unique means the uniqueness flag is being read backwards")
+
+	if !isSQLite000036() {
+		assertRefreshTokensClientIdIndex000036(t, h, "at 000035")
+		return
+	}
+
+	require.False(t, describeIndex(t, h, "refresh_tokens", refreshTokensClientIdIndex000036).Exists,
+		"%s must not exist on SQLite at 000035: that absence is the divergence #282 reports",
+		refreshTokensClientIdIndex000036)
+
+	require.NoError(t, h.Migrator.Migrate(36), "apply 000036")
+	assertRefreshTokensClientIdIndex000036(t, h, "after apply")
+
+	require.NoError(t, h.Migrator.Migrate(35), "roll back 000036")
+	assert.False(t, describeIndex(t, h, "refresh_tokens", refreshTokensClientIdIndex000036).Exists,
+		"%s must be gone after rolling back to 000035", refreshTokensClientIdIndex000036)
+
+	require.NoError(t, h.Migrator.Migrate(36), "re-apply 000036")
+	assertRefreshTokensClientIdIndex000036(t, h, "after down/up round trip")
+}
+
+// assertRefreshTokensClientIdIndex000036 checks the index's shape in the catalog: present,
+// covering exactly client_id, and not unique. Not unique matters as much as present, since
+// a client holds many refresh tokens and a unique index here would refuse the second one.
+func assertRefreshTokensClientIdIndex000036(t *testing.T, h *isolatedDB, phase string) {
+	t.Helper()
+
+	shape := describeIndex(t, h, "refresh_tokens", refreshTokensClientIdIndex000036)
+
+	require.Truef(t, shape.Exists, "[%s] %s is missing on %s",
+		phase, refreshTokensClientIdIndex000036, dbType())
+	assert.Equalf(t, []string{"client_id"}, shape.Columns,
+		"[%s] %s must cover exactly client_id", phase, refreshTokensClientIdIndex000036)
+	assert.Falsef(t, shape.Unique,
+		"[%s] %s must NOT be unique: a client holds many refresh tokens",
+		phase, refreshTokensClientIdIndex000036)
+}
+
+func isSQLite000036() bool {
+	return dbType() == "" || dbType() == "sqlite"
+}
--- a/src/authserver/tests/data/migration_000037_audit_logs_details_default_test.go
+++ b/src/authserver/tests/data/migration_000037_audit_logs_details_default_test.go
@@ -1,0 +1,113 @@
+package datatests
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// TestMigration000037_AuditLogsDetailsDefault is goal 2 of #282 stated as a test:
+// audit_logs.details defaults to '{}' on all four engines. It runs against an ISOLATED
+// database of the configured dialect (see migration_testdb_helper.go).
+//
+// The migration file exists on MySQL alone, which was the engine without the default, so
+// this test takes two shapes: MySQL goes 000035 to 000037 and back, and the other three
+// assert the state they already have at 000035 without calling Migrate past it, since
+// Migrate refuses a target version the engine's source does not carry.
+//
+// What is asserted, and what deliberately is not:
+//
+//   - The default's EFFECT: insert a row naming every column except details, read details
+//     back, expect {}. §5 of the agreement rejects asserting the recorded text, because
+//     MySQL can only carry this default as an expression and stamps it with the DDL
+//     connection's charset, so information_schema reports _utf8mb4'{}' where the other
+//     three report '{}'. An assertion on the text would encode a driver setting rather
+//     than a schema fact.
+//   - The before state as "omission cannot yield {}", capturing both the error and the
+//     value rather than choosing between them. Under STRICT_TRANS_TABLES, which the dev
+//     container's MySQL sets globally and the Go DSN does not override, the insert is
+//     refused outright with ERROR 1364; a non-strict server would store an empty string.
+//     Either way it is not {}, and that is the whole claim.
+//   - The catalog's Default is checked only for EMPTINESS on MySQL, which is the one thing
+//     about it that is engine-independent. That is what distinguishes "the migration added
+//     a default" from "some other write happened to put {} in the column".
+//
+// Run per dialect via: ./run-tests.sh --type data --db <sqlite|mysql|postgres|mssql>
+//
+//	--run TestMigration000037_AuditLogsDetailsDefault
+func TestMigration000037_AuditLogsDetailsDefault(t *testing.T) {
+	h := newIsolatedDB(t)
+
+	require.NoError(t, h.Migrator.Migrate(35), "migrate to 000035")
+
+	if dbType() != "mysql" {
+		details, err := insertAuditLogWithoutDetails000037(t, h, "already-defaulted")
+		require.NoErrorf(t, err, "on %s, audit_logs.details already carries a default at 000035", dbType())
+		assert.Equalf(t, "{}", details,
+			"on %s, omitting details at 000035 must already yield {}", dbType())
+		return
+	}
+
+	// Before: no default in the catalog, and omission cannot produce {}.
+	assert.Empty(t, dumpTable(t, h, "audit_logs").column(t, "details").Default,
+		"MySQL's audit_logs.details must carry no default at 000035: that absence is the divergence #282 reports")
+	assertDetailsOmissionCannotYieldEmptyObject000037(t, h, "at 000035", "before-migration")
+
+	require.NoError(t, h.Migrator.Migrate(37), "apply 000037")
+
+	assert.NotEmpty(t, dumpTable(t, h, "audit_logs").column(t, "details").Default,
+		"audit_logs.details must carry a default after 000037; its recorded TEXT is deliberately not asserted")
+	assertDetailsOmissionYieldsEmptyObject000037(t, h, "after apply", "after-migration")
+
+	require.NoError(t, h.Migrator.Migrate(35), "roll back 000037")
+	assert.Empty(t, dumpTable(t, h, "audit_logs").column(t, "details").Default,
+		"the default must be gone after rolling back to 000035")
+	assertDetailsOmissionCannotYieldEmptyObject000037(t, h, "after roll back", "after-down")
+
+	require.NoError(t, h.Migrator.Migrate(37), "re-apply 000037")
+	assertDetailsOmissionYieldsEmptyObject000037(t, h, "after down/up round trip", "after-reapply")
+}
+
+func assertDetailsOmissionYieldsEmptyObject000037(t *testing.T, h *isolatedDB, phase, event string) {
+	t.Helper()
+
+	details, err := insertAuditLogWithoutDetails000037(t, h, event)
+	require.NoErrorf(t, err, "[%s] an insert omitting details must be accepted once the default exists", phase)
+	assert.Equalf(t, "{}", details, "[%s] omitting details must yield {}", phase)
+}
+
+func assertDetailsOmissionCannotYieldEmptyObject000037(t *testing.T, h *isolatedDB, phase, event string) {
+	t.Helper()
+
+	details, err := insertAuditLogWithoutDetails000037(t, h, event)
+	if err != nil {
+		return // refused, which is what a strict server does with a NOT NULL column that has no default
+	}
+	assert.NotEqualf(t, "{}", details,
+		"[%s] with no default declared, omitting details must not yield {}; got it from a non-strict server storing ''", phase)
+}
+
+// insertAuditLogWithoutDetails000037 inserts an audit_logs row naming every column except
+// details and reads details back. Literals rather than placeholders because the dialects
+// disagree on placeholder syntax and every value here is test-controlled, following
+// seedKeyPair000030. The timestamp literal parses on all four engines.
+//
+// The insert's error is returned rather than asserted, because whether omission is refused
+// or silently stored is exactly what the caller is deciding about.
+func insertAuditLogWithoutDetails000037(t *testing.T, h *isolatedDB, event string) (string, error) {
+	t.Helper()
+
+	_, err := h.SQL.Exec(fmt.Sprintf(
+		`INSERT INTO audit_logs (created_at, audit_event) VALUES ('2026-01-01 00:00:00', '%s')`, event))
+	if err != nil {
+		return "", err
+	}
+
+	var details string
+	require.NoErrorf(t, h.SQL.QueryRow(fmt.Sprintf(
+		`SELECT details FROM audit_logs WHERE audit_event = '%s'`, event)).Scan(&details),
+		"read back details for audit event %s", event)
+	return details, nil
+}
--- a/src/authserver/tests/data/migration_000038_nvarchar_columns_test.go
+++ b/src/authserver/tests/data/migration_000038_nvarchar_columns_test.go
@@ -1,0 +1,212 @@
+package datatests
+
+import (
+	"fmt"
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// oidcClaimsDefaultConstraint000038 is the name the migration gives the default constraint
+// it has to drop and recreate, lowercase df_<table>_<column> like the nine named default
+// constraints mssqldb already carries.
+const oidcClaimsDefaultConstraint000038 = "df_clients_include_open_id_connect_claims_in_id_token"
+
+// nvarcharColumns000038 is the exact set of columns the migration retypes: table, column,
+// the type the catalog reports before, the type it must report after, and whether the
+// column is nullable. The nullability is recorded here because it must be UNCHANGED, which
+// is the trap this migration is written around: ALTER COLUMN with no NULL or NOT NULL
+// keyword makes the column nullable whatever it was.
+var nvarcharColumns000038 = []struct {
+	table    string
+	column   string
+	before   string
+	after    string
+	nullable bool
+}{
+	{"codes", "code_challenge", "varchar(256)", "nvarchar(256)", true},
+	{"codes", "code_challenge_method", "varchar(10)", "nvarchar(10)", true},
+	{"user_profile_pictures", "content_type", "varchar(64)", "nvarchar(64)", false},
+	{"client_logos", "content_type", "varchar(64)", "nvarchar(64)", false},
+	{"clients", "include_open_id_connect_claims_in_id_token", "varchar(10)", "nvarchar(10)", false},
+}
+
+// TestMigration000038_NvarcharColumns is goal 5 of #282 stated as a test: the five SQL
+// Server columns declared VARCHAR become NVARCHAR, and the auto-named default constraint
+// on clients.include_open_id_connect_claims_in_id_token is replaced by a named one. It
+// runs against an ISOLATED database of the configured dialect (see
+// migration_testdb_helper.go).
+//
+// SQL Server only. The other three engines have no non-Unicode string type to migrate away
+// from: SQLite has one TEXT type, MySQL's tables are utf8mb4 throughout, and PostgreSQL's
+// text is Unicode by definition. The migration file exists on mssql alone and Migrate
+// refuses a target version the engine's source does not carry.
+//
+// The assertions, and why each is here:
+//
+//  1. The five columns' type changes and their NULLABILITY DOES NOT. Three of the five are
+//     NOT NULL, and ALTER COLUMN written without the keyword would silently make them
+//     nullable, trading one divergence for another that nothing else in the tree would
+//     notice.
+//  2. Everything else about the four affected tables is byte-identical, computed as a
+//     difference over stage 1's dumps rather than as a hand-written list of what should
+//     have stayed. That catches a column the migration dropped even when the test author
+//     forgot the column existed, which is the property decision 5 exists for, applied here
+//     to an ALTER rather than to a rebuild.
+//  3. The seeded row's value survives. A shape dump cannot tell ALTER COLUMN apart from
+//     DROP COLUMN followed by ADD COLUMN, and the second would silently blank the column
+//     for every existing client.
+//  4. The default constraint's NAME: auto-generated before (SQL Server's DF__ prefix),
+//     the name above after. dumpTable deliberately carries no constraint name, and the
+//     name is the entire reason a later migration can drop this constraint without
+//     repeating the catalog lookup this one has to do.
+//
+// Run via: ./run-tests.sh --type data --db mssql --run TestMigration000038_NvarcharColumns
+func TestMigration000038_NvarcharColumns(t *testing.T) {
+	if dbType() != "mssql" {
+		t.Skipf("SQL Server only: %s has no non-Unicode string type to migrate away from, so it has no 000038 file", dbType())
+	}
+
+	h := newIsolatedDB(t)
+
+	require.NoError(t, h.Migrator.Migrate(35), "migrate to 000035")
+
+	clientId := seedClient000035(t, h, "mig38-client")
+	setOidcClaimsSetting000038(t, h, clientId, "on")
+
+	before := dumpAffectedTables000038(t, h)
+	assertNvarcharColumnTypes000038(t, h, before, "at 000035", func(c int) string { return nvarcharColumns000038[c].before })
+	assert.Truef(t, strings.HasPrefix(oidcClaimsConstraintName000038(t, h), "DF__"),
+		"at 000035 the default constraint must still be the auto-generated one 000013 left; got %q",
+		oidcClaimsConstraintName000038(t, h))
+
+	require.NoError(t, h.Migrator.Migrate(38), "apply 000038")
+
+	after := dumpAffectedTables000038(t, h)
+	assertNvarcharColumnTypes000038(t, h, after, "after apply", func(c int) string { return nvarcharColumns000038[c].after })
+	assertOnlyTheFiveTypesMoved000038(t, before, after, "after apply")
+	assert.Equalf(t, "on", readOidcClaimsSetting000038(t, h, clientId),
+		"after apply, the seeded client's setting must survive the retype")
+	assert.Equal(t, oidcClaimsDefaultConstraint000038, oidcClaimsConstraintName000038(t, h),
+		"after apply, the default constraint must carry the name a later migration can drop it by")
+
+	require.NoError(t, h.Migrator.Migrate(35), "roll back 000038")
+
+	down := dumpAffectedTables000038(t, h)
+	assertNvarcharColumnTypes000038(t, h, down, "after roll back", func(c int) string { return nvarcharColumns000038[c].before })
+	assertOnlyTheFiveTypesMoved000038(t, before, down, "after roll back")
+	assert.Equalf(t, "on", readOidcClaimsSetting000038(t, h, clientId),
+		"after roll back, the seeded client's setting must survive the retype")
+	assert.Truef(t, strings.HasPrefix(oidcClaimsConstraintName000038(t, h), "DF__"),
+		"the down migration must restore an UNNAMED default, so 000035 is the shape 000013 left")
+
+	require.NoError(t, h.Migrator.Migrate(38), "re-apply 000038")
+
+	reapplied := dumpAffectedTables000038(t, h)
+	assertNvarcharColumnTypes000038(t, h, reapplied, "after down/up round trip", func(c int) string { return nvarcharColumns000038[c].after })
+	assertOnlyTheFiveTypesMoved000038(t, before, reapplied, "after down/up round trip")
+	assert.Equal(t, oidcClaimsDefaultConstraint000038, oidcClaimsConstraintName000038(t, h),
+		"after the round trip, the default constraint must be named again")
+}
+
+// affectedTables000038 is the four tables the migration touches, in a fixed order so the
+// before and after dumps line up.
+var affectedTables000038 = []string{"client_logos", "clients", "codes", "user_profile_pictures"}
+
+func dumpAffectedTables000038(t *testing.T, h *isolatedDB) map[string]tableShape {
+	t.Helper()
+
+	shapes := map[string]tableShape{}
+	for _, table := range affectedTables000038 {
+		shapes[table] = dumpTable(t, h, table)
+	}
+	return shapes
+}
+
+// assertNvarcharColumnTypes000038 checks each of the five columns against the type want
+// returns for it, and checks its nullability against the fixed value in the table, which
+// never changes in either direction.
+func assertNvarcharColumnTypes000038(t *testing.T, h *isolatedDB, shapes map[string]tableShape, phase string, want func(int) string) {
+	t.Helper()
+
+	for i, c := range nvarcharColumns000038 {
+		got := shapes[c.table].column(t, c.column)
+		assert.Equalf(t, want(i), got.Type, "[%s] %s.%s", phase, c.table, c.column)
+		assert.Equalf(t, c.nullable, got.Nullable,
+			"[%s] %s.%s changed nullability: ALTER COLUMN written without an explicit NULL/NOT NULL keyword makes a column nullable",
+			phase, c.table, c.column)
+	}
+}
+
+// assertOnlyTheFiveTypesMoved000038 is the exhaustive half. It compares the before and
+// after dumps of all four tables and requires that the ONLY difference anywhere is the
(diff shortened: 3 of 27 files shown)


leodip:
## Deferred decisions: none

Nothing was deferred, and nothing is waiting on you. All six decisions in the agreement are
`Decided`, none is `Open`, and the run raised no new decision and escalated nothing to this thread
across four stages and two review gates. So there is no assumption in this branch that a later
reader has to accept on trust.

Recorded here so the list is legible without the agreement, which is gitignored and does not
survive the run. Five were settled with you at the interview; the sixth was raised by the spec
session and settled there.

| # | Question | Answer |
|---|---|---|
| 1 | Which of the five divergences does this change carry? | **All five.** A divergence left standing is the goal unmet. D3 and D5 share a SQLite rebuild, so taking one without the other buys nothing |
| 2 | D3: which direction, and does it belong to #282 or #259? | **Relax SQLite to nullable, here.** The only direction that reaches parity while keeping PKCE-optional working. #259 section 2 is struck |
| 3 | D5: level the other three down, or leave the schema and keep the test? | **Level down.** SQL Server refuses the CASCADE (`Msg 1785`), so parity is only reachable downward |
| 4 | Migration numbering | **Four contiguous numbers, 000036 to 000039**, with no file on an engine that needs no change, which is the established convention |
| 5 | How is the rebuild proved to reproduce the current table? | **A catalog shape dumper, before and after, over two tables, on all four engines.** Asserts the difference is exactly the intended set, so it catches a dropped column the test author also forgot |
| 6 | D2: is a default whose recorded text differs from the other three worth adding at all? | **Yes, via `MODIFY`.** Exact parity would mean rebuilding `audit_logs`, typically the largest table in a deployment, to remove a default nothing reads |

### The two things worth revisiting are exceptions, not deferrals

Both are decided and implemented, and both are in the goal by name rather than discovered later by a
diff. They are the only two places where "the four engines build the same schema" is qualified, so
they are what to look at if you want to reopen anything:

- **D2, MySQL's default is equal in effect, not in recorded text.** MySQL can only store a `TEXT`
  default as the expression `(_utf8mb4'{}')` where the other three store `'{}'`. Decision 6 above.
  The test asserts the default's effect, inserting a row that omits `details` and reading back `{}`,
  and deliberately never asserts the recorded text, which would encode a driver's charset setting
  rather than a schema fact.
- **D5, the cascade invariant now lives in Go on all four engines rather than in the schema on
  three.** Reversing this means reversing decision 3, and SQL Server cannot be part of that reversal
  at all. What holds the invariant is that `DeleteUser` and `DeleteClient` contain the only two
  statements in the tree that delete a user or client row, and both clear that row's refresh tokens
  first. A future deletion path that skips them would now fail loudly with a foreign key violation
  on all four engines instead of silently orphaning on three, which is the safer failure.

### One correction the agreement carries, for the record

Decision 5's parenthetical above says `codes` has 21 columns. It has 22: `auth_state_generation`
(000024) and `revoked` (000026) are the two it omits. The plan review caught this before any code
was written and the plan was built against 22, so nothing was implemented against the wrong number.
It is a counting error in the sealed text, not a design question, and it changes no decision.

leodip:
## Follow-ups: none filed, and one issue this change edits

**No follow-up issues were drafted, at any of the four stages or either review gate.** That is the
run's own finding, not an omission: the plan review's response returned empty `## Follow-ups` and
`## Bookkeeping` sections, and the final review of the whole branch returned zero findings, zero
`unchecked` entries and zero follow-ups. Nothing was discovered beside the work that belonged to a
different change and had to be left.

Tracker re-checked at 2026-08-28 00:58 UTC, since it can move while a run is working. The five
related issues are all still open and none has been overtaken: #249, #259, #268, #283, #284.

### One issue needs your edit: #259

Not a follow-up to file, a fold-in to show you, because the run changed another issue's scope and
you cannot object to that if you were never told.

**#259 section 2 is struck.** #259 is "a client-scoped revocation cannot reach ROPC refresh grants,
and PKCE-optional is broken on SQLite". Its section 2 is the same defect as divergence 3 here, and
it recommends the **opposite** fix: restoring `NOT NULL` on all four engines. That premise is
refuted, `models.Code` holds `sql.NullString` and `CreateAuthCode` leaves it invalid, so the insert
really does send NULL, and taking #259's advice would have turned a one-engine 500 into a
four-engine one. Decision 2 settled this at the interview, before any code existed.

The fix landed here in `6c6f0416` because it shares a SQLite table rebuild with divergence 5, and
doing the two separately would mean rebuilding the same two tables twice.

**What is left of #259 is its section 1**, the ROPC client-scoped revocation security defect, which
is untouched by this branch and explicitly out of scope in the goal. #259's own text says section 1
is the one to do if only one gets done, so what remains is the more important half. Someone should
edit #259 to strike section 2 and point it at this PR; the run does not edit issues.

### Two other things to know, neither of them work owed

- **Migrations 000036 to 000039 are reserved by this branch.** #249, #259 and #268 are all open and
  all want a migration number, so whichever of them lands second renumbers. This matters more than
  it looks: the long-lived `goiabada_data` and `goiabada_integration` databases make a reused
  migration number fail **silently**, so the run that renumbers has to drop them.
- **#284 is the parity test that would have caught all five of these**, and this change deliberately
  does not absorb it. What landed here is a catalog shape dumper scoped to two tables, which is
  #284's step 1 in miniature; #284 keeps generalising to all 25 tables, the golden files and the
  cross-engine allowlist. Its allowlist now has two entries waiting for it, D2's expression default
  and D5's Go-held invariant, both named in the goal.

leodip:
## The code review, in full

This is the run's complete review account, pasted rather than linked: the verdicts live in a
gitignored directory that the next round overwrites, so without this comment the fix commits would
have no record of what prompted them.

**Two gates, one round each, `gpt-5.6-sol` at effort `xhigh` in a fresh Codex session both times.**
Under `LEO_REVIEW_MODE=end` no stage was reviewed on its own, so the final gate is the only reading
the code ever got. The plan gate ran before any code existed.

|---|---|---|---|---|---|
| plan | 1 | 4 | 0 | 4 | all fixed in `plan.md`, no code existed yet |
| final, whole branch | 1 | 0 | 0 | 0 | clean; `static-review-exhausted`, zero `unchecked` |

---

## Gate 1: the plan, round 1

Verdict `static-review-exhausted`; `security`, `conformance` and `quality` all `reviewed`. Scope was
`plan.md` against the sealed contract, plus the probes, the migration helper and the precedent
migration test.

**What it executed rather than asserted:** `check-anchors.sh` (26 anchors, each resolving once);
`tree-hash.sh --split` before and after, reporting `production` unmoved; direct catalog queries
against all four engines for table cardinalities, column metadata and foreign-key metadata; RFC 9700
section 2.1.1 fetched from the IETF publisher; a scratch `golang-migrate` absent-target probe;
`probe/sqlite_rebuild_fk.sh` re-run, reproducing variants A to D silently deleting the code-linked
refresh token and variant E preserving both rows; and the repository-backed `CreateCode` probe,
reconfirming that `CodeChallenge.Valid=false` still failed on SQLite as diagnosed.

**Four findings, all `significant`, all `confidence: high`, all fixed.** Each carried a populated
`forced_answer`, which marks a finding the code settles rather than a choice for you, and each
mechanically restored something the agreement had already decided. None was refuted, deferred or
escalated. No production code existed, so every fix landed in `plan.md` and nothing was committed.

| # | Axis | Finding | Independent verification | Outcome |
|---|---|---|---|---|
| 1 | quality | Stage 1's control required 21 `codes` columns; the table has 22 | 22 in all four `schema.sql` snapshots; the extras are `auth_state_generation` (000024) and `revoked` (000026). `refresh_tokens` is 17 as planned | Fixed: step 4 requires 22 and names both |
| 2 | conformance | Stage 1 required exact `fk_refresh_tokens_*` constraint names on all four engines; SQLite cannot report one | SQLite 3.53.4: a table declared `CONSTRAINT fk_named FOREIGN KEY ...` gives `PRAGMA foreign_key_list` without the name. The agreement separately rejects DDL-string comparison, so no legal route existed | Fixed: name dropped; foreign keys keyed by local column, referenced table, referenced column and delete action |
| 3 | quality | The control asserted no column type or default and had no nullable-polarity arm on SQLite, so a branch returning a constant or an empty string would have passed | Verified at 000035 on all four engines: `created_at` nullable, `refresh_token_jti` `NOT NULL`, `auth_state_generation` carrying a real per-engine type and `0` default including mssql's named constraint | Fixed: both nullability polarities on every engine plus per-dialect type and default; mutations went from one to five, one per projection |
| 4 | quality | The tests were anchored to versions that do not exist: stage 1 ran against a moving head while asserting the very divergences stage 3 removes | `golang-migrate` v4.19.1 `read()` opens with `versionExists(suint(to))`, so an absent target fails outright. `migration_000021_countries_test.go` already abandoned plain `Migrate(20)` for `Up()` plus `Force(20)`. Version lists confirm the gaps | Fixed: stage 1 pinned at `Migrate(35)`, the last version all four share and none moves; stage 3 carries per-engine anchors; stage 2 applies each migration only where the file exists |

**Two errors in the sealed agreement surfaced here, both resolved without asking.** The 21-column
count above, and decision 5's "the test dumps both tables at 000038", which is unreachable because
decision 4 makes 000038 mssql-only and mssql has no 000039. Neither is a design question: 22 is a
fact with no alternative, and where two `Decided` items conflict the specific per-engine enumeration
governs, leaving exactly one implementation. Every `Recount with` command in the fanout was re-run
and all nine still matched, so these are annotation errors and not a contradiction in the design.

**`unchecked`: four entries, all answered.** Two dismissed as correct for a gate with no code. One
dismissed as covered by the run's shape, since executed evidence arrives at the final gate. **One
was a real gap and changed the plan:** a shape dump cannot detect two preserved columns whose values
were *swapped* by the hand-written refill, and variant E only proved the rows survived. Stage 3
therefore gained a case that seeds a distinct recognisable value per column and asserts each is
still in its own column after the apply, the down and the re-apply. That case went on to catch three
independent mutations at the final gate.

---

## Gate 2: the whole branch, round 1

Scope `f1c77bd0..48b4fb80`, all three axes, started 00:39:46Z and completed 00:53:07Z.

**Verdict `static-review-exhausted`. Zero findings, zero `unchecked` entries, zero follow-ups.** The
reviewer states plainly that nothing was held back, so there is no finding to fix, defer or refute.

A clean verdict is worth exactly what its coverage is, so here is the coverage.

**Conformance.** Read the agreement, plan, plan review, four stage logs, the closing log, the probes
and every file in the diff. Matched every settled deliverable to its migration, snapshot and test:
the SQLite `refresh_tokens(client_id)` index, the MySQL `audit_logs.details` effective `{}` default,
nullable SQLite PKCE columns, the five SQL Server `NVARCHAR` columns, both refresh-token owner
foreign keys at `NO ACTION` on all four engines, and the six removed integration skips. Confirmed
the sparse migration fanout is deliberate, 000036 to 000039 existing only on the engines each repair
needs while all four snapshots converge. Confirmed nothing out of scope entered: no collation work,
no parity test, no general ROPC revocation, no unrelated foreign-key indexing, no legacy SQL Server
down-migration repair. Confirmed no stale claim survives anywhere in the repository.

**Quality.** Reviewed the catalog dumper's query, scan, normalization, ordering, nullability,
default, type, index and foreign-key handling on all four engines. Reviewed every up and down
migration, specifically that SQLite 000039's parking order preserves refresh-token rows while
`codes` is rebuilt, that both tables' columns are copied in the right order, that the intended
indexes are restored, and that the downgrade's `NULL`-to-empty PKCE conversion matches what existing
readers expect. Ran the focused helper, migration, deletion and cleanup tests on all four engines.
Verified `git diff --check`, all 26 anchors, the exact branch head and the live split tree hashes,
with a clean worktree afterwards.

**Independent mutations, beyond the four stages' own.** This is what a clean verdict most needs to
carry, because it is the evidence the tests are worth their green:

| Mutation | Result |
|---|---|
| Removed user refresh-token cleanup | SQLite user-deletion and transaction tests failed on the foreign key. **Caught** |
| Removed client refresh-token cleanup | Both client-deletion tests failed. **Caught** |
| Swapped SQLite 000039 copy mappings, `codes` up path | `TestMigration000039_RowValuesSurviveTheRebuild` failed. **Caught** |
| Swapped copy mappings, `codes` down path | Same test failed. **Caught** |
| Swapped copy mappings, `refresh_tokens` down path | Same test failed. **Caught** |

With stage 3's own recorded mutation of the refresh-token up path, **all four rebuild copy blocks
are exercised**. Every mutated file was restored byte for byte by the wrapper. Worth recording: one
attempt first used a nonexistent test suffix, so the runner reported no tests; it was repeated
immediately with the real name and failed as required. A run that reports no tests looks exactly
like a survived mutation if nobody notices.

**Security.** Traced PKCE from authorize validation through code issuance to token redemption:
public clients still require it, confidential clients may omit it as decided, and redemption still
rejects a verifier when no challenge was bound. Existing unit coverage distinguishes SQL `NULL` from
an empty stored challenge, and the six restored integration cases exercise the public-client refusal
and confidential-client success paths including token and refresh exchanges. Compared against RFC
9700 section 2.1.1: mandatory PKCE for public clients is preserved and verifier binding for
challenged codes is not weakened. Traced every production user and client deletion entry point
across the four adapters: all delegate to the common transaction-aware implementations, remove owned
refresh tokens before the principal, propagate cleanup errors and roll back on failure, so
`NO ACTION` prevents orphaning without turning a legitimate deletion into a partial write. Checked
the migration SQL for injection and data-loss hazards: values are fixed migration text or
catalog-derived identifiers, and the SQLite parking table is transaction-local. No changed route,
handler, template, credential comparison, token signature, session, authorization, rate-limit, audit
or secret-handling surface exists on this schema-only branch.

**What it did not check, and why.** Nothing on the branch was left unreached. No browser pass: the
branch changes no UI and no HTTP response. It did not repeat the full green tiers locally, on the
grounds that exact-head CI and the live tree hashes already establish them, and spent the time on
focused cross-engine tests and independent mutations instead.

**Restore verified rather than believed.** The reviewer mutated production files, so this session
checked the tree afterwards: `git status --porcelain` empty, `HEAD` still `48b4fb80`, and
`tree-hash.sh --split` returning `production e58087916560`, `unit-tests f99c8a13ae64`,
`data-tests fa388342484b`, `integration-tests a47ff55c470b`, `docs 8dd009829b08`,
`all edb1fd47f377`, each identical to the value recorded before the review ran. Production has not
moved, so every mutation is provably gone.

The review budget was 4 rounds and 1 was spent.

Issue #288, opened 2026-08-28
Nothing checks that a new migration was written the same way on all four engines


leodip:
Nothing in this repository checks that a new migration was written the same way on all four
engines. The migrations are written four times by hand — `src/core/data/{mysqldb,postgresdb,mssqldb,sqlitedb}/migrations/` — and the only thing that
compares the results is a person reading four diffs.

That has already failed twice, both times for months:

- **#282** found five structural divergences by diffing four fully migrated catalogs: an index on
  three engines and not the fourth, a default on three, two columns still `NOT NULL` on one, five
  non-Unicode `varchar` columns on one, and a foreign key action that differed. One of them (#259
  section 2) was a live 500 on SQLite, the default database.
- **#283** found the same shape in string comparison: `=` over a `client_id` was case-sensitive on
  SQLite and PostgreSQL and case-insensitive on MySQL and SQL Server. RFC 6749 section 1.9 makes
  that a conformance defect rather than a preference.

## What is proposed

A `scripts/check-migrations.sh` in this repository, and a job in `.github/workflows/check.yml` that
runs it. It reads the four migration directories as text, so it needs no database and no container
and finishes in well under a second.

Rules, each traceable to a defect above:

| Rule | Prevents |
|---|---|
| filenames are `NNNNNN_lower_snake.{up,down}.sql` | a version that sorts or parses unexpectedly |
| every `.up.sql` has its `.down.sql` | an irreversible migration |
| one number names one change across engines | an engine silently skipping the change it did not get |
| a migration on fewer than four engines carries a `-- parity:` line saying which and why | #282's partial fixes are legitimate; forgetting three files is not, and they look identical |
| SQL Server: no bare `VARCHAR` | #282 D4 |
| SQL Server: every string column spells `COLLATE Latin1_General_100_CS_AS_KS_WS_SC_UTF8` | #283. `NewMsSQLDatabase` creates `IF NOT EXISTS`, so a pre-created database keeps its own default and an unpinned column lands case-insensitive there |
| SQL Server: default constraints are named | #282 D4. An unnamed one gets a per-database generated name, so a later `ALTER COLUMN` cannot drop it by name |
| MySQL: every `COLLATE` is `utf8mb4_0900_as_cs`, and every `CREATE TABLE` spells one | #283. The `_ci` and `_ai` forms fold |
[...]

Issue #283, opened 2026-08-27
String comparison is case-insensitive on MySQL and SQL Server and case-sensitive on SQLite and PostgreSQL, so an identifier means different things per engine


leodip:
Nothing in the schema pins a collation, so each engine compares strings by its own default. The result is that `=` and every UNIQUE index mean two different things depending on where Goiabada is deployed. Measured on a fresh, fully migrated database on each engine:

| engine | `'Alice' = 'alice'` | collation in force |
|---|---|---|
| SQLite | `0` | BINARY, the default |
| PostgreSQL | `false` | `en_US.utf8`, deterministic |
| MySQL | `1` | `utf8mb4_0900_ai_ci` |
| SQL Server | `1` | `Latin1_General_100_CI_AI_SC_UTF8` |

MySQL's is pinned by us. `NewMySQLDatabase` issues:

```sql
CREATE DATABASE IF NOT EXISTS %v CHARACTER SET utf8mb4 COLLATE utf8mb4_0900_ai_ci;
```

`ai_ci` is accent-insensitive and case-insensitive, so at least that one is consistent from deployment to deployment.

SQL Server's is not pinned at all. `NewMsSQLDatabase` issues a bare `CREATE DATABASE [%s]`, so the database inherits whatever the server default happens to be. The value in the table above is what the dev container's server uses. **Two operators running SQL Server can get different uniqueness semantics out of the same release**, which is worse than any single choice would have been.

MySQL is also not internally consistent: `audit_logs`, `client_logos` and `user_profile_pictures` carry `utf8mb4_unicode_ci` while the other 17 tables carry `utf8mb4_0900_ai_ci`.

## What it actually changes

Every UNIQUE index on an identity column, and every lookup through one: `users.email`, `users.subject`, `users.username`, `clients.client_identifier`, `groups.group_identifier`, `resources.resource_identifier`, `permissions.permission_identifier`, `web_origins(origin, client_id)`.

Two of those are already neutralised by application code, by convention rather than by schema. Email is lowercased at every entry point (`handler_auth_pwd`, `handler_forgot_password`, `handler_account_register`, and the account and admin APIs). Origins are canonicalised through `urlutil` before they are stored or compared.

Identifiers are not. `IdentifierValidator.ValidateIdentifier` accepts `^[a-zA-Z]([a-zA-Z0-9_-]*[a-zA-Z0-9])?$` and folds nothing, so:

[...]

PR #290, opened 2026-08-28
feat(core): the four database engines build one schema, and it is checked

### LINKED ISSUE #284: Nothing checks that the four database engines build the same schema

leodip:
Closes #284.

## What this is for

Goiabada's migrations are written four times by hand, once each for SQLite, MySQL, PostgreSQL and
SQL Server, and nothing has ever compared the results. #282 diffed four fully migrated catalogs by
hand and found five structural divergences that had been shipping for months; #283 then found the
same shape in string comparison, with `=` meaning case-sensitive on two engines and
case-insensitive on the other two. Both were invisible in review, and both were cheap to prevent at
the moment the file was written.

This builds the instrument that prevents them.

## The rule this establishes

Each engine's fully migrated catalog is dumped into a normalised structure and committed beside
that engine's migrations as `src/core/data/<engine>db/schema.golden`. Two tests then hold it:

- **Per engine, in the data tier.** Migrate an isolated database to head, dump it, compare against
  that engine's committed golden file. A migration that builds something other than what was
  intended fails on the engine it was written for.
- **Across engines, in the unit tier, with no database.** One test reads all four golden files and
  asserts they describe the same schema, modulo an explicit allowlist of engine idioms. Each
  allowlist rule states the idiom, why it is deliberate, and **how many places it currently
  excuses**, and the test fails when that number moves. Without the count, "SQLite records no length
  for a string column" is written once to excuse 89 columns and goes on excusing the 200th,
  including the one where the missing length was a mistake.

Adding a migration therefore has a stated, discoverable way to refresh the four files, and the
parity test stays red until it is done.

Two things fall out of building it. `schema_migrations` itself is brought into line: three engines
already agree on `version bigint not null primary key, dirty boolean not null` and SQLite alone
creates it nullable and keyless, so Goiabada pins the shape on new installs and a SQLite-only
migration rebuilds it on existing ones. And the four `schema.sql` snapshots are deleted, replaced by
the golden files. They are documentation nothing loads, they cannot be held to being correct, and
SQL Server's is wrong today: it names 25 primary key constraints where the migrations write 0, and
declares two indexes whose names occur only in the MySQL files.

### Stage 1: the dumper moves to core, exported, and learns to read a whole schema

`830be233`. The table-scoped catalog dumper that found #282's and #283's divergences lived
privately inside the data tier's test package, where a command cannot reach it and where it can
only compare a table against itself on one engine. It is now `src/core/data/schemadump`, and it
does three things it could not:

- **Enumeration.** `Tables(db, dialect)` lists a whole schema, sorted in Go rather than by the
  server, because `ORDER BY` on a string column is decided by four different collations. An empty
  list is an error, because an empty dump compared against an empty golden file reads as "nothing
  changed" and passes.
- **Index origin.** Whether a statement named an index or the engine did, read from each catalog
  directly and never matched against a name pattern: SQL Server's generated names carry a random
  per-database suffix, so a name pattern is the one thing that cannot be trusted here.
- **Column generation.** Whether the engine numbers a column. Nothing else in the shape can see
  it: two MySQL tables differing only by `AUTO_INCREMENT` report identical rows for every other
  projection, and so do two SQLite tables through `pragma_table_info`.

It also refuses, rather than silently drops, a construct the shape cannot record: composite foreign
keys, `CHECK` constraints, computed columns, `ON UPDATE` actions, partial indexes, `INCLUDE`
columns, descending keys, `DEFERRABLE`, and an index MySQL named for itself. That matters more here
than elsewhere because the same code will write the golden files and check them, so anything
dropped during the dump is dropped from the committed record too. All nine are absent from all four
chains today, so each test case builds its own table to fire at all.

The data tier keeps thin wrappers, so the ten migration test files that lean on these helpers are
untouched.

**Tiers:** unit 1263 pass; data on all four engines locally, sqlite 498, mysql 505, postgres 505,
mssql 508, all green. Eight mutations against the new seams, every one caught. Integration is not
applicable: no handler, route or schema moved.

**One deviation worth knowing.** The MySQL guard for computed columns first read `EXTRA LIKE
'%GENERATED%'`, and the mysql data tier refused it: MySQL 8 puts `DEFAULT_GENERATED` in `EXTRA` for
any column whose `DEFAULT` is an expression, which `audit_logs.details` has been since 000037. It
reads `GENERATION_EXPRESSION` instead. The four-engine tier caught it before the commit.

### Stage 2: `schema_migrations` gets one shape on all four engines

`505b453b`. golang-migrate creates `schema_migrations` itself, and its four drivers do not agree.
MySQL, PostgreSQL and SQL Server all build `version bigint not null primary key, dirty boolean not
null`; the SQLite driver builds `(version uint64, dirty bool)`, both columns nullable, no primary
key, and a separate `version_unique` index. The parity check reads that table like any other, so
the four have to agree on it, and the shape should be Goiabada's rather than whichever one the
pinned dependency happens to build.

Every driver creates the table only if it is absent, so each `NewMigrator` now issues its own
`CREATE TABLE` first and the driver's statement becomes a no-op. That settles a new install.
Migration `000041` rebuilds the table on a SQLite install created before this existed; the other
three need no file and the up migration says so on its `-- parity:` line.

Three details are load-bearing and none of them is obvious:

- **`INTEGER` and not `BIGINT` on SQLite.** Only `INTEGER PRIMARY KEY` is a rowid alias. Spelled
  `BIGINT`, SQLite builds its own index for the key and the driver's unconditional
  `CREATE UNIQUE INDEX IF NOT EXISTS version_unique` lands on top of it, leaving two unique indexes
  on one column where the other three have one.
- **SQL Server's pre-create takes golang-migrate's own `sp_getapplock`**, on a pinned connection,
  around its check and its create. That engine has no atomic `CREATE TABLE IF NOT EXISTS`, so two
  processes starting against one empty database can both pass the check and the loser dies with
  Msg 2714. The driver's own `ensureVersionTable` runs the same sequence and is safe only because
  it holds that lock, so issuing the pre-create before `WithInstance` without it would *remove* a
  property the driver already had.
- **The `NOT NULL` that carries the weight is `dirty`'s.** golang-migrate's `Version()` scans both
  columns into Go values and swallows a scan error, reporting `NilVersion`, which is the value that
  makes it run the whole chain from `000001` against a populated database. A NULL `dirty` does
  exactly that. A NULL `version` cannot be stored either, but by a different mechanism: SQLite
  *replaces* a NULL in an `INTEGER PRIMARY KEY` with a generated rowid, `NOT NULL` or not, so the
  column never rejects one and never holds one.

Tiers: modules 1264 pass; data 1508 pass and 20 skip on all four engines; integration 1012 pass on
SQLite. Two mutations caught: `INTEGER` to `BIGINT` in the SQLite pre-create, and `NOT NULL` dropped
from the migration.

**Worth knowing.** A first attempt asserted that the SQL Server migration lock is released by the
time `NewMigrator` returns. Two mutations, deleting the release and deleting the connection's return
to the pool, both left it green, because `go-mssqldb` implements `driver.SessionResetter` and
`database/sql` fires it on every pooled reuse: SQL Server drops session-scoped application locks on
that reset, so a leaked one is reclaimed immediately. The assertion was removed rather than kept
under a weaker claim. The lock stays, because connection reset says nothing about two separate
processes racing, which is the case that produced Msg 2714. Asserting that race would need a
barrier the production code does not have.

### Stage 3: the golden files, the generator that writes them, and the per-engine assertion
Landed in `b36c5127`.

Each engine's fully migrated catalog is recorded in `src/core/data/<engine>db/schema.golden`, and
`src/authserver/tests/data/schema_golden_test.go` migrates a fresh database, dumps it and holds it
to that file on the engine it is running on. That is the half of this change that runs on all four
engines and the thing that turns CI red when a migration lands without a regeneration.

Regenerate with `go run ./cmd/schemadump` from `src/core`, inside the dev container. The command
dumps all four engines before writing any file, so a run that dies on the third leaves the
committed set as it was rather than three files refreshed against one stale, and each file is
renamed into place so an interrupted run cannot leave a truncated record for the assertion to
read as a schema diff. Credentials default to the dev container's compose services and each field
takes a `GOIABADA_SCHEMADUMP_<ENGINE>_*` override.

**Names the engine invented are masked to `(engine-named)` rather than recorded**, which is forced
rather than stylistic: SQL Server's generated primary key and unique constraint names carry a
random per-database suffix, so a file recording them would fail on every regeneration and could
not be committed at all. Whether a name was the engine's or a migration's is read from each
catalog's own origin and never guessed from the shape of the name, and indexes are ordered by key
columns before name so two masked entries on one table stay apart. Two independent generations
produced byte-identical files on all four engines.

The parser landed here with the encoder rather than in stage 4, as the plan had sketched: the same
package writes the golden file and checks it, so a fact the encoding drops is dropped from every
consumer with nothing able to recover it, and round-tripping a dump through both is the only
evidence the format carries what the shape holds.

Tiers: unit 1271 pass, data 1516 pass across mysql, postgres, mssql and sqlite, 0 fail. Six
mutations against the new tests, all caught. Integration is not applicable: no handler, route or
template moves.

**The instrument has already found its first defect**, left for stage 5 per decision 6:
`browser_sessions.data` is declared `longtext` on SQLite, where every other string column on that
engine is `TEXT`. SQLite accepts any type name, so a MySQL spelling reached a SQLite migration and
nothing noticed until now.

### Stage 4: the cross-engine comparison, table-driven against synthetic shapes
Commit `214646a2`. Two test-only files in `src/core/data/`, adding the half that reads all four
golden files and says whether they describe one schema. It runs in the modules tier with no
database, so it runs once rather than four times across the database jobs.

Two halves, and the split is what makes an allowlist reviewable:

- **Canonicalisation maps spelling and never meaning.** `TEXT`, `varchar(40)`,
  `character varying(40)` and `nvarchar(40)` are one type written four ways, and the four collation
  names all say case-sensitive after #283. But a column's declared length, its signedness and its
  fractional-second precision stay comparable, so SQLite's missing length and MySQL's
  `bigint unsigned` survive to be counted on the allowlist rather than silently vanishing. The
  vocabulary is a closed per-engine table: a spelling absent from it is a failure naming it, and no
  allowlist rule may excuse one, because every other answer about that column would be a guess too.
- **The allowlist excuses the rest, and each rule carries how many places it covers and a digest
  over which places those are.** The count stops a rule written for 89 columns going on to excuse
  the 200th. The digest stops one accepted exception being corrected while a different one appears
  under the same rule at the same count, which a count alone cannot see. A divergence two rules both
  excuse is a failure rather than a first-match win, since otherwise neither count means anything.

An index is identified by its ordered key columns and its uniqueness, never by its name: every SQL
Server primary key and inline `UNIQUE` here is named by the engine with a random per-database
suffix that differs between two databases built from identical DDL. The name is still compared
where all four engines got theirs from a migration, because a name is what a later migration says
when it drops an index.

Exercised against synthetic shapes only: an idealised fixture of one table spelled four ways that
must report nothing, then 21 cases each exactly one difference wide, plus two tables pinning
canonicalisation directly. 63 subtests. Wiring it to the four committed files is stage 5.

Tiers: unit 1274 passed on all modules, data 1516 passed on all four engines. Integration is not
applicable, since the stage touches no handler, route or schema. Six mutations against the new
seams, all caught: the `DEFAULT NULL` fold, the membership digest, index identity by tuple, the
gate on comparing index names, the refusal to excuse an unreadable spelling, and PostgreSQL's bare
`timestamp` being microsecond precision.

**Measured at this gate, as a throwaway that was deleted afterwards:** pointing the comparison at
the four committed golden files reports **346 divergences over fourteen shapes**, with **no**
divergence at all on the table set, the column set, foreign keys, on-delete actions or generation.
Eleven shapes are deliberate engine idioms and become allowlist rules in stage 5. Three are not:
MySQL's catalog cannot distinguish `DEFAULT ''` from no default, which is a gap in the dumper's
own projection rather than a schema defect; MySQL names three foreign-key indexes `fk_*` where the
other three engines name them `idx_*`, which breaks a later migration that drops one by name; and
`audit_logs.details` is `text` on MySQL, capped at 64 KiB, where the other three are unbounded.

### Stage 5: the four real golden files compared, and every divergence fixed
Commit `c922f621`. The comparison was pointed at the four committed golden files for the first
time and reported **346 divergences**. Sorted: 339 deliberate engine idioms, one gap in the
instrument, three defects. Under decision 6 the three are fixed here rather than filed, so `main`
is never knowingly divergent.

**The instrument gap.** `ColumnShape.Default` cannot say whether a column has a default at all:
`""` is both "no default" and what MySQL's `information_schema.COLUMN_DEFAULT` reports for
`DEFAULT ''`. Four columns declaring `DEFAULT ''` on all four engines therefore read as having none
on MySQL alone. `HasDefault` is now read from every engine's catalog rather than inferred on three
of them from an empty expression, and the golden file format moves to version 2 to carry it. A fact
taken from one engine and guessed on three is how the gap arrived; guessing it again would leave
the same shape behind.

**Migration `000042`, MySQL only, for the two schema defects.**

- `audit_logs.details` was `TEXT`, a 65,535-byte ceiling, where SQLite, PostgreSQL and SQL Server
  leave it unbounded. It holds an audit event's JSON payload, so the cap is reachable, and past it
  MySQL 8's default strict `sql_mode` refuses the write: the audit record of a security event is
  what is lost. Widened to `LONGTEXT`, which `browser_sessions.data` already uses for the same
  reason.
- `codes`, `refresh_tokens` and `user_sessions` carried their foreign-key indexes under the
  constraint's name, `fk_codes_user` and its two siblings, where the other three engines created
  `idx_codes_user_id` and siblings in their own `000024`. InnoDB requires an index over a foreign
  key's columns, so the initial migration declared the inline `KEY` to satisfy it and `000024`
  skipped MySQL. A later migration saying `DROP INDEX idx_codes_user_id` would succeed on three
  engines and fail on the fourth, which is exactly what a name comparison exists to catch. Renamed
  in place, which InnoDB accepts for an index a foreign key depends on where a drop would not be.

**The allowlist, eleven rules over 339 places**, each with its reason, its count and a digest over
its membership:

| Rule | Excuses |
|---|---|
| sqlite: a string column carries no declared length | 91 |
| sqlite: a datetime carries no declared fractional-second precision | 66 |
| mysql: an id column is bigint unsigned where the other three are signed | 50 |
| sqlite: no index object behind a rowid primary key | 25 |
| sqlite: the rowid primary key reports as nullable | 25 |
| postgres: BIGSERIAL leaves a nextval() default the other three do not have | 25 |
| sqlite: a boolean is declared numeric | 19 |
| mysql: InnoDB indexes every foreign key and the other three do not | 18 |
| sqlite: a boolean is declared INTEGER | 11 |
| sqlite: a 32-bit integer is declared INTEGER, which is 64-bit | 8 |
| mysql: audit_logs.details records its default with a charset introducer | 1 |

Every rule requires the three engines it does not name to agree with each other, so two engines out
of line is never an idiom. That is the boundary the `audit_logs.details` defect crossed, and a test
pins it against the defect as it stood rather than against a shape.

Decision 11's prediction held: no divergence at all on the table set, the column set, foreign keys,
on-delete actions or the generation property, so modelling generation cost the allowlist nothing.

Tiers: unit 4270 passed, data 1520 passed on all four engines, integration 1012 passed on sqlite.
Seven mutations across the new seams, all caught, two of them against the migration file itself.

Deviations: the allowlist and its test went into a new `schema_parity_golden_test.go` rather than
into `schema_parity_test.go`, which is already 881 lines of machinery; and the migration is
`000042` rather than the `000041` the design named, because `000041` is stage 2's SQLite file.

### Stage 6: the snapshots are deleted and the agent files corrected
Landed in `ab4e13a8`.

The four `schema.sql` snapshots are deleted, 3,080 lines and 108 KB. Nothing consumed them: every
`//go:embed` in the four database packages names `migrations/*.sql` and nothing else, and no script,
CI step or test read them. They were also unverifiable by construction, and SQL Server's was wrong
at the time of writing, naming 25 primary key constraints where the migrations write 0 and declaring
two indexes whose names occur only in the MySQL files. What replaces them is the checked record: the
four `schema.golden` files, generated by exactly the code that checks them.

This was deliberately last. Until the golden files existed and both tests were green on all four
engines, the snapshots were the only record of what the chain builds, so there is no window in which
the repository has neither. The precondition was met by CI rather than by a local claim: green on
`c922f621` across all four database jobs and all three unit jobs.

Also in the stage: `CLAUDE.md`'s Database Pattern line points at `sqlitedb/schema.golden`; four code
comments that described `schema.sql` as a file that exists are corrected, the one that mattered
being `migration_dumptable_test.go`'s recount command for the table count, since the comment above
it says a mismatch is a finding rather than a number to adjust and that instruction stops being
followable when its command names no file; and `AGENTS.md` is a byte-identical copy of `CLAUDE.md`
again.

Tiers: modules 1276 passed 0 failed; data 1520 passed 0 failed across mysql, postgres, mssql and
sqlite, with the per-engine golden assertion green on all four *after* the deletion. Integration not
applicable, since no handler, route or schema moved and no production Go changed. `where.sh
migrations` passes every rule including `snapshot`. No mutation is owed: the stage adds no test and
pins no new behaviour.

Nothing here reaches an operator: the migrations, the database and every endpoint are unchanged, so
the change asks nothing of anyone on upgrade.

Deviations: `AGENTS.md` had drifted in four hunks, not the three the agreement stated, the fourth
being #213's Key Logic bullet that goes with the routing row; and the `CLAUDE.md` bullet was written
at 20 words and cut back to 10 before the gate, to sit with its siblings at 10 and 9.

## One thing this owes outside the repository

`check-migrations.sh`, the four-engine migration checker in `~/code/ai`, carries a `snapshot` rule
whose target is `SNAP = "src/core/data/{}/schema.sql"`. After this change that path names nothing.
This branch still passes it, because the rule keys on `git diff --name-only` and a deletion appears
there, but **the next branch to touch a migration fails a rule it cannot satisfy**. The rule needs
repointing at `schema.golden`. That script is not in this repository, so the run could not make the
edit.

## What the review changed

The whole branch was read once, after every stage had landed. Round 1 returned **two significant
conformance findings and no security findings**. Both were verified against the code, both stood,
and both are fixed here.

### Stage 1's guard did not refuse an expression index

Decision 11 settles that the dumper fails when it meets a construct the shape cannot represent
rather than dropping it silently, because the same code writes the golden file and checks it: a
fact discarded during the dump is discarded from the committed record too. An expression
(functional) index is such a construct and the guard had no check for it.

On PostgreSQL it was genuinely silent. `dumpIndexes` joins `indkey` onto `pg_attribute`, an
expression's `indkey` entry is `0`, that matches no attribute, so the row is dropped by the join and
a wholly expression-based index disappears from the dump altogether. A migration adding one on one
engine would have passed both the per-engine assertion and the cross-engine comparison. On MySQL and
SQLite the key column name comes back NULL and reached the scan as `converting NULL to string is
unsupported`, naming neither the construct nor the remedy.

Fixed in `432725f0`: the guard carries the construct on the three engines that can express it, each
read from its own catalog, and `guardTable` moves ahead of the index and foreign key projections so
those two engines get the guard's sentence instead of the driver's. It stays behind the column read,
so a table that does not exist is still answered by the existing error.

`e3fb542e` then closes the rest of the family, which the finding's own reasoning had left open. The
two constructs it named as unenumerated turned out to be covered, but establishing that turned up
two real members: an index built with a **non-default method** (a GIN, hash, full-text, spatial or
columnstore index over one column was the same shape as a plain B-tree over it) and a key with a
**non-default null ordering**.

### A defect the run recorded and then never fixed

`browser_sessions.data` was declared `longtext` on SQLite. On MySQL that name is real and correct,
being the engine's largest string; SQLite has no such type and assigns affinity by substring, so the
column behaved as `TEXT` while being recorded in another engine's vocabulary, standing beside ninety
that say `TEXT`. It arrived when 000035 created the table on all four engines.

Stage 3 found it and wrote that it "is exactly the class of thing #284 exists to find, and decision 6
puts it on this branch, in stage 5". Stage 5 fixed three other defects and left this one, without
allowlisting it and without recording a reclassification.

It was invisible to the check that should have held stage 5 to it: `parityTypes` mapped SQLite's
`text` and `longtext` onto one value, so the cross-engine comparison could not see the difference at
all. That is the more important half.

Fixed in `719a9e75`, in both halves. Migration `000043` rebuilds the table on SQLite alone, since
SQLite cannot ALTER a column's type; nothing holds a foreign key onto `browser_sessions` on any
engine, so the implicit DELETE behind `DROP TABLE` has nowhere to cascade and no table needs parking
first. And the SQLite `longtext` entry is **removed** from `parityTypes` rather than given a distinct
value, so the spelling is refused by name if it ever returns, which is what that table's contract
says an unrecognised type does. `sqlitedb/schema.golden` regenerated with `go run ./cmd/schemadump`:
it moves by exactly one line, which is also the evidence the rebuild carried every other column,
index and foreign key across untouched.

### The index-name comparison was all-or-nothing across the four engines

Round 2's single finding, and it is decision 2's intent against decision 2's wording.

`compareIndexes` compared an index name only when **every** engine's index for that key came from a
migration. One engine-named index in the set, an inline `UNIQUE`, a SQLite rowid primary key or a SQL
Server unique constraint, and the name comparison switched off entirely, so a name a migration
misspelled on one of the other three went uncaught. The decision's headline says "wherever all four
names came from a migration" and the code implemented that literally; its rationale two paragraphs
down is the misspelled-name class, `DROP INDEX idx_codes_user_id` succeeding on three engines and
failing on the fourth, which the all-four rule defeats on exactly the keys where engines spell an
index differently.

Measured before changing anything, because a widened check that reported on the current schema would
be a divergence to fix rather than a test change. Over the 76 index keys in the four committed golden
files: 48 already compared, 25 with nothing to compare either way, and 3 mixed, each with exactly one
migration-named engine and so a singleton. The fix reports nothing new today.

Fixed in `77dc366a`, test-only. `compareIndexes` builds the set of engines whose indexes for the key
are all migration-named and compares names across that set, filling the rest of the failure line with
`-` so it still carries a column per engine. All six shapes of the family are now pinned by name as
cases, including the two the reviewer had left unprobed.

## Tiers

Run locally, against the final tree:

| Tier | Engines | Result |
|---|---|---|
| unit / modules | n/a | **1276 pass, 0 fail** |
| data | mysql, postgres, mssql, sqlite | **2059 pass, 0 fail** |
| integration | sqlite | **1012 pass, 0 fail** |
| `where.sh migrations` | all four | every rule passes, next free number 000044 |
| `check-anchors.sh` | n/a | 22 of 22 resolve |

The **four-engine integration** result, and the `Lint`, `Vulnerabilities`, `Versions` and
`Build validation` jobs that nothing local runs, are the check suite's. It is **green on all eleven
jobs at `77dc366a`**, the final commit:
https://github.com/leodip/goiabada/actions/runs/33231211883

Twelve mutations were run against the new tests across the two code review rounds and eleven were
caught. The single survivor was a guard clause no mutation could falsify, because the loop it
protected does not execute on an empty set; it was removed rather than recorded, and the two boundary
cases stay as the statement of the behaviour.

## Deviations from the plan, at the run level

Each stage's entry above carries its own; these are the four a reader of the whole change should
know.

- **Three migrations, where the agreement predicted one.** `000041` was planned, for
  `schema_migrations` on SQLite. `000042` (MySQL) and `000043` (SQLite) came from divergences the
  instrument found on its first runs, which decision 6 puts on this branch rather than on follow-ups.
- **The golden format is at version 2.** Stage 5 found that `ColumnShape.Default` could not tell "no
  default" from `DEFAULT ''`, because MySQL's catalog reports both as the empty string, so four
  columns read as having no default on MySQL alone. `HasDefault` is projected from all four catalogs
  and the format version was raised.
- **The golden parser landed in stage 3, with the encoder, not in stage 4 as sketched.** An encoding
  whose reader is unwritten cannot be shown lossless, and the same package writes the file and checks
  it, so a dropped fact is unrecoverable downstream.
- **Section 4.1's "the four per-engine branches move verbatim" is amended twice**, each recorded as a
  deviation where it landed: `Generated` from decision 11, and `is_system_named` from the plan
  review's finding 3.

## For the release notes

Three migrations run on upgrade, all automatic, and none of them asks an operator to do anything
before or after. Two things are worth stating anyway.

- **SQLite rebuilds two tables in place**, `schema_migrations` (`000041`) and `browser_sessions`
  (`000043`), because SQLite cannot alter a column's type or add a primary key. Both copy their rows
  forward inside the migration's transaction. Take the backup you would normally take before a
  schema-changing upgrade. One consequence is accepted and recorded in `000043`: the
  `sqlite_sequence` row goes with the table and is re-established at the highest id copied across, so
  `AUTOINCREMENT`'s never-reused guarantee breaks once for `browser_sessions.id`. That column is a
  surrogate key, a session is found by `(owner, session_id_hash)`, and no table references it.
- **On MySQL, `audit_logs.details` is widened from `TEXT` to `LONGTEXT`** (`000042`). It was capped
  at 65,535 bytes against an unbounded column on the other three engines, and past the cap MySQL 8's
  default strict `sql_mode` refuses the insert, so an oversized audit record of a security event was
  lost rather than truncated. The same migration renames three foreign-key indexes from `fk_*` to
  `idx_*` to match the other three engines.

For contributors rather than operators: **the four `src/core/data/*/schema.sql` snapshots are
deleted**. Adding a migration now means running `cd src/core && go run ./cmd/schemadump` in the dev
container, which regenerates all four `schema.golden` files at once, and committing them. Skip it and
every database job in CI goes red. `CLAUDE.md` and `AGENTS.md` say so.

## What the run has for you, in three comments

- **[Decisions](https://github.com/leodip/goiabada/pull/290#issuecomment-5460054672)** — nothing is
  deferred. One question was escalated (decision 11) and you answered it; three plan-review findings
  were resolved rather than escalated, with why; and what reversing decisions 6 and 10 would now
  cost.
- **[Follow-ups](https://github.com/leodip/goiabada/pull/290#issuecomment-5460054898)** — one issue
  ready to file with its `gh issue create` command, one tooling edit owed outside this repository,
  and one triage question about #288. Nothing was folded in.
- **[The code review, in full](https://github.com/leodip/goiabada/pull/290#issuecomment-5460055116)**
  — every round, every finding, the evidence, the mutations and the tier results. The verdicts
  themselves live in a gitignored directory the next round overwrites, so that comment is the only
  durable copy.

### Review

Reviewed by `deepseek/deepseek-v4-pro-0813` at effort `xhigh`, **two rounds over the whole branch**.
`LEO_REVIEW_MODE=end` was in force, so no stage was reviewed on its own. The plan was reviewed
separately, before any code existed, by `gpt-5.6-sol` at effort `xhigh`.

|---|---|---|---|---|
| plan | 1 | 4 | 2 | 3 resolved into stages 1, 2 and 4; 1 escalated as decision 11, answered `A` |
| final | 1 | 2 | 0 | both fixed, `432725f0`, `e3fb542e`, `719a9e75` |
| final | 2 | 1 | 0 | fixed, `77dc366a` |

- **blocking, conformance (plan):** the normalised shape could not see column generation, so two
  MySQL tables differing only by `AUTO_INCREMENT` produced byte-identical dumps. Escalated as
  decision 11 and answered `A`; the property landed in stage 1.
- **blocking, conformance (plan):** count-only allowlist rules pass an equal-count membership
  substitution. Fixed in stage 4: each rule now carries a digest over its sorted instance keys
  beside the count.
- **significant, conformance (plan):** SQL Server's default projection omitted `is_system_named`,
  which section 4.2's name-masking rule needs when 42 of 64 defaults are unnamed. Fixed in stage 1.
- **significant, quality (plan):** pre-creating `schema_migrations` before `sqlserver.WithInstance`
  moved a check-then-create outside golang-migrate's advisory lock. Fixed in stage 2 by taking the
  driver's own `sp_getapplock` around it.
- **significant, conformance (round 1):** the guard dropped a PostgreSQL expression index silently
  instead of refusing it. Fixed in `432725f0`, and the family closed in `e3fb542e`, which turned up
  two more real members.
- **significant, conformance (round 1):** `browser_sessions.data` was still `longtext` on SQLite, a
  defect the run itself recorded and then dropped, hidden by a vocabulary fold that made it
  invisible to the check meant to catch it. Fixed in `719a9e75`, in both halves.
- **significant, conformance (round 2):** the index-name comparison was all-or-nothing across the
  four engines. Fixed in `77dc366a`.

**Nothing was refuted and nothing was deferred.** Round 2 moved no production code, so under the
round budget a third round is not funded: the reviewer has read the branch twice and the second
reading found one test-level gap.

[this comment](https://github.com/leodip/goiabada/pull/290#issuecomment-5460055116).

## Status

**Complete and ready for review.** All six stages have landed, the whole branch has been read twice
by a reviewer with every finding fixed, and the check suite is green on all eleven jobs at the final
commit `77dc366a`.

The dumper is a package in core and reads a whole schema. `schema_migrations` has one shape on all
four engines. Each engine's catalog is recorded in a golden file the data tier checks on that engine,
and the cross-engine comparison runs over the four committed files with an eleven-rule allowlist and
no unexcused divergence. Every divergence the instrument found is fixed on this branch, four defects
in all. The four `schema.sql` snapshots it replaces are gone, with `CLAUDE.md` and `AGENTS.md` back
in line.

One thing is owed outside this repository and the run could not make it: `check-migrations.sh`'s
`snapshot` rule still points at `src/core/data/{}/schema.sql`, and wants repointing at
`schema.golden` before the next branch touches a migration.


leodip:
## Decision 11: what should the new schema check notice, and what should it do about the parts it cannot describe?

**What this is about.** Goiabada's database schema is written four times by hand, once each for
SQLite, MySQL, PostgreSQL and SQL Server, as one numbered migration file per engine. Nothing in the
repository checks that the four end up building the same thing, which is why #282 found five
structural divergences that had shipped for months and #283 found `=` meaning case-sensitive on two
engines and case-insensitive on the other two. This issue builds that check. It connects to each
engine, reads back the schema the engine actually built, and writes that into a file committed next
to that engine's migrations. Two tests then read those files: one per engine, which fails when the
file was not refreshed after a migration, and one across all four, which fails when the engines
disagree about anything that is not on a written allowlist.

**Where you would meet it.** Adding a migration. You write the four `.up.sql` files as you do today,
run one command that regenerates the four committed records, and the tests tell you whether the
engines have stopped agreeing.

**What decides what the check can see.** The committed record is not the raw DDL. It is a normalised
description, inherited from #282, where it only ever described one table at a time. Per column it
holds the name, the type in that engine's own spelling, whether it is nullable, its default, its
collation, and on SQL Server the name of its default constraint. Per index: the name, whether it is
unique, and the key columns in order. Per foreign key: one row of column, referenced table,
referenced column, and what happens on delete.

**Anything a table has that is not on that list is invisible to both tests.** The same code writes
the record and checks it, so nothing later in the run can recover a fact it dropped. The plan review
found this and I measured it today. Two things are missing:

- **Which columns the database numbers for you.** All 26 tables have an auto-numbered `id` on every
  engine: 26 `AUTO_INCREMENT` on MySQL, 26 `IDENTITY(1,1)` on SQL Server, 26 `BIGSERIAL` on
  PostgreSQL, `AUTOINCREMENT` throughout SQLite. None of it is recorded. A migration that dropped it
  on one engine would change nothing in that engine's file, and both tests would stay green.
- **Whether a foreign key spans one column or several.** A foreign key is stored as one row per
  column, so a single constraint over two columns is indistinguishable from two separate
  one-column constraints. There are no multi-column foreign keys today on any engine.

Plus seven kinds of thing the description has no room for at all, none of which any engine's
migrations currently use: CHECK constraints, computed columns, `ON UPDATE` rules on a foreign key,
indexes with a `WHERE` clause, indexes with included columns, descending index keys, and deferrable
constraints.

**One part happens whichever option you pick**, so it is not part of the question: the reader errors
out when it meets something it cannot describe, rather than quietly dropping it. That costs nothing,
only ever refuses more than it does today, and fires on nothing currently in the repository.

### The options

The axis is what the check notices when a migration is written wrong.

**A. Also record which columns are auto-numbered.** One property per column, read from each engine's
own catalog. A migration that drops auto-numbering on one engine then fails on that engine.

**B. Record nothing further.** The committed files stay silent about auto-numbering. Cheapest, and
it leaves the four engine-specific readers exactly as #282 wrote and reviewed them.

**C. Record auto-numbering, and multi-column foreign keys, and each column's position in the table.**
Widest coverage now.

### What that means in practice

Say you add migration 000042 and, writing the four files by hand, you leave `AUTO_INCREMENT` off the
MySQL `id` and get the other three right.

- Under **B**, every test passes and the four committed records look exactly as expected. The first
  symptom is inserts failing on MySQL, in whatever environment runs MySQL.
- Under **A** and **C**, the MySQL test fails as soon as you regenerate, naming the column.

Now say a future migration adds a two-column foreign key. Under **A** and **B** it is recorded as
two one-column foreign keys, which is wrong but consistently wrong on all four engines, so the check
stays green and nobody learns anything. Under **C** it is recorded correctly. Today that case does
not exist.

|  | A | B | C |
|---|---|---|---|
| Notices auto-numbering dropped on one engine | yes | no | yes |
| Notices a multi-column foreign key correctly | no | no | yes |
| Work in stage 1 | one extra reader per engine | none | one extra reader per engine, plus a foreign-key change |
| Could force a migration on this branch | yes, unlikely | no | yes, unlikely |

That last row is the only real risk and it is worth stating plainly. This issue already decided that
any divergence the check finds on its first run gets fixed here rather than deferred. Recording
auto-numbering could in principle surface such a divergence. All four engines declare all 26 tables
auto-numbered, so I expect none, but that is not proven until the generator runs for the first time.

**What does not differ.** None of the three changes the running server, any interface, or anything
an operator sees or configures. All three land in the same stage and the same package. Each reverses
for the cost of one field and its tests. Whichever you pick, the error-on-anything-undescribable
guard above is written.

### Recommendation: A

The stated goal of this whole change is that "a migration that builds something other than what was
intended fails on the engine it was written for". A column quietly losing its auto-numbering is the
plainest instance of that, and 26 columns per engine currently depend on it. C's foreign-key work is
designing a representation against zero existing instances, and the guard already covers that case
for free by refusing to describe what it cannot.

**Why I am asking rather than just doing it.** The design says the four engine-specific readers move
across unchanged and that "nothing here improves them", and A edits all four. And because a
divergence found on the first run is fixed on this branch rather than filed, A is the option that
can grow the work in a way B cannot. If you would rather this check stayed narrow and cheap, B is a
legitimate answer and now is when it is cheap to say.

This blocks stage 1, which writes the description and its tests. Nothing else in the plan depends on
the answer, and the rest of the plan review is already resolved: the allowlist now pins which
exceptions it accepts rather than only how many, the SQL Server reader learns to tell an
engine-invented default-constraint name from one a migration chose, and the SQL Server
`schema_migrations` pre-create takes golang-migrate's own advisory lock so two replicas starting at
once cannot collide. Reply with **A**, **B** or **C**.

leodip:
A

leodip:
## Decisions: nothing is deferred

Every decision in the agreement reached `Decided` before the stage that depended on it ran, and no
step was built on an assumption the run took forward unsettled. **There is nothing in this slot for
you to reverse.**

That is worth spelling out rather than leaving as a bare "none", because the run did settle things
on its own and you should see which.

### One question was escalated, and you answered it

**Decision 11: what does the parity check model, and what happens to a construct it cannot
represent?** Posted here on 2026-08-28, answered `A` ten minutes later.

The reviewer found at the plan gate that the normalised shape carried over from #282 could not see
column generation at all: it built two MySQL tables differing only by `AUTO_INCREMENT` and the
planned projection returned byte-identical rows. 26 `AUTO_INCREMENT` on MySQL, 26 `IDENTITY` on SQL
Server, 26 `BIGSERIAL` on PostgreSQL and `AUTOINCREMENT` throughout SQLite were all invisible.
Because the same code writes the golden file and checks it, a fact discarded during the dump is
discarded from the committed record too, and no later stage can recover it.

`A` added the property, read from each engine's own catalog. A migration that leaves `AUTO_INCREMENT`
off one engine's `id` now fails on that engine naming the column, which is goal 1 stated exactly.

**One thing was measured only after the answer arrived, because `A` is what made it load-bearing**,
and it is the trap worth knowing about: `sqlite_sequence` holds no row for a table until that table's
first insert. The generator dumps a scratch database migrated to head and holding no rows, so reading
generation from there would have recorded all 26 SQLite tables as not auto-numbered. Under decision 6
the run is instructed to fix a divergence with a migration on this branch, so an unattended run could
have written a migration to "fix" a property that was never broken. The declaration in `sqlite_schema`
is the only surviving source and is where the projection reads from (`probe/sqlite_autoincrement.out`).

### Three plan-review findings were resolved rather than escalated

Each because nothing was being chosen between materially different outcomes:

| Finding | Why it was not yours to answer | Landed in |
|---|---|---|
| Count-only allowlist rules pass an equal-count membership substitution, and canonicalising signedness and string length erases two exceptions before they can be counted | The agreement contradicted itself. Decision 5's prose sends `bigint unsigned` and SQLite's `TEXT` through canonicalisation and says such differences "never reach the allowlist"; decision 5's own table then counts "SQLite records no length" at 89. Canonicalising spelling while leaving length and signedness comparable is the only reading under which both sentences do something | stage 4 |
| MySQL's catalog cannot tell a migration-chosen index name from an engine-invented one, and SQL Server's default projection omits `is_system_named` | The MySQL half is a correct refutation that does not bite (every index in `mysqldb/migrations/` is named). The SQL Server half is forced by section 4.2, which masks engine-generated default names and cannot do so from `dc.name` alone when 42 of 64 defaults are unnamed | stage 1 |
| Pre-creating `schema_migrations` before `sqlserver.WithInstance` moves a check-then-create outside golang-migrate's advisory lock | Decision 7 already decided the pre-create; only its SQL Server spelling was open. Taking the driver's own `sp_getapplock` around check and create restores a property golang-migrate already had | stage 2 |

### Two of your own decisions, and what reversing them would now cost

Neither is deferred. Both had consequences large enough to be worth naming next to the change.

**Decision 6, that any divergence the instrument finds is fixed on this branch rather than filed.**
The instrument's first run over the real schema reported **346 divergences**: 339 deliberate engine
idioms across eleven allowlist rules, one gap in the instrument itself, and **three defects**. All
three are fixed here, by migration `000042` (MySQL) and in the dumper. The reviewer later found a
fourth the run had recorded and then dropped, fixed by `000043` (SQLite). Reversing this would mean
four open issues and a parity test shipping green over a schema known to diverge.

**Decision 10, that the four `schema.sql` snapshots are deleted.** 3,080 lines and 108 KB, gone in
`ab4e13a8`, once the golden files existed and both tests were green on all four engines. Restoring
them is `git checkout ab4e13a8^ -- src/core/data/*/schema.sql`, and the cost of doing so is the dual
maintenance the change removed: the SQL Server snapshot was not stale but a different document,
naming 25 primary key constraints where the migrations write 0, and declaring two index names that
occur only in the MySQL files (`probe/snapshot_drift.out`).

leodip:
## Follow-ups: one issue to file, one tooling edit outside this repository, one triage question

Nothing was folded in. Every defect the run found inside its own scope was fixed in the stage that
found it, under decision 6, and is described in the body above rather than here.

---

### 1. Ready to file: `NewMsSQLDatabase` can fail one replica on concurrent first startup

**Found** at the plan gate, beside the `NewMigrator` finding that stage 2 fixed. Verified against
`src/core/data/mssqldb/db.go` before drafting.

**Why it was not folded in.** Stage 2 opens `NewMigrator` (`db.go:149`), not `NewMsSQLDatabase`
(`db.go:38`), and the remedy is not the same few lines: this guard runs against `master` before the
application database exists, so the lock has to be taken in a different connection scope. Folding it
in would also make an issue about schema parity change how the database itself is created, which its
goal does not cover.

**Duplicate search re-run just now**, over every state, four queries (`mssql CREATE DATABASE`,
`concurrent startup replica`, `sp_getapplock`, `NewMsSQLDatabase`). **No duplicate.** The only hits
were #105 and #127, both closed and unrelated, and #283/#288 matching on the word "migration".

**Title:** `fix(core): concurrent first startup can fail one replica creating the SQL Server database`
**Labels:** `bug`

**Body:**

> `NewMsSQLDatabase` creates the application database when it is absent, with a check followed by a
> create and no serialisation between them:
>
> ```sql
> IF NOT EXISTS (SELECT * FROM sys.databases WHERE name = N'%s')
>     CREATE DATABASE [%s]
> ```
>
> SQL Server has no atomic `CREATE DATABASE IF NOT EXISTS`, so two Goiabada instances starting at
> the same time against a server where the database does not yet exist can both pass the check. One
> creates the database; the other fails with `Msg 2714, There is already an object named ... in the
> database` and the process exits.
>
> Multi-replica deployment is documented in
> `site/src/content/docs/production-deployment/kubernetes.mdx`. The generated manifest starts at one
> replica, so this needs a deliberate scale-up plus a first start against an empty server to hit,
> which is why it has not been seen. It is a startup crash, not data loss: restarting the failed
> replica succeeds, because the database exists by then.
>
> The same shape one layer down was found in `NewMigrator` while building #284 and fixed there by
> taking `sp_getapplock` around the check and the create, which is what golang-migrate's own
> `ensureVersionTable` does. The fix here is the same idea in a different scope: the lock has to be
> held on a connection to `master`, since the application database is the thing being created.
> Alternatively, tolerate `Msg 2714` and re-check that the database exists.
>
> The other three engines are not affected: MySQL, PostgreSQL and SQLite all have a create form that
> tolerates a concurrent creation.

Files it:

````bash
cat > /tmp/mssql-race.md <<'ISSUE'
`NewMsSQLDatabase` creates the application database when it is absent, with a check followed by a
create and no serialisation between them:

```sql
IF NOT EXISTS (SELECT * FROM sys.databases WHERE name = N'%s')
    CREATE DATABASE [%s]
```

SQL Server has no atomic `CREATE DATABASE IF NOT EXISTS`, so two Goiabada instances starting at the
same time against a server where the database does not yet exist can both pass the check. One
creates the database; the other fails with `Msg 2714, There is already an object named ... in the
database` and the process exits.

Multi-replica deployment is documented in
`site/src/content/docs/production-deployment/kubernetes.mdx`. The generated manifest starts at one
replica, so this needs a deliberate scale-up plus a first start against an empty server to hit,
which is why it has not been seen. It is a startup crash, not data loss: restarting the failed
replica succeeds, because the database exists by then.

The same shape one layer down was found in `NewMigrator` while building #284 and fixed there by
taking `sp_getapplock` around the check and the create, which is what golang-migrate's own
`ensureVersionTable` does. The fix here is the same idea in a different scope: the lock has to be
held on a connection to `master`, since the application database is the thing being created.
Alternatively, tolerate `Msg 2714` and re-check that the database exists.

The other three engines are not affected: MySQL, PostgreSQL and SQLite all have a create form that
tolerates a concurrent creation.
ISSUE

gh issue create \
  --title "fix(core): concurrent first startup can fail one replica creating the SQL Server database" \
  --label bug \
  --body-file /tmp/mssql-race.md
````

---

### 2. Not filed, because the file is not in this repository: `check-migrations.sh` needs repointing

`~/.claude/skills/leo-spec/scripts/check-migrations.sh` carries a `snapshot` rule holding
`SNAP = "src/core/data/{}/schema.sql"`: a branch whose migrations moved without refreshing that
engine's snapshot fails it. This change deletes all four snapshots.

**This branch still passes the rule**, because it keys on `git diff --name-only` and a deletion
appears there, and MySQL's snapshot was refreshed by stage 5 in the ordinary way before stage 6
removed it. **Every branch after this one fails it**, because a migration will land with no
`schema.sql` left to refresh.

The rule wants repointing at `schema.golden`, in `~/code/ai`. It is a one-line change to a path
constant. Nothing in this repository can make it and nothing in CI depends on it: the rule gates the
agent workflow, not the build.

---

### 3. A triage question, not an action

**#288 is open** ("Nothing checks that a new migration was written the same way on all four
engines"), filed the same day as #284, and #284's agreement never mentions it. This change builds a
catalog-shape check across the four engines, which is close enough that #288 may be wholly or partly
answered by it: the data tier now holds each engine's freshly migrated catalog to a committed golden
file, and the modules tier compares the four files against each other with an eleven-rule allowlist.

Judging whether that closes #288, narrows it, or leaves it untouched is yours. The run did not widen
its scope to absorb it.

**Comment by leodip** [comment]
## The code review, in full

`LEO_REVIEW_MODE=end` was in force for this run, so **no stage was reviewed on its own**: the whole
branch was read from a fresh session after every stage had landed, twice. The plan was reviewed
separately before any code existed. The verdicts themselves live in a gitignored directory that the
next round overwrites, so this comment is the durable account of them.

Reviewer: `gpt-5.6-sol` at effort `xhigh` for the plan gate, `deepseek/deepseek-v4-pro-0813` at
effort `xhigh` for both code rounds, taken from `.review/reviewer.env` rather than inferred.

|---|---|---|---|---|
| plan | 1 | 4 | 2 | 3 fixed in stages 1, 2 and 4; 1 escalated as decision 11 and answered `A` |
| final | 1 | 2 | 0 | both fixed: `432725f0`, `e3fb542e`, `719a9e75` |
| final | 2 | 1 | 0 | fixed: `77dc366a` |

Nothing was refuted and nothing was deferred. Twelve mutations were run against the new tests across
the two code rounds; eleven were caught and the one survivor changed the code rather than being
recorded and left.

---

## The rounds

### Plan gate, round 1, `gpt-5.6-sol` at effort `xhigh`

Request `284-20260828T220429Z`, `type: plan-review`. Verdict: `findings`, all three axes reviewed.
Four findings: two blocking conformance, two significant. Three carried `needs_human: true`. No
production code existed at this gate; the branch was one empty commit ahead of `origin/main`.

|---|---|---|---|---|
| 1 | conformance | blocking | The normalised shape cannot see column generation and collapses a composite foreign key into two scalar ones, so different schemas produce identical golden files | **Escalated** as decision 11, open |
| 2 | conformance | blocking | Count-only allowlist rules pass an equal-count membership substitution, and canonicalising signedness and string length erases two exceptions before they can be counted | **Fixed** in stage 4 |
| 3 | conformance | significant | MySQL's catalog cannot tell a migration-chosen index name from an engine-invented one, and SQL Server's default projection omits `is_system_named` | **Fixed** in stage 1 |
| 4 | quality | significant | Pre-creating `schema_migrations` before `sqlserver.WithInstance` moves a check-then-create outside golang-migrate's advisory lock | **Fixed** in stage 2 |

**Finding 1** is the one the run would not answer for the user. `forced_answer` was empty, which
`decisions.md` section 1 reads as a choice rather than a forced repair, and it is schema territory,
which is never auto-resolved. It was escalated as the class rather than the instance: what the model
covers, and what happens to a construct outside it. Measured before asking, so the question carries
numbers rather than possibilities. Column generation is used on every table by every engine (26
`AUTO_INCREMENT`, 26 `IDENTITY`, 26 `BIGSERIAL`, `AUTOINCREMENT` throughout SQLite) and is invisible
to the shape. The composite-foreign-key collision is real and currently empty: no engine has one.
Seven further constructs the shape cannot represent are absent from all four engines today (CHECK,
generated columns, `ON UPDATE`, partial indexes, `INCLUDE`, descending keys, `DEFERRABLE`). One part
did not wait for the answer, because it is common to every option offered: the dumper now errors on
a construct it cannot represent instead of dropping it silently.

**Finding 2** was resolved on the agreement contradicting itself rather than on preference. Decision
5's prose sends `bigint unsigned` and SQLite's `TEXT` through canonicalisation and says such
differences "never reach the allowlist"; decision 5's own table then counts "SQLite records no
length for a string column" at 89, and section 2 says #282's left-alone list, which names both,
"become allowlist entries, not fixes". Canonicalising spelling while leaving length and signedness
comparable is the only reading under which all three sentences do something. The membership half, a
digest over each rule's sorted instance keys beside the count it already carries, only narrows what
is accepted and costs one line per rule, so it does not resurrect the instance-scoped form decision
5 rejected for being unreadable. The reviewer reached the same answer independently via
`forced_answer`.

**Finding 3** split. The MySQL half is a correct refutation of something the plan asserted, and it
does not bite: MySQL invents an index name only for an index declared without one, and every index
in `src/core/data/mysqldb/migrations/` is named, so the guard enforces a property the schema already
has instead of inferring one the catalog cannot report. The SQL Server half is forced by section
4.2, which masks engine-generated default names and cannot do so from `dc.name` alone when 42 of 64
defaults are unnamed; the projection reads `is_system_named`, recorded as a deviation from section
4.1's "move verbatim".

**Finding 4** was resolved rather than escalated because nothing was being chosen between materially
different outcomes: decision 7 already decided the pre-create, and only its SQL Server spelling was
open. Taking the driver's own `sp_getapplock` around check and create restores a property
golang-migrate already had. The alternative, tolerating `Msg 2714` and re-reading the shape, is more
machinery for the same invariant.

All four `unchecked` entries were answered, two covered and two dismissed with reasons, in
`log/plan.md`. Both reviewer follow-ups were dispositioned: the `check-migrations.sh` repoint is
already section 4.7's and stage 6 records it in the PR body; the `NewMsSQLDatabase` race is drafted
above as F1.

Round 1 changed the plan and the agreement and no production code, which is the expected shape of a
plan gate. The round cap is 2 and no second round was requested: three findings are resolved in
writing and the fourth is the user's to answer, so a second review before the answer would read the
same plan with one question still open on it.

---

## Section 10, final review round 1: request written (2026-08-29)

`LEO_REVIEW_MODE=end` was in force for the whole run, so no stage was reviewed on its own and no
code on this branch has been read by a reviewer. `.review/request.md` is now written for the first
and only reading: `type: code-review`, `gate: final`, `round: 1`, `status: awaiting-review`, scoped
to `git diff origin/main...HEAD` in full, 39 files, +6960 / -3571 across 7 commits. All three axes
in scope. The plan-gate review is named in it for what it was: a reading of `plan.md` before any
code existed, whose one open question became decision 11.

**The check suite is green on the final commit `ab4e13a8`**, all eleven jobs:
`Tests / sqlite`, `Tests / mysql`, `Tests / postgres`, `Tests / mssql`, `Unit / core`,
`Unit / authserver`, `Unit / adminconsole`, `Lint`, `Vulnerabilities`, `Versions` and
`Build validation`. Run: https://github.com/leodip/goiabada/actions/runs/33227029729

That is the authoritative four-engine data and integration result for this branch, and it satisfies
section 5's precondition for marking the pull request ready. It does not release the finish on its
own: `.review/final-review.md` is still present, so the review has to conclude first.

**`tree-hash.sh --verify` says 5 of 6 stage entries record results against a tree that has since
moved**, which no session inside a stage was placed to notice. Stages 1 to 5 are all stale on
`production`, and with it every mutation those stages recorded; only stage 6's entry is live, and it
carries modules 1276 passed and four-engine data 1520 passed against the current tree. The full
block is pasted into the request's `## Tiers` section rather than summarised, so the reviewer can see
per tier which claims are still free to believe and which cost a re-run. Stale is not a finding: it
is what landing five more stages does.

## Section 10, final review round 1

**Reviewer** `deepseek/deepseek-v4-pro-0813` at effort `xhigh`, fresh session, started
2026-08-29T01:50:14Z, taken from `.review/reviewer.env` rather than inferred. Gate `final`,
round 1 of at most 4. Scope was the whole branch against its base, 39 files, +6960 / -3571,
because `LEO_REVIEW_MODE=end` meant no stage was reviewed on its own. Axis coverage: security,
conformance and quality all `reviewed`.

**Verdict: findings.** Two, both `significant`, both `conformance`, both `confidence: high`,
both `needs_human: true`. No security findings. Both were verified against the code, both stood,
and both are fixed on the branch. Neither was refuted and neither was deferred.

### Finding 1, significant, conformance: the guard does not refuse an expression index

*Claim.* Decision 11 settles that the dumper fails when it meets a construct the shape cannot
represent rather than dropping it silently. An expression (functional) index is such a
construct, `guardTable` had no check for it, and on PostgreSQL it is dropped from the dump with
no error at all, so a migration adding one on one engine would pass both the per-engine
assertion and the cross-engine comparison.

*Verified, and it stands.* Read directly out of `dump/indexes`: the PostgreSQL branch joins
`unnest(ix.indkey)` onto `pg_attribute` on `a.attnum = k.attnum`, and an expression's `indkey`
entry is `0`, which matches no attribute on any ordinary table. So the row is dropped by the
join. A wholly expression-based index disappears from the dump entirely; a mixed `(a,
lower(email))` index keeps only `a` and is recorded as single-column. The reviewer's own probe
found the same thing by building the DDL and running the dumper's query, and reported zero rows.

The finding also claimed MySQL and SQLite fail rather than dropping. That half is right about
the direction and understates the problem: they fail, but opaquely, and after the guard would
have spoken. Demonstrated rather than assumed, by neutering the new SQLite predicate and
reading what surfaced in its place:

```
schemadump: scan index catalog row on sqlite.guard_probe: sql: Scan error on column
index 3, name "name": converting NULL to string is unsupported
```

That names neither the construct nor the remedy, and a migration author meeting it has nothing
to act on.

*Fixed* in `432725f0`. `unrepresentableChecks` gains "an expression (functional) index" on the
three engines that can express it, each reading its own catalog's answer: `indkey` holding `0`
on PostgreSQL, `COLUMN_NAME IS NULL` on MySQL, `pragma_index_info` reporting a NULL name on
SQLite. SQL Server is deliberately absent, with the reason in the code: it has no expression
index, and its nearest equivalent is an index over a computed column, which the
generated-column check already refuses.

`guardTable` also **moves ahead of** `dumpIndexes` and `dumpForeignKeys` in `DumpTable`. That
is what turns the MySQL and SQLite halves from the scan error above into the guard's sentence,
and it is not in the finding: it came out of the fix. It stays behind the column projection, so
a table that does not exist is still answered by the existing read-no-columns error rather than
by guard queries returning zero of everything, which keeps
`TestSchemaDump_RefusesWhatItCannotRead` meaning what it meant.

One guard case per engine is added to `TestSchemaDump_RefusesUnrepresentableConstructs`, in the
existing table-driven shape.

*The finding's `why_exhaustive` conceded its own boundary was not established*: it checked the
nine listed constructs and probed expression indexes, but did not enumerate every other index
construct each engine can express, naming MySQL JSON multi-valued indexes and SQL Server
spatial indexes. That is honest and it is not closed here. See section 9.

### Finding 2, significant, conformance: a defect the run itself recorded was never fixed

*Claim.* `parityTypes` maps SQLite `text` and `longtext` onto one value, so the cross-engine
comparison cannot see `browser_sessions.data`'s SQLite-only spelling drift. Stage 3 recorded
that column as the first defect the instrument found and said decision 6 put it on this branch
in stage 5. Stage 5 did not fix it, did not allowlist it, and recorded no reclassification.

*Verified on every claim, and it stands.* `src/core/data/sqlitedb/migrations/000035_browser_sessions.up.sql:34`
declares `data longtext`. Stage 3's log entry says, in its own words, that this "is exactly the
class of thing #284 exists to find, and decision 6 puts it on this branch, in stage 5, rather
than on a follow-up". Stage 5's entry fixes three defects by migration `000042` and never
mentions this one. The committed `sqlitedb/schema.golden` still read `type="longtext"`. And
`parityTypes` gave SQLite's `text` and `longtext` the identical
`{family: parityString, bound: parityBoundUndeclared}`, so the two spellings were the same
value and the comparison could not have seen the difference.

This is the most serious kind of finding this gate can produce, because it is the run failing
to deliver something the agreement required and reporting success: decision 6 is the user's own
answer and it says nothing is deferred to a follow-up and no defect goes on the allowlist.

*Fixed* in `719a9e75`, in both halves, because either alone would leave the hole.

The **column**: migration `000043_browser_sessions_data_text`, SQLite only, rebuilding the
table in `000041`'s shape, since SQLite cannot ALTER a column's type. Safe in the way decision
9 requires: nothing holds a foreign key onto `browser_sessions` on any of the four engines, so
the implicit DELETE behind `DROP TABLE` has nowhere to cascade and no table needs parking
first, which is why this is the plain rebuild and not `probe/sqlite_rebuild_fk.out`'s variant
E. Both indexes are recreated. One consequence is accepted and recorded in the file: the
`sqlite_sequence` row goes with the table and is re-established at the highest id copied
across, so AUTOINCREMENT's never-reused guarantee breaks once. `id` is a surrogate key, a
session is found by `(owner, session_id_hash)`, `BrowserSession.Id` reaches no cookie or token,
and no table references it.

The **vocabulary**: the SQLite `longtext` entry is **removed** from `parityTypes` rather than
given a distinct value, which is where this departs from the finding's `forced_answer`. The
finding proposed keeping it distinct so the divergence would reach the allowlist. Once the
column is fixed there is no SQLite `longtext` left to excuse, and the table's own contract is
that "a spelling that is not here is a failure naming it rather than a guess". Absent, the
spelling is refused by name the moment it reappears, which is strictly stronger than an
allowlist rule and needs no count to maintain. A comment says so in place, because the entry's
absence is a fix and adding one back would undo it.

`sqlitedb/schema.golden` regenerated with `go run ./cmd/schemadump` in the dev container, all
four engines in one process. It moved by **exactly one line**, `longtext` to `TEXT`, and the
other three files did not move at all. That is also the evidence that the rebuild carried every
other column, index and foreign key across untouched, since the golden file records all of them.

### The reviewer's `unchecked` list, answered

- *"Did not re-run the full four-engine data tier or any integration tier."* **Covered.** Both
  run this session, below.
- *"Did not execute `go run ./cmd/schemadump` end to end because it writes the real golden files
  in the working tree."* **Covered.** Run here, where writing them is the point. All four
  engines succeeded and the diff was the single line above.
- *"Did not independently re-derive catalog shape across four live databases beyond the
  committed golden files."* **Covered** by the four-engine data tier, which is exactly that:
  `TestSchemaGolden_MatchesTheCommittedFile` migrates a fresh isolated database to head on each
  engine and compares the dump against the committed file. It passed on all four.
- *"Did not fetch an external specification: the change makes no new normative claim that needs
  one."* **Dismissed**, and agreed. Section 1 records that no normative source governs a parity
  test, and the one axis with specification behind it, collation, was settled by #283 and is
  inherited rather than chosen here.

### Tiers, after both fixes

| Tier | Command | Engines | Result | Time |
|---|---|---|---|---|
| unit | `where.sh test --type modules` | n/a | **1276 pass, 0 fail** | 11s |
| data | `where.sh test --type data` | mysql, postgres, mssql, sqlite | **2043 pass, 0 fail**, all four completed | 2m9s |
| integration | `where.sh test --type integration --db sqlite` | sqlite | **1012 pass, 0 fail** | 3m |
| migrations gate | `where.sh migrations` | all four | all rules pass, next free number 000044 | instant |
| anchors | `check-anchors.sh` | n/a | 22 of 22 resolve | instant |

Integration was run locally despite the change touching no handler or route, because it adds a
migration, and the schema is the other thing that rule names.

Tree at the time of these results, `tree-hash.sh --split`:

```
production 1a713589aae8   unit-tests f647fc3877bf   data-tests 3b859502de88
integration-tests 7b69ee4ead08   docs 83265345312f   all 511c319f1be8
```

`production` was checked against this value after the last mutation restored, per section 4.4,
before anything was committed. `docs` and `integration-tests` are unchanged from the request's
`--verify` block, which is correct: neither fix touches documentation or an integration test.

### Mutations, four, all caught

|---|---|---|---|
| `schemadump/guard.go` | `AND k.attnum = 0` to `AND k.attnum = -1` | data, postgres | **CAUGHT** |
| `schemadump/guard.go` | `WHERE ii.name IS NULL` to `WHERE 1 = 0` | data, sqlite | **CAUGHT** |
| `schemadump/guard.go` | `AND COLUMN_NAME IS NULL` to `AND 1 = 0` | data, mysql | **CAUGHT** |
| `000043_..._data_text.up.sql` | `data TEXT` back to `data longtext` | data, sqlite | **CAUGHT** |

One per engine that can express an expression index, and one for the migration. The SQLite one
is also what produced the scan-error evidence quoted under finding 1.

### One correction to the record, made here

The round-1 request told the reviewer that `browser_sessions.data` was among what stage 5 had
settled, by saying section 3's decisions were all `Decided` and that all defects found were
fixed. That was wrong in the way finding 2 describes, and the finding is right that
`log/closing.md` carried the same claim. This entry is the correction, and section 9 carries no
follow-up for it because the defect itself is now fixed rather than deferred.

**Both findings fixed, no follow-up filed for either.** Round 2 requested, because a table
rebuild and a widened guard are new production code that has not been read by a reviewer.

## Section 10, final review round 1, addendum: the family closed

Finding 1's `why_exhaustive` stated its own boundary and did not reach it: it had checked the
nine listed constructs and probed expression indexes, but had "not enumerate[d] every other
index construct each engine can express (for example MySQL JSON multi-valued indexes or SQL
Server spatial indexes), so the family boundary is not fully established beyond expression
indexes."

Section 4.4 says resolve the family rather than the instance, so the members were enumerated
here instead of being left for round 2 to rediscover one at a time.

**The two the reviewer named are covered, and neither needed new code.** A MySQL multi-valued
index is a functional key part, so `information_schema.statistics.COLUMN_NAME` is NULL and the
expression check refuses it. A SQL Server spatial index is `sys.indexes.type = 4`, which the
first of the two checks below refuses.

**Two genuine members turned up while establishing that**, both fixed in `e3fb542e`:

1. **An index built with a non-default method.** `IndexShape` records key columns and
   uniqueness and nothing about how the index is built, so a GIN, hash, full-text, spatial or
   columnstore index over one column is the same shape as a plain B-tree over that column.
   Replacing one with the other on a single engine compared equal on every axis this change
   has. Not exotic for this schema either: full-text search over `users`, or a JSON index on
   `audit_logs.details`, are both things a later migration might reach for. Refused on the
   three engines with a second method, each from its own catalog: `pg_am.amname`,
   `information_schema.statistics.INDEX_TYPE`, `sys.indexes.type`. SQLite has only a B-tree.
2. **An index key with a non-default null ordering.** The sibling of the descending-key check
   already present, on the one engine where null ordering is choosable apart from direction.

**One defect in the first draft of that, caught by the data tier and worth recording**, because
it is the same class of mistake as the finding that started this. The null-ordering predicate
was first written as `(o.opt & 2) = 2` alone, on the reasoning that a plain `DESC` key would be
refused by the descending check first. That reasoning was about slice order and it was wrong:
the new check sits earlier in `unrepresentableChecks`, so `a DESC` (which sets both `indoption`
bits, NULLS FIRST being the default for a descending key) was reported as a null-ordering
problem and `TestSchemaDump_RefusesUnrepresentableConstructs/a_descending_index_key` failed on
PostgreSQL naming the wrong construct. The predicate now carries `AND (o.opt & 1) = 0`, so it
depends on the index rather than on the order of the checks. Between the two checks all four
direction-and-nulls combinations are refused.

**Tiers after this**: data on all four engines **2047 pass, 0 fail**; unit **1276 pass, 0
fail**. The four-engine golden assertion passing is also the evidence that neither new check
false-positives on the real schema, since it dumps every table on every engine.

**Four more mutations, all caught**, taking round 1 to eight:

|---|---|---|---|
| `schemadump/guard.go` | `AND am.amname <> 'btree'` to `<> 'nonesuch'` | data, postgres | **CAUGHT** |
| `schemadump/guard.go` | `AND (o.opt & 2) = 2 AND (o.opt & 1) = 0` to `AND 1 = 0` | data, postgres | **CAUGHT** |
| `schemadump/guard.go` | `AND INDEX_TYPE <> 'BTREE'` to `<> 'FULLTEXT'` | data, mysql | **CAUGHT** |
| `schemadump/guard.go` | `type NOT IN (1, 2)` to `type NOT IN (1, 2, 6)` | data, mssql | **CAUGHT** |

Tree at these results: `production caba8bd29009  unit-tests f647fc3877bf  data-tests
88c9c09c4481  integration-tests 7b69ee4ead08  docs 83265345312f  all f461254d252b`. Checked
after the last mutation restored and before the commit.

**No follow-up is filed for the family.** It is closed for index constructs as far as the four
catalogs can express them, and what remains outside it is the shape's known and deliberate
limits, which decision 11 settled and section 9 already records.

## Section 10, final review round 2

`deepseek/deepseek-v4-pro-0813` at `xhigh`, fresh session, over the whole branch against its
base. **One finding, significant, conformance. Verified, it stood, and it is fixed in
`77dc366a`.** Round 1's three fix commits, which were the thing round 2 most needed read, drew
no finding: the SQLite table rebuild in `000043`, the widened guard in `schemadump/guard.go`,
and the removed `parityTypes` entry all checked out.

### Finding 1: the index-name comparison was all-or-nothing across the four engines

**The claim.** `compareIndexes` compared an index name only when every engine's index for that
key had `Origin == OriginCreated`. One engine-named index in the set, an inline `UNIQUE`, a
SQLite rowid primary key or a SQL Server unique constraint, and the name comparison was skipped
entirely, so a name a migration misspelled on one of the other three went uncaught.

**Verified, and it stood.** The predicate was a single `migrationNamed` boolean ANDed over all
four engines, and the code comment beside it argued only the half that is true: an engine-named
index says nothing about what a migration spelled. It says nothing about the engines beside it
either, and those are the ones a later `DROP INDEX` breaks on, which is the entire reason
decision 2 compares names at all rather than keying on the tuple alone.

**This is decision 2's intent against decision 2's wording**, and that is worth stating plainly
rather than burying. The decision's headline reads "with the name compared as well wherever all
four names came from a migration", and the code implemented that literally. Its rationale, two
paragraphs down, is the misspelled-name class: "`idx_email` misspelled on one engine breaks that
migration on that engine only, and nothing else would catch it." The all-four rule does not
serve that rationale, it defeats it on exactly the keys where engines spell an index differently.
Resolved rather than escalated, under `decisions.md` section 1: the remedy only narrows what the
parity check accepts, changes no interface, touches no migration, no schema and no production
code, costs nothing at runtime, and arrived with a populated `forced_answer` the reviewer
independently demonstrated with a scratch probe.

**Measured before changing anything**, because a widened check that reports on the current schema
would be a divergence to fix under decision 6 rather than a test change. Over the 76 index keys
in the four committed golden files:

| Shape | Keys |
|---|---|
| every engine migration-named, already compared | 48 |
| no engine migration-named, nothing to compare either way | 25 |
| mixed, **each with exactly one** migration-named engine | 3 |

The three are `client_logos.unique index(client_id)` and
`user_profile_pictures.unique index(user_id)`, MySQL naming the unique index by hand where the
other three declare it inline, and `schema_migrations.unique index(version)`, SQLite's
`version_unique` against a primary key on the other three. Every one is a singleton with nothing
to compare, so the fix reports nothing new today. The four-engine golden assertion passing after
it is the evidence, since it dumps every table on every engine.

**The fix**, in `77dc366a`, test-only. `appendIfDisagreedAmong` compares an axis over a subset of
the engines and fills the rest of `Says` with `-`, so the failure line still carries a column per
engine; `appendIfDisagreed` delegates to it with all four. `compareIndexes` builds the set of
engines whose indexes for the key are all `OriginCreated` and compares names across that.

**Where the fix departed from the `forced_answer`, so it can be argued with.** The reviewer
proposed guarding on the set having at least two members. That guard was written, and then
removed: one engine agrees with itself and none agree vacuously, so the loop already returns
nothing for both, and the mutation below proves the branch was unreachable. A branch no mutation
can falsify is worse than no branch. The intent moved into the doc comment instead.

### The family, and the sibling the reviewer did not probe

`siblings_checked` listed four members and marked the fourth unprobed: "two or more engines
engine-named with the migration-chosen names disagreeing". It is now a case, and it passes. All
four members are pinned by name:

| Member | Case | Reports |
|---|---|---|
| all four migration-named, one misspelled | `an index name is compared when every engine's came from a migration` | yes |
| one engine-named, the rest agreeing | `an engine that invented its own name drops out and the rest still agree` | no |
| one engine-named, one of the rest misspelled | `but it does not switch the comparison off for the engines beside it` | **yes, new** |
| two engine-named, the other two disagreeing | `nor when two engines invented theirs and two migrations disagree` | **yes, new** |
| three engine-named, a singleton left | `and one migration-named engine alone is a singleton with nothing to compare` | no |
| all four engine-named, nothing left | `and no migration-named engine at all is no comparison` | no |

The last is not reachable from the four golden files today and the case says so: the one index
every engine names for itself is the primary key, and SQLite builds none at all for a rowid
alias, so presence disagrees before a name is ever compared. A table declaring an inline `UNIQUE`
on all four engines would reach it.

### Mutations: three caught, one survived and changed the code

|---|---|---|---|
| `schema_parity_test.go` | `if len(among) < 2 {` to `if len(among) < len(parityDialects) {` | modules | **CAUGHT** |
| `schema_parity_test.go` | `if named {` to `if false {` | modules | **CAUGHT** |
| `schema_parity_test.go` | `says[d] == says[among[0]]` to `says[d] == says[parityDialects[0]]` | modules | **CAUGHT** |
| `schema_parity_test.go` | `if len(among) < 2 {` to `if len(among) < 0 {` | modules | **SURVIVED** |

The first restores the defect exactly and the new cases catch it. The third is the one that
distinguishes comparing within the subset from comparing against the full set: with it, the
singleton case reports `mysql=idx_widgets_name` against `sqlite=-`.

**The survivor was a finding about the change, and it was acted on rather than recorded and
left.** It survived twice: first against the real golden files, and again after a case was added
putting all four engines' index in the engine-named shape. The second run is what showed why. The
loop dereferences `among[0]` inside its body, which does not execute on an empty set, so the
guard was not preventing a panic and not changing any answer. It was removed, and the two
boundary cases stay as the statement of the behaviour.

### The `unchecked` list, answered in full

1. **CI pending on `e3fb542e`.** *Covered.* It has since concluded **green on all eleven jobs**:
   `Build validation`, `Lint`, `Vulnerabilities`, `Versions`, `Unit / core`, `Unit / authserver`,
   `Unit / adminconsole`, and `Tests /` on mysql, postgres, mssql and sqlite. That is the
   four-engine claim for round 1's fixes, which no local run could have produced.
2. **The full data and integration tiers were not re-run, only spot-checked.** *Covered.* Both
   tiers were run here after the fix: unit **1276 pass, 0 fail**, data **2059 pass, 0 fail** on
   all four engines with `TestSchemaGolden_MatchesTheCommittedFile` green on each. Integration
   was not re-run and does not need to be: `production` and `integration-tests` are byte-identical
   to the tree the recorded 1012-pass sqlite run executed against, so that result still describes
   this tree. The fix touches two `_test.go` files in `src/core/data`, which the modules tier runs.
3. **No external specification fetched.** *Dismissed.* Section 1 records that no normative source
   governs a parity test, and the one axis with an external authority behind it, collation, was
   settled by #283 and is inherited rather than chosen here. The reviewer's reason is the right one.
4. **Security scoped to `NewMigrator` and the three migrations.** *Covered*, and it is a coverage
   statement rather than a gap: the diff touches no auth, token, session, PKCE, redirect, scope,
   rate-limit or audit surface, and this round's change is test-only.

### Tiers and tree

|---|---|---|---|
| unit | `where.sh test --type modules` | n/a | **1276 pass, 0 fail** |
| data | `where.sh test --type data` | mysql, postgres, mssql, sqlite | **2059 pass, 0 fail** |
| migrations gate | `where.sh migrations` | all four | all rules pass, next free 000044 |
| anchors | `check-anchors.sh` | n/a | 22 of 22 resolve |

Tree after the last mutation restored and before the commit: `production caba8bd29009
unit-tests aabcba511a7a  data-tests 88c9c09c4481  integration-tests 7b69ee4ead08  docs
83265345312f  all f58a6efbc5a2`. **`production` is byte-identical to the value round 2 was
reviewed at**, which is both the proof that every mutation was restored and the statement that
this round changed no production code.

### Round 2 is the last round

Nothing is left open: the single finding is fixed, its family is enumerated and pinned, and every
`unchecked` entry is answered above. No follow-up is filed and nothing is deferred. Round 2 also
moved no production code, so under section 4.4's budget a round 3 is not funded, and that is the
right answer rather than a constraint worked around: the reviewer has now read the branch twice
and the second reading found one test-level gap.

PR #294, opened 2026-08-29
test(core): the four migration directories are held to eleven source-text rules

### LINKED ISSUE #288: Nothing checks that a new migration was written the same way on all four engines

leodip:
Closes #288.

Goiabada's migrations are written four times by hand, once each for MySQL, PostgreSQL, SQL Server
and SQLite, and nothing in this repository reads their source text. #282 found five structural
divergences that had shipped for months. #283 found `=` meaning case-sensitive on two engines and
case-insensitive on the other two, which is a conformance failure rather than a style problem: RFC
6749 section 1.9 makes protocol parameter values case-sensitive, section 3.3 says the same of
`scope`, and OIDC Core section 2 says it of `sub`. All three are stored in string columns and
compared with `=`, so the column's collation is where the requirement is met or missed.

Both defect classes were invisible in review and both were cheap to prevent at the moment the file
was written. This adds the check that would have caught them.

## The rule it establishes

A pull request cannot land a migration that breaks four-engine parity in a way that is decidable
from the source text, and the check runs on the same commits a person hand-writes.

It is a Go test under `src/core/data`, beside the cross-engine golden comparison #284 landed, so it
runs in the core tier on every CI job with no new job, no new host tooling and no database. Eleven
rules over the four migration directories:

| Rule | Engines | Applies above | Refuses |
|---|---|---|---|
| naming | all | every number | a filename that is not `NNNNNN_lower_snake.{up,down}.sql` |
| pairing | all | every number | an `.up.sql` with no `.down.sql`, or the reverse |
| one number, one change | all | every number | a number whose slug differs between engines |
| coverage | all | every number | a partial migration that does not declare its engine set, and a complete one that declares anything |
| no bare `VARCHAR` | mssql | 000014 | `VARCHAR(n)`, which loses anything outside the code page under a non-UTF-8 server collation |
| collation pinned | mssql | 000038 | a string column that does not spell `Latin1_General_100_CS_AS_KS_WS_SC_UTF8` |
| named defaults | mssql | 000018 | `ADD ... DEFAULT` with no `CONSTRAINT` name, which gets a per-database generated one |
| nullability restated | mssql | every number | `ALTER COLUMN c <type>` with no `NULL` or `NOT NULL`, which silently drops `NOT NULL` |
| one collation | mysql | 000035 | a `COLLATE` that is not `utf8mb4_0900_as_cs`, and a `CREATE TABLE` that spells none |
| reversibility declared | all | every number | a `.down.sql` with no statement that does not declare itself an intentional no-op |
| golden records its version | all | every number | a committed `schema.golden` whose recorded migration version is not that engine's highest |

Every cutoff is measured rather than chosen: it names the last migration in the tree that breaks
that rule, and a companion test requires that migration to actually break it, so a cutoff cannot be
inflated to step over a migration being written now. Nothing in the tree is being repaired here.
PostgreSQL and SQLite carry no content rule, because both compare byte-wise already and neither has
a per-column collation to forget.

The eleventh rule replaces the git-diff form the issue proposed, which cannot work in CI:
`check.yml` checks out at depth 1, so on a pull request there is no local `main` and
`git merge-base HEAD main` fails outright, and on a push to `main` the merge base is `HEAD` and the
diff is empty. Instead the golden file records the migration version of the catalog it was dumped
from, and the lint holds that number to the highest migration file on disk for that engine. No git,
no database, and it fires in a tarball and a worktree alike.

Green here is a floor. Whether the four migrated catalogs actually agree on columns, indexes,
defaults and foreign keys is #284's, landed in #290, and this does not touch it.

### Stage 1: the golden file records its engine's migration version

Landed in `4b5e8f5c`.

The golden format goes to version 3, whose header carries `migrated=<n>` beside `version=` and
`engine=`. `schemadump.Golden` now holds the three things a golden file records, and `Encode` takes
it where `Parse` returns it, so a header field cannot be dropped on the way through. `Encode`
refuses a non-positive version, and the parser refuses a header with no `migrated=` field, one whose
value is not an integer, and one that is not positive, each with its own message: a field the writer
never wrote is repaired by regenerating against a newer binary, and a corrupt value by restoring the
file.

`schemadump.MigratedVersion` reads the number off a live database with golang-migrate's own query,
`SELECT version, dirty FROM schema_migrations`, which is identical on all four engines because #284
pinned that table to one shape. Every failure is an error and never a benign zero, a dirty row and
an empty table included: a reader that answered 0 when the table was missing or half-applied would
make the version rule compare 0 against 0 and pass on a file nobody regenerated.

The cross-engine parity comparison deliberately does not read the number. The four legitimately
differ, `mssql 40`, `postgres 39`, `mysql 42`, `sqlite 43`, because a migration is free to land on
fewer than four engines.

All four golden files regenerated. Exactly one line moved in each, the header, which is what a
change that alters no catalog shape should produce.

**Tiers:** unit 1277 pass, 0 fail. Data on mysql, postgres, mssql and sqlite: 1528 pass, 0 fail.
`where.sh migrations` green. Integration not applicable, nothing here is reachable through an
endpoint. Seven mutations against the new seams, all caught; two survived on the first attempt and
were fixed in the tests, where two adjacent refusals overlapped so that asserting only "an error
occurred" left one branch unexercised.

### Stage 2: the nine undeclared partial migrations declare themselves

Landed in `5871fe8d`.

Ten migration numbers exist on fewer than all four engines. Four of them already opened with a
`-- parity: <engines> only.` line saying so; the other nine files, across six numbers, said nothing,
so a reader had to diff four directories to learn whether a missing file was deliberate or
forgotten. The nine now carry the same declaration: `000002` on MySQL and SQLite, `000015` and
`000038` on SQL Server, `000036` on SQLite, `000037` on MySQL, and `000039` on MySQL, PostgreSQL and
SQLite.

Each declaration names the engines that carry the number and then says, in its own sentence, why the
others need no file, so the line is a declaration rather than a label. Seven of the nine already
explained themselves in prose, which is left exactly as it was below a `--` separator. Every reason
was verified against the tree rather than asserted: PostgreSQL and SQL Server were both written
after v0.7, so their `000001_initial_create` already builds the unique index and all three columns
`000002` adds and neither ever created the two `settings` columns it drops; SQL Server has had
`refresh_tokens` on `ON DELETE NO ACTION` since its `000011` and nullable PKCE columns since its own
`000007`, which is why it needs no `000039`.

No DDL byte moves, and that is what makes this safe on a deployed database. golang-migrate v4.19.1
records only `(version, dirty)` and no checksum of the migration file, so a comment added to an
already-applied migration is inert, where a DDL edit would make a fresh database differ from every
existing one.

The rules that will enforce this arrive in stages 3 and 4. The same scan was run by hand here: all
14 partial `.up.sql` files declare, every declaration's engine list equals the set of engines
carrying its number, every one is followed by prose, and no migration present on all four engines
declares anything.

**Tiers:** unit 1277 pass, 0 fail. Data on mysql, postgres, mssql and sqlite: 1528 pass, 0 fail,
with `TestSchemaGolden_MatchesTheCommittedFile` green on all four, which is the direct evidence that
a comment changes no catalog and owes no golden regeneration. `where.sh migrations` green on all ten
of its rules. Integration not applicable, no handler, route or schema touched. No mutation is owed:
the stage adds no test.

### Stage 3: the rule engine and the five structural rules

Landed in `200eb5de`.

The machinery and the first five rules, as a Go test in `package data` beside the cross-engine
golden comparison. It reads files and nothing else: no database, no git, no network, so it runs in
the core tier on every CI job rather than only the four database ones, and gives the same answer in
a tarball, in a worktree and on a shallow clone.

`scanSQL` reads a migration once and returns it with every comment byte blanked to a space, the same
length as the input so every offset still names its own line, plus one mark per `;`, `,`, `(` and
`)` that sits outside a comment, a single-quoted literal and a quoted identifier, each carrying its
parenthesis depth. Statements and `CREATE TABLE` / `ALTER TABLE` declarations then come back as
addressable units, which is what will let a content rule bind to one column rather than to a whole
file: a second `CREATE TABLE` in a file whose first one spells its collation cannot pass on the
first one's strength.

The lexical vocabulary is measured off the four directories rather than assumed. Identifiers are
`[...]` on SQL Server and backticks on MySQL and on SQLite, which uses them 109 times for MySQL
compatibility; PostgreSQL quotes no identifier anywhere. A doubled delimiter escapes on every
engine, which `mssqldb/000029` spans two lines with. A backslash escapes inside a literal on MySQL
alone, which `000034` forces, since MySQL writes `'\\'` where the other three write `'\'`, and
reading SQL Server's form under MySQL's rules would leave the quote open and the rest of the file
would read as string interior. Four forms are refused rather than guessed at, none of which any
migration uses: a block comment, a top-level `$`, a `"`, and unbalanced state at end of file.

The five rules. **naming**: six digits and a `lower_snake` slug, with SQLite's five-digit
`00001_initial_create` its only named exemption, so no new five-digit name can land beside it.
**pairing**: every `.up.sql` has its `.down.sql` and back. **number identity**: one number names one
change on every engine carrying it. **coverage**: a partial migration declares which engines carry
it and why, with the declared set required to *equal* the actual one, and a complete migration
declares nothing; the declaration is confirmed to be a real comment through the scanner rather than
by pattern alone. **reversibility**: a `.down.sql` holding no statement declares itself an
intentional no-op under its own number, which all twelve statement-free downs already do.

The migration directories are **discovered** and compared against the four dialects in both
directions, rather than iterated from the four constants. Requiring each of the four catches a
directory that went missing; refusing a name that is not one of them catches a directory that was
added, which a loop over four constants cannot see at all. A fifth engine skipped in silence is the
exact failure this rule exists to prevent.

**Tiers:** core tier, 4323 pass, 0 fail, 10s, including the five new test functions and their 43
subtests and the lint against the four committed directories, which pass as they stand. Data and
integration are not applicable: nothing here runs a query, touches the schema or is reachable
through an endpoint, and the new tests run in the core tier that every CI job runs. `where.sh
migrations` green on all ten of its rules; `check-anchors` green.

**Mutations:** eleven run, eleven caught. Ten break this stage's own code, one per rule plus the
scanner's literal tracking, its unbalanced-state check, its MySQL backslash handling and the
symmetric directory comparison. The eleventh breaks a real migration file, `sqlitedb/000036`'s
`-- parity:` line, and requires the lint to notice. That last one is the only evidence the lint
reads the committed tree rather than passing because it read nothing, and the synthetic half cannot
establish it. Every file was restored and confirmed by hash.

### Stage 4: the six content and version rules, their cutoffs, and the guard's description

Commit `08be5533`. The rest of the rule table, and the guard's description in both catalogs.

Five content rules over the statements and declarations stage 3 splits, and the crucial thing about
them is what each one binds to. The SQL Server rules bind to one **declaration**, one column, and
the MySQL rules to one **statement**, one table. Never the file. A check asking whether the
collation pin appears anywhere in the file passes a `CREATE TABLE` whose second string column
spells none, because the first one did, and passes a second `CREATE TABLE` that silently inherits
the database default. Both shapes are reachable: the tree already holds a file with 22
`CREATE TABLE`s in it.

| Rule | Engine | Binds to | Cutoff | Refuses |
|---|---|---|---|---|
| `mssql/nvarchar` | SQL Server | declaration | 000014 | `VARCHAR(n)`, which loses anything outside the code page under a non-UTF-8 server collation |
| `mssql/collate` | SQL Server | declaration | 000038 | a string column not pinned to `Latin1_General_100_CS_AS_KS_WS_SC_UTF8` |
| `mssql/named-default` | SQL Server | declaration | 000018 | `ADD ... DEFAULT` with no `CONSTRAINT` name, which gets a per-database generated one |
| `mssql/nullability` | SQL Server | declaration | none | `ALTER COLUMN c <type>` with no `NULL` or `NOT NULL`, which drops `NOT NULL` in silence |
| `mysql/collation` | MySQL | statement | 000035 | a `COLLATE` that is not `utf8mb4_0900_as_cs`, and a `CREATE TABLE` that spells none |
| `golden version` | all | golden file | none | a `schema.golden` whose recorded migration version is not that engine's highest |

Grandfathering is **one cutoff per rule**, each naming the last migration that actually breaks it,
never one number for the tree. A shared baseline couples the rules: a rule added later whose
violations reach higher forces it up, and every rule already there silently stops checking the
migrations below. Per-rule, bare `VARCHAR` is checked from 000015 rather than from 000040.
`mssql/nullability` takes no cutoff at all, because no `ALTER COLUMN` in the tree has ever omitted
the keyword, including all 92 in `000040`.

`TestMigrationCutoffs_AreTight` is what stops a cutoff being inflated to step over the migration
being written right now, which is the one way to use this check that would be worse than not
running it. It lowers each cutoff by one and requires the migration it names to report. The number
is read off `migrationCutoffsDefault` rather than written again, so the test cannot agree with
itself.

The eleventh rule is decision 2's replacement for the `git diff` rule the issue proposed: each
engine's `schema.golden` must record its own highest migration number, read from the header stage 1
added. It touches no git and no database, so unlike a merge base against `main` it gives an answer
on CI's depth-1 checkout, where the proposed rule would have been silently inert on every run.

The two collation pins are constants held by `TestMigrationPins_AreTheParityVocabulary` to being
members of `parityCollations` with the value `case-sensitive`. One package, one vocabulary: the
lint says a column must spell the name and the cross-engine comparison says what the name decides,
so a typo in either place fails rather than passing unnoticed in both.

**The rules reproduce the prior art.** Run at cutoff 0 against the committed tree they report 5
bare `VARCHAR`s, 100 unpinned SQL Server string columns, 18 unnamed defaults and 26 folding
`COLLATE` clauses, each last violated at exactly the number its cutoff names. The one deliberate
difference is the `CREATE TABLE` form, which finds 26 tables where a file-wide check found 5 files.

Nine mutations, and one of them earned its keep. Widening `mssql/collate` from the declaration to
the statement, and the MySQL `CREATE TABLE` form from the statement to the file, are both caught by
the mixed-target cases. Inflating the `mssql/collate` cutoff from 000038 to 000039 **survived**:
`TestMigrationCutoffs_AreTight` carried the four numbers twice, once where the rules read them and
once where the assertion did, so an inflated cutoff moved the first and left the second, and the
test passed while the rules exempted a migration nobody had looked at. That is precisely the
failure the test exists to prevent. Fixed by reading the number off the struct; the mutation is now
caught, and so is the same one on `mysql/collation`. The suite was green and the companion test was
decorative, and mutation is the only thing that found it.

Documentation: one paragraph in `AGENTS.md` and `CLAUDE.md`, byte-identical, beside **Schema golden
files** and in the voice of the **gofmt guard** paragraph above it. 66 words against measured
neighbours of 58 and 71. (Review round 1 found it omitted three obligations the rules enforce and
expanded it to 101 words; see **Review** below.)

Tiers: unit 1288 pass, 0 fail. Data 1528 pass, 0 fail on all four engines, run although the stage
did not owe it, because the changed files live under `src/core/data`. `where.sh migrations` green.
Integration is not applicable: nothing here is reachable through an endpoint.

### After the stages: two review rounds

The whole branch was reviewed once every stage had landed, and two rounds of findings were fixed on
top of them.

- **`a5aa4839`** answers round 1's three findings: the scanner now masks string-literal and
  quoted-identifier bytes so no content rule can read a value or a name, the MySQL collation check
  binds to the table's own options rather than the whole statement, the SQL Server string-column rule
  recognises length-less and unpinnable types, the number index no longer drops a duplicate version,
  and both catalogs gained the three author obligations they had omitted.
- **`ab4e318a`** answers round 2's one finding, seven members of it: a `#` line comment on MySQL, the
  SQL Server type read by position rather than searched for in text, a closed list of string types
  classified against a live server, `CREATE TYPE` refused so no alias type can escape it, and both
  collation pins compared case-insensitively.

Both rounds are recorded in full in the review comment linked below. Neither touched production code,
because on this branch there is none to touch: the lint is a Go test.

## Tiers

**Locally**, on the final tree: core tier **2809 passing, 0 failing**, which is the tier that runs
this lint against the four committed migration directories. All three modules **4579 passing, 0
failing**, including `TestGoSourcesAreGofmted`. The data tier was run on mysql, postgres, mssql and
sqlite during stages 1, 2 and 4, **1528 passing, 0 failing**, which is where the `schema_migrations`
reader and the golden comparison are exercised against real databases. `where.sh migrations` green on
every rule, and all 15 code anchors resolve.

**Integration is not applicable** and was not run: nothing in this change is reachable through an
endpoint. No handler, route or template moved.

**The four-engine result is the check suite's**, and it has concluded. All eleven jobs pass on
`ab4e318ac37c632a091ff4470a25ad84624a34e6`: Build validation, Lint, Versions, Vulnerabilities, the
three unit jobs and the four database jobs, each of which migrates a live engine and runs the data
tier against it.

Run: https://github.com/leodip/goiabada/actions/runs/33256561450

**Mutation testing**, since a lint that has quietly stopped matching anything passes a green suite:
**42 mutations across the branch, every one caught**. Three survived on first attempt and each one
was a real gap. The most valuable was in stage 4: inflating the `mssql/collate` cutoff from 000038 to
000039 survived, because the companion test carried the four numbers twice, once where the rules read
them and once where the assertion did. An inflated cutoff moved the first and left the second, so the
test passed while the rules exempted a migration nobody had looked at. That is precisely the failure
that test exists to prevent, the suite was green, and mutation is the only thing that found it.

## Deviations worth knowing

- **The migration directories are discovered, not enumerated.** The design said the lint "enumerates
  exactly the four and requires each directory". The plan review pointed out that this catches a
  directory that went missing and is blind to one that was added, which is the failure the rule exists
  to prevent. The lint compares the discovered set against the four dialects **in both directions**.
  A deviation from the mechanism, not from the guarantee.
- **The scanner refuses four lexical forms, not two.** The sketch named block comments and dollar
  quoting; `"` was added after measuring, since it is an identifier quote on postgres and sqlite and a
  string delimiter on mysql, the scanner models neither, and no statement on any engine carries one. A
  refusal that names the line is actionable; a scanner silently picking one of two meanings is not.
- **The guard's paragraph in `AGENTS.md` and `CLAUDE.md` is 101 words against a budget of "four or
  five lines".** The budget was set before the rule count was final and could not hold three more
  author obligations. Two shapes at 7 lines were tried and rejected: both split a backtick code span
  across a line break.
- **The parity back-fill uses a `--` separator** between the declaration and the prose below it,
  matching the five pre-existing declarations, rather than letting the two paragraphs butt together as
  the plan described. The two `000002` files keep a blank line instead, because what follows them is a
  stale filename header rather than prose.
- **`golden_test.go` reports 15 in-package `Encode`/`Parse` calls where the fanout recounted 13.** 13
  moved and 2 are new, in the refusal case the plan itself asked for. The recount command counts calls
  in a file the stage adds cases to.
- **`replaceVersion` re-anchored on `\tversion=`** rather than `version=`, since the header gained a
  second `=`-carrying field.

## Review

Reviewed by `openai-codex/gpt-5.6-sol` at effort `xhigh`, **three rounds: one over the plan and two
over the whole branch**. No stage was reviewed on its own, so the two final rounds are the only
reading this code got, and they read it with everything built on top of it.

|---|---|---|---|---|
| 1 | plan | 2 | 0 | both fixed in the plan before stage 1 |
| 1 | final | 3 | 0 | all fixed, `a5aa4839` |
| 2 | final | 1 | 0 | fixed, `ab4e318a` |

All six findings were confirmed against the code before anything was touched. None was contested,
deferred or refuted, and none is outstanding.

- **significant, quality (plan):** approximating SQL boundaries as raw text loses rule locality, so a
  file-wide pin check accepts a second unpinned table. Worse against this tree than the synthetic
  fixtures showed, since MySQL alone treats a backslash as an escape and `000034` forces it. Fixed by
  specifying the scanner's measured lexical vocabulary and requiring every content rule to bind to one
  table or one column.
- **significant, quality (plan):** a one-directional engine check cannot see an added directory. Fixed
  by comparing discovered directories against the four dialects both ways.
- **significant, security (final 1):** content rules matched against string-literal and
  quoted-identifier bytes, and the MySQL pin check read the whole statement. `DEFAULT 'varchar(64)'`
  was a false finding while `DEFAULT 'constraint'`, `NTEXT` and a length-less `NVARCHAR` were all
  missed. Fixed in `a5aa4839`: the scanner returns a masked copy that rules match against, identifiers
  keep their shape but not their spelling, and the type is recognised without requiring a length.
- **significant, conformance (final 1):** the number index kept one slug per engine, so a second up
  migration at the same number vanished and no rule could report it. Fixed in `a5aa4839`.
- **minor, conformance (final 1):** both catalogs omitted three obligations the rules enforce. Fixed
  in `a5aa4839`.
- **significant, security (final 2):** seven legal SQL spellings moved the pin token out of its
  semantic role or the type out of the rule's reach and returned clean, including `[c] [NVARCHAR](32)`,
  `[c] sysname`, `national character varying` and a MySQL `#` comment. Fixed in `ab4e318a` by bounding
  the space rather than patching the seven: type read by position, a closed list of string types
  classified against a live SQL Server by a new probe, and `CREATE TYPE` refused.

The gate closed after round 2 of a funded 4 on the round budget, whose trigger is "the last round
changed no production code". **That fired on a technicality here**, since the deliverable is test code
and the production hash could never move. Round 2 was productive. Closing is still right on three
independent grounds, set out in the review comment, the load-bearing one being that this lint is the
outer of two guards: a collation the text rules miss still fails the cross-engine golden comparison,
which errors on any collation name outside the four case-sensitive pins.

## The three closing comments

- [Decisions](https://github.com/leodip/goiabada/pull/294#issuecomment-5462889832): all seven settled
  at the seal, none deferred, plus the one standing cost this change introduces.
- [Follow-ups](https://github.com/leodip/goiabada/pull/294#issuecomment-5462891337): none filed, with
  the one fix folded in and the two known limits deliberately not filed.
- [The review](https://github.com/leodip/goiabada/pull/294#issuecomment-5462897465): every round and
  every finding, with the evidence and the mutation tables.

## Status

**Complete.** All four stages landed, both review rounds are answered, and the check suite is green on
`ab4e318a`, the branch head and the tree both rounds read.

`TestMigrationSource_TheFourCommittedDirectories` now holds the four migration directories to eleven
rules on every CI job, in the core tier, reading files and nothing else. The tree passes every rule as
it stands: nothing here repairs a historical divergence, and every cutoff was measured rather than
chosen.


leodip:
## Decisions: nothing was deferred

Seven decisions were settled at the seal, and all seven were still `Decided` when the run finished.
The run raised none of its own, deferred none, and carried no assumption forward that a later reader
has to unpick. There is no reversal cost to weigh, which is the short version of this comment.

What the seven settled, in one line each, since the agreement is gitignored and this is the durable
copy:

1. **The checker is a Go test under `src/core/data`**, not `scripts/check-migrations.sh` plus a CI
   job. It runs in the core tier on every job, adds no python3 to the host or to CI during the
   dependency-reduction crusade, and gets tests of its own, which the script form has none of.
2. **The ninth rule became the golden's recorded version**, replacing the proposed `git diff` against
   `git merge-base HEAD main`. That form cannot work in CI: `check.yml` checks out at depth 1, so on
   a pull request there is no local `main` and the merge base fails outright, and on a push to `main`
   the diff is empty. It would have been silently inert on every run.
3. **One cutoff per rule**, each naming the last migration that actually breaks it, rather than a
   single shared baseline of `000040`. A shared cutoff couples the rules: a rule added later whose
   violations reach higher forces it up, and every rule already there silently stops checking below.
4. **The nine undeclared partial migrations were back-filled** rather than exempted, so the coverage
   rule applies to every number with no cutoff at all.
5. **A `-- parity:` line is checked for content**, not just presence: the declared engine set must
   equal the set that actually carries the number. Presence alone passes a line naming the wrong
   engines, which is exactly what copying a neighbouring migration produces.
6. **Both unenforced obligations became rules**, SQL Server `ALTER COLUMN` nullability and the
   statement-free `.down.sql` declaration, because the tree already satisfies both and adding either
   later would cost another issue, agreement and run.
7. **Declarations are found by statement, not by line.** A gate whose failure mode is a false alarm
   on a legitimate migration is the wrong shape.

### The one standing cost, restated

Decision 2 means **every migration from now on regenerates all four `schema.golden` files**, including
one that changes no catalog shape at all. `000019`, `000021` and `000033` are that kind and currently
need no regeneration. The header's version moved, so the file did.

That was accepted at the seal and it is the only part of this change that makes future work more
expensive: one `cd src/core && go run ./cmd/schemadump` in the dev container per migration, which
`AGENTS.md` and `CLAUDE.md` already required for schema-changing migrations. Reversing it means
dropping rule eleven and going back to `data/golden` catching staleness only after four databases
have been migrated.

### One judgment the run made on your behalf

All four review findings, across both rounds, arrived with `needs_human: true` and a **populated
`forced_answer`**. The run triaged all four as fixes rather than escalating them, on `decisions.md`
section 1: a remedy that mechanically restores an invariant the agreement already decided, narrows
only what a check accepts, changes no interface and costs nothing at runtime is not a choice. The
collation pins were already `Decided`; each finding was that legal SQL can spell a string column in a
way the rules did not recognise.

If you would have wanted a say on any of them, the review comment on this PR records each one: what
it claimed, what verifying it against the code found, and what changed.

leodip:
## Follow-ups: none filed, and one fix folded in

**No issue is drafted for you to file.** Both review rounds' findings were fixed rather than
deferred, no verdict is outstanding, and the run found no defect next to the work that it left
undone. `gh issue list --state all --search "migration lint collation source text"` was re-run at
the finish and returns nothing, so nothing here is already tracked either.

**Two test-hardening fixes, in stage 1, commit `4b5e8f5c`.** Mutation found two adjacent refusals
overlapping in `parseHeader` and again in `MigratedVersion`: `strconv.Atoi("")` errors, so the
missing-field branch was covered by the not-a-positive-integer branch, and an empty
`schema_migrations` leaves `version` at 0, so the empty-table branch was covered by the non-positive
branch. In both places a test asserting only that an error occurred passed against code with the
first branch deleted.

Fixed in the tests rather than the code, because both refusals are correct and are genuinely
different diagnoses: a missing field is repaired by regenerating against a newer binary, a corrupt
value by restoring the file. The four `migrated=` cases now assert the message that distinguishes
them, and a new data case records version 0 so that branch is reached on its own.

It fit inside the stage under the fold-in test: it rides on surfaces stage 1 already touched, there
was nothing to settle, and it is the whole of the fix. Filing it would have spent a triage entry, an
agreement, a run and a review on three assertions.

### Two known limits, considered for filing and deliberately not filed

Both are stated in the source beside the rule they qualify, which is the `credential_read_lint_test.go`
precedent decision 1 cites: a source-text lint states its own lexical limits where a reader of the
rule will find them.

1. **A computed column declares no type in type position**, so `[c] AS <expression>` is never asked
   for a collation. It inherits one from the expression, and the alternative is evaluating SQL
   expressions.
2. **The grammar of a non-string declaration is not enumerated**, so an unfamiliar type reads as
   not-a-string.

Neither is filed, for three reasons that hold together:

- **The tree exercises neither.** No migration on any engine uses either shape.
- **Both fail in the safe direction**, a missed check rather than a false alarm on a valid
  migration. That is the design's stated preference, and a lint that can refuse correct input is
  worse than one with a documented blind spot.
- **Both are caught downstream by a stricter guard.** Verified rather than assumed:
  `canonicalCollation` in `src/core/data/schema_parity_test.go` returns an *error* for any collation
  name outside `parityCollations`, which carries exactly the four case-sensitive pins. A SQL Server
  string column left under the server's folding default therefore reports a name the vocabulary does
  not carry and fails the cross-engine golden comparison in the core tier, from the four committed
  files, with no database. Decision 2's version rule is what makes that reliable, since it forces a
  regeneration on every migration.

An issue reading "the outer of two guards has a documented blind spot that the inner guard covers"
would be filed only to be closed on reading. Filing it would also push the backlog the wrong way,
which is the standing instruction: a run that files issues while closing one has not reduced
anything.

### Nothing was encroached on

Section 2's out-of-scope items are unchanged. Catalog shape stays #284's, landed as #290. No
historical divergence was repaired: the tree passes every ported rule as it stands. The agent
tooling's copy of the checker, outside this repository, was not touched. #209, the same
"nothing checks that..." shape for audit event constants, was not started.

leodip:
## The review

Reviewed by `openai-codex/gpt-5.6-sol` at effort `xhigh`. **Three rounds in total: one over the plan
before any code existed, and two over the whole branch after every stage had landed.** No stage was
reviewed on its own: this run used end-of-branch review, so the two final rounds are the only reading
the code got, and they read it with everything built on top of it.

| Gate | Round | Findings | Blocking | Significant | Minor | Outcome |
|---|---|---|---|---|---|---|
| plan | 1 | 2 | 0 | 2 | 0 | both fixed in the plan before stage 1 |
| final | 1 | 3 | 0 | 2 | 1 | all fixed, `a5aa4839` |
| final | 2 | 1 | 0 | 1 | 0 | fixed, `ab4e318a` |

All six findings were confirmed against the code before anything was touched. **None was contested,
deferred or refuted.** Every one arrived `needs_human: true` with a populated `forced_answer`, and
all six were triaged as fixes rather than escalations, per the decisions comment above.

`production` never moved across either final round, `2f299776674a` throughout, because on this branch
the deliverable **is** test code: the lint is a Go test. That matters for how the gate closed, below.

---

## Plan gate, round 1: two findings, both fixed before a line was written

Verdict `static-review-exhausted`, which ends plan reviewing, so no round 2 was requested.

**1. significant, quality: rule locality lost when SQL boundaries are approximated as raw text.** The
plan required neither quote-aware splitting nor mixed-target cases, so a raw split false-rejects a
valid migration and a file-wide pin check accepts a second unpinned table.

Confirmed, and **worse against this tree than the reviewer's synthetic fixtures showed**: a backslash
escapes inside a string literal on MySQL and on none of the other three engines, which `000034`
forces by writing `'\\'` on mysql where mssql, postgres and sqlite write `'\'`. Getting that one
dialect rule wrong makes a scanner read the rest of `mssqldb/000034` as string interior, and that
file is above the bare-`VARCHAR` cutoff, so the rule would have silently reported nothing.

Fixed by specifying the scanner's measured lexical vocabulary in stage 3, refusing block comments and
dollar quoting (neither occurs in the tree) rather than guessing, failing closed on unbalanced
lexical state, and requiring every stage 4 content rule to bind to one table or one column with a
mixed-target case and a widening mutation apiece.

**2. significant, quality: closed-world engine drift hidden by a one-directional check.** The design
promised that "a fifth engine added without updating it is a failure rather than a silent skip" and
then named a mechanism, iterating four constants and requiring each directory, that catches a missing
directory and is blind to an added one. Confirmed.

Fixed by having stage 3's loader **discover** `src/core/data/*db/migrations` and compare against the
four `schemadump.Dialect` values in both directions, with an extra-engine case and a mutation that
reverts to a missing-only check. Recorded as a deviation from the design's mechanism sentence, not
from its guarantee.

No production code existed at this gate, so no tier ran.

---

## Final gate, round 1: three findings, all confirmed, all fixed in `a5aa4839`

Fresh session, all three axes reviewed. `production` `2f299776674a` before and after: all three fixes
are confined to test files and the two catalogs.

### Finding 1, security, significant. Content rules read value and name bytes. CONFIRMED.

**Claim:** `scanSQL` blanks comments but leaves string-literal and quoted-identifier bytes in the text
every content regex matches against, and the MySQL check searches the whole statement for the
collation pin rather than the table's own options.

Verified independently against the code before touching it:

- `scanSQL` wrote only `clean[i] = ' '` in the comment arm. The `'` and `[`/backtick arms advanced
  past the span and wrote nothing, so literal bytes survived into the text every rule matched.
- The MySQL unpinned arm asked `strings.Contains` over the **whole statement**.
- `mssqlStringColumnRe` was `...n?(?:var)?char\s*\(`, which recognises char, varchar, nchar and
  nvarchar **with a length** and nothing else. `NTEXT` is a string type that cannot carry a UTF-8
  collation at all, and `[x] NVARCHAR` with no length is a legal one-character column. Neither was a
  string column as far as the rule was concerned.

The file's own header comment already claimed the behaviour that was missing: "the scanner treats a
literal's interior as opaque, so a column type built up inside an `EXEC('...')` is never read by any
rule below". It did not. **The repair makes the comment true rather than introducing a new idea.**

**Fixed:** `scanSQL` now returns a second, masked copy of the same length. Literal interiors become
spaces, quoted-identifier interiors become `x`, and newlines survive both so an offset still names its
line. Identifiers keep their *shape*, because a column declaration is `<name> <type>` and a blanked
name would stop the type being recognised; masking their spelling is what keeps a column called
`[constraint]` or `[varchar]` from reading as the keyword it looks like. Statements and fragments
gained `Match` beside `Text`: `Text` is what a finding quotes back to the author, `Match` is what
every rule matches, and all eight regex call sites moved to `Match`. The MySQL arm now reads only the
suffix after the column list closes, which is the only place a table-level collation can be spelled,
and a `CREATE TABLE` whose body the scanner cannot find reports as unpinned rather than passing.
`mssqlStringColumnRe` lost its required `\(`, and a new rule refuses `TEXT` and `NTEXT` by type with a
message saying a UTF-8 collation is only valid on char, varchar, nchar and nvarchar, rather than
demanding a `COLLATE` clause SQL Server would reject.

Ten new cases, **including both directions of the confusion**: `DEFAULT 'varchar(64)'` and a folding
collation named inside a value are now accepted (they were false findings), while
`DEFAULT 'constraint'`, a column named `[constraint]`, a pin on one column standing in for the
table's, a pin spelled inside a default value, `NTEXT`, and a length-less `NVARCHAR` are now refused
(they were all missed).

### Finding 2, conformance, significant. The number index dropped a duplicate version. CONFIRMED.

**Claim:** the index stored `number -> dialect -> slug`, one slug, so a second up migration at the
same number on one engine overwrote the first and whichever filename sorted last was the only one any
rule saw.

Verified: `numbers[f.Number][f.Dialect] = f.Slug` over sorted filenames, so a duplicate slug sorting
*before* the ordinary one disappeared entirely. Pairing works off stems and passes a fully paired
duplicate; naming passes a well-formed name; coverage sees the number as carried. **Nothing in the
file could report it.** golang-migrate's own `source/iofs` refuses such a tree with
`ErrDuplicateMigration`, so a build never starts, but the four directories are hand-written and
nothing reads them before a database job does. The rule table promises "one number, one change" and
this was the on-one-engine half of it.

**Fixed:** the index keeps `[]string` per engine, and the identity rule reports any engine holding two
up migrations at one number **before** flattening, since comparing slugs across engines when one
engine has two of them compares an arbitrary pick. Same rule id, so the count stays at eleven. Two new
cases, both using a slug that sorts before `initial_create`, which is the shape the old index lost
completely.

### Finding 3, conformance, minor. The guidance omitted three obligations. CONFIRMED.

`AGENTS.md` and `CLAUDE.md` listed naming, pairing, number identity, the collation pins, the golden
version and the parity declaration, and omitted three enforced author actions: a named `CONSTRAINT` on
an added SQL Server `DEFAULT`, an explicit `NULL`/`NOT NULL` on an SQL Server `ALTER COLUMN`, and the
declaration a statement-free `.down.sql` owes. Verified against the rules themselves. Both files
amended and still byte-identical.

**Deviation, recorded rather than absorbed.** The agreement budgeted this paragraph at "one paragraph,
four or five lines, matching its two neighbours". It was 5 lines and 66 words; it is now 8 lines and
101 words, against neighbours of 5 lines (gofmt guard) and 6 (schema golden files). The budget was set
before the rule count was final and it cannot hold three more author obligations. Two shapes at 7
lines were tried and rejected: both split a backtick code span across a line break.

### Both `unchecked` entries answered

1. *"The complete SQL grammar boundary beyond the demonstrated shapes was not established."*
   **Covered, in the direction available.** The masking makes the boundary lexical rather than
   grammatical: a rule can no longer read a value or a name, whatever the surrounding grammar.
   Fail-closed was applied at the two places a grammar question remained.
2. *"Endpoint authentication, token validation, session lifecycle, authorization, rate limiting and
   audit emission were not dynamically exercised."* **Dismissed.** The branch changes no handler,
   route, template or runtime auth code.

### Mutations: five, all caught

| Replacement | Reverts | Result |
|---|---|---|
| `if buf[i] != '\n' {` to `if false {` | the mask itself | CAUGHT |
| `return s.masked[close+1 : s.end], true` to `return s.Match, true` | table-option scope | CAUGHT |
| `if len(slugs) < 2 {` to `if len(slugs) < 99 {` | duplicate detection | CAUGHT |
| `isMSSQLUnpinnableString(decl.Match)` to `false` | the `NTEXT` refusal | CAUGHT |
| `n?(?:var)?char\b` to `n?(?:var)?char\s*\(` | length-less string columns | CAUGHT |

**Tiers:** core tier 2791 pass, 0 fail, ~35s. All modules 4561 pass, 0 fail, including
`TestGoSourcesAreGofmted`. Anchors 15 of 15. Data and integration not re-run and not applicable: no
production code, no migration file, no `schema.golden` and no `schemadump` source moved, and both
categories hashed unchanged.

---

## Final gate, round 2: one finding, seven members, fixed in `ab4e318a`

Resumed session, all three axes reviewed. One `security`, `significant`, `confidence: high` finding.
`production` `2f299776674a` before and after, the same value both rounds were reviewed at.

### Finding 1, security, significant. A token outside the type's role decides the type. CONFIRMED.

**Claim:** valid SQL can move the pin token out of its semantic role, or move the type out of the
rule's reach, and the declaration returns clean. **Seven members**, each verified against the code
here, with **the engine's own behaviour measured rather than recalled**: a new probe declared
twenty-one spellings with no `COLLATE` against the dev container's SQL Server and read each one's
resolved type and collation back out of `sys.columns`.

| # | Member | Why it returned clean | Engine's answer |
|---|---|---|---|
| 1 | `[c] [NVARCHAR](32)` | the rule needed the type after the name; the mask turns `[NVARCHAR]` into `[xxxxxxxx]`, so the type read as a second name | `nvarchar`, folding |
| 2 | `[c] [VARCHAR](32)` | same, and the bare-`VARCHAR` rule excluded a `[` before the word on purpose, to keep a column called `[varchar]` out | `varchar`, folding, and narrow |
| 3 | `[c] NVARCHAR(32) ... DEFAULT CONVERT(...)` | a veto list rejected the whole declaration on the word `convert` | a real `nvarchar` column, unpinned |
| 4 | `[c] sysname` | in the same veto list, and not spelled `char` | `sysname`, folding, and **pinnable**: `sysname COLLATE <pin>` is accepted |
| 5 | `[c] national character varying(32)` | not in the rule's four-name vocabulary | `nvarchar`, folding |
| 6 | MySQL `# COLLATE=utf8mb4_0900_as_cs` after the column list | `scanSQL` modelled `--` and not `#`, so the comment's bytes stayed in the table options | MySQL reads a comment; the table inherits the database default |
| 7 | MySQL pin as a bare identifier in the table options | the arm asked `strings.Contains(opts, pin)`, which a partition name satisfies | nothing is collated |

All seven confirmed. The reviewer's own `why_exhaustive` said the family was **not** closed, naming
user-defined alias types as the unbounded tail, **so the fix closes the boundary rather than the
seven**:

- **`scanSQL` models MySQL's `#` line comment, and MySQL's alone.** On SQL Server a `#` opens a
  temporary table's name, so blanking the rest of the line there would swallow a real statement. This
  protects the scanner rather than only the collation rule, since an apostrophe or a parenthesis
  inside such a comment would otherwise open a literal or unbalance the depth.
- **The type is read by position, and the veto list is gone.** The type is the token after the column
  name; the column name is the token after an optional `ALTER TABLE <name>` and an optional
  `ALTER COLUMN` or `ADD`. Everything else in a declaration is a value, a keyword or a constraint, and
  none of them can reach type position, which retires member 3 as a class rather than as a word
  removed from a list. The boundary is found on the masked copy so no literal can move it, and the
  type's bytes are read off the unmasked copy, the only place a bracketed type name survives. **The
  mask is lifted in type position and nowhere else**, so a column called `[varchar]` still reads as a
  name.
- **`mssqlStringTypes` is a closed list:** the six built-in character types, every ANSI synonym the
  engine accepts for them, and `sysname`, each classified by what the probe resolved it to. `text` and
  `ntext` are unpinnable, and the finding now quotes the engine refusing them. **`CREATE TYPE` is
  refused** rather than followed, which is what makes the closed list the whole answer instead of a
  sample of one: an alias type is the only way a migration could put a string column's type outside
  it.
- **The MySQL options arm reads a `COLLATE` clause** instead of asking whether the pin's name appears.
  Both pin comparisons moved to `strings.EqualFold`, since a collation name is a case-insensitive
  identifier on both engines and refusing a differently-cased spelling of the right collation would be
  a false alarm.

Seventeen new cases, **each member paired with an accepting form**: a bracketed type that spells the
pin passes, a `sysname` that spells the pin passes, a `#` comment beside a pinned table passes, a
differently-cased pin passes on both engines, and a column named `[nvarchar]` declaring a `BIGINT` is
still not asked for a collation. Two scanner cases as well.

**Stated limit, recorded in the source beside the rule:** a computed column, `[c] AS <expression>`,
declares no type, so its collation is not checked. It comes from the expression, the tree has none,
and the alternative is evaluating SQL expressions.

### Mutations: ten, all caught

| # | Replacement | Result |
|---|---|---|
| 1 | a bracketed token read from the masked copy instead of the raw one | CAUGHT |
| 2 | `case c == '#' && d == schemadump.MySQL` never fires | CAUGHT |
| 3 | the MySQL options arm goes back to `strings.Contains(opts, pin)` | CAUGHT |
| 4 | `"sysname"` deleted from `mssqlStringTypes` | CAUGHT |
| 5 | `"national character varying"` deleted | CAUGHT |
| 6 | `if createTypeRe.MatchString(s.Match)` becomes `if false` | CAUGHT |
| 7 | the `convert` veto reinstated over the positional read | CAUGHT |
| 8 | the SQL Server pin compared with `!=` instead of `EqualFold` | CAUGHT |
| 9 | the MySQL pin compared with `!=` instead of `EqualFold` | CAUGHT |
| 10 | `"character varying"` deleted from `mssqlStringTypes` | CAUGHT |

### All three `unchecked` entries answered

- *"The complete SQL Server user-defined type space and complete legal grammar of all four dialects
  were not enumerated."* **Covered.** The alias-type space is bounded at zero by refusing
  `CREATE TYPE`, and the built-in string space is a closed list whose every member was resolved
  against a live server. What remains unbounded is the grammar of a *non-string* declaration, which is
  safe in the direction that matters: an unrecognised type reads as not-a-string.
- *"The schemadump generator was not run because it writes four tracked golden files."* **Dismissed.**
  Nothing in this round touched `schemadump`, a migration file or a `schema.golden`.
- *"Endpoint authentication, tokens, sessions, rate limiting, audit, templates and browser behaviour
  were not dynamically exercised."* **Dismissed**, and it agrees with the design: integration is not
  applicable, since nothing here is reachable through an endpoint.

**Tiers:** core tier 2809 pass, 0 fail, ~35s, up from 2791. All modules 4579 pass, 0 fail, up from
4561. Migration gate green against the real four directories under the new classifier. Anchors 15 of
15.

---

## How the review closed, and why that is right

**Two of four funded rounds ran at the final gate.** A round 3 request was written and never served:
the driver declined to fund it, on the rule that a gate whose last round changed no **production**
code has had its answer.

**That trigger fired on a technicality here, and saying so is the honest version.** The budget rule
tests the `production` hash, and on this branch that hash can never move, because the deliverable
*is* test code. `last_round_changed_production: no` was structurally true from the first commit and
says nothing about whether round 2 was productive. It was: it found a real `security`, `significant`
defect, seven legal SQL spellings that returned clean from the content rules.

Closing here is nonetheless right, on three grounds that do not depend on the budget:

1. **The findings converged**, three then one, and round 2's fix bounded the space rather than
   patching the instances it was handed: a closed list classified against a live server, `CREATE TYPE`
   refused so the alias space is bounded at zero, and the type read by position rather than searched
   for in text. That is the boundary argument round 1's verdict said finding 1 had not established.
2. **The residual region fails safe.** What stays unbounded is the grammar of a non-string
   declaration, and an unrecognised type reads as not-a-string, so the rule stays silent rather than
   raising a false alarm on a legitimate migration.
3. **This lint is a floor with a stricter guard under it.** Verified rather than assumed:
   `canonicalCollation` in `src/core/data/schema_parity_test.go` returns an *error* for any collation
   name outside `parityCollations`, which carries exactly the four case-sensitive pins. A SQL Server
   string column left under the server's folding default fails the cross-engine golden comparison in
   the core tier, from the four committed files, with no database. Decision 2's version rule is what
   makes that reliable, since it forces a regeneration on every migration.

Ground 3 is the load-bearing one, and it is why a further round would have been buying polish on the
outer of two guards.

**The tree the rounds read is the tree that stands.** Re-verified at `ab4e318a`: `production`
`2f299776674a`, `unit-tests` `e109cc2fa8ed`, `data-tests` `2bd0a7326232`, `integration-tests`
`7b69ee4ead08`, `docs` `09ff9f93a983`. All identical to the values recorded when round 2 was ingested,
so no mutation was left behind and nothing moved after the round was written.

PR #295, opened 2026-08-29
fix(core): startup can be told the database exists, and stops racing when it creates one

### LINKED ISSUE #293: fix(core): concurrent first startup can fail one replica creating the SQL Server database
`NewMsSQLDatabase` creates the application database when it is absent, with a check followed by a
create and no serialisation between them:

```sql
IF NOT EXISTS (SELECT * FROM sys.databases WHERE name = N'%s')
BEGIN
    CREATE DATABASE [%s]
    COLLATE Latin1_General_100_CS_AS_KS_WS_SC_UTF8
END
```

SQL Server has no atomic `CREATE DATABASE IF NOT EXISTS`, so two Goiabada instances starting at the
same time against a server where the database does not yet exist can both pass the check. One
creates the database; the other fails with `Msg 2714, There is already an object named ... in the
database` and the process exits.

Multi-replica deployment is documented in
`site/src/content/docs/production-deployment/kubernetes.mdx`. The generated manifest starts at one
replica, so this needs a deliberate scale-up plus a first start against an empty server to hit,
which is why it has not been seen. It is a startup crash, not data loss: restarting the failed
replica succeeds, because the database exists by then.

The same shape one layer down was found in `NewMigrator` while building #284 and fixed there by
taking `sp_getapplock` around the check and the create, which is what golang-migrate's own
`ensureVersionTable` does. The fix here is the same idea in a different scope: the lock has to be
held on a connection to `master`, since the application database is the thing being created.
Alternatively, tolerate `Msg 2714` and re-check that the database exists.

The other three engines are not affected: MySQL, PostgreSQL and SQLite all have a create form that
tolerates a concurrent creation.

Found at #284's plan gate, drafted in
https://github.com/leodip/goiabada/pull/290#issuecomment-5460054898 and not folded in there:
that change opens `NewMigrator`, not `NewMsSQLDatabase`, and an issue about schema parity should not
change how the database itself is created.

leodip:
Closes #293.

Starting Goiabada creates the application database when it is absent. That is done by a
maintenance connection to `master`, to `postgres`, or to the MySQL server with no database
selected, opened on every single start, followed by a create statement. Two deployments are broken
by it.

**Two instances starting at once against an absent database.** On SQL Server the create is a check
followed by a create, not one atomic statement, so racers can both pass the check and the losers
fail. Measured here: 5 of 8 processes failed, reproduced at 7/8 and 2/6. PostgreSQL is worse and
the issue did not know it, because its `CREATE DATABASE` is not `IF NOT EXISTS` at all: 7 of 8
racers died on `duplicate key value violates unique constraint "pg_database_datname_index"`
(`23505`), which the existing tolerance does not match, so the process exits. MySQL is immune
structurally, not by luck: it serialises on the schema metadata lock and demotes the duplicate to
`Note 1007`, which the driver never raises. SQLite has no create statement to race.

**A dedicated database user on PostgreSQL.** With the database already created and the application
role owning it but holding no `CREATEDB`, startup's `CREATE DATABASE` fails with
`permission denied to create database` (`42501`) and the process exits, although the role can do
everything Goiabada actually needs inside its own database. The production checklist tells readers
"Don't use root/admin accounts", which on PostgreSQL currently produces a server that will not
start.

## What this change does

`GOIABADA_DB_CREATE` / `--db-create`, default **true**, so nothing changes for anybody who sets
nothing. Set it to `false` and Goiabada issues no create statement **and opens no maintenance
connection at all**: not `master`, not `postgres`, not the no-database MySQL DSN. Each exists for
nothing but the create statement, and a managed deployment is exactly where the maintenance
database is out of reach. It applies to the three server engines; SQLite has no create step, and
its equivalent is `mode=rw` in the DSN.

On the default path, the create is serialised. PostgreSQL takes `pg_advisory_lock` on the
maintenance connection and SQL Server takes `sp_getapplock`, each held across the existence check
and the create on one connection pinned out of the pool, so at most one process ever issues
`CREATE DATABASE`. That is the same lock SQL Server already takes one layer down around
`schema_migrations` (#284). MySQL is unchanged.

## Two departures from the issue text

The issue names `Msg 2714` as the loser's error and proposes tolerating it. `Msg 2714` is the
duplicate-**object** error; creating a **database** gives `Msg 1801`, and most losers get no error
number and no message at all (`Number=0`, "Request failed but didn't provide reason"). Over 5
rounds of 8: 20 reason-less, 4 `Msg 1801`, 16 succeeded. With the driver's own message logging on,
the server sends nothing for those losers, so a tolerate-shaped remedy could not have been written
as an error-code test.

The issue also says the other three engines all tolerate a concurrent creation. PostgreSQL does
not, as above.

### Stage 1: The setting, wired end to end and inert

`072eb43d`. `config.DatabaseConfig` gains `Create bool`, read from `GOIABADA_DB_CREATE` (default
`true`) and `--db-create`, passed through `NewDatabase` to the mysql, postgres and mssql engine
configs. **Nothing reads it yet**: no constructor branches on it, so behaviour is identical on all
four engines. The wiring lands and is proved green before stage 2 gives it meaning.

The field is positive-sense, so its zero value does not create and every literal building an engine
config must set it. All 10 sites do: nine `true`, and the pre-created-database fixture in
`migration_testdb_helper.go` `false`, which is what that fixture has always meant. A missed site
fails loudly rather than quietly, since the database is then absent and the migration fails at once.

`getEnvAsBoolDefault` is new because `getEnvAsBool` takes no default and can only express
default-false; its 7 existing callers are untouched. SQLite gets no field and a comment saying so:
there is no create statement and no maintenance connection there, and `mode=rw` in the operator's
own DSN is the equivalent.

Tiers: unit 1290 pass, data 1528 pass across mysql, postgres, mssql and sqlite. Integration not
applicable, no handler, route or schema is touched. Two mutations against the new helper's default
and its parsing, both caught.

### Stage 2: The skipping path, and the documentation

`f144f788`. `GOIABADA_DB_CREATE=false` now means it. Each of `NewMySQLDatabase`,
`NewPostgresDatabase` and `NewMsSQLDatabase` wraps its maintenance connection and its create
statement in `if dbConfig.Create`, bodies unchanged, and logs the mode on the other arm. On SQL
Server that takes the `master` open, its `Ping` and the check-then-create batch, which is the point:
an Azure SQL contained user cannot reach `master` at all, so the ping alone would stop a start that
has everything it needs inside the application database. On PostgreSQL it fixes a plain bug: with
the database present and the role owning it but holding no `CREATEDB`, `CREATE DATABASE` gives
`SQLSTATE 42501`, which the `already exists` tolerance does not match, so the server would not
start. The production checklist's "don't use root/admin accounts" was advice a PostgreSQL reader
could not follow.

MySQL and PostgreSQL also validate the application handle on the skipping arm. `sql.Open` only
parses a DSN, so both used to return a usable-looking handle and a `nil` error against a database
that was not there, and the failure surfaced later inside the migrator as somebody else's. Both now
`Ping` there, so `Error 1049 (42000): Unknown database` and `database "x" does not exist (SQLSTATE
3D000)` reach the caller from the constructor. The creating arm is untouched: its create statement
already forces first use, which is what keeps existing deployments identical.

Data tests build a real restricted login per engine, because this tier connects as `root`, `sa` or
`postgres` and a superuser cannot observe the difference: MySQL a user granted only `ALL PRIVILEGES
ON <db>.*`, PostgreSQL a `LOGIN NOCREATEDB NOSUPERUSER` role owning the database, SQL Server a login
mapped `db_owner` into it and denied `VIEW ANY DATABASE`. Five cases: the least-privilege start on
three engines, the absent-database error on three engines with the catalog checked afterwards, an
operator's pre-created database left alone on the **creating** arm, and the two SQLite cases.

MySQL is the one engine where a restricted login proves nothing, and that is measured rather than
assumed: it has no per-schema `CONNECT` privilege, so no credential denies the maintenance DSN while
allowing the application one, and the `if dbConfig.Create` → `if true` mutation survives. The MySQL
case therefore reads `performance_schema.accounts.TOTAL_CONNECTIONS` for a fixture account created
for that test alone and requires exactly 1.

Docs: a new **[Database setup](https://github.com/leodip/goiabada/blob/issue-293-database-creation/site/src/content/docs/production-deployment/database.mdx)**
page, 252 words against a 350-word budget, with the `CREATE DATABASE` statement per engine and one
line on what the login must do inside it, an `Aside` for the one irreversible thing (SQL Server's
`COLLATE` clause), a sidebar entry after "Overview", a `GOIABADA_DB_CREATE` row in the reference
table linking to the page, and the checklist bullet doing the same.

Tiers: unit 4403 pass; data on mysql, postgres, mssql and sqlite, 518 / 519 / 521 / 508 pass, 0
fail. Integration not applicable, no handler, route or schema is touched. Seven mutations, all
caught.

One deviation worth knowing: the SQL Server fixture denies `VIEW ANY DATABASE` rather than `CONNECT`
in `master`, which the plan had asked for. `DENY CONNECT` on a `master` user locks the login out of
**every** database, the application's own included; denying sight of the catalog leaves it reachable
and makes the creating path fail the way a least-privilege deployment really fails, with `CREATE
DATABASE permission denied in database 'master'`.

### Stage 3: The lock on the creating path

Landed in `38e82928`. Goal 1: two instances starting at the same moment against a server where the
application database is absent both start.

`NewPostgresDatabase` and `NewMsSQLDatabase` no longer issue their create statement through the
pooled maintenance handle. Each now pins one connection out of that pool, takes an exclusive lock
on it, and holds it across the existence check and the create, so at most one process ever issues
`CREATE DATABASE`. PostgreSQL uses `pg_advisory_lock` keyed by a 64-bit FNV-1a of the database
name, computed in Go rather than through `hashtextextended` so it needs no minimum server version
and every racer reaches the same key without asking the server anything. SQL Server uses
`sp_getapplock` at `Exclusive` / `Session` / `LockTimeout = -1`, which is the shape the migration
lock one layer down already uses. Both locks are untimed, so a stuck holder blocks startup rather
than failing it; that is the trade the migration lock already made, and what each lock spans is a
catalog read and one `CREATE DATABASE`.

**The SQL Server lock resource carries no database name, deliberately.** `sp_getapplock` compares
its resource as binary, while `sys.databases.name` is compared under `master`'s collation, which
folds case, trailing space, Unicode normalization form and character width. A resource of
`"goiabada:create-database:" + name` would therefore give two instances configured `goiabada` and
`Goiabada` two *different* locks over what SQL Server treats as one database: both pass
`IF NOT EXISTS`, both create, which is the original defect with a lock bolted on. Normalizing in Go
cannot close that, because the equivalences are whatever the instance's collation says they are.
The constant resource is strictly wider than the thing it guards, which is the safe direction; its
cost, stated in the comment, is that creating two *different* Goiabada databases on one instance
now serializes against each other. PostgreSQL keeps its name-derived key, because
`pg_database.datname` is compared byte-exact and the key is then exactly as precise as the check.

MySQL is unchanged and gains only a comment saying why it needs no lock: it is immune structurally
rather than by luck, serialising on the schema metadata lock and demoting the duplicate to
`Note 1007`, which the driver never raises. 288 full constructor sequences at 24-way concurrency,
zero failures. SQLite has no create statement to race.

Tiers: unit 4404 pass; data on mysql, postgres, mssql and sqlite, 1546 pass, 0 fail. Integration not
applicable, no handler, route or schema is touched.

Four mutations, all caught, and each one reproduces the engine's own race error rather than merely
failing an assertion about a lock. Neutralising `pg_advisory_lock` brings back
`duplicate key value violates unique constraint "pg_database_datname_index" (SQLSTATE 23505)`;
neutralising `sp_getapplock` brings back `mssql: Request failed but didn't provide reason`, the
reason-less loser that is why this serializes instead of tolerating an error and re-checking;
restoring the database name into the SQL Server lock resource makes the case-variant pair create
twice; changing the advisory-lock namespace is caught by the unit case that pins the key literal.

One deviation worth knowing: `createDatabaseUnderAppLock` takes the database *name* and builds the
check-then-create batch itself, rather than being handed the finished statement. That is what makes
the lock-resource mutation writable at all, and it keeps the `IF NOT EXISTS` collation comment next
to the lock that now guards it.

### Folded in: the name it creates is the name it connects to

`7d931556`. This was drafted as a follow-up issue and folded in instead.

`GOIABADA_DB_NAME` reaches two places that have to agree: the create statement and the connection
string. On PostgreSQL they did not. `CREATE DATABASE` folds an unquoted identifier, so
`GOIABADA_DB_NAME=Goiabada` created `goiabada`, while the same string sits in the connection URL's
path as a literal connection parameter and is **not** folded, so the handle then asked for
`Goiabada`. The deployment could never start, it never self-corrected, and the database it did
create sat on the server next to the one the operator asked for. The constructor returned **no
error** for it: `sql.Open` only parses the URL and the creating arm has no `Ping`, the create
statement being what forces first use there, and on this path the create succeeded, against the
wrong name. The failure surfaced on the handle's first query, inside the migrator, as somebody
else's problem.

The three server engines now quote the identifier in the same shape: PostgreSQL `"name"`, MySQL
`` `name` ``, SQL Server `[name]`, each doubling its own quote character, and SQL Server's `N'...'`
catalog literal doubling the single quote. That closes the fold on PostgreSQL and, on all three,
the neighbouring case of a name needing quotes for some other reason, a hyphen or a space, which
was a syntax error. **Nothing changes for a lower-case name**, which is what every existing
deployment has.

MySQL and SQL Server were not broken: MySQL does not fold an unquoted identifier and SQL Server was
already bracketing. They are here so the three engines answer the question the same way, and
because the name reaches every one of these statements by interpolation, an identifier not being
something a bind parameter can carry. `GOIABADA_DB_NAME` is operator-supplied configuration rather
than user input, so the doubling is hardening and not a live hole.

`pg_database.datname` is compared byte-exact, so `databaseExists` and `AdvisoryLockKey` are
untouched and stay exactly as precise as the statement they guard. SQL Server's lock resource still
carries no name, for the reason it never did.

The data tier constructs at a mixed-case name and at a hyphenated one on all three engines, then
asks the returned handle which database it is in and the server how many databases answer to that
name ignoring case. Each engine package's unit tier holds the doubling itself, which no realistic
deployment would exercise. The two drop helpers in the fixtures now quote too: unquoted,
`dropPostgres` folded a mixed-case fixture's name and silently dropped nothing.

One line was added to the **Database setup** page's PostgreSQL section, saying to quote the name in
your own `CREATE DATABASE` when it is not all lower case.

Tiers: unit 4435 pass; data on mysql, postgres, mssql and sqlite, 2092 pass, 0 fail. Six mutations,
all caught, and three of them reproduce the engine's own error rather than merely failing an
assertion: unquoting PostgreSQL's create brings back
`FATAL: database "Goiabada_MixedCase_..." does not exist (SQLSTATE 3D000)` from a constructor that
returned `nil`, unquoting MySQL's brings back `Error 1064 (42000)` on the hyphenated name, and each
of the four escape helpers loses its doubling.

## Status

Complete. All three stages landed, the review found and closed one blocking defect, the drafted
follow-up was folded in rather than filed, and the check suite is green on the branch head.

| Stage | Commit | What it did |
|---|---|---|
| 1. The setting, wired end to end and inert | `072eb43d` | `GOIABADA_DB_CREATE` / `--db-create`, default true, carried to the three server engine configs. Nothing read it yet. |
| 2. The skipping path, and the documentation | `f144f788` | `false` skips both the create statement and the maintenance connection; new **Database setup** page. |
| 3. The lock on the creating path | `38e82928` | PostgreSQL and SQL Server serialise the check and the create on a pinned connection. |
| — review fix | `95dc8b3e` | The catalog check moved ahead of the lock, so the default path no longer reaches for a shared maintenance-database lock on every start. |
| — fold-in | `7d931556` | The three server engines quote the database identifier, so the name Goiabada creates is the name it connects to. |

All four goals are met: concurrent starts against an absent database all succeed on SQL Server and
PostgreSQL; an operator can tell Goiabada the database exists and it then opens no maintenance
connection on any engine; a least-privilege login starts the server on all four engines; and anybody
who sets nothing sees identical behaviour.

## Tiers

**Locally**, per stage, again after the review fix, and again after the fold-in: the unit tier
across all three modules (4435 pass), and the data tier on sqlite, mysql, postgres and mssql
(2092 pass, 33 skip, 0 fail on the last run). Twenty mutations across the run, every one caught.

**Integration was not run and is not applicable**, per decision 8: the diff reaches no handler,
route, template, schema or HTTP flow. Seventeen files, four of them documentation, the rest
`src/core/config`, `src/core/data`, `src/core/cmd/schemadump` and the data tier's fixtures.

**The four-engine result for the committed tree is the check suite's**, not a local run's. Green on
`95dc8b3e`, and re-running on `7d931556`, all eleven jobs: `Build validation`, `Vulnerabilities`, `Versions`, `Lint`,
`Unit / {core,authserver,adminconsole}` and `Tests / {sqlite,mysql,postgres,mssql}`, each of those
four being that engine's data tier.
[Run 33264170616](https://github.com/leodip/goiabada/actions/runs/33264170616).

### Review

Reviewed by `openai-codex/gpt-5.6-sol` at effort `xhigh`, 2 rounds over the whole branch, plus 1
round at the plan gate before any code existed. No stage was reviewed on its own.

|---|---|---|---|---|
| plan | 1 | 3 | 0 | all fixed in the plan, before any code existed |
| final | 1 | 1 | 1 | fixed, `95dc8b3e` |
| final | 2 | 0 | 0 | clean |

- **blocking, security:** both creation locks were taken in the shared maintenance database
  (`postgres`, `master`) *before* the existence check, so every start on the default path reached for
  them, and any principal able to connect to that maintenance database, holding no rights at all
  inside Goiabada's own database, could stall every restart indefinitely. Reproduced live on both
  engines by the reviewer. Fixed in `95dc8b3e`: an unlocked catalog check runs first and the lock is
  reached only when there is something to create, with a regression test that holds the production
  lock identity from an unrelated session and requires an ordinary restart to return inside a bounded
  wait.
- **significant, conformance (plan gate):** on `Create: false` against an absent database, MySQL and
  PostgreSQL would have returned a handle and a `nil` error, because `sql.Open` is lazy. Stage 2
  validates the application handle on that arm, so the engine's own `1049` / `3D000` reaches the
  caller.
- **significant, quality (plan gate):** a restricted login cannot observe the skipped maintenance
  connection on MySQL, which has no per-schema `CONNECT`. Stage 2's MySQL case asserts
  `performance_schema.accounts.TOTAL_CONNECTIONS` is exactly 1 for a dedicated fixture account.
- **minor, conformance (plan gate):** the SQL Server lock resource would have been narrower than the
  identity it guards, since `sp_getapplock` compares binary while `sys.databases.name` follows
  `master`'s collation. Stage 3 uses a constant resource.

Full account, with the evidence, in [this comment](https://github.com/leodip/goiabada/pull/295#issuecomment-5463722470).

### Decisions and follow-ups

No decision was deferred; all ten were settled at the seal and none is open.

**One follow-up was drafted and has been folded in rather than filed**, in `7d931556`: PostgreSQL
folded a mixed-case `GOIABADA_DB_NAME` at creation but not at connection, so such a deployment
created one database and connected to another.
[The original draft](https://github.com/leodip/goiabada/pull/295#issuecomment-5463717633) stands as
the account of the defect; what shipped is the "quote the identifier" remedy it named, taken across
all three server engines rather than PostgreSQL alone. Nothing is left to file.

### DIFF
--- a/site/astro.config.mjs
+++ b/site/astro.config.mjs
@@ -80,6 +80,7 @@ export default defineConfig({
 					label: 'Production deployment',
 					items: [
 						{ label: 'Overview', slug: 'production-deployment' },
+						{ label: 'Database setup', slug: 'production-deployment/database' },
 						{ label: 'Cloudflare Tunnel', slug: 'production-deployment/cloudflare-tunnel' },
 						{ label: 'Cloudflare + Nginx', slug: 'production-deployment/cloudflare-nginx' },
 						{ label: 'Reverse proxy (no Cloudflare)', slug: 'production-deployment/reverse-proxy' },
--- a/site/src/content/docs/production-deployment/database.mdx
+++ b/site/src/content/docs/production-deployment/database.mdx
@@ -1,0 +1,50 @@
+---
+title: Database setup
+description: Pre-create the Goiabada database and run with a least-privilege database user.
+---
+
+import { Aside } from '@astrojs/starlight/components';
+
+By default Goiabada creates its database at startup if it is not already there, which requires a
+login allowed to create databases on the server. Set `GOIABADA_DB_CREATE=false` when you have
+created the database yourself, or when the login is not allowed to: Goiabada then issues no
+`CREATE DATABASE` and opens no connection to `master` or `postgres`.
+
+Create the database with the statement for your engine, then give the login full rights inside it.
+
+## PostgreSQL
+
+```sql
+CREATE DATABASE goiabada ENCODING 'UTF8';
+```
+
+Make the login the owner, or grant it full rights in the database. It needs no `CREATEDB`
+attribute and no access to the `postgres` database.
+
+## MySQL
+
+```sql
+CREATE DATABASE goiabada CHARACTER SET utf8mb4 COLLATE utf8mb4_0900_as_cs;
+```
+
+`GRANT ALL PRIVILEGES ON goiabada.* TO 'goiabada'@'%';` is enough. No server-wide privilege is
+needed.
+
+## SQL Server
+
+```sql
+CREATE DATABASE goiabada COLLATE Latin1_General_100_CS_AS_KS_WS_SC_UTF8;
+```
+
+Map the login into the database and add it to `db_owner`. It needs no access to `master`.
+
+<Aside type="caution">
+On SQL Server, spell the `COLLATE` clause. A database default cannot be changed afterwards, because
+`ALTER DATABASE ... COLLATE` blocks while the application is connected, and every text column a
+future Goiabada migration adds inherits it.
+</Aside>
+
+## SQLite
+
+There is nothing to pre-create, and `GOIABADA_DB_CREATE` does not apply: the driver creates the file.
+To refuse an absent file instead of creating one, put `mode=rw` in `GOIABADA_DB_DSN`.
--- a/site/src/content/docs/production-deployment/database.mdx
+++ b/site/src/content/docs/production-deployment/database.mdx
@@ -21,6 +21,10 @@ CREATE DATABASE goiabada ENCODING 'UTF8';
 Make the login the owner, or grant it full rights in the database. It needs no `CREATEDB`
 attribute and no access to the `postgres` database.
 
+Quote the name if `GOIABADA_DB_NAME` is not all lower case. PostgreSQL folds an unquoted
+identifier, so `CREATE DATABASE Goiabada` makes a database called `goiabada`, which is not the one
+Goiabada connects to.
+
 ## MySQL
 
 ```sql
--- a/site/src/content/docs/production-deployment/production-checklist.mdx
+++ b/site/src/content/docs/production-deployment/production-checklist.mdx
@@ -52,7 +52,7 @@ Only enable `TRUST_PROXY_HEADERS=true` if Goiabada is actually behind a trusted 
 ## Database
 
 - [ ] **Production database** - Use MySQL or PostgreSQL for production (not SQLite)
-- [ ] **Dedicated database user** - Don't use root/admin accounts
+- [ ] **Dedicated database user** - Don't use root/admin accounts; see [database setup](/production-deployment/database/)
 - [ ] **Database backups configured** - Regular automated backups with tested restore procedures
 - [ ] **Database encryption** - Enable TLS for database connections if over network
 
--- a/site/src/content/docs/reference/environment-variables.mdx
+++ b/site/src/content/docs/reference/environment-variables.mdx
@@ -87,6 +87,11 @@ Used only on first startup:
       <td>Database name</td>
     </tr>
     <tr>
+      <td style="white-space: nowrap;"><code>GOIABADA_DB_CREATE</code><br/><code>--db-create</code></td>
+      <td><code>true</code></td>
+      <td>Create the database if it does not exist (server engines only). Set to <code>false</code> for a <a href="/production-deployment/database/">pre-created database</a></td>
+    </tr>
+    <tr>
       <td style="white-space: nowrap;"><code>GOIABADA_DB_DSN</code><br/><code>--db-dsn</code></td>
       <td><code>file::memory:</code></td>
       <td>SQLite DSN (only for sqlite)</td>
--- a/src/authserver/tests/data/database_create_test.go
+++ b/src/authserver/tests/data/database_create_test.go
@@ -618,3 +618,136 @@ func dropServerDatabase(t *testing.T, cfg *config.DatabaseConfig, name string) {
 		dropMsSQL(t, cfg, name)
 	}
 }
+
+// TestNewDatabase_CreateTrue_TheNameItCreatesIsTheNameItConnectsTo is the case the identifier
+// quoting exists for.
+//
+// GOIABADA_DB_NAME reaches two places that have to agree: the create statement and the
+// connection string. On PostgreSQL they did not. CREATE DATABASE folds an unquoted identifier,
+// so GOIABADA_DB_NAME=Goiabada created `goiabada`, while the same string sits in the connection
+// URL's path as a literal connection parameter and is not folded, so the handle then asked for
+// `Goiabada`. The deployment could never start, it never self-corrected, and the database it did
+// create sat on the server next to the one the operator asked for.
+//
+// The constructor did NOT return an error for it, which is why the assertion has to go through
+// the handle. sql.Open only parses the URL, and the creating arm has no Ping (the create
+// statement is what forces first use there, and on this path the create succeeded, against the
+// wrong name). So NewPostgresDatabase returned a nil error and a handle that failed on its first
+// query with `database "Goiabada" does not exist (SQLSTATE 3D000)`, inside the migrator, as
+// somebody else's problem.
+//
+// Two names, because they fail differently and on different engines. A mixed-case name is the
+// reported defect and is PostgreSQL's alone: MySQL does not fold an unquoted identifier and SQL
+// Server was already bracketing. A hyphenated name was a plain syntax error on PostgreSQL and
+// MySQL both, so it is what makes MySQL's backticks load-bearing here rather than only in the
+// unit tier. SQL Server needed neither and is here to stay unbroken.
+//
+// What this tier cannot witness is the doubling inside the quoting, a name carrying a double
+// quote, a backtick or a `]`. Those live in each engine package's own unit test, because a
+// database named that way is not a deployment anybody has and building the fixture would be the
+// only thing the case proved.
+//
+// SQLite is skipped: it has no create statement and no name, only a path.
+func TestNewDatabase_CreateTrue_TheNameItCreatesIsTheNameItConnectsTo(t *testing.T) {
+	if dbType() == "sqlite" || dbType() == "" {
+		t.Skip("sqlite has no database name, only a DSN path, so there is no identifier to quote")
+	}
+
+	for _, tc := range []struct {
+		label string
+		name  string
+	}{
+		// Mixed case on purpose, and unmistakably so: isolatedDBName is all lower case, which is
+		// exactly the shape that hid this for as long as it was hidden.
+		{"mixed case", "Goiabada_MixedCase_" + strings.TrimPrefix(isolatedDBName(), "goiabada_mig_")},
+		{"a hyphen", "goiabada-hyphen-" + strings.TrimPrefix(isolatedDBName(), "goiabada_mig_")},
+	} {
+		t.Run(tc.label, func(t *testing.T) {
+			cfg := config.GetDatabase()
+			name := tc.name
+
+			var sqlDB *sql.DB
+			var currentDatabase string
+			switch dbType() {
+			case "mysql":
+				db, err := mysqldb.NewMySQLDatabase(&mysqldb.DatabaseConfig{
+					Type: "mysql", Username: cfg.Username, Password: cfg.Password,
+					Host: cfg.Host, Port: cfg.Port, Name: name, Create: true,
+				}, false)
+				require.NoErrorf(t, err, "NewMySQLDatabase at %s", name)
+				t.Cleanup(func() { _ = db.DB.Close(); dropMySQL(t, cfg, name) })
+				sqlDB, currentDatabase = db.DB, "SELECT DATABASE()"
+			case "postgres":
+				db, err := postgresdb.NewPostgresDatabase(&postgresdb.DatabaseConfig{
+					Type: "postgres", Username: cfg.Username, Password: cfg.Password,
+					Host: cfg.Host, Port: cfg.Port, Name: name, Create: true,
+				}, false)
+				require.NoErrorf(t, err, "NewPostgresDatabase at %s", name)
+				t.Cleanup(func() { _ = db.DB.Close(); dropPostgres(t, cfg, name) })
+				sqlDB, currentDatabase = db.DB, "SELECT current_database()"
+			case "mssql":
+				db, err := mssqldb.NewMsSQLDatabase(&mssqldb.DatabaseConfig{
+					Type: "mssql", Username: cfg.Username, Password: cfg.Password,
+					Host: cfg.Host, Port: cfg.Port, Name: name, Create: true,
+				}, false)
+				require.NoErrorf(t, err, "NewMsSQLDatabase at %s", name)
+				t.Cleanup(func() { _ = db.DB.Close(); dropMsSQL(t, cfg, name) })
+				sqlDB, currentDatabase = db.DB, "SELECT DB_NAME()"
+			default:
+				t.Fatalf("unsupported db type %q", dbType())
+			}
+
+			// The load-bearing assertion, and it has to be a query rather than the constructor's
+			// error, because on the broken path there was no error. This is the first use of the
+			// handle, which is where PostgreSQL used to answer `database "..." does not exist`.
+			var got string
+			require.NoError(t, sqlDB.QueryRow(currentDatabase).Scan(&got),
+				"first use of the handle the constructor returned: this is where an unquoted CREATE DATABASE surfaced, as the migrator's problem rather than the constructor's")
+			require.Equal(t, name, got,
+				"the handle must be connected to the database the constructor created, spelled the way the operator spelled it")
+
+			// And the other half of the symptom: the deployment must not be left with a database
+			// it did not ask for sitting next to the one it did. Counted case-insensitively,
+			// because that is the only comparison that can see both at once, and because on
+			// PostgreSQL, the one engine where the two are distinct rows, it is the only one that
+			// would have caught the pair.
+			require.Equal(t, 1, serverDatabasesMatchingFold(t, name),
+				"exactly one database on the server may answer to this name; two means the create statement and the connection string spelled it differently")
+		})
+	}
+}
+
+// serverDatabasesMatchingFold counts the databases whose name matches name ignoring case.
+//
+// serverDatabaseExists above cannot answer this. Its predicate is the engine's own comparison,
+// which folds case on MySQL and SQL Server and does not on PostgreSQL, so on those two it cannot
+// distinguish one database from a folded pair and on PostgreSQL it would only ever see one of
+// them. Folding explicitly on all three asks the same question everywhere: how many databases on
+// this server answer to this name (#293).
+func serverDatabasesMatchingFold(t *testing.T, name string) int {
+	t.Helper()
+	cfg := config.GetDatabase()
+
+	var dsn, driver, query string
+	switch dbType() {
+	case "mysql":
+		driver, dsn = "mysql", mySQLServerDSN(cfg.Username, cfg.Password, cfg)
+		query = "SELECT COUNT(*) FROM information_schema.SCHEMATA WHERE LOWER(SCHEMA_NAME) = LOWER(?)"
+	case "postgres":
+		driver, dsn = "pgx", postgresMaintenanceDSN(cfg.Username, cfg.Password, cfg)
+		query = "SELECT COUNT(*) FROM pg_database WHERE LOWER(datname) = LOWER($1)"
+	case "mssql":
+		driver, dsn = "sqlserver", msSQLMasterDSN(cfg)
+		query = "SELECT COUNT(*) FROM sys.databases WHERE LOWER(name) = LOWER(@p1)"
+	default:
+		t.Fatalf("%s has no server catalog of databases", dbType())
+	}
+
+	sqlDB, err := sql.Open(driver, dsn)
+	require.NoError(t, err, "open the server to read its catalog")
+	defer func() { _ = sqlDB.Close() }()
+
+	var n int
+	require.NoError(t, sqlDB.QueryRow(query, name).Scan(&n), "count databases matching %s", name)
+	return n
+}
--- a/src/authserver/tests/data/database_create_test.go
+++ b/src/authserver/tests/data/database_create_test.go
@@ -1,0 +1,224 @@
+package datatests
+
+import (
+	"database/sql"
+	"os"
+	"path/filepath"
+	"reflect"
+	"testing"
+
+	"github.com/leodip/goiabada/core/config"
+	"github.com/leodip/goiabada/core/data/mssqldb"
+	"github.com/leodip/goiabada/core/data/mysqldb"
+	"github.com/leodip/goiabada/core/data/postgresdb"
+	"github.com/leodip/goiabada/core/data/sqlitedb"
+	"github.com/stretchr/testify/require"
+)
+
+// The three engine constructors that open a maintenance connection and issue a create
+// statement. GOIABADA_DB_CREATE decides whether they do, and these tests own both answers at
+// the constructor boundary (#293).
+//
+// Why the boundary and not data.NewDatabase: the factory also migrates and runs the startup
+// data tasks, so a failure there could be any of three things. The constructor's contract is
+// "given a config, return a usable handle or an error", and that is what an operator's
+// deployment either satisfies or does not.
+
+// TestNewDatabase_CreateFalse_StartsUnderALeastPrivilegeLogin is the case #293 exists for, and
+// the one that cannot be faked.
+//
+// This tier connects as root, sa or postgres, so a test that pre-created the database and set
+// Create: false would pass with the maintenance connection still opened and the create still
+// issued: a superuser can do both. The restricted login is what observes their absence from
+// outside, which is why the fixture goes to the trouble of building one per engine.
+//
+// PostgreSQL is where this used to be a hard failure rather than a nicety: with the database
+// present and the role owning it but holding no CREATEDB, the old unconditional CREATE DATABASE
+// gave "permission denied to create database (SQLSTATE 42501)", which the constructor's
+// "already exists" tolerance does not match, so the server would not start. The production
+// checklist's "don't use root/admin accounts" was advice a PostgreSQL reader could not follow.
+func TestNewDatabase_CreateFalse_StartsUnderALeastPrivilegeLogin(t *testing.T) {
+	if dbType() == "sqlite" || dbType() == "" {
+		t.Skip("sqlite has no login and no create statement, so there is nothing to restrict")
+	}
+
+	r := newRestrictedLoginDB(t)
+	db, sqlDB := r.constructRestricted(t)
+
+	if dbType() == "mysql" {
+		// Read before anything else touches the server. MySQL cannot deny the maintenance DSN
+		// to a login it allows the application DSN to, so the privilege above proves nothing
+		// here and this counter is what does: one connection is the skipping path, two is a
+		// constructor that still opened the no-database DSN.
+		require.Equal(t, 1, r.connectionCount(t),
+			"with Create false the constructor must open exactly one connection, to the application database; a second means the maintenance DSN was still opened")
+	}
+
+	// Migrating the full chain and then reading through the handle is the rest of the claim:
+	// the restricted login does not merely connect, it can do everything Goiabada needs.
+	h := newIsolated(t, db, sqlDB)
+	require.NoError(t, h.Migrator.Up(), "migrate the full chain as the restricted login")
+
+	var clients int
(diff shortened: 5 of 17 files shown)


leodip:
## Decisions, follow-ups and fold-ins

**No decision was deferred.** All ten decisions in the agreement were settled at the seal, none was
raised by the run, and none is open. There is nothing here for you to adjudicate before merging.

Nothing was folded in either: the run found no defect on the surfaces it touched that it fixed
outside the agreed scope. The one corrective commit on the branch, `95dc8b3e`, answers the review's
blocking finding and is accounted for in the review comment below, not here.

One follow-up was drafted, verified against the code, and is ready to file as it stands. It is
pre-existing and this change does not touch name handling. Duplicate search re-run just now over
every state (`gh issue list --state all --search "database name case postgres"`, again for
`GOIABADA_DB_NAME`, and for `mixed-case identifier CREATE DATABASE`): nothing tracks it.

---

### PostgreSQL: a mixed-case `GOIABADA_DB_NAME` creates one database and then connects to another

**Labels:** `bug`

**Body:**

`NewPostgresDatabase` in `src/core/data/postgresdb/db.go` uses the configured database name two ways
that do not agree:

- it interpolates it into an **unquoted** statement, `fmt.Sprintf("CREATE DATABASE %v;", dbConfig.Name)`,
  and PostgreSQL folds an unquoted identifier to lower case;
- it puts the same raw string in the connection URL's path,
  `fmt.Sprintf("postgres://%v:%v@%v:%v/%v", ...)`, where the database name is a literal connection
  parameter and is **not** folded.

So with `GOIABADA_DB_NAME=Goiabada` the first start creates a database called `goiabada` and then
fails to open `Goiabada`. Reproduced against PostgreSQL in the dev container:

```
CREATE DATABASE Issue293Case            -> err=<nil>
catalog now holds                       -> "issue293case"  (configured "Issue293Case")
sql.Open(app DSN)                       -> err=<nil>  (lazy, so nil)
first use of that handle                -> err=FATAL: database "Issue293Case" does not exist (SQLSTATE 3D000)
```

Every later start repeats it: the create is tolerated, because `database "goiabada" already exists`
contains the tolerated text, and the connection fails the same way. The deployment can never start,
it never self-corrects, and the error names a database the operator can see they asked for while a
database they did not ask for sits next to it.

MySQL and SQL Server are not affected. MySQL's `CREATE DATABASE IF NOT EXISTS %v` does not fold, and
SQL Server brackets the name (`CREATE DATABASE [%s]`) while its DSN carries the same string, so both
halves agree.

**What closing it would take.** Quoting the identifier, `CREATE DATABASE "%v";`, makes the two halves
agree and also settles the neighbouring case of a name needing quotes for any other reason, a hyphen
or a space, which is a syntax error today. Refusing a mixed-case name at config load is the smaller
change. Either way it wants a data-tier case per engine constructing at a mixed-case name and reading
back through the returned handle, plus a note on whether an operator who already has a folded database
is left with an unreachable one. Note separately that the identifier is interpolated rather than
parameterised on all three server engines, which is its own hardening question about
`GOIABADA_DB_NAME` as an input.

**Not folded into #293:** that change is about *whether* the create statement runs and whether it can
be raced; this is about what the statement spells.

**File it with:**

```bash
gh issue create --label bug \
  --title 'PostgreSQL: a mixed-case GOIABADA_DB_NAME creates one database and then connects to another' \
  --body-file <the body above>
```

leodip:
## The review account

This is the code review, in full. Under `LEO_REVIEW_MODE=end` no stage was reviewed on its own: the
whole branch was read once, after everything had landed, so this is the only independent reading the
code got. Two rounds of a budget of four. The verdicts themselves live in a gitignored directory and
are overwritten at each gate, so this comment is the durable copy.

**Reviewer:** `openai-codex/gpt-5.6-sol`, effort `xhigh`, fresh sessions, `MAX_ROUNDS=4`.

|---|---|---|---|---|
| plan | 1 | 3 | 0 | all fixed in the plan, before any code existed |
| final | 1 | 1 | 1 | fixed in `95dc8b3e` |
| final | 2 | 0 | 0 | clean |

---

### Plan gate, round 1

Scope: `plan.md` against the sealed agreement, all three axes. 3 findings, all judged true, each with
a populated `forced_answer`. The reviewer independently re-ran `check-anchors.sh` (19 green), two
committed probes, and seven probes of its own against live MySQL, PostgreSQL and SQL Server.

|---|---|---|---|---|
| 1 | conformance | significant | On `Create: false` against an absent database, MySQL and PostgreSQL return a handle and a `nil` error, because both use lazy `sql.Open` and touch the application handle nowhere before returning. §5 seam 1's third table requires the engine's own "does not exist" error at the constructor. | **Fixed** in the plan, stage 2. Verified by reading all four constructors: MySQL ends at `sql.Open(dsnWithDBname)`, PostgreSQL opens `dbURL` and never uses it, SQL Server `Ping`s and closes on failure, SQLite runs pragmas and `PingContext`. Stage 2 validates the application handle on the `Create: false` arm only, preserving `1049` and `3D000`; the `Create: true` arm is untouched, so goal 4 holds. |
| 2 | quality | significant | The MySQL restricted-login test cannot observe the skipped maintenance connection, so the stage's own `if dbConfig.Create` → `if true` mutation survives on that engine. | **Fixed** in the plan, stage 2. §1 already carried the measurement: a login with only `ALL PRIVILEGES ON <db>.*` connects with no database selected and its `CREATE DATABASE IF NOT EXISTS` succeeds. MySQL has no per-schema `CONNECT`, so no credential separates the two DSNs. Stage 2 adds a `performance_schema.accounts.TOTAL_CONNECTIONS` assertion on a MySQL-only fixture account, required to be exactly 1. This is not the "counting connections" side channel §5 rejected: the rejection's stated reason was that it reaches into internals, and a server-side connection counter for a dedicated account does not. Decision 8's restricted-login test is kept unchanged on all three engines. |
| 3 | conformance | minor | §4's SQL Server lock resource `"goiabada:create-database:" + name` is compared as binary while `sys.databases.name` follows `master`'s collation, so equivalent spellings take different locks, both pass the absence check and both create. | **Fixed** in the plan, stage 3. The reviewer measured it for case, trailing space, Unicode normalization and width, with accent-only as a negative control. A lock identity *narrower* than the identity it protects is the unsafe direction, and no Go-side normalization closes it across arbitrary collations. The resource became the constant `"goiabada:create-database"`; its cost (creations of different databases on one instance serialize) is stated in the comment and bounded. The race test gained a case-variant pair. PostgreSQL keeps its name-derived FNV key, with the comment saying why. |

Three `unchecked` entries, all answered: two were accurate statements of the gate's limits and needed
nothing (no code existed yet; no tier was run because nothing had changed), and the third, that the
documentation prose did not exist, was stage 2's work under decision 9's hard 350-word budget and was
read by the final review.

Ground the reviewer covered and put out of scope became the drafted follow-up in the comment above.

The gate closed at one round of a funded two: all three findings were resolved by applying the
reviewer's own `forced_answer`, so a second round would mostly have read back its own conclusions.

---

### Final gate, round 1: one blocking finding

Request `293-20260829T162832Z`. Scope: the whole branch against `origin/main`, `git diff
origin/main...HEAD`, 15 files, +1203 −93, over `4a892079`, `072eb43d`, `f144f788`, `38e82928`. All
three axes recorded as reviewed.

#### Finding 1 — blocking, security. CONFIRMED, fixed in `95dc8b3e`.

**Family:** indefinite startup dependence on cooperative maintenance-database locks that unrelated
principals can acquire.

**The claim.** Both `Create: true` paths took their lock in the shared maintenance database *before*
checking whether the application database existed. `pg_advisory_lock` lives in a key space scoped to
`postgres`; `sp_getapplock` at `LockOwner = 'Session'` lives in `master`. Both spaces are shared with
every session on the server, so the resource can be held by a principal with no rights whatsoever
inside Goiabada's own database. The reviewer reproduced it live on both engines: a fresh PostgreSQL
`NOCREATEDB NOSUPERUSER` role with no grant on the application database took the exact FNV-derived
key, and a fresh SQL Server login with only ordinary `public` took the constant `sp_getapplock`
resource. In each case the real constructor, run against an **already-existing** `goiabada_data`, was
still blocked after 500 ms and returned immediately on release.

`siblings_checked` covered all four engines: MySQL adds no lock (decision 6) and SQLite has no
maintenance connection at all. `why_exhaustive` rests on a repository-wide search for
`pg_advisory_lock` and `sp_getapplock`, which finds exactly the two sites the branch adds. No earlier
verdict carries this family, so there was no self-contradiction to challenge.

**Verified against the code before resolving.** True as written.
`createDatabaseUnderAdvisoryLock` pinned a connection and executed `SELECT pg_advisory_lock($1)` as
its first statement, with the `pg_database` count after it; `createDatabaseUnderAppLock` did the same
with `sp_getapplock`, the `IF NOT EXISTS` batch being wholly inside the lock. Nothing read the catalog
first, so on the default path every start of every deployment reached for the lock forever, not just
the one start that creates.

**Not covered by decision 5's accepted cost, which is why it stands.** Decision 5 accepted that "both
locks are untimed, so a stuck holder blocks startup rather than failing it", reasoning that the lock
spans "one catalog read and one CREATE DATABASE" and is "taken once per database in the life of a
deployment". The second half was false of the code as committed: the lock was taken on every start,
and its holder could be any stranger with a login on the server. So this is a defect against decision
5's own stated premise, not a relitigation of it.

**The fix, `95dc8b3e`.** An unlocked catalog check now runs first on both engines, and the lock is
reached only when there is something to create.

- PostgreSQL: a new `databaseExists(ctx, rowQuerier, name)` asks `pg_database` the same question at
  both sites, once on the maintenance pool before the lock and once on the pinned connection under it.
  One function deliberately: the outer check is only sound while it is no weaker than the inner one,
  and two spellings could drift.
- SQL Server: a new `databaseExists(ctx, masterDB, name)` counts `sys.databases` through a
  parameterised nvarchar, which folds case, trailing space, normalization form and width under
  `master`'s collation exactly as the batch's `IF NOT EXISTS` does. Two instances configured
  `goiabada` and `Goiabada` against an absent database therefore both still find nothing here, both
  still take the constant resource, and the batch still collapses them to one database.

Decision 5 is untouched: the remedy is still serialise rather than tolerate, the locks are still
untimed at `LockTimeout = -1`, and the check that decides whether `CREATE DATABASE` runs is still the
one held under the lock. What changed is only whether the lock is on the path at all.
`postgresdb.AdvisoryLockKey` and `mssqldb.CreateDatabaseResource` are now exported so the data tier,
a separate module, can hold the exact lock rather than a restatement of it.

**Residual, inside decision 5's accepted cost rather than beside it.** A hostile holder can still
stall the one start that genuinely has to create the database. That is "a stuck holder blocks startup
rather than failing it", now bounded to the creation window instead of the life of the deployment.
Recorded rather than fixed, because timing the lock is what decision 5 rejected.

**Regression test.** `TestNewDatabase_CreateTrue_AnUnrelatedLockHolderDoesNotBlockAnOrdinaryRestart`
in `src/authserver/tests/data/database_create_test.go`, at seam 1. It creates the database, holds the
production lock identity from an unrelated session on the maintenance database, then requires a
`Create: true` constructor to return inside a bounded 30s wait, bounded so a regression fails rather
than hanging the tier. PostgreSQL and SQL Server; skipped on MySQL and SQLite with the reason. Passed
at 0.04s and 0.14s with the lock held.

#### Round 1's `unchecked`, all four answered

1. *"No integration tier was run."* — **Dismissed.** Decision 8 records integration as not applicable
   and the diff reaches no handler, route, template, schema or HTTP flow.
2. *"Azure SQL contained-user and managed PostgreSQL deployments were not live-tested."* —
   **Dismissed** for this change, with the reason stated: both are decision 3's motivation for the
   *skipping* path, which opens no maintenance connection at all, so there is nothing engine-hosted
   for a managed deployment to refuse. The creating path is what needs `master`/`postgres`, and a
   managed deployment that cannot reach them is exactly the deployment expected to set
   `GOIABADA_DB_CREATE=false`. The new database setup page says so.
3. *"Token validation, redirect and PKCE handling, one-shot claims, revocation reach, session
   lifecycle, endpoint authorization, rate limiting and audit emission were not re-reviewed."* —
   **Dismissed.** The diff does not reach those surfaces; the reviewer says as much itself.
4. *"The full four-engine data tier and Astro build were not rerun."* — **Covered.** The data tier ran
   on all four engines twice after the fix, green both times (1548 pass, 32 skip, 0 fail on the
   second), plus the unit tier across all three modules. The docs tree hash is unchanged by this round
   (`2952b07dd534` before and after), so stage 2's Astro build still stands against the committed tree.

#### What round 1 ran

|---|---|---|---|
| unit | `where.sh test --type modules` | n/a | green, all three modules |
| data | `where.sh test --type data` | sqlite, mysql, postgres, mssql | green, 1548 pass, 32 skip, 0 fail, ~175s |
| integration | not run | n/a | not applicable, decision 8: no handler, route or schema is touched |

Mutations against the fix, all three **caught**:

| File | Replacement | Case that caught it |
|---|---|---|
| `src/core/data/postgresdb/db.go` | `databaseExists(ctx, maintenanceDB, name)` → `...name+"_no_such_db"` | `..._AnUnrelatedLockHolderDoesNotBlockAnOrdinaryRestart` |
| `src/core/data/mssqldb/db.go` | `databaseExists(ctx, masterDB, name)` → `...name+"_no_such_db"` | `..._AnUnrelatedLockHolderDoesNotBlockAnOrdinaryRestart` |
| `src/core/data/postgresdb/db.go` | `databaseExists(ctx, maintenanceDB, name)` → `true, error(nil)` | `..._ConcurrentConstructorsAgainstAnAbsentDatabase` |

The third pins the other direction: a pre-check that always claims the database is there would skip
creation entirely, and goal 1's own case fails on it. The early return is guarded from both sides.

Tree at commit: `production 5c1d2a50b8d0`, `unit-tests e7ed874bf957`, `data-tests 408084cc7f7b`,
`docs 2952b07dd534`, checked with `tree-hash.sh --split` after the mutation runs, with `git status`
showing exactly the four intended files, so no mutation survived into the commit.
`check-anchors.sh`: all 19 rows pass.

The fix landed as a new commit on the branch rather than being folded back into stage 3, and
`log/stage-3.md` points at it.

---

### Final gate, round 2: clean

Request `293-20260829T170500Z`, resumed session, scope the whole branch against `origin/main`, five
commits, `4a892079` through `95dc8b3e`. Round 2 existed for one reason: round 1's blocking fix was
production code no reviewer had read.

**Verdict: clean, zero findings.** All three axes reviewed. What it ran, from the verdict's own list:

- Re-read the agreement sections 0 to 5, the plan, the three stage logs, the closing log, the round 1
  account, and the whole `95dc8b3e` corrective diff.
- `git diff origin/main...HEAD` and `git diff 38e82928...95dc8b3e`, plus repository-wide references to
  `databaseExists`, `AdvisoryLockKey`, `CreateDatabaseResource`, `pg_advisory_lock` and `sp_getapplock`.
- `git diff --check origin/main...HEAD`; `check-anchors.sh`; `tree-hash.sh --verify`.
- `where.sh ci 95dc8b3e`, reported all eleven jobs green.
- The data tier on postgres for
  `TestNewDatabase_CreateTrue_(AnUnrelatedLockHolderDoesNotBlockAnOrdinaryRestart|ConcurrentConstructorsAgainstAnAbsentDatabase)`,
  and on mssql for those two plus `TestNewMsSQLDatabase_CreateTrue_CaseVariantNamesRaceToOneDatabase`.
- Replayed its own round 1 reproduction against the fixed constructors with unrelated low-privilege
  holders, including a SQL Server case-variant database name.
- Confirmed `git status` and the tracked diff were unchanged after all of its experiments.

Its note, verbatim: *"Round 1's ordinary-restart lock denial is fixed on both engines. The unlocked
prechecks fail closed, preserve the under-lock creation decision and case-equivalent SQL Server
identity, and are covered from both early-return directions."*

Both `unchecked` entries were dismissed on stated evidence: no integration tier (decision 8, and the
diff's fifteen files are four documentation files plus `src/core/config`, `src/core/data`,
`src/core/cmd/schemadump` and the data tier's fixtures, with nothing under `internal/handlers`,
`internal/server`, `oauth`, `validators`, `web/template` or any `migrations/` directory), and the
OAuth surfaces, every one of which is downstream of a database handle this change only changes the
construction of.

Nothing moved during round 2: the working tree was clean before it and after it. The claim that
carries the branch is therefore the check suite rather than a local run, which is also what settles
the staleness `tree-hash.sh --verify` reported: all three stage entries recorded their tiers against
trees that `95dc8b3e` then moved, so the committed tree is judged as a whole by the suite that runs
data on all four engines in parallel.

Issue #266, opened 2026-08-27
Replace ChunkedCookieStore with a server-side SQL session store: the real ceiling is the reverse proxy's header limit, not the 50-chunk cap


leodip:
Both modules keep **all** session state in the browser. `sessionstore.ChunkedCookieStore` gob-encodes `session.Values`, hands the whole thing to securecookie for encryption and HMAC, then `Save` splits the result into `DefaultChunkSize` (3800 byte) pieces and writes one cookie per piece plus a metadata cookie. `DefaultMaxChunks = 50` advertises a ceiling of roughly 190 KB.

**That ceiling does not exist.** The limit that binds in production is the reverse proxy's maximum header size, and it is about a tenth of it. Goiabada never sees the failure: Go's `net/http` allows 1 MB of headers, so the request dies upstream and the operator gets an opaque 400 or 502 with nothing in the Goiabada log to explain it.

## Measured

Encoding realistic payloads through `ChunkedCookieStore.Save` against an `httptest.ResponseRecorder`, then summing `len(name) + 1 + len(value) + 2` over the cookies it sets, which is what the browser sends back on the next request:

| payload | cookies set | `Cookie:` request header |
|---|---|---|
| auth server, `AuthContext` ~1.2 KB | 2 | ~2.7 KB |
| auth server, `AuthContext` ~2.5 KB | 3 | ~5.1 KB |
| admin console, tokens ~1.5 KB | 2 | ~3.7 KB |
| admin console, tokens ~4.8 KB | 4 | ~9.6 KB |
| admin console, tokens ~9.3 KB | 6 | ~17.7 KB |

Expansion from payload to request header is about **1.9x**: gob, then AES-CTR, then base64, then HMAC, then base64 again.

## Why 50 chunks is a fiction

| | Default limit | On exceeding |
|---|---|---|
| **nginx** | `large_client_header_buffers 4 8k`, so one `Cookie:` line must fit in 8 KB | 400 |
| **Apache httpd** | `LimitRequestFieldSize` 8190 | 400 |
| **HAProxy** | `tune.bufsize` 16384 for the whole request head | 400 |
| **Envoy** | `max_request_headers_kb` 60 | 431 |
| **Goiabada itself** | Go's `DefaultMaxHeaderBytes`, 1 MB | never reached |

[...]

PR #296, opened 2026-08-29
refactor(adminconsole): two env vars stop configuring what the auth server owns

### LINKED ISSUE #285: Two admin console env vars configure values the auth server owns: remove GOIABADA_ADMINCONSOLE_OAUTH_CLIENT_ID and GOIABADA_ADMINCONSOLE_ISSUER
Follows #266 / PR #267, decision 23.

Two settings in the admin console's environment name a value the auth server owns. Neither has a
second correct value, both can be set to something the auth server will not agree with, and in both
cases the disagreement locks administrators out of the admin console with no way in through the admin
console itself. Remove both and derive the value instead.

They are independent changes and can land in either order.

# 1. `GOIABADA_ADMINCONSOLE_OAUTH_CLIENT_ID`

Replace with `constants.AdminConsoleClientIdentifier`, the constant that already holds the only value
it can usefully take. The `--adminconsole-oauth-client-id` flag goes with it.

## Why

Three facts, all in the tree today:

1. **The seeder can only ever create that one name.** The admin console client is built in
   `database_seeder.go` with `ClientIdentifier: constants.AdminConsoleClientIdentifier`. Configuration
   is not consulted, and cannot be: the seeder runs on the auth server.
2. **The name cannot be changed afterwards.** `Client.IsSystemLevelClient` covers it, and the clients
   API in `handler_api_clients.go` has exactly two guards keyed on that, one refusing deletion and one
   refusing an identifier change.
3. **Nothing a custom client offers is unavailable on the seeded one.** Its secret (there is a
   generate-new-secret action on the authentication page), redirect URIs, web origins, default ACR
   level, consent and enabled flows are all editable. What the variable buys is a different string in
   the clients list.

Against that, the variable has a standing cost. It lives in the admin console's environment, while
the seeder and the migrations run on the auth server, which cannot read it. So every feature that
needs the auth server to provision something for "the admin console's client" has to choose between a
fixed name and a configured one, with no way to honour the configured one. #266 hit this first:
migration `000035_browser_sessions` grants `authserver:browser-sessions` and enables the client
credentials flow for the literal `admin-console-client`, so a deployment pointing the admin console at
a different client is locked out of it completely after upgrading. This will not be the last feature
to meet that wall.

The environment variables reference makes it worse by listing `admin-console-client` in a column
headed "Default", which reads as one valid value among many. There is one valid value.

## What to change

Code:

- `src/core/config/config.go`: drop `AdminConsoleConfig.OAuthClientID`, its `getEnv` call and its
  `flag.StringVar`.
- `src/adminconsole/cmd/goiabada-adminconsole/main.go`: the emptiness check, and the argument passed
  to `apiclient.NewSessionTokenSource`.
- `src/adminconsole/internal/server/server.go`: `Server.Start`'s startup validation now covers the
  secret alone, and its message needs rewording.
- `src/adminconsole/internal/server/routes.go`: the argument passed to `NewMiddlewareJwt`.
- `src/adminconsole/internal/handlers/handler_auth_callback.go`: the `clientID` used for the code
  exchange.
- `src/core/middleware/middleware_jwt.go`: the `clientID` field and constructor parameter, and the
  fallback in `JwtSessionHandler` that already reaches for the constant when the configured value is
  blank. That branch becomes unconditional and can go.
- `src/core/data/database_seeder.go`: the bootstrap env file it writes no longer needs the client id
  line.
- `src/authserver/cmd/goiabada-authserver/main.go`: the two operator instruction blocks that list the
  variable.
- `src/cmd/goiabada-setup/main.go`: `OAuthClientID` in the wizard's config struct, its default, and
  the three generators that emit it (compose file, shell exports, Kubernetes manifest).

Shipped artifacts:

- `src/cmd/goiabada-setup/docker-compose.yml` and `src/cmd/goiabada-setup/goiabada-k8s.yaml`
- the five `src/build/docker-compose-*.yml` files, where it appears as a commented line

Docs:

- `site/src/content/docs/reference/environment-variables.mdx`
- `site/src/content/docs/production-deployment/production-checklist.mdx`
- `site/src/content/docs/production-deployment/native-binaries.mdx`

## Compatibility

- Every deployment produced by `goiabada-setup` or by the shipped compose files sets the variable to
  `admin-console-client`, so for all of them this is a no-op. The value stops being read and a stale
  line in an environment file is ignored. Operators need change nothing.
- Breaking for anyone running the admin console under a client of their own. They must point it back
  at `admin-console-client`, which always exists and cannot have been renamed, and move their redirect
  URIs and secret onto it. Release note required.
- One escape hatch closes with it: a database whose admin console client is not named
  `admin-console-client`, from a rename made before the guard existed or from a hand-built
  installation. After this change no configuration reaches such a client and the failure appears at
  token request time. Worth deciding whether that deserves a startup check or only a release note.
- If PR #267 lands a startup warning comparing the configured id against the constant (its decision
  23, option B), that comparison becomes dead code here and should be removed along with the rest.

# 2. `GOIABADA_ADMINCONSOLE_ISSUER`

Replace with the issuer the auth server reports, carried on the public settings payload the admin
console already fetches. The `--adminconsole-issuer` flag goes with it.

## Why

This one is worse than the client id, because no unusual configuration is needed to fall into it. The
product's own UI is enough.

- The issuer is a **database setting**, `settings.Issuer`, seeded to the auth server base URL and
  stamped into every token by `token_issuer.go`.
- It is **editable from the admin console**, under Settings > General, through
  `handler_api_settings_general.go`.
- The admin console validates the `iss` of every token it holds against `settings.Issuer`, in
  `MiddlewareJwt.JwtSessionHandler`. But the `models.Settings` it reads is assembled in
  `MiddlewareSettingsCache`, and while `AppName`, `UITheme` and `SMTPEnabled` come from the auth
  server, `Issuer` alone comes from `config.GetAdminConsole().Issuer`, which is the environment
  variable, defaulting to the auth server base URL.

The two agree at install time and disagree the moment anyone edits the issuer. What follows:

1. An administrator changes the issuer in Settings > General.
2. `handler_admin_settings_general.go` notices the change, deliberately clears the session and sends
   them to log in again, which is the right thing to do since tokens carrying the old `iss` are now
   stale.
3. They log in. The new token carries the new `iss`.
4. `JwtSessionHandler` compares it against the environment variable, which still holds the old value,
   finds `hasInvalidIssuer`, clears the session and redirects to `/`.
5. `/` needs a session. Back to step 3.

There is no way out from inside the admin console. Somebody has to edit the environment and restart
the process. The setting is offered in the UI as an ordinary field with no warning attached.

## What to change

- `src/authserver/internal/dtos/public_settings.go` and its copy at
  `src/adminconsole/internal/dtos/public_settings.go`: add `Issuer`.
- `src/authserver/internal/handlers/handler_public_settings.go`: map `settings.Issuer` onto the
  response. That handler's comment records the DTO as the whole boundary between an anonymous caller
  and the 32 fields of `models.Settings`, and `handler_public_settings_test.go` fails when the
  boundary widens. Widening it here is deliberate and safe: the same value is already served
  unauthenticated at `/.well-known/openid-configuration`. Update the guard test as a conscious act,
  not as a fix to a red build.
- `src/adminconsole/internal/middleware/middleware_settings_cache.go`: take `Issuer` from the fetched
  payload alongside the other three fields.
- `src/core/config/config.go`: drop `AdminConsoleConfig.Issuer`, its `getEnv` call and its
  `flag.StringVar`.
- `site/src/content/docs/reference/environment-variables.mdx`: remove the row. It appears in no
  compose file, no Kubernetes manifest and no other page.

Considered and rejected: fetching `/.well-known/openid-configuration` instead. The admin console
already calls the public settings endpoint on every request through `SettingsCache`, so reusing that
path costs one field rather than a second HTTP client and a second cache.

## Compatibility

- No deployment artifact in the repository sets this variable, so for anyone using the tooling this
  is a no-op, and a stale line in a hand-written environment file is ignored.
- Anyone who did set it was, at best, repeating the auth server base URL. Afterwards the admin console
  uses the value tokens actually carry, so it can only be more correct.
- `handler_admin_settings_general.go` already calls `settingsCache.Invalidate()` after an update, so
  the instance that made the change sees the new issuer at once. Other replicas keep the old one for
  up to `cacheTTL`, currently 30 seconds, and bounce sessions to `/` during that window. That is a
  brief interruption immediately after an issuer change, in place of today's permanent lockout.
- Worth a regression test: change the issuer through the API, then complete a sign-in and confirm the
  session survives. Nothing covers that path today, which is why the loop exists.

# Not in scope

`GOIABADA_ADMINCONSOLE_OAUTH_CLIENT_SECRET` stays. It is per deployment, genuinely secret, and has no
single correct value.

leodip:
Closes #285.

## What this is for

Two environment variables let an operator configure the admin console with values the auth server
owns, and both can be set to something the auth server disagrees with.

`GOIABADA_ADMINCONSOLE_ISSUER` is the damaging one. The admin console validates the `iss` claim of
every token against it, while the auth server stamps tokens with `settings.issuer` from its database.
Change the issuer in Settings > General and the admin console locks its administrator out
permanently: the settings handler clears the session on purpose, the new token carries the new
issuer, the JWT middleware compares it against the stale configured value, clears the session again,
and every attempt lands back at `/`. The only way back in is editing the environment or direct SQL,
neither reachable from the console you can no longer sign in to.

`GOIABADA_ADMINCONSOLE_OAUTH_CLIENT_ID` is the quieter one. The seeder creates the admin console's
client as the literal `admin-console-client` and never reads configuration, a rename guard has
refused to change that identifier since before the first release, and migration 000035 provisions the
client-credentials switch and the browser-sessions permission against that same literal. So the
variable has exactly one correct value, presented in the documentation under a column headed
"Default" as though it were one of many.

## The rule this establishes

Neither value is configurable on the admin console. The client id is derived at compile time from
`constants.AdminConsoleClientIdentifier`; the issuer is derived at run time from the auth server that
stamps it into tokens, fetched on the existing `/api/public/settings` call the admin console already
makes on every request.

OpenID Connect Core 1.0 section 3.1.3.7 requires the relying party's Issuer Identifier to *exactly*
match the `iss` claim, so the current middleware is a correct implementation of a wrong input. After
this change that match is between two copies of one value with a single writer, instead of between a
database column and an environment variable that nothing keeps in agreement. Discovery 1.0 section 3
describes the same model: the Issuer Identifier is something the OP asserts and the RP obtains from
it.

### Stage 1: The issuer comes from the auth server

`ed3caa71`. The issuer now has one writer.

`issuer` joins both `PublicSettingsResponse` copies (they are hand-kept mirrors in different modules,
and the admin console's now says so) and the public settings handler maps `settings.Issuer` onto it.
That widening discloses nothing new: the same string is already served to anonymous callers at
`/.well-known/openid-configuration`, which OIDC Discovery 1.0 section 3 requires. The endpoint's
field allowlist gains the entry, which is exactly the deliberate decision that guard exists to force
into review.

`MiddlewareSettingsCache` reads the fetched issuer instead of `config.GetAdminConsole().Issuer`, and
answers 500 naming the auth server when the payload carries no issuer. `IsIssuerValid` is plain
string equality, so an expected issuer of `""` would reject every token the auth server can mint and
reproduce the lockout silently; the only live route to it is an auth server too old to serve the
field.

`AdminConsoleConfig.Issuer`, `GOIABADA_ADMINCONSOLE_ISSUER` and `--adminconsole-issuer` are gone,
along with the row that listed them in the environment variables reference.

New `middleware_settings_cache_test.go`, the first test file that package has had: the issuer and the
other three fields both come from the payload (the second case is what stops the first being
satisfied by a hardcoded `models.Settings`), an absent issuer and an explicitly empty one are each
refused, and an unreachable auth server is refused.

Tiers: unit 4440 pass on all three modules, integration 1333 pass on sqlite. Data is not applicable,
since the stage adds no query, interface method, migration or schema change. Three mutations against
the new seams, all caught.

### Stage 2: The client id is the constant

`1b84b4a1`. The admin console authenticates as one identifier, and it is not configurable.

`GOIABADA_ADMINCONSOLE_OAUTH_CLIENT_ID` and `--adminconsole-oauth-client-id` are gone. All five
readers take `constants.AdminConsoleClientIdentifier`: the client credentials source for the session
store, the authorization code exchange, the JWT middleware's construction, and the two startup
guards, which now check only the secret. That constant is what the seeder writes and what migration
000035 grants the client credentials switch and the browser-sessions permission against, so nothing a
deployment can set will point the admin console at a client the auth server has not provisioned.

Nothing is lost by removing it. `IsSystemLevelClient` covers the identifier and the admin API's
rename guard has refused to change it since `37566d31`, which predates `v0.0.1`, so no released
version ever let the seeded name and the configured name disagree.

`NewMiddlewareJwt` keeps its `clientID` parameter: the admin console passes the constant and the auth
server keeps passing `""`, which leaves `core/middleware` naming no module's identity. What goes is
the fallback `RequiresScope` applied when that field was blank. The auth server constructs the
middleware for `JwtAuthorizationHeaderToContext` alone and reaches neither the fallback nor the
refresh path, and `refreshToken` keeps its "both credentials or no refresh" guard unchanged.

Three tests in `middleware_jwt_test.go` constructed with `""` while asserting `RedirToAuthorize`
received the constant, so they were passing by way of the fallback and step 2 would have left them
asserting nothing about the caller. They construct with the constant now. A mutation replacing
`m.clientID` with `""` is caught by all three.

Two comments that had become false were rewritten rather than left: `refusalMessage` and its test
described a deployment that pointed the variable at its own client, which can no longer exist, so
they now describe the refusal's remaining route, an administrator turning client credentials off on
the seeded client or taking the browser-sessions permission away from it. `settings_client.go`'s
promise of "appName, uiTheme, smtpEnabled" was falsified by stage 1 and now names `issuer`.

Docs drop the variable from the environment variable reference, the production checklist and the
native-binaries export block. Three deletions, nothing added.

Tiers: unit 4440 pass, 0 fail on all three modules; integration 1333 pass, 0 fail on sqlite. Data not
applicable, no migration, schema or data-layer method moved. One mutation, caught.

### Stage 3: Operator surfaces stop naming either variable

Landed in `fffacdda`. The sweep of everything that told an operator to set
`GOIABADA_ADMINCONSOLE_OAUTH_CLIENT_ID`, or set it for them. Nothing a running deployment reads
changes.

The legacy two-step bootstrap file stops carrying the variable, and the banner announcing its
contents names the secret alone rather than "OAuth client ID and secret". Both of the auth server's
operator instruction blocks drop it. The setup wizard drops the config field, its hardcoded default
and all three emitters, and its two committed outputs, `docker-compose.yml` and `goiabada-k8s.yaml`,
are edited to match what the emitters now produce. The five shipped compose files drop the commented
line and the two statements it left false: the `BOOTSTRAP:` banner calling `bootstrap.env` "OAuth
client ID/secret + session keys", and "uncomment the 4 lines below", now three. The devcontainer
stops setting it, and the admin console Makefile's `serve` guard falls back to the client secret
alone.

`GOIABADA_ADMINCONSOLE_ISSUER` needed nothing here: it appears nowhere in the tree outside one
explanatory comment in a test stage 1 added. That absence is what makes stage 4's refusal on any
value of it safe.

**Deviation from the plan.** The plan described this stage as a sweep with no new tests. It landed
with one refactor and one new test file, because the stage's only real behaviour change, the
bootstrap file no longer carrying the client id, was built inline inside `Seed`, which needs a live
database. Its only reachable tier was therefore the data tier, which this stage does not run, so the
change would have shipped as a claim nothing could check. The content moved into a pure
`bootstrapEnvContent` and is now pinned by two unit cases in `core/data`: no client id variable
survives and no `admin-console-client` literal under any other name, and the five remaining
variables each carry their own value with a count assertion so a sixth cannot arrive silently.

Tiers: unit 4442 pass, 0 fail, 57 packages on all three modules; `go build` clean in
`src/cmd/goiabada-setup`, which has no test tier of its own. Integration and data not applicable, no
handler, route, schema or database access moves. Two mutations against the new cases, both caught.
The `make serve` guard was exercised in all three arms with `air` stubbed: it refuses without a
secret, starts with one, and still loads a bootstrap file when present.

### Stage 4: Startup refuses a stale configuration

Commit `53f608d9`. Stages 1 to 3 removed both variables; this one makes a deployment that still
carries either of them say so at startup instead of failing later and elsewhere.

`config.ValidateRemovedAdminConsoleVars` sits beside the other fail-fast validations and is called
from the admin console's `main()` right after the configuration loads. It reads the raw environment,
because neither value is held in config any more, sharing the existing `deprecatedEnvVarsPresent`
helper so that a variable set to an empty string still counts as present.

The two arms are deliberately asymmetric:

- **`GOIABADA_ADMINCONSOLE_OAUTH_CLIENT_ID`** refuses only when its trimmed value differs from
  `admin-console-client`. Present and equal starts silently, which is every deployment the setup
  wizard has ever produced. A different value would point the admin console at a client the auth
  server never provisioned, and that deployment breaks on the next release anyway, since migration
  000035 provisions the client credentials switch and the browser sessions permission only on the
  seeded client.
- **`GOIABADA_ADMINCONSOLE_ISSUER`** refuses whenever it is present at all. Nothing in the tree ever
  set it, so any value is a hand-written line, and leaving it silently unread is the state this
  change exists to end.

Each message names the variable, the value found, why it is no longer configuration and what to
remove. The client id is checked first, so an operator carrying both wrong values is told about the
client id, which is the one that would otherwise fail as an unrecognised client rather than as a
lockout.

It refuses rather than warning, which is the difference from the removed-setting warning loop it
borrows its helper from: an operator can miss a log line in a running deployment, and missing it is
what the two silences cost until now.

Tiers: unit 4452 pass, 0 fail, 57 packages on all three modules. Data and integration not
applicable, no migration, model, query, handler, route or schema moves and the new function is
reachable only from `main()`. Eight cases over both variables, including set-empty, surrounding
whitespace and both-wrong-at-once. Five mutations, all caught.

### After the review

Two commits landed after stage 4, both from the single review of the whole branch. Neither changes
what the admin console does.

`7a1a6959` gives the setup wizard a tier that can fail. `src/cmd/goiabada-setup` had no test file at
all, so `go test ./...` answered `[no test files]` and exited 0, and no unit job in `check.yml` ran
that module anyway, so a test added there would not have executed. Stage 3 edited three generators an
operator deploys from and could only claim `go build` clean. The module now has `main_test.go`, a table
over four generators and the two committed wizard outputs, each case requiring both that the removed
variables are absent and that a named line is still present, so nothing passes vacuously; and a
`Unit / setup` job runs it. The same commit adds a comment in `middleware_settings_cache.go` recording
that its English error text follows the same convention as the error above it and the CSRF rejection
beside it, all three of which answer before locale resolution.

`ae05d6aa` closes the one gap round 2 found in that new test: the committed-artifact half asserted only
the `# Generated by goiabada-setup` banner, which is the first line of each file and so survives any
truncation, meaning a banner-only file passed. Each artifact now names a line it must still carry.

### After the pull request was out of draft

`06cb74d0` is not #285's work at all. It is the first of the two follow-ups this run drafted, fixed
here rather than filed.

Two compiled binaries were tracked, both outputs of the same `main` package:
`src/cmd/goiabada-setup/goiabada-setup` (11 MB) and
`src/cmd/goiabada-setup/build/goiabada-setup-linux-amd64` (10 MB), the only compiled files in the
tree. The first was a trap. A `main` package's default build output path is the package directory
itself, so a plain `go build ./...` in that module wrote to exactly the path of the tracked file,
silently replacing a committed artifact with one built on the developer's machine and staging a
multi-megabyte diff. Nothing warned. It fired during stage 3, where checking that the wizard still
compiled after a five-line source change produced an 11 MB to 12 MB binary diff alongside it, caught
only by reading `git status` before committing.

Nothing read either copy. Releases are built from source: release.yml runs `build-binaries.sh` in both
`src/build` and `src/cmd/goiabada-setup` and uploads what those write, the docs link to the release
downloads, and `Build validation` cross-compiles the same way. So the committed binaries were
artifacts of unknown provenance, with nothing in the repository recording which commit or platform
produced them.

Both are untracked now and `.gitignore` covers both paths, plus the release archives
`src/build/build-binaries.sh` writes beside itself. No output path needed changing: the Makefile
already builds to `build/goiabada-setup` and `build-binaries.sh` to
`build/goiabada-setup-<os>-<arch>`, and `go build -o` creates the directory, so a fresh clone that no
longer gets `build/` from git still builds. An ignore rule stops the accident and nothing else, so the
Lint job gains a step that fails on any tracked ELF, Mach-O or PE file, covering `git add -f` and a
`main` package added in some other directory.

Verified by rebuilding: `rm goiabada-setup && go build ./...` produces a 12.7 MB binary and leaves
`git status` clean, which is the exact sequence that staged a diff during stage 3. The new guard prints
nothing against the tree, prints the path when one binary is force-added back, and goes quiet when it
is removed again. `make build` works with `build/` absent, the setup module's tier passes, and the
workflow file parses.

This is the second of the three findings [#187](https://github.com/leodip/goiabada/issues/187)'s title
promises. That issue stays open for its third.

### The second follow-up landed here too

`647bffff`. The other of the two follow-ups this run drafted, also fixed rather than filed.

Ten responses in the admin console were hardcoded English, and they were the ones an administrator is
least able to act on. Two different causes, and only the first is what the round 1 reviewer reported.

**Nothing had resolved a locale yet.** `Server.initMiddleware` mounted `i18n.MiddlewareLocale` last on
the application branch, so both of `MiddlewareSettingsCache`'s refusals ran with no localizer to reach
for, and `MiddlewareCsrf` sits on the root router above that branch, earlier still. An administrator
whose browser asks for pt-BR was told in English that their auth server needs upgrading, which is the
message stage 1 of this branch added.

The locale middleware goes first on the branch now. Nothing depended on the old order: `resolveLocale`
reads `?ui_locales`, the AuthContext and `Accept-Language`, and touches no settings, no session and no
database. The authserver resolves locale last for a reason that does not carry over, that it passes a
real `AuthHelper` and so must run after the session is decoded; this module passes `nil`. CSRF stays on
the root router, because moving it onto the branch to gain a localizer would leave any route registered
outside that branch unprotected; it resolves a tentative localizer of its own through a new
`i18n.ResolveRequestLocale`, which is what `MiddlewareLocale` already does with a nil reader.

**The other seven were never translated.** They sit below the locale middleware and always had one;
what they lacked was a message. Each answered with `http.Error` and the Go error text, so "unable to
cast the session value to TokenResponse" reached the browser while the log, where that detail belongs,
got nothing. They render through `HttpHelper.InternalServerError` now, the localized error page every
handler in the module already uses, which logs the cause against the request id the page displays.
`MiddlewareJwt` takes a `ServerErrorRenderer` for it, in the same narrow-interface idiom as the
`ErrorRenderer` beside it, and `SessionIdentifierToContext` takes the same.

Three new catalog keys in both locales, none for the seven: the error page's own entries already exist.
Two refusals stay plain text rather than rendering that page, deliberately, since it reads settings off
the context for its layout and settings are the thing that is missing when they fire.

Two details fall out of the same change. The CSRF reason is no longer interpolated into the response,
where it described the deployment's origin handling to whoever was refused, an attacker included;
`explainCsrfFailure` already carries it into the log. And the settings-fetch error is now logged at
all, which it never was, as a single line rather than slog's `%+v` stack, since a down auth server
produces one per request.

The chain order is what a middleware unit test cannot see, so `TestInitMiddleware_RefusalsAreLocalized`
drives the real `initMiddleware` and asserts both refusals in pt-BR, in English and through
`?ui_locales`; every localized row first requires the two catalog entries to differ, so a chain that
resolved nothing cannot pass. The six `MiddlewareJwt` sites each require the cause at the renderer and
absent from the body, since a 500 alone was already true before. `SessionIdentifierToContext` had no
test at all and has three now, its happy path included, so the refusal case cannot be satisfied by a
middleware that refuses everything.

One thing host-only testing would have missed: the integration tier asserted the old
`"Forbidden - "` prefix, so this would have gone red on all four database jobs. It reads the catalog
entry now.

Tiers: unit 4668 pass, 0 fail, 57 packages on all three modules; integration 1333 pass, 0 fail on
sqlite, which is the tier that assertion lives in. Data not applicable. gofmt, vet, unparam and
golangci-lint clean on all three modules. Eight mutations, all caught.

## For the release notes

**Both variables are gone, and the admin console refuses to start if either is still set.**

- `GOIABADA_ADMINCONSOLE_ISSUER` is refused whenever it is present, with any value, including empty.
  Remove the line. The issuer now comes from the auth server on every request.
- `GOIABADA_ADMINCONSOLE_OAUTH_CLIENT_ID` is refused when its value is anything other than
  `admin-console-client`. Set to that literal it starts silently, which is what every version of the
  setup wizard has ever emitted, so a deployment produced by the tooling needs no action. Removing the
  line is still the right thing to do; it is no longer read.
- `--adminconsole-issuer` and `--adminconsole-oauth-client-id` are gone too. `flag.Parse` exits 2 on
  an undefined flag, so a start script passing either fails loudly.

**What will no longer work:** pointing the admin console at an OAuth client you created yourself. No
released version ever permitted it, since the seeder writes `admin-console-client` and the admin API's
rename guard has refused to change that identifier since before `v0.0.1`, and migration 000035
provisions the client credentials switch and the browser sessions permission only on the seeded client.
A deployment that did it by direct SQL now stops at startup with a message naming the variable instead
of failing later as an unrecognised client.

**Upgrade ordering.** The admin console reads the issuer from `issuer` on the auth server's
`/api/public/settings`, which is new in this release. If the two roll independently, an admin console
on the new version talking to an auth server on the old one answers 500 naming the auth server until
the auth server rolls. The compose files and the Kubernetes manifest pin both images to one tag, so
this affects only a deployment that upgrades them separately.

**Changing the issuer no longer locks you out.** It used to be permanent: the settings handler clears
the session on purpose, and the new token then failed the JWT middleware's comparison against the
stale configured value forever. In a deployment running more than one admin console replica, sessions
served by a replica whose settings cache has not expired still bounce to `/` for up to 30 seconds after
the change. A single instance never sees it.

## Tiers

**Locally, per stage:** unit on all three modules at every stage, ending at 4642 pass, 0 fail, 57
packages, 1304 top-level tests; the setup module's own tier at 8 pass, 0 fail; integration on sqlite at
1333 pass, 0 fail for stages 1 and 2. Data is not applicable anywhere on this branch: no migration,
schema, model, query or data-layer interface method moves. Integration is not applicable to stages 3
and 4 or to either review commit, none of which moves a handler, route or schema. Eleven mutations
across the stages and nine more during the review, every one caught. `check-anchors.sh` resolves all 27
rows.

**The four-engine result is the check suite's, and it is green on `ae05d6aa`:**
[run 33279030107](https://github.com/leodip/goiabada/actions/runs/33279030107). All twelve jobs, which
is data and integration on sqlite, mysql, postgres and mssql as four parallel jobs, the four unit jobs
including the new `Unit / setup`, plus Lint, Vulnerabilities, Build validation and Versions. Nothing
local ran mysql, postgres or mssql. `06cb74d0` re-runs the same suite: it edits `.gitignore` and
`check.yml` and removes two binaries, so no Go source, template, schema or migration moves, and the new
Lint step is the one thing in it that CI had not already run. `647bffff` re-runs it again, and does move
Go source in all three modules: it was verified locally at unit 4668 pass, 0 fail, 57 packages and
integration 1333 pass, 0 fail on sqlite, with gofmt, vet, unparam and golangci-lint clean.

## Deviations from the plan

- **Stage 3** was planned as a sweep with no new tests and landed with a refactor and a new test file.
  Its one real behaviour change, the bootstrap file no longer carrying the client id, was built inline
  inside `Seed`, which needs a live database, so its only reachable tier was one the stage does not
  run. The content moved into a pure `bootstrapEnvContent` and is pinned in the core unit tier.
- **Stage 4** was sketched with each rejection message saying the admin console always authenticates as
  `admin-console-client`. That is the remedy for the client id and a non sequitur for the issuer, so
  each message carries its own remedy instead.
- **The setup module's test tier and its CI job** were in no stage. They came out of the review, which
  found that the module had no executed test at all.
- **The committed binaries** were in no stage and are not #285's subject. They were drafted as a
  follow-up on #187 and fixed here on request, after the pull request was already out of draft.
- **The English-only middleware rejections** were the round 1 reviewer's deferred finding, recorded at
  the site as a decision and drafted as a follow-up. Fixed here on request instead, which reverses that
  deferral: the comment stage 4 left saying the message follows the English convention is gone, and the
  ordering question the deferral called a decision was settled by reading what `resolveLocale` actually
  depends on.

### Review

Reviewed by `openai-codex/gpt-5.6-sol` at effort `xhigh`, 2 rounds over the whole branch. No stage was
reviewed on its own.

|---|---|---|---|
| 1 | 2 | 0 | one fixed, `7a1a6959`; one deferred as a follow-up, with a comment left at the site |
| 2 | 1 | 0 | fixed, `ae05d6aa` |

- **significant, quality:** the setup wizard's three changed generators had no executed test, and the
  module had no test file and no CI job, so reintroducing the removed variable left it green. Fixed in
  `7a1a6959`: the module's first test file, a `Unit / setup` job, five mutations all caught.
- **significant, quality:** that new test's committed-artifact half accepted a file truncated to its
  banner line. Fixed in `ae05d6aa`: each artifact now names a line it must still carry. Four mutations,
  all caught.
- **minor, conformance:** the new missing-issuer error is English only, because it answers before
  locale resolution. Deferred at review time, then fixed on request in `647bffff` along with the nine
  other responses in the same family, so the comment recording the convention is gone with it.

No round 3: round 2's fix left `production` byte-identical, and nothing was left open.

## The run's three closing comments

- [Decisions, and what reversing them costs](https://github.com/leodip/goiabada/pull/296#issuecomment-5465328773)
- [Follow-ups and folded-in fixes](https://github.com/leodip/goiabada/pull/296#issuecomment-5465328838), two follow-ups drafted. That comment says neither was filed, which was true when it was written: both were fixed here instead, in `06cb74d0` and `647bffff`, so neither needs filing
- [The code review in full](https://github.com/leodip/goiabada/pull/296#issuecomment-5465328907), both rounds with the evidence

## Status

Complete. Four stages landed plus two review commits, the whole branch reviewed twice, the check suite
green on `ae05d6aa`, and the pull request out of draft. `06cb74d0` landed after that, on request: one
of the run's two drafted follow-ups, fixed here rather than filed, and `647bffff` after it with the
other one.


leodip:
## Decisions

Seven decisions, all **Decided**, none open and none deferred with an assumption standing in for an
answer. The run raised one of them (7) and settled it against the tree; the other six were settled at
the seal. Four of them have a consequence that outlives the branch, so here they are with what
reversing each would cost.

### 3 and 7: the admin console refuses to start when either removed variable is still set

The client id refuses only when its trimmed value differs from `admin-console-client`; the issuer
refuses whenever it is present at all. The asymmetry is deliberate: every deployment the setup wizard
has ever produced sets the client id to that literal and must keep starting, while nothing in the
tree has ever set the issuer, so any value there is a hand-written line.

**What this costs:** a deployment carrying a stale `GOIABADA_ADMINCONSOLE_ISSUER`, or a client id
naming a client somebody created themselves, stops at startup instead of starting and failing later.
That is the point, but it is a hard failure on upgrade and it needs the release note in the PR body.

**Reversing it:** ignore both variables and carry only a release note. The operator then reconciles an
unfamiliar client id from a 500, or an issuer lockout from a redirect loop, which is the state this
change exists to end.

### 6: the 30 second replica window after an issuer change is accepted

`MiddlewareSettingsCache` has a 30 second TTL and the settings handler invalidates only the cache of
the instance that served the request. A multi-replica deployment therefore bounces sessions to `/` for
up to 30 seconds after an issuer change, in place of today's permanent lockout. A single instance
never sees it.

**Reversing it:** force a settings refetch when the issuer comparison fails and re-check. That closes
the window and puts an auth server call on a path any holder of a stale issuer token can drive, in the
admin console's most frequently run middleware. Rejected for that reason, not for effort.

### 5: an auth server that reports no issuer fails the request

`IsIssuerValid` is plain string equality, so an expected issuer of `""` rejects every token the auth
server can mint and reproduces the lockout silently. The middleware answers 500 naming the auth server
instead. The only live route to it is an auth server too old to serve the field, since an absent
settings row already answers 500 earlier and the admin API's minimum length stops the stored value
being blanked.

**Reversing it:** fall back to the auth server base URL, which is today's default and so keeps the
smaller copy of the defect for exactly the deployment that changed its issuer.

### 4: `/api/public/settings` now carries `issuer`

The endpoint's field allowlist gains the entry, which is the guard working rather than a repair. The
same string is already served to anonymous callers at `/.well-known/openid-configuration`, which
OIDC Discovery 1.0 section 3 requires, so nothing new is disclosed.

**Reversing it:** leave the boundary at three fields and fetch discovery from the admin console, which
buys a narrower DTO at the price of a second HTTP client and a second cache for one string.

### The assumption the branch runs on

> Nobody runs the admin console against an auth server built from a different commit for longer than a
> deployment takes.

The compose files and the Kubernetes manifest pin both images to one tag, so this holds by
construction there. Two Kubernetes Deployments can roll independently for a minute or two, and during
that minute an admin console on the new image talking to an auth server on the old one gets the 500
from decision 5 rather than a lockout.

leodip:
## Follow-ups and folded-in fixes

Two follow-ups, neither filed: filing is yours. Duplicate searches were re-run just now, over every
state. One fix was folded into a stage rather than deferred, and it is listed here too so the list is
the whole of what the run found beside the work.

### 1. The committed `goiabada-setup` binary, and its own build overwrites it

**This is not a new issue. It belongs on [#187](https://github.com/leodip/goiabada/issues/187),** which
is open and whose title already promises "two committed binaries" but whose body describes only the
first of its three findings, the citations to a gitignored doc. Anyone picking #187 up today would
close the citation finding and have no description of the other two. The duplicate search returns #187
for "goiabada-setup binary committed" and for "tracked binary gitignore build output", and nothing
else in any state.

Confirmed by listing tracked executables over 1 MB, which returns exactly two and so matches #187's
count:

```
src/cmd/goiabada-setup/goiabada-setup                     11 MB
src/cmd/goiabada-setup/build/goiabada-setup-linux-amd64   10 MB
```

Both are build outputs of the same `main` package. The second is what `build-binaries.sh` writes. The
first is what a plain `go build ./...` in that directory writes, and that is the trap: because it is a
`main` package, the module's own build command writes to exactly the path of the tracked file, so

```
cd src/cmd/goiabada-setup && go build ./...
```

silently replaces the committed binary with one built on the developer's machine, staging a
multi-megabyte diff of a build output. Nothing warns. It happened during this run's stage 3, where
verifying the wizard still compiled after a five-line source change produced an 11 MB to 12 MB binary
diff alongside it, caught only by reading `git status` before committing. Since then this branch has
added a `Unit / setup` CI job, so `go test` is now the ordinary way to check that module too, but
`go build ./...` remains the reflex and still fires the trap.

Two things are wrong independently. The binary should not be tracked at all: it is a build output,
releases are built by `build-binaries.sh` so nothing depends on the committed copy being current, and
there is no way to tell from the repository which commit it was built from or on what platform, so
anyone who checks out the tree and runs it is running an artifact of unknown provenance. And if it
must stay tracked, the collision should not: a `main` package whose default build output path is a
tracked file is a trap that fires on the most ordinary command in Go.

**What closing it would take:** `git rm --cached src/cmd/goiabada-setup/goiabada-setup`, a
`.gitignore` entry for it, and a line in that module's `Makefile` or in `build-binaries.sh` making the
output path explicit (`go build -o build/goiabada-setup .`) so the ignore rule and the build agree.
Check whether any documentation or release step reads the committed path before removing it.

**Why it did not fit inside #285:** it is unrelated to either environment variable, it touches release
tooling and `.gitignore` rather than the admin console's configuration surface, and removing a tracked
binary is a history and packaging decision rather than contained work inside a stage.

To file it, as a comment on #187 rather than a new issue:

```bash
gh issue comment 187 --body "$(cat <<'BODY'
### The second finding this issue's title names: two committed binaries

`src/cmd/goiabada-setup/goiabada-setup` (11 MB) and
`src/cmd/goiabada-setup/build/goiabada-setup-linux-amd64` (10 MB) are tracked build outputs of the
same `main` package. Listing tracked executables over 1 MB returns exactly these two.

Because `cmd/goiabada-setup` is a `main` package, `cd src/cmd/goiabada-setup && go build ./...`
writes to exactly the path of the first tracked file, silently replacing it with a binary built on
the developer's machine and staging a multi-megabyte diff. Nothing warns. Found during #285, where
verifying the wizard still compiled after a five-line source change produced an 11 MB to 12 MB
binary diff alongside it, caught only by reading `git status` before committing.

Two independent problems. The binaries should not be tracked: they are build outputs, releases are
built by `build-binaries.sh`, and there is no way to tell from the repository which commit they were
built from or on what platform, so anyone who runs the checked-out copy runs an artifact of unknown
provenance. And if one must stay tracked, the collision should not: a `main` package whose default
build output path is a tracked file is a trap that fires on the most ordinary command in Go.

Closing it: `git rm --cached src/cmd/goiabada-setup/goiabada-setup`, a `.gitignore` entry, and an
explicit output path (`go build -o build/goiabada-setup .`) in that module's `Makefile` or in
`build-binaries.sh` so the ignore rule and the build agree. Check whether any documentation or
release step reads the committed path first.
BODY
)"
```

### 2. Admin console middleware rejections render in English regardless of the request locale

**Title:** Admin console middleware rejections render in English regardless of the request locale
**Labels:** `bug`, `go`. The drafted labels were `bug`, `i18n`, `adminconsole`; neither `i18n` nor
`adminconsole` exists in this tracker, so the command below uses the two that do.

`Server.initMiddleware` in `src/adminconsole/internal/server/server.go` mounts the application branch
in this order: `MiddlewareSettingsCache`, `MiddlewareCookieReset`, then `i18n.MiddlewareLocale(nil)`.
Locale resolution is last, so every middleware that answers the request before it has no localizer to
reach for, and the CSRF pair mounted on the root router above runs earlier still.

Three responses are affected today, all `http.Error` with an English literal: the settings-fetch
failure and the missing-issuer branch in `MiddlewareSettingsCache`, and the `MiddlewareCsrf`
rejection page, whose limitation `initMiddleware` already acknowledges in a comment and accepts for
"an infrequent, transient failure mode". None of these strings exists in either active catalog, so an
administrator whose browser asks for `pt-BR` gets English for all three. Each is a genuine failure the
operator has to read and act on, and the missing-issuer one specifically tells them their auth server
needs upgrading, which is the case where the reader is least likely to be the person who chose the
deployment's language.

This is the round 1 reviewer's finding 2, verified as stated and left deliberately. The missing-issuer
branch is new on this branch, but the property is not: the error one line above it has behaved this
way for as long as it has existed. Fixing only the new string would leave the file inconsistent with
itself and with the CSRF precedent, and moving locale resolution ahead of the settings cache reorders
a chain #266 restructured deliberately, which is a decision rather than work and nothing section 2 of
this agreement asks for. What landed instead is a comment at the site recording the convention and
pointing at `initMiddleware`, so the next reader finds a decision rather than an oversight.

**Duplicate search:** `gh issue list --state all --search` over "locale middleware English error",
"localization admin console http.Error", "locale middleware error English", "localization middleware"
and "http.Error localized" returns nothing, in every state.

```bash
gh issue create --title "Admin console middleware rejections render in English regardless of the request locale" \
  --label bug --label go --body "$(cat <<'BODY'
`Server.initMiddleware` in `src/adminconsole/internal/server/server.go` mounts the application branch
in this order: `MiddlewareSettingsCache`, `MiddlewareCookieReset`, then `i18n.MiddlewareLocale(nil)`.
Locale resolution is therefore last, so every middleware that answers the request before it has no
localizer to reach for, and the CSRF pair mounted on the root router above runs earlier still.

Three responses are affected today, all of them `http.Error` with an English literal:

- `MiddlewareSettingsCache`, "Unable to fetch settings from authserver: ..." when the settings fetch
  fails.
- `MiddlewareSettingsCache`, "The authserver did not report an issuer; it may be running an older
  version" when the auth server is too old to serve the field. Added by #285.
- `MiddlewareCsrf`, whose rejection page carries a comment in `initMiddleware` acknowledging the same
  limitation and accepting it for "an infrequent, transient failure mode".

None of these strings exists in either active catalog, so an administrator whose browser asks for
`pt-BR` gets English for all three. Each is a genuine failure the operator has to read and act on:
the second one specifically tells them their auth server needs upgrading, which is the case where the
reader is least likely to be the person who chose the deployment's language.

There are two shapes for a fix and they are not equivalent. Resolving the locale earlier, by mounting
`i18n.MiddlewareLocale` ahead of the settings cache, makes a localizer available to all three but
reorders a chain that #266 restructured deliberately, so it needs checking that the locale middleware
has no dependency on the settings it would then precede. Alternatively each middleware could resolve a
tentative localizer for itself from `Accept-Language` and `?ui_locales`, which is what
`MiddlewareLocale` already does with a nil auth helper, at the cost of doing it twice per request on
the error paths only.

Found while reviewing #285, whose new missing-issuer branch is the second string above. That change
did not create the property, and fixing only its own string would have left the file inconsistent with
the error one line above it, so the branch recorded the convention in a comment at the site and left
the family here.
BODY
)"
```

### Folded in rather than filed: the bootstrap file's contents became testable

**Commit `fffacdda`** (stage 3). Stage 3's one real behaviour change is that the legacy two-step
bootstrap file stops carrying `GOIABADA_ADMINCONSOLE_OAUTH_CLIENT_ID`. That content was built inline
inside `Seed`, which needs a live database, so its only reachable tier was the data tier, which that
stage does not run, and the change would have shipped as a claim nothing could check. The content
moved into a package-level pure function `bootstrapEnvContent`, called from `Seed` exactly where the
literal was, and is now pinned in the core unit tier: the rendered file names no client id variable
and carries no `admin-console-client` literal under any other name, and the five remaining variables
are each present with their value plus a count assertion so a sixth cannot arrive silently. Two
mutations against it, both caught.

**Why it fit:** it rides on `src/core/data/database_seeder.go`, which stage 3 was already editing;
there was nothing to settle; extracting the function is the whole of the fix; and section 2 of the
agreement stays true as written. It is a deviation from the plan's sketch, which described stage 3 as
a sweep with no new tests, and it is recorded as one in the stage log, in the commit message and in
the pull request body.

leodip:
## The code review

This run used `LEO_REVIEW_MODE=end`, so **no stage was reviewed on its own**. The whole branch was
reviewed twice, after every stage had landed, by `openai-codex/gpt-5.6-sol` at effort `xhigh`. Round 1
started 2026-08-29T22:07:53Z in a fresh session; round 2 at 22:25:38Z in a resumed one. Both rounds
read all three axes with nothing excluded.

| Round | Findings | Blocking | Significant | Minor | Outcome |
|---|---|---|---|---|---|
| 1 | 2 | 0 | 1 | 1 | one fixed in `7a1a6959`, one deferred as a follow-up with a comment left at the site |
| 2 | 1 | 0 | 1 | 0 | fixed in `ae05d6aa` |

There is no round 3, and the reason is the round budget rather than fatigue: rounds past the second
are funded only by a round that changed production code, and round 2's fix moved `unit-tests` while
leaving `production` byte-identical. Nothing was left open at that point. The one finding was fixed,
`unchecked` was empty, and round 1's deferred finding is a drafted follow-up.

---

## Round 1

**What the reviewer did.** Read the agreement, the plan, all four stage logs, the closing log, the
probe and every file in `git diff origin/main...HEAD`. Ran `tree-hash.sh --verify` and
`check-anchors.sh`; the hashes matched and all 27 anchors resolved. Read `where.sh ci` for `53f608d9`
and found 11 jobs green. Ran the branch's new targeted tests. Ran five scratch mutations of its own,
covering issuer mapping, empty-issuer refusal, client-id forwarding, bootstrap output and stale-env
refusal, all caught. Ran the real admin console binary with a stale client id and with a stale issuer,
and both were refused at startup. Fetched OIDC Core 1.0 and RFC 9700 from their publishers. Its own
note: "The current issuer and client-id implementation is fail closed and the checked runtime behavior
is sound."

### Finding 1, significant, quality: the setup wizard's generators have no executed test

**The claim.** `generateAdminConsoleService`, `generateEnvFile` and `generateKubernetesManifests` are
pure functions whose output an operator deploys from, and stage 3 changed all three. Reintroducing
`GOIABADA_ADMINCONSOLE_OAUTH_CLIENT_ID` in each still left the setup module green, so nothing in CI
pinned the operator-visible requirement. The reviewer ran that reintroduction in a scratch copy and
wrote a scratch table test that failed in each arm.

**Verified, and worse than the claim as written.** `src/cmd/goiabada-setup` is a module of its own with
no test file at all, so `go test ./...` answered `[no test files]` and exited 0, which is
indistinguishable from a pass. And CI never ran it either way: `check.yml`'s `Lint` and
`Vulnerabilities` jobs loop over `core authserver adminconsole cmd/goiabada-setup`, but the unit jobs
are per module and there was no setup one, so a test file added there would not have been executed.
Stage 3's own log entry recorded "`go build` clean in the setup wizard module, which has no test tier"
and stopped there, which is exactly the gap.

**Fixed in `7a1a6959`, in two parts, because either alone closes nothing.**

- `src/cmd/goiabada-setup/main_test.go`, the module's first test file. A table over four generators,
  the three the stage changed plus `generateDockerCompose`, which composes the admin console service
  into the whole file. Each must emit neither removed variable **and** must still emit a named
  surviving line, so a generator that returned nothing cannot pass the absence check vacuously. A
  second test holds the two committed wizard outputs to the same rule; they are hand-kept copies of
  generator output and nothing regenerates them, so they drift on their own schedule. Standard library
  only: the module has four dependencies and testify is not among them.
- A `Unit / setup` job in `.github/workflows/check.yml` running `go test ./... -count=1` in that
  module. No `setup-test-container`: the module has no templates and no CSS, so the Tailwind CLI and
  the OS packages `run-tests.sh` needs would cost more than the run.

**Five mutations, all caught**, run by hand rather than through `where.sh mutate`, which dispatches on
`run-tests.sh` tiers and the setup module is not one. Each was backed up to a temp directory and
restored from a shell `trap`, with `git status` confirming afterwards that only the new untracked test
file remained.

| Mutation | Result |
|---|---|
| `generateAdminConsoleService` re-emits `GOIABADA_ADMINCONSOLE_OAUTH_CLIENT_ID` | CAUGHT |
| `generateEnvFile` re-emits the `export` line | CAUGHT |
| `generateKubernetesManifests` re-emits the ConfigMap entry | CAUGHT |
| committed `docker-compose.yml` drifts the line back | CAUGHT |
| `generateAdminConsoleService` re-emits `GOIABADA_ADMINCONSOLE_ISSUER` | CAUGHT |

The family the reviewer named, "pure setup-output generators can reintroduce removed deployment
configuration", is closed for both variables and for the committed artifacts, not just for the one
variable and the three functions the finding listed.

### Finding 2, minor, conformance: the missing-issuer message is English only

**The claim.** The new `publicSettings.Issuer == ""` branch writes a browser-visible English sentence
that appears in neither catalog, and it runs before locale resolution, so a `pt-BR` request gets
English. The reviewer proved it with a scratch test against the production middleware order.

**Verified as stated, and it is a pre-existing property of the position rather than a new defect.**
`Server.initMiddleware` mounts `MiddlewareSettingsCache` first on the application branch and
`i18n.MiddlewareLocale(nil)` last, so no localizer exists when this middleware answers. The
settings-fetch failure on the line above the new branch has the identical property and predates this
branch, and the CSRF pair two positions earlier carries a comment saying so in as many words: "CSRF
runs before the locale middleware below, so no localizer exists yet when a request is rejected... This
is acceptable for an infrequent, transient failure mode."

**Deferred, and the reason is that fixing only the new string is the wrong shape.** The options were to
localize one of two adjacent errors in one function, which leaves the file inconsistent with itself and
with the CSRF precedent, or to move locale resolution ahead of the settings cache, which reorders the
chain #266 deliberately restructured and is nothing section 2 of this agreement asks for. What landed
in `7a1a6959` is a comment at the site recording that the English is by the same convention as its
neighbours and pointing at `initMiddleware`, so the next reader finds a decision rather than an
oversight. The family is a drafted follow-up, in the follow-ups comment on this pull request.

- *"No real browser sign-in after changing the issuer was executed; the repository has no admin console
  integration harness."* **Dismissed.** Section 5 of the agreement rejected exactly this test candidate
  at the seal, for this reason, and named the pieces that close the loop instead: the public settings
  seam and the settings-cache seam, plus the existing `TestJwtSessionHandler_ValidSession` (accept arm,
  matching `iss`) and `TestJwtSessionHandler_InvalidIssuer` (reject arm). Building the harness is a
  change of its own.
- *"No live mixed-version or multi-replica deployment was exercised; the 30 second replica window was
  reviewed statically."* **Covered, by decision 6**, which accepted that window with its reasoning and
  its rejected alternative, and which goes to you in the decisions comment. Nothing was left to check:
  the window is the accepted behaviour, not an unknown.
- *"Unrelated replay, OTP, rate limiting, session lifetime and exact-comparison surfaces were not
  re-audited because this branch has no call path into them."* **Dismissed.** Correct, and out of scope
  by section 2.

### Tiers re-run after the round 1 fixes

| Tier | Command | Result |
|---|---|---|
| unit, three modules | `where.sh test --type modules` | 4642 pass, 0 fail, 57 packages, 1304 top-level tests |
| unit, setup module | `go test ./... -count=1` in `src/cmd/goiabada-setup` | 8 subtests pass, 0 fail, the module's first |
| anchors | `check-anchors.sh` | all 27 rows resolve |

Data and integration were not applicable to the fix: no handler, route, schema or database access
moved. Tree after the fixes, `tree-hash.sh --split`: production `99515c22fad2`, unit-tests
`d4de63f49ec3`, data-tests `8c5a637bd44d`, integration-tests `7b69ee4ead08`, docs `8bc2850d207e`.
`production` moved from the value round 1 was reviewed at, by the one comment in
`middleware_settings_cache.go`, and every mutation above was verified gone by `git status` before the
commit.

---

## Round 2

Request `285-20260829T223500Z`, the whole branch against `origin/main`, six commits, all three axes.
Verdict: one finding, significant. `unchecked` was empty, so nothing was left unreached and there was
no work list to answer.

**What the reviewer did.** Re-read the request, the agreement, the plan, all four stage logs, the
closing log, the probe and the whole diff. Ran `git diff --check`, `tree-hash.sh --verify` and
`check-anchors.sh`. Polled `where.sh ci` until all twelve jobs were green on `7a1a6959`. Recounted
every removal the agreement's fanout table names. Ran ten of its own scratch-copy mutations across the
setup generators. Its note records that the round 1 coverage work is otherwise sound and that
production behaviour is fail closed throughout.

### Finding 1, significant, quality: the committed-artifact test accepts a banner-only file

**The claim.** `TestCommittedOutputsOmitTheRemovedAdminConsoleVars` made two assertions per artifact:
that it sets neither removed variable, and that it still carries `# Generated by goiabada-setup`.
Replacing either `docker-compose.yml` or `goiabada-k8s.yaml` with that banner line alone satisfies
both, so the absence checks could pass on output that is no longer deployable.

**Confirmed, by reading and by mutation.** The banner was the only positive assertion, and it is the
first line of each file, so it survives any truncation. The reviewer ran the truncation on each
artifact separately and recorded the production test exiting 0 both times; it also wrote a scratch-only
per-file marker test and showed that one fails on each. Reproduced here before fixing: with
`docker-compose.yml` replaced by its banner, the test passed.

The finding is precise about what was *not* vacuous, and that half checks out too. All four
dynamic-generator cases already required a named surviving line, empty admin console output was
rejected, removing the service's surviving client-secret marker was rejected, and a deleted or renamed
artifact already failed at `os.ReadFile`. The gap was exactly the two committed artifacts and nothing
else in the file.

**Fixed in `ae05d6aa`.** The committed-artifact test becomes a table like the generated-output test
beside it, each case naming a line the artifact must still carry, chosen from the admin console block
the removed line was deleted from: `- GOIABADA_ADMINCONSOLE_LISTEN_PORT_HTTP=9091` in the compose
service, and `GOIABADA_ADMINCONSOLE_BASEURL: "https://admin.goiabada.dev"` in the ConfigMap. The banner
check stays, since it says the file is still the wizard's output rather than something hand-written.

**Family: vacuous absence tests for committed generated artifacts.** It has exactly two members, the
two files the test enumerates, and both are now covered. Four mutations, all caught:

| Mutation | Result |
|---|---|
| `docker-compose.yml` replaced by `# Generated by goiabada-setup` alone | CAUGHT, `does not contain "- GOIABADA_ADMINCONSOLE_LISTEN_PORT_HTTP=9091"` |
| `goiabada-k8s.yaml` replaced by `# Generated by goiabada-setup` alone | CAUGHT, `does not contain "GOIABADA_ADMINCONSOLE_BASEURL: ..."` |
| `GOIABADA_ADMINCONSOLE_OAUTH_CLIENT_ID=admin-console-client` appended to `docker-compose.yml` | CAUGHT |
| `GOIABADA_ADMINCONSOLE_ISSUER` appended to `goiabada-k8s.yaml` | CAUGHT |

The first two are the reviewer's own mutations, now failing where they used to pass. The last two exist
because the restructure rewrote the loop that had been catching them, and a fix that silently dropped
an existing check would be worse than the gap it closed. Each ran under a backup-and-trap wrapper that
restores the file and compares it by hash, because `where.sh mutate`'s tiers do not cover the setup
module, which has its own CI job instead.

### Tiers, and the tree they ran against

Unit on the three modules: 4642 pass, 0 fail, 57 packages, about 90 seconds. The setup module: 8 pass,
0 fail. `check-anchors.sh`: all 27 rows resolve. Data and integration are not applicable to a commit
that moves one test file and no handler, route, schema or database access; the four-engine tiers are
the check suite's on the pushed commit either way.

`tree-hash.sh --split` after the fix: production `99515c22fad2`, **unchanged** from the value the round
was reviewed at, unit-tests `d4de63f49ec3` to `201559f288ff`, data-tests `8c5a637bd44d`,
integration-tests `7b69ee4ead08`, docs `8bc2850d207e`. Checked rather than assumed: no production file
moved during the mutation work.

PR #297, opened 2026-08-30
fix(authserver): order code issuance against session termination

leodip:
Closes #139.

A code inserted between a session termination's sweep and its commit escapes both revocation
paths, and the code that comes out of it redeems into an offline refresh token that outlives the
session it was issued through.

`TerminateUserSessionTx` runs three writes in one transaction: sweep the session's codes, sweep its
refresh tokens, delete the session row. `/auth/issue` reads the session for liveness, inserts a
code, then runs a compensating `UPDATE codes ... WHERE NOT EXISTS (SELECT 1 FROM user_sessions ...)`,
each statement on its own connection with no transaction. A code that lands after the sweep and
before the commit is missed by the sweep, because the row did not exist yet, and missed by the
compensating statement, because that statement still sees the uncommitted-deleted session.

The window is open on **two** of the four engines, not the one the issue claims. Measured directly
against the data layer with two independent handles, in the order `TerminateUserSessionTx` really
uses:

| Engine | Window |
|---|---|
| PostgreSQL | open |
| SQL Server | open |
| MySQL | closed, the insert blocks on the sweep's next-key lock |
| SQLite | closed, the insert blocks on the writer lock |

SQL Server's incidental protection comes from the compensating read blocking on the exclusive lock
the `DELETE` holds, and in the real order that `DELETE` has not run yet.

## The rule this establishes

**Issuance and termination both write to the same `user_sessions` row before touching anything
else**, so on every engine one of them waits for the other, and the one that waited reads its answer
after the wait rather than from a snapshot taken before it. That leaves exactly two outcomes and no
third, because that row is the only object both sides touch and neither takes any other lock first:

- **The termination waits.** Its code sweep then runs after the insert committed, so it marks the
  new code and redemption answers `invalid_grant`.
- **The issuance waits.** Its acquisition then matches zero rows, so it refuses and no code is
  written at all. The browser is sent back to level 1, and a `prompt=none` ceremony is answered
  `login_required`, which is the same answer the liveness check a few statements earlier already
  gives for the same condition.

`user_sessions.session_identifier` is UNIQUE on all four engines, so one unconditional
`UPDATE user_sessions SET updated_at = ? WHERE session_identifier = ?` is a single-row lock
everywhere. No schema change, no migration, no new column. It is the primitive `AcquireClientRow`
already uses for clients (#245).

The termination adds no acquisition of its own: `DeleteUserSession` moves to the front of
`TerminateUserSessionTx` and is itself the statement that takes the row. Measured on all four
[...]

Issue #106, opened 2026-07-25
Security: password reset, password change and account disable do not invalidate live sessions or refresh tokens


leodip:
## Summary

Three state changes that are meant to cut off access do not terminate anything that is already live:

1. **Password reset** (forgot-password flow)
2. **Self-service password change**
3. **Admin disabling a user account**

All three write the `users` row and stop. Existing `UserSession` rows, refresh tokens and unexpired access tokens survive untouched.

Severity: **Medium** (OWASP A07:2025 Authentication Failures; OWASP ASVS 5.0 V3 Session Management)

## Affected code

**Password reset** - `src/authserver/internal/handlers/handler_reset_password.go:252-264`

```go
user.PasswordHash = passwordHash
user.ForgotPasswordCodeEncrypted = nil
user.ForgotPasswordCodeIssuedAt = sql.NullTime{Valid: false}
err = database.UpdateUser(nil, user)
// no session deletion, no refresh token revocation
```

**Self-service password change** - `src/authserver/internal/handlers/apihandlers/handler_api_account_password.go:85-89`

```go
user.PasswordHash = passwordHash
user.ForgotPasswordCodeEncrypted = nil
user.ForgotPasswordCodeIssuedAt = sql.NullTime{Valid: false}
if err := database.UpdateUser(nil, user); err != nil { ... }
// same omission
```

**Admin disable** - `src/authserver/internal/handlers/apihandlers/handler_api_users_crud.go:509-512` (`HandleAPIUserEnabledPut`)

```go
user.Enabled = req.Enabled
err = database.UpdateUser(nil, user)
// same omission
```

## Why disable is not fully mitigated by the existing checks

`user.Enabled` **is** checked when new tokens are minted: `token_validator.go:173` (authorization_code), `:398` and `:419` (refresh_token), `:634` (ROPC), and `handler_auth_completed.go:163`. So a disabled user cannot obtain *new* tokens.

But the per-request gate for the API surface, `RequireValidSession` in `src/authserver/internal/middleware/api_auth.go:123-181`, only verifies that the `sid` claim resolves to a `UserSession` that is within its idle and max-lifetime bounds. It never loads the user or checks `Enabled`. A disabled user therefore keeps working access to `/api/v1/account/*` (and `/api/v1/admin/*` if they held those scopes) for the remainder of the access token lifetime, which is `Settings.TokenExpirationInSeconds`.

## Impact

[...]

PR #302, opened 2026-09-05
refactor(core): drop the global lock order, retry deadlock victims instead

### LINKED ISSUE #139: Security: a code insert can land between a termination's UPDATE and its COMMIT, escaping both marks

### LINKED ISSUE #301: Transactions: keep atomicity and the #139 session-row ordering, drop the global lock order
## Summary

#139 (PR #297) closed a real integrity gap: an authorization code could be inserted between a session termination's sweep and its commit, and redeem into an offline refresh token that outlived the session. The fix that closes it is two statements: the termination deletes the session row first, and issuance writes that same row before inserting its code, inside one transaction.

Most of what the branch carries beyond that is not about integrity. It is a global lock order (users, then clients shared, then user_sessions, then grants) applied at nine transaction sites, with four new acquisition methods on the `Database` interface, a client-deletion discovery barrier, and about 6000 lines of tests that pin the order per site. That layer exists to prevent deadlocks, and a deadlock is not an integrity failure: the engine detects it, rolls one side back atomically, and that one request answers 500. The cycles it prevents need an admin action or a credential change on one account to land inside the few milliseconds of that same account's code issuance.

This issue asks to keep the integrity fix and the atomicity, remove the lock-order discipline, and document the residual in one paragraph. The end state is the code as it stood on `main` before #139, plus the two-statement ordering and the harmless parts of #297.

## The core problem, separated into three

**1. Atomicity.** Several changes are multi-statement and must land together or not at all: terminating a session (delete the row, revoke its codes, revoke its refresh tokens), a credential change (bump the generation, end the sessions, revoke the tokens), creating a session (the row and its association), deleting a user or a client (the children SQL Server cannot cascade, then the parent). A half-applied termination leaves live tokens. This is what transactions are for, it is cheap, and it matters in ordinary use. Keep.

**2. The #139 read-then-write race.** Issuance observed the session on one connection and inserted on another, so a termination could commit in between, and a code landing after the termination's sweep but before its commit escaped both the sweep and the old compensating read. The fix needs both parties to write the same `user_sessions` row before anything else. That is `DeleteUserSession` first in `TerminateUserSessionTx` and `AcquireUserSessionRow` then `CreateCode` in one transaction in `HandleIssueGet`. Measured on all four engines in both orderings, and pinned by `TestIssuanceOrdering_AgainstTermination` and its RCSI twin. Keep.

**3. Deadlocks between the atomic transactions.** Once issuance holds a row across statements, it can form a cycle with another multi-statement transaction on the same user or client. The branch answered with a global order. Each round of that work found a lock no statement names (a foreign-key parent lock, SQL Server re-checking a foreign key on any `UPDATE` that lists the column, a cascade) and had to extend the order: decision 10 hoisted four sites, decision 11 put `users` on top, stage 7 put `clients` shared below it and gave `DeleteClient` a session-row loop. The review brief of stages 7 and 8 names the structural weakness itself: the order is documented, not enforced; nine sites have a test each and site ten deadlocks in production, on some engines only, under a race.

## Why deadlocks are not the thing to spend this on

A deadlock victim is rolled back. Nothing is half done. The engine's own detection (PostgreSQL `40P01`, MySQL `1213`, SQL Server `1205`) is the resolution, and the loser's request answers 500. On SQLite there is one connection for the whole process (`SetMaxOpenConns(1)` in `sqlitedb`), so no two transactions ever overlap and none of this applies.

How often the cycles happen without the order, in typical use, with issuance holding the session row across its insert:

| Concurrent with a code issuance for account A | Cycle? |
|---|---|
| Another login for A, a session bump, a refresh, a redemption, either background reaper | No. None holds a row issuance wants while wanting one issuance holds. |
| A termination of that session | No. Both start on the same row by design; one waits. |
| A password change, reset or admin password set for A | Yes, on MySQL and SQL Server only, if it lands inside issuance's transaction. PostgreSQL does not conflict a foreign key's `FOR KEY SHARE` with the sweep's `FOR NO KEY UPDATE`. |
| Deleting user A | Yes, on PostgreSQL, MySQL and SQL Server, same window. |
| Deleting the client A is signing in to | No cycle. Issuance waits on the client row and then fails its foreign key; a 500 today and on `main` alike. |

So the residual is an administrator acting on one account, or that account changing its own password, within one transaction's lifetime of the same account completing an OAuth login. The loser is told 500 and retries. That is the edge case, and the lock order is the price of never seeing it.

Between the admin transactions themselves (a user deletion racing a credential sweep on the same user, a client deletion racing a termination whose session holds that client's tokens) latent cycles existed on `main` before #139 and exist after it. They are admin-versus-admin on the same object at the same millisecond.

### Keep

- Transactions for atomicity, at every site that has one today.
- `TerminateUserSessionTx` deleting the session row first, and `TestTerminateUserSessionTx_SweepsAfterTheSessionRowIsDeleted`.
- In `HandleIssueGet`, one transaction: `AcquireUserSessionRow` then `CreateAuthCode(tx, ...)` then commit, with the existing explicit rollback before `refuseIssuanceUnusableSession`, because that path reaches the database on a nil transaction and on SQLite would otherwise wait on its own connection.
- `AcquireUserSessionRow` on the `Database` interface, with its four wrappers.
- `AcquireClientRow`, which predates this work (#245) and serves the client update paths.
- `CreateAuthCode` taking a leading `tx`, and `ErrIssuingClientGone` in place of the nil dereference.
- `refuseIssuanceUnusableSession` and the three refusal shapes.
- The session-first order inside `RevokeUserAuthState` and `commondb.disableAndRevoke`, and the sort by id in their loops. Consistent among themselves, harmless, and already tested.
- The `dont-update` tags on `RefreshToken.{CodeId,UserId,ClientId}`, `UserSession.UserId` and `UserSessionClient.{UserSessionId,ClientId}`, with `dont_update_keys_test.go`. Nothing reassigns those keys, and the tags remove an invisible SQL Server parent lock at zero cost.
- Migration 000044 and the index it names on all four engines, with its test.
- The `READ_COMMITTED_SNAPSHOT` paragraph in `production-deployment/database.mdx`, which stays true: the kept ordering runs under RCSI in `TestIssuanceOrdering_AgainstTermination_RCSI`.
- The data-tier helpers the kept tests use: `secondDatabase` in `second_handle_test.go`, `goBlocked` and the blocked-party harness in `blocked_party_test.go` and `blocked_party_harness_test.go`, and the RCSI fixture in `rcsi_test.go`.

### Remove

- `AcquireUserRow`: the interface declaration, four wrappers, the mock, and its calls in `HandleIssueGet`, `user.StartNewUserSession` and `commondb.DeleteUser`.
- `AcquireClientRowShared` and `AcquireClientRowSharedWith`: the interface declaration, four dialect wrappers, the mock, and its calls in `HandleIssueGet`, `user.StartNewUserSession` and `user.BumpUserSession`.
- `AcquireUserSessionRowById` and `GetUserSessionClientsByClientId`: declarations, wrappers, mock, and their use in `commondb.DeleteClient`, together with `sortedDistinctSessionIds`. `DeleteClient` returns to its `main` shape: delete the client's refresh tokens, delete the client. `AcquireClientRow` at its top may stay or go; it orders nothing that matters once the session loop is gone.
- The leading `AcquireUserSessionRow` in `revokeOnAuthCodeReuse`. The function returns to its `main` shape: read the tokens, revoke, delete the session if anything was revoked.
- `commondb.DeleteUser` returns to its `main` shape: delete the refresh tokens, delete the user, cascade removes the rest. The explicit session loop may stay if it reads better; it orders nothing that matters.
- The tests that pin the removed order: `tests/data/lock_order_test.go`, `lock_order_bump_test.go`, `lock_order_user_row_test.go`, `lock_order_client_delete_test.go`; `handlers/lock_order_test.go`; `commondb/client_delete_order_test.go`; `user/usersession_manager_lock_order_test.go`; and the acquire expectations inside `handler_auth_issue_test.go`, `usersession_manager_start_test.go`, `usersession_manager_test.go`, `revocation_test.go`, `handler_token_test.go` and the affected `tests/data` files.
- The lock-order paragraphs in the doc comments of every site above, and the rule as stated in the `Database` interface comments. What replaces them is the residual paragraph below.

### Accept, and write down

One paragraph, in `CLAUDE.md` under Important Patterns and as the doc comment on the issuance transaction, saying:

> Transactions exist for atomicity. Issuance holds the session row across its code insert so that a termination can never slip between its sweep and its commit (#139). No global lock order is imposed. A credential change, a user deletion or a client deletion landing inside the few milliseconds of that same account's code issuance can deadlock on MySQL, PostgreSQL or SQL Server; the engine rolls one side back and that request answers 500, with nothing half applied. SQLite has one connection and cannot deadlock.

### Deferred, and only on evidence

A retry on the engine's deadlock error inside `commondb.inTransaction` and the three `*Tx` helpers in `revocation.go` is about thirty lines and four error codes, and would cover every cycle, present and future, without anyone knowing an order. Every transaction here is safe to rerun after a rollback: its effects are inside the transaction, and the audit log sits after the commit. It is deliberately not part of this issue. Add it if the 500s described above are actually observed on the demo or reported, not before.

## Related issues

Each was told on its own thread (2026-09-05) that this work is next in priority and what it changes for them.

| Issue | State | Boundary with this work |
|---|---|---|
| #139 | closed by PR #297 | The parent. Its fix (delete-first termination, issuance writes the session row before inserting) is kept in full and its two ordering tests must pass unchanged. What goes is decisions 10 and 11 and stage 7 of its agreement. |
| #131 | open | Wants rotation coordinated with user-scoped generation changes inside a transaction. #139 told it to reuse the ordering primitive; under this issue it inherits atomicity and the documented residual, not a `users`-first rule. `AcquireUserRow` and `AcquireClientRowShared` will not exist for it to call. |
| #132 | open | Wants family containment atomic with rotation. Same boundary as #131: an insert into `refresh_tokens` reaches `users`, `clients` and `codes` through foreign keys, which under #297's rule would have obliged it to take `users` first, and under this issue obliges nothing. |
| #248 | open | Part 5 (a client deleted mid-ceremony dereferences nil) was fixed on #297 by `ErrIssuingClientGone`, which this issue keeps. Removing the shared client acquisition makes losing that race narrow again rather than reliable, as on `main`; the foreign-key failure on the insert itself stays a 500 as on `main`. Nothing for part 5 to change. |

## Alternatives considered

**Keep the lock order as is.** It is correct and tested on four engines. Its cost is that every future transaction touching sessions or grants must know a rule that lives in comments, and a mistake fails only on some engines and only under a race. #131 and #132 want refresh rotation inside a transaction; an insert into `refresh_tokens` reaches `users`, `clients` and `codes` through foreign keys, so under the current rule they would have to take `users` first too, and the rule would grow by another site each. Under this proposal they inherit atomicity and the accepted residual, and nothing else.

**Replace the issuance transaction with one conditional locking insert** (`INSERT INTO codes ... SELECT ... FROM user_sessions WHERE session_identifier = ? FOR KEY SHARE`, with `FOR SHARE` on MySQL and `WITH (UPDLOCK, ROWLOCK)` on SQL Server). It closes the window with issuance holding nothing across statements. Rejected here: it costs four hand-written dialects of a fifteen-column insert for the same integrity result, and within its one statement it still takes the session row before the foreign key's parent, so it still cycles with a same-user credential sweep on MySQL and SQL Server. Recorded so nobody re-derives it.

**Retry only, no ordering at all.** Deferred as above rather than rejected. It is the one addition that replaces the whole discipline, and it should be earned by an observed failure.

## Acceptance criteria

1. `TestIssuanceOrdering_AgainstTermination` and its RCSI twin pass unchanged on all four engines. That is the #139 property, and this issue must not weaken it.
2. `HandleIssueGet`'s authorization code branch issues exactly `BeginTransaction`, `AcquireUserSessionRow`, `CreateAuthCode`, `CommitTransaction` against the database, pinned in `handler_auth_issue_test.go` by call order.
3. `AcquireUserRow`, `AcquireClientRowShared`, `AcquireUserSessionRowById` and `GetUserSessionClientsByClientId` no longer exist on the `Database` interface, in any engine package, or in the mock.
4. `DeleteClient`, `DeleteUser`, `revokeOnAuthCodeReuse`, `StartNewUserSession` and `BumpUserSession` carry no acquisition and no lock-order comment.
5. The residual paragraph is present in `CLAUDE.md` and on the issuance transaction, and no doc comment anywhere states a lock order.
6. Unit, data (four engines) and integration tiers green.

## Not in scope

- The deadlock retry (deferred, above).
- #131 and #132.
- The SQLite pool wait with no deadline on a nil-transaction call while a transaction is open. It is unchanged by this issue; the kept issuance transaction already rolls back before any such call.

leodip:
Closes #301.

## What this changes

PR #297 closed #139 by making one property true: a termination deletes its session row first, and
issuance takes that row before inserting its authorization code, in one transaction, so a code can
never slip between a termination's sweep and its commit. Around those two statements it also built a
**global lock order**, users then clients-shared then user_sessions then grants, carried by four new
acquisition methods on the `Database` interface and applied at nine call sites.

The order is what this change removes. It bought the absence of engine-detected deadlocks, and it
cost every future author a rule they have to know before they can open a transaction anywhere in the
tree. In its place:

- **`Database.RunInTransaction(fn)`**, one implementation, which reruns `fn` when the engine aborts
  it as a deadlock victim: bounded at three attempts with a short backoff, then the error surfaces.
  Every transaction owner in the codebase goes through it, and a lint test fails the build on a bare
  `BeginTransaction` in production code outside the helper.
- **A deadlock classifier per engine**, driven off the driver's own error: PostgreSQL SQLSTATE
  `40P01`, MySQL `1213`, SQL Server `1205`. A lock-wait timeout is not a deadlock and is not
  retried. SQLite runs the whole process on one connection and cannot deadlock.
- **#139's property kept in full**, and kept as an integrity rule rather than a deadlock rule,
  pinned by the two ordering tests it already had.

Retrying is correct for every transaction here because each one's effects are inside it and its
audit event is written after its commit, so a rerun repeats no side effect a caller can observe.

## The rule this establishes

No lock order is imposed. Open your transaction through `RunInTransaction` and, if you are writing a
session and the grants hanging off it, write the session row first so you and a termination of the
same session serialize on that row rather than reaching the retry. Concurrent transactions on the
same account can still deadlock on MySQL, PostgreSQL or SQL Server; the loser is rolled back with
nothing half applied, and rerun. Do not add ordering to prevent a deadlock. Add a test that forces
it and shows the retry resolves it.

### Stage 1: `RunInTransaction` and the deadlock classifier
Landed in `98fe6646`. `Database.RunInTransaction(fn)` opens a transaction, runs `fn`, commits on
nil and rolls back otherwise, and reruns the whole body when the engine aborts it as a deadlock
victim: three attempts, 25ms then 100ms between them, one warning per rerun, the last deadlock
wrapped at exhaustion. Anything else `fn` returns, and any commit failure that is not a declared
deadlock, comes back after one attempt, because a commit whose outcome the client cannot know is
never replayed. Each dialect installs its classifier on the shared handle: PostgreSQL `40P01`,
MySQL `1213`, SQL Server `1205` in both shapes of `mssql.Error`, SQLite always false. Lock-wait
timeouts are deliberately not deadlocks. `inTransaction` goes through it when handed nil, so its
three callers already retry; handed a transaction it is unchanged, so only an owner ever reruns.
The one caller that retrying broke, the collision e-mail backfill, now builds its revocation audit
per attempt and publishes only the attempt that committed.

Tiers run locally: unit on three modules (57 packages, 1345 pass), data on four engines (1664
pass, 64 skip), including a forced two-row deadlock between two handles that the retry resolved on
MySQL, PostgreSQL and SQL Server. Seven mutations, each caught, one of which showed two audit cases
insensitive to the defect they pin; they were reshaped before the commit. Integration not
applicable, nothing reachable through an endpoint changed. The four-engine check suite for
`98fe6646` is the CI run's to report.

### Stage 2: the `core` module's five transaction owners go through the helper
Landed in `78d0ff13`. `StartNewUserSession`, `BumpUserSession`, `UserCreator.CreateUser`,
`SigningKeyRotator.Rotate` and `ReencryptDataToNewKey` each replace their `BeginTransaction` /
deferred rollback / commit triple with a closure passed to `RunInTransaction`. Each body was read
for rerun safety first: the ids the inserts assign onto a model are reassigned by the next attempt
before anything reads them, the session manager's association loops range by value so an attempt
never mutates the slice the next attempt decides from, the rotator generates its key before the
helper opens and its two sentinels are not deadlocks, and re-encryption is entirely inside its
closure. No owner here writes an audit event. The two #139 acquisitions inside the session manager
stay, inside their closures, until stage 4 removes them; `core` now has no bare `BeginTransaction`
outside the helper and the four dialect wrappers.

Four `core` mock-based test files were re-stubbed from `On("BeginTransaction")` to
`EXPECT().RunInTransaction(...)` through a package-level stub that records what the body handed the
helper, so "nothing committed" is asserted as "the body returned its error". `user_creator_test.go`
is new, since `CreateUser` had no unit test; one of its five cases runs the body twice and shows
each attempt's permission row names that attempt's user id. Folded in: the admin key-rotation
handler test drives the real rotator over a mock database and broke with it, so it is re-stubbed
here through a package-level stub the stage 3 conversions reuse.

Tiers run locally: unit on three modules (57 packages, 4772 pass), data on four engines (2304
pass, 65 skip), integration on sqlite (1333 pass), the last because the session manager is on every
login path and its unit tests are mock-backed. Five mutations, one per converted owner, all caught.
The four-engine check suite for `78d0ff13` is the CI run's to report.

### Stage 3: the ten `authserver` owners, and the lint that keeps them there
Landed in `18d1bfc5`. Issuance, the auth-code replay, `RevokeUserAuthStateTx`,
`TerminateUserSessionTx`, `RevokeClientGrantsTx`, OTP enable and disable, the admin enabled-put's
disabling closure, the client update that does not own the authentication mode, and the web-origins
PUT each replace their `BeginTransaction` / deferred rollback / commit triple with a closure passed
to `RunInTransaction`. With these, every transaction owner in the repository is on the helper, and
`TestNoBareBeginTransaction` in `core/data` keeps it so: it parses every non-test Go file under
`src/` and reports any call expression selecting `BeginTransaction` outside the helper's body and
the four dialect wrappers. It parses rather than greps because the `Database` interface and the
generated mock declare the method and must keep it; a synthetic fixture tree proves the checker
tells a call from a declaration, a comment and a string.

Two owners return from inside the transaction on a condition that is not a fault and now do so
through a sentinel. Issuance returns `errIssuanceRefused` for a gone session and for a client whose
registration went away, and runs the refusal only after the helper returned: every path out of the
refusal touches the database on a nil transaction, and on SQLite the process shares one connection,
so a refusal issued while the transaction was open would wait on itself. The enabled-put returns
`errUserAlreadyDisabled` when its compare-and-set found no transition, so the empty transaction
rolls back rather than commits. The web-origins PUT keeps the driver error in the chain it hands the
helper, through a typed `webOriginsWriteFailure` whose `Unwrap` is the database error, so a real
deadlock is still recognised and rerun; every log line and error response moved below the helper,
because an attempt that has already written a 500 cannot be rerun into a 200. Each body was read for
rerun safety first: the revocation helpers' write callbacks are compare-and-sets or idempotent
column writes, every read is inside its closure, and every audit event is written after the helper
returns.

Thirteen `authserver` mock-based test files were re-stubbed to `EXPECT().RunInTransaction(...)`
through package-level stubs that note begin, commit and rollback, so the ordering tests keep their
edges. Four new web-origins cases: a body rerun after a deadlock-shaped failure answers once (one
200, one audit event), an exhausted retry is one 500, a failed load keeps its own message, and a
failed write names its origin.

Tiers run locally: unit on three modules (57 packages, 4777 pass), data on four engines (2304 pass,
65 skip), integration on sqlite (1333 pass), the last because the converted owners sit on the login,
token, account and admin API paths. Five mutations, one per seam the stage made observable, all
caught. The four-engine check suite for `18d1bfc5` is the CI run's to report.

### Stage 4: the four acquisition methods and the global lock order come out
Landed in `0002a70b`. `Database` loses `AcquireUserRow`, `AcquireClientRowShared`,
`AcquireUserSessionRowById` and `GetUserSessionClientsByClientId`, in the interface, `commondb`
(`AcquireClientRowSharedWith` and `sortedDistinctSessionIds` with them), all four dialects and the
regenerated mock. The order those methods imposed, users then clients shared then user_sessions
then grants, existed only to prevent engine-detected deadlocks, and with `RunInTransaction`
rerunning the victim it was cost without a property behind it: an extra UPDATE on every code issued
and every session started, a shared lock on every bump, and a rule every future transaction had to
remember.

The call sites return to their shapes before the order. Issuance is `AcquireUserSessionRow` then
`CreateAuthCode`, in one transaction, which is the one ordering the repository keeps on purpose:
the termination deletes the session row first and issuance takes it before inserting, so a code can
never slip between a termination's sweep and its commit, and the two tests that measure that pair on
a real catalog keep every assertion they had. `StartNewUserSession` and `BumpUserSession` are their
two writes. `DeleteClient` is tokens then the client. `DeleteUser` keeps its explicit session loop
and drops the user-row acquisition: starting on the session rows is what lets it wait behind a
termination, the replay or a ceremony instead of deadlocking with them, measured clean against
issuance in both orderings on all four engines; the one pair its shape can still deadlock with, a
credential change for the same user, is the retry's and is stage 5's pair 2. The replay keeps its
session acquisition for the same local reason. Every comment at a kept-order site now says why that
transaction's first statement is first and obliges no other site; no comment in production code
states a lock order, and the doc comment on the issuance transaction carries the paragraph that is
now item 6 of `CLAUDE.md`'s Important Patterns.

Seven test files went (2534 lines): the four `tests/data/lock_order*_test.go`,
`commondb/client_delete_order_test.go`, `user/usersession_manager_lock_order_test.go` and
`handlers/lock_order_test.go`. The four tests that pin a kept order survive next to what they test:
the sweep's session-first order in `revocation_test.go`, the replay's in `handler_token_test.go`,
the replay-versus-termination pair in a new `tests/data/replay_ordering_test.go`, and the backfill's
first-statement pin where it was. The issuance call-order pin, one `RunInTransaction`, the
acquisition then the insert and nothing between, is a subtest of
`TestHandleIssueGet_TheAcquisitionOrdersTheInsert`. The blocked-party harness takes its rows with
`AcquireClientRow` on a test client, the same single-row UPDATE shape.

Tiers run locally: unit on three modules (57 packages, 1345 pass lines), data on four engines
(mysql, postgres, mssql, sqlite; 1605 pass lines), integration on sqlite (1012 pass lines), the last
because the stage touched the issuance and token handlers. Five mutations against the kept-order
pins, all caught: the issuance acquisition removed, the insert taken off the transaction, the
replay's acquisition removed, the sweep's sort inverted, and the acquisition answering "live" for a
terminated session at the data tier. The four-engine check suite for `0002a70b` is the CI run's to
report.

### Stage 5: the forced-deadlock pairs

Landed in `925946fb`. One new test file, `tests/data/deadlock_pairs_test.go`, and no production
change. It pins the residual: each same-account deadlock the dropped lock order leaves behind is
forced as two real production transactions, both through `RunInTransaction`, and shown to resolve
by the retry rerunning the engine's victim. A barrier decorator over `data.Database` parks one
party holding its first rows while the blocked-party harness confirms the other is genuinely queued
behind it, then releases to close the cycle. The pairs: the credential sweep against issuance
holding the session row; `DeleteUser` against the credential sweep; `DeleteClient` against a
termination whose session holds that client's tokens. Each asserts the end state through the public
getters and records per engine which pairs deadlocked.

Ran on all four engines locally: sqlite skips (one connection, no overlap), mysql, postgres and
mssql pass with the retry firing once per real deadlock. Pair 1 cycles on MySQL and SQL Server and
is clean on PostgreSQL by design (the code insert's `FOR KEY SHARE` does not conflict with the
sweep's `FOR NO KEY UPDATE`); pairs 2 and 3 cycle on all three. Pair 3's cycle is anchored on the
`codes -> clients ON DELETE CASCADE`, so it is deterministic. Mutation-checked by neutralising the
classifier gate in `commondb.deadlock`: MUTATION CAUGHT on all three engines, because with nothing
classified as a deadlock the raw engine error surfaces and every pair's assertion fails. The
four-engine check suite for `925946fb` is the CI run's to report.

**Corrected after the review, in `b9cb8a0a`.** Round 1 found two real defects in this stage's test
work, both fixed there and neither touching production code. The credential-sweep pair asserted only
issuance's error, so it accepted an impossible result (a live session with no minted code); worse,
its captured result lived outside the `RunInTransaction` closure, so a rerun could inherit the
aborted attempt's code. It now resets per attempt and branches on the outcome: live requires a code
that re-reads through `GetCodeById` with the right session, user, client, `Used == false` and
`Revoked == false`; gone requires no code at all. And the barrier had four unbounded waits with no
cleanup release, so a `t.Fatal` between the park and the release would have stranded a worker holding
database locks until the package timed out, presenting as a hang in an unrelated test. The release is
now a `sync.Once`-guarded `t.Cleanup` and all four waits are bounded at `lockWaitCeiling`.

## Tiers

Every stage ran unit on three modules and data on all four engines locally, plus integration on
sqlite where it touched a handler, route or schema. Per-stage numbers are in each stage's entry
above. **The four-engine integration result is the check suite's alone**, by design: locally it costs
five and a half minutes to learn strictly less than CI learns in four.

The check suite on `b9cb8a0a`, which is HEAD, is **green on all twelve jobs**: Build validation;
Tests on sqlite, mysql, mssql and postgres, each running the data tier then the integration tier;
Unit on authserver, adminconsole, core and setup; Lint; Vulnerabilities; Versions.
[Run 34034455618](https://github.com/leodip/goiabada/actions/runs/34034455618).

Twenty-two mutations were run across the five stages against the seams each stage made observable,
and every one was caught. Two more were run during the review, both caught.

## Where this departs from the issue

Four places, each settled with the repository owner before the agreement was sealed and each recorded
in the decisions comment below.

- **The auth-code replay keeps its leading session acquisition.** The issue removed it. Against the
  delete-first termination #139 keeps, the `main`-shape replay deadlocks on MySQL and SQL Server with
  the replay the victim, which is a cycle `main` never had, and a replayed code would answer 500
  instead of `invalid_grant` with no reuse audit event written.
- **`DeleteUser` keeps its explicit session loop**, dropping only the user-row acquisition. Starting
  on the session rows is what lets it wait behind a termination, the replay or issuance instead of
  deadlocking with them. `DeleteClient` does return to its `main` shape exactly.
- **The two #139 ordering tests could not "pass unchanged"** as AC1 asked: they called two of the
  four methods being removed. They drop those two calls and keep every assertion.
- **The retry landed here rather than being deferred on evidence.** The issue left it open; the
  owner's stated worry about deadlocks and about future authors having to remember a lock order is
  the evidence, and the retry is what makes removing the order safe rather than merely cheaper.

### Review

Reviewed by `openai-codex/gpt-5.6-sol` at effort `high`, 2 rounds over the whole branch, after a
plan-gate round by `openai-codex/gpt-6-astra` at effort `high` before any code existed. No stage was
reviewed on its own.

|---|---|---|---|---|
| plan | 1 | 7 | 1 | all 7 applied to `plan.md`, across four of the five stages |
| final | 1 | 4 | 1 | 2 fixed in `b9cb8a0a`, 2 refuted with evidence |
| final | 2 | 0 | 0 | clean pass |

- **blocking, security (plan gate):** `disableAndRevoke` builds its revocation audit outside its
  transaction callback, so making `inTransaction(nil)` retry would have made a deadlocked-then-retried
  collision backfill emit `AuditRevokedUserAuthState` naming sessions the committed attempt never
  terminated. The reviewer reproduced it in a Go overlay against a real engine. Fixed in stage 1
  (`98fe6646`): the audit is built per attempt and only the committing attempt publishes.
- **blocking, security (final round 1):** the review request itself directed the reviewer to open a
  URL, crossing the review trust boundary. Refuted as a finding about this change (the request is
  gitignored and not in the diff) and accepted as a note about the request template, which lives in
  another repository.
- **significant, conformance (final round 1):** the credential-sweep pair accepted an impossible
  result and let a rerun inherit the aborted attempt's captured code. Fixed in `b9cb8a0a`; the
  reviewer's own mutation went from `MUTATION SURVIVED` to `MUTATION CAUGHT`.
- **significant, quality (final round 1):** the barrier could strand a database worker on a
  `t.Fatal`. Fixed in `b9cb8a0a`.
- **significant, security (final round 1):** rollback failures are logged rather than returned.
  Refuted on three grounds, the weightiest being that the forced answer disables the retry on MySQL,
  where InnoDB rolls the victim back server-side and the client's rollback then errors. The reviewer
  conceded on round 2 after independently reading Go's `database/sql` rollback path and all four
  installed drivers.

Full account, with the evidence and the `unchecked` lists,
[in this comment](https://github.com/leodip/goiabada/pull/302#issuecomment-5559438343).

## Status

**Complete, and ready for review.** All five stages landed (`98fe6646`, `78d0ff13`, `18d1bfc5`,
`0002a70b`, `925946fb`), the review of the whole branch closed clean at round 2 with its two round-1
findings fixed in `b9cb8a0a`, and the check suite is green on `b9cb8a0a`, which is HEAD.

Three comments carry what does not fit here:

- [Deferred decisions](https://github.com/leodip/goiabada/pull/302#issuecomment-5559433699), of
  which there are none. All eight of the agreement's decisions were settled before the seal and the
  run reopened none, so nothing in this change rests on a bet you have not seen.
- [Follow-ups and fold-ins](https://github.com/leodip/goiabada/pull/302#issuecomment-5559433827):
  no issue was drafted, one fix was folded into stage 2, and the change leaves no `ceiling:` markers.
- [The full review account](https://github.com/leodip/goiabada/pull/302#issuecomment-5559438343):
  every round, every finding, and what answered it.

Nothing here asks an operator to do anything on upgrade. No schema, configuration, route, response
or template changed; the retry is invisible to a caller except that a transaction the engine would
have failed now succeeds.


leodip:
## Deferred decisions: none

Nothing in this change rests on an assumption the run took because it could not ask you. Section 3 of
the agreement holds **eight decisions, all settled by you at the interview before the seal**, and the
run reopened none, escalated none and raised none of its own. The guard's count at the finish is
`decisions: 8 total, 0 open, 0 raised by the run`.

So there is no list of reversible bets here, and nothing below needs your answer. It is recorded for
the reader who wants to know why the change looks like this rather than like the issue text, since
four of the eight depart from what #301 asked for.

| # | Question | Your answer | Where it shows |
|---|---|---|---|
| 1 | Does `revokeOnAuthCodeReuse` keep its leading session acquisition? | **Keep it** ("keep it, option A"). The issue removed it. Grounding found that removing it reopens a cycle `main` never had: against the delete-first termination #139 keeps, the `main`-shape replay deadlocks on MySQL and SQL Server with the replay the victim, so a replayed code would answer 500 instead of `invalid_grant` and write no reuse audit event. | `handler_token.go`; pinned by `TestRevokeOnAuthCodeReuse_TakesTheSessionRowFirst` and `TestLockOrder_ReplayResponseAgainstTermination` |
| 2 | What shapes do `DeleteUser` and `DeleteClient` return to? | **"B for users, drop for clients."** `DeleteUser` keeps #297's explicit session loop and drops only the user-row acquisition, because starting on the session rows is what lets it wait behind a termination, the replay or issuance instead of deadlocking with them. `DeleteClient` returns to `main` exactly. | `commondb/user.go`, `commondb/client.go` |
| 3 | The residual paragraph: as the issue quoted it, or amended? | **Rewritten around the retry** ("Good"). The issue's draft predates decision 7, and with the retry there is no victim a reader needs to know about. | `CLAUDE.md` Important Patterns item 6, and the doc comment on the issuance transaction |
| 4 | Which tests of a kept order survive, and where? | **"A": keep them, move them next to what they test, rewrite every comment to the local reason.** The alternative left each kept order an unpinned habit whose inversion would show up only as retries nobody sees. | four tests moved into `revocation_test.go`, `handler_token_test.go`, `tests/data/replay_ordering_test.go`, and one left in place |
| 5 | The kept #139 tests change in three mechanical ways, against AC1's "pass unchanged" | **Accepted.** They drop the two removed acquisitions and the paragraph explaining them; two helpers move to their only remaining caller; the blocked-party harness holds rows with `AcquireClientRow` instead. Every assertion stays. | `tests/data/issuance_ordering_test.go`, `blocked_party_harness_test.go` |
| 6 | Is the residual pinned by a test, or only written down? | **Pinned** ("I want it covered with solid tests that prove the implementation"). A paragraph in `CLAUDE.md` is not evidence for a claim about four engines, and a future change could falsify it on one engine with every tier green. | `tests/data/deadlock_pairs_test.go`, three forced pairs |
| 7 | The deadlocks that remain: accept, retry, or reorder more sites? | **Retry, in this issue** ("(B) here please! It's important"). Rejected: accept-and-document, which does nothing for a future site that deadlocks on one engine under a race; and reordering the credential paths, which is the order discipline this issue exists to remove. | `RunInTransaction` |
| 8 | Where does the retry live, and which owners go through it? | **"A": one entry point on `Database`, every owner through it**, with a lint that fails the build on a bare `BeginTransaction`. Rejected: the loop at only the six session-and-grant sites, which covers today's pairs and leaves next year's site to a rule in a comment. | `commondb/db.go`, `begin_transaction_lint_test.go`, 15 converted owners |

The one thing the run decided on its own was not a section 3 decision: the plan gate's blocking
finding (the collision e-mail backfill audited rolled-back attempts once `inTransaction(nil)` became
retrying) was a defect with one correct answer, so it was fixed in stage 1 rather than escalated. It
is in the review comment.

leodip:
## Follow-ups: none

No gate in this run drafted an issue. The plan gate's seven findings and the final gate's four were
all inside the deliverable: five were fixed in the stage that owned them, two in the round-1 fix
commit, one refuted with evidence and one accepted as a process note that belongs in another
repository (below). Nothing was verified as real, in scope for the project and out of scope for this
change, which is the only thing that becomes a follow-up.

Nothing to file, so nothing to duplicate-search.

## Folded in rather than filed

One, and it is why stages 2 and 3 have a shared stub rather than two.

- **Stage 2, `78d0ff13`: the admin key-rotation handler test.**
  `src/authserver/internal/handlers/apihandlers/handler_api_settings_keys_test.go` drives the real
  `oauth.SigningKeyRotator` over a mock `data.Database` with `BeginTransaction` / `CommitTransaction`
  / `RollbackTransaction` stubs, so converting the rotator to `RunInTransaction` in stage 2 broke it.
  The plan missed it because its sketch counted mock-based test files per module and this one lives
  in `authserver` while its subject lives in `core`.

  Fixed in the same stage rather than filed: it rides on a surface the stage already touched (the
  rotator's own tests), there was nothing to settle, it is the whole of the fix, and §2 stayed true
  as written. The fix is a package-level `apihandlers/run_in_transaction_stub_test.go` that hands the
  body the transaction, returns what the body returned and records it as `bodyErr`, plus
  `stubRotateRead` returning that stub so the refusal cases assert the body handed its error to the
  helper where they used to assert no commit. Stage 3 reuses it for that module's own ten owners, so
  the fold-in paid for itself one stage later.

## Ceilings left by this change

**None.** `git diff --unified=0 origin/main...HEAD | grep '^+.*ceiling:'` returns nothing, and that
is the right answer rather than an omission. The two things a reader might take for provisional are
decisions rather than simplifications: the retry's bound (three attempts, 25ms then 100ms) is
decision 7's answer, and the classifier's three error codes are the complete set for the engines that
can deadlock here, PostgreSQL `40P01`, MySQL `1213`, SQL Server `1205`, with SQLite unable to
deadlock at all on one connection.

## One process note, for `~/code/ai` and not this tracker

The final review's round 1 opened with a **blocking security finding against the review request
itself**, not against the code: `.review/request.md` imperatively directed an unsandboxed reviewer to
open `https://github.com/leodip/goiabada/pull/302`, which crosses the review trust boundary. The
reviewer declined to follow the URL and said so, which was the right handling.

Refuted as a finding about this change, since `.review/request.md` is gitignored, is not on the
branch, is not in the diff, and nothing in the deliverable moves either way. Accepted as a real point
about the request template: the round-2 request dropped the URL and named the branch and the diff
command instead. The template lives in the `leo-spec` skill in `~/code/ai`, so this is written up
here and reported to you rather than drafted as a goiabada issue.

leodip:
## Code review, the full account

This is the review of the change. There were two gates: the plan gate, which read `plan.md` against
the sealed agreement before any code existed, and the final gate, which read the whole branch against
`origin/main` after all five stages had landed. **No stage was reviewed on its own** (the run ran
under `LEO_REVIEW_MODE=end`), so the final gate's two rounds are the only reading the shipped code
got, and they are pasted here in full because the verdicts themselves live in a gitignored directory
that the next run overwrites.

**Totals: three rounds across two gates, eleven findings.** One blocking and five significant plus
one minor at the plan gate, all seven applied to the plan. One blocking and three significant at the
final gate's round 1: two fixed in `b9cb8a0a`, one refuted with evidence and conceded by the reviewer
on re-reading, one refuted as a note about the review harness rather than the code. Round 2 passed
clean with no findings. No production code changed after `925946fb`.

---

### Plan gate, round 1 of 2, 2026-09-05

Reviewer `openai-codex/gpt-6-astra` at effort `high`, Codex session resumed, request
`301-20260905T225504Z`. All three axes `reviewed`. Scope was `plan.md` against the sealed agreement,
with no production code in existence: the branch was one empty commit ahead of `main` and the
reviewer's final tracked diff was empty (SHA-256 `e3b0c442…52b855`).

**Seven findings: one blocking, five significant, one minor. All seven confirmed against the code,
all seven applied to `plan.md`, none refuted and none deferred.** No decision moved and none was
escalated: section 3 stays at eight decided, zero open.

The reviewer earned this by writing programs rather than re-reading. It ran Go overlays carrying the
proposed helper over unchanged production bodies, and real engines in the dev container, which is
what makes a finding credible at a gate where the thing under review is prose.

|---|---|---|---|---|
| 1 | security | blocking | `disableAndRevoke` audits rolled-back attempts once `inTransaction(nil)` retries | fixed: new stage 1 step 6, cases in step 8, mutation in step 10 |
| 2 | conformance | significant | web-origins database errors would become non-retryable sentinels, and responses would be written inside the callback | fixed: stage 3 sketch, owner leaves the refusal group |
| 3 | conformance | significant | the `DeleteUser` pair cannot require two successes | fixed: stage 5 oracle stated per victim |
| 4 | quality | significant | two lock-order test files still run in stages 2 and 3 after their subjects convert | fixed: 16 re-stubbed files becomes 18 |
| 5 | quality | significant | the lexical lint rejects the `Database` interface and the generated mock | fixed: `go/parser` call expressions, not text |
| 6 | quality | significant | the commit-error branch has no discriminating test | fixed: three scripted commit cases plus a mutation |
| 7 | quality | minor | the backoff assertion is not scheduling-independent | fixed: unexported sleeper seam, assert requested durations |

**Finding 1 is the one worth reading.** `src/core/data/commondb/email_backfill.go:disableAndRevoke`
declares its `revocationSweep` outside the callback and appends to both result slices inside it.
Stage 1 step 5 makes `inTransaction(nil)` retry, which recruits this function as a retrying owner
without anyone converting it, so a deadlocked-then-successful collision backfill would emit
`AuditRevokedUserAuthState` naming sessions the committed attempt never terminated and tokens it
never revoked. The reviewer reproduced it in an overlay: one successful retry audited
`[session-one, session-one]`. The plan could not have caught this on its own, because its conversion
sweep is scoped to the fifteen explicit `BeginTransaction` owners and this function contains no
`BeginTransaction`.

The run closed that family independently rather than on the reviewer's assurance:
`grep -rn "d.inTransaction(" src/core` returns exactly three callers, and the two besides
`disableAndRevoke` are `return d.inTransaction(tx, ...)` with nothing declared ahead of the call, so
they carry no cross-attempt state. Inside `disableAndRevoke`, `transitioned`, `NewGeneration` and
`OldGeneration` are assigned rather than appended and already self-correct; the two slices are the
entire defect.

**Finding 3 is the one that changed a test's meaning rather than its coverage.**
`IncrementUserAuthStateGeneration` refuses with "user not found when incrementing auth state
generation" when it matches no row, deliberately. Stage 5's sketch asserted both parties of each
forced-deadlock pair return nil, which on the `DeleteUser`-versus-credential-sweep pair is
satisfiable only by weakening that refusal. The reviewer produced the counter-example on real
PostgreSQL: one deadlock, the credential owner ran twice, the deletion committed, the retry
correctly refused. The oracle is now stated per victim, which is the honest form of the property the
retry actually promises: an operation still applicable succeeds, and one whose precondition the
winner removed still refuses.

**Findings 2, 4 and 5 were each a way the plan would have produced a red stage or an exempted
owner**, and all three are the kind only a reader without the author's picture finds. Finding 5 was
confirmed in a single command against the current tree.

`unchecked` carried five entries. All five are dismissed as plan-gate artefacts, each naming
something that does not exist yet and is already owned by a stage: implementation verification
(stage 1 steps 8 to 10), per-engine deadlock measurement including the mssql initialization stall
(stage 5), a live-engine backfill deadlock (unnecessary; the defect is captured state, which the
scripted driver settles), the unwritten conversions and integration pass (stages 3 and 5), and a
whole-server security audit (outside §2; this change adds no route, field, permission or template).
Reasoning per entry is in `log/plan.md`.

The reviewer also noted, as bookkeeping, that the copied #139 probe transcript shows PostgreSQL
deadlocking on the replay pair while §1's context table names only MySQL and SQL Server. Recorded,
not acted on: it strengthens decision 1 rather than reopening it, and §1 is frozen at the seal.

No follow-up was verified outside scope. The plan gate closes at round 1 without spending round 2:
every finding is applied, and what the reviewer says remains open is executable assurance the code
gates own.

---

### Final gate opened, 2026-09-06

Round 1's request covered the whole branch against `origin/main`: 6 commits `f39b7df6..925946fb`,
82 files, +3536/-4495, all three axes in scope, no part of it excluded. The check suite was green on
`925946fb`, all twelve jobs. The only earlier review was the plan gate above.

### Final gate, round 1 of 4, 2026-09-06

Reviewer `openai-codex/gpt-5.6-sol` at effort `high`, fresh session, `gate: final`, `round: 1` of
`max_rounds: 4`, request `301-20260906T095022Z`. Verdict `changes-requested`, four findings, all
`confidence: high`, all `judgement: true`. Two fixed, one refuted with evidence, one out of the
deliverable. All three axes reported `reviewed`.

Fixes landed in one commit against stage 5's work; production is untouched by this round, so the
branch's shipped behaviour is exactly what the check suite already judged green on `925946fb`.

#### Finding 1, blocking, security, `.review/request.md:Read this first`, REFUTED as a finding, accepted as a process note

Claim: the request imperatively directs an unsandboxed reviewer to open
`https://github.com/leodip/goiabada/pull/302`, crossing the review trust boundary. Family:
untrusted review-input instructions that direct external navigation.

Refuted as a finding about this change: `.review/request.md` is not on the branch, is not in the
diff, is gitignored, and nothing in the deliverable changes either way. Nothing to fix in the
repository, so no code moved.

Accepted as a process note, and the reviewer's own handling was right: it declined to follow the
URL and said so. The round-2 request drops the "Read this first" URL and names the branch and the
diff command instead. That is a `leo-spec` request-template matter and its home is `~/code/ai`,
not this tracker, so it is written up here and reported on the pull request rather than drafted as
a goiabada issue.

#### Finding 2, significant, conformance, `deadlock_pairs_test.go:TestDeadlockRetry_CredentialSweepAgainstIssuance`, FIXED

Claim: the pair asserts only issuance's error and accepts an impossible result, a live session with
no minted code, rather than requiring either a valid mint or a valid session-gone refusal. Evidence
offered: neutralising `mintCode` to a benign nil left the targeted PostgreSQL test passing,
`MUTATION SURVIVED`.

Verified and confirmed by reading. `issuanceOut.live` and `.code` were captured and never asserted;
the doc comment said so in as many words ("Whether issuance minted a code depends on which party
won, so it is not asserted"). Section 5 seam 3 asks the pair to observe "the password changed and
the code issued or refused", so the test was under-asserting against its own seam rather than
against a preference.

A second defect fell out of writing the oracle, and it is why the shape mattered. `out` lives
outside the `RunInTransaction` closure, so a rerun inherited the aborted attempt's values: an
attempt that set `live=true, code=X` and then lost the commit to a deadlock left `code=X` behind,
and a rerun finding the session gone would set `live=false` while `code` stayed non-nil. That is
precisely the hazard `RunInTransaction`'s own doc comment names ("A body that keeps state outside
the closure and appends to it sees the rolled-back attempt's rows on top of its own"), and the test
proving the retry was committing it. `out = issuanceOut{}` now opens every attempt.

Fixed: the closure resets its captured result per attempt, and the test now branches on `out.live`.
Live requires a non-nil code, re-reads it through `GetCodeById`, and pins its session identifier,
user, client, `Used == false` and `Revoked == false`, the last of which says the sweep invalidates
a code that won the race by advancing the user's generation rather than by writing to the code row
(#106). Gone requires no code at all.

The generation comparison is deliberately NOT asserted: `mintCode` issues under
`AuthContext{}`, so the code's `AuthStateGeneration` is zero whatever the user's is, and
`0 < post-sweep generation` would pass for a reason the test does not set up. The retained #139
ordering tests own the refusal shape.

Mutation, the reviewer's own, re-run after the fix:

```
where.sh mutate src/authserver/tests/data/deadlock_pairs_test.go \
  --replace 'code, err := mintCode(other, tx, client, user, session.SessionIdentifier)' \
  --with 'code, err := (*models.Code)(nil), error(nil)' \
  --type data --db postgres --run TestDeadlockRetry_CredentialSweepAgainstIssuance
MUTATION CAUGHT (exit 0)
```

Tree restored byte-identically; `production` stayed `00dc174302b4` across the run.

#### Finding 3, significant, quality, the new barrier harness, FIXED

Claim: unbounded barrier receives and no cleanup release, so a fatal setup or timing assertion can
strand workers holding database locks until the whole package times out. Target named
`blocked_party_harness_test.go:deadlockPair`; there is no such symbol, the code is `barrier` in
`deadlock_pairs_test.go`. Misnamed target, real substance.

Verified by reading the control flow. Four unbounded waits existed: `<-b.reached` and `<-b.release`
inside `arriveBefore`, and the two worker result channels. Between the park and
`b.releaseParked()`, pairs 2 and 3 call `requireBlocked` and `requireStillWaiting`, both of which
end in `t.Fatal`. A Fatal unwinds the test goroutine with `runtime.Goexit`, so `releaseParked` is
never reached, the parked worker waits forever on a release nobody sends while holding its
transaction, and every later test on that engine wanting one of those rows blocks until the
package timeout. What a reader would see is a hang in an unrelated test rather than the assertion
that failed, which this repository has already paid for once
(`reference_mssql_memory_exhaustion`, a hang presenting on an unrelated test).

Fixed, all four of the reviewer's forced answers:

- `newBarrier(t, what)` registers `t.Cleanup(b.releaseParked)`, so the release happens on every
  exit path including a Fatal;
- `releaseParked` is `sync.Once`-guarded, because the cleanup also fires on the success path after
  the test has already released and closing a closed channel panics;
- `awaitParked(t)` bounds the wait for the park at `lockWaitCeiling` and fails naming the worker;
- `awaitWorker(t, what, ch)` bounds the two result-channel receives the same way. The other party
  already had this: `blockedParty.await` has bounded at `lockWaitCeiling` since #139.

No mutation. Stranding a live database worker to demonstrate it costs the package timeout on four
engines, which is the failure being prevented; the reviewer declined for the same reason. The
control flow above is the evidence, and the `sync.Once` is exercised on every passing run, since
the cleanup fires after the explicit release in all three pairs.

#### Finding 4, significant, security, `commondb/db.go:runTransactionOnce`, REFUTED

Claim: rollback failures on the callback-error path are logged and discarded, so a caller continues
without knowing cleanup was not confirmed and a deadlock is retried despite an uncertain rollback.
Forced answer: surface rollback failures except `sql.ErrTxDone`, and let any rollback failure
prevent a deadlock retry.

Refuted on three grounds, in order of weight.

1. **The forced answer breaks the case the retry exists for, and a committed test says so.**
   `TestRunInTransaction_AVictimTheEngineAlreadyRolledBackIsStillRerun`
   (`commondb/run_in_transaction_test.go:164`) is MySQL's shape: InnoDB rolls a deadlock victim
   back server-side, the driver reports the client's subsequent rollback as an error, and the
   test's own comment reads "This is the case that fails if rollback errors are returned". Making
   a rollback failure prevent a retry disables the retry on MySQL, on the exact pair the three new
   data-tier pairs force. The reviewer's scratch assertion measured the behaviour correctly; its
   remedy inverts it.
2. **There is no uncertain rollback to surface.** `database/sql` sets `Tx.done` before it calls
   the driver, and `Tx.rollback` closes the transaction and either returns its connection to the
   pool or discards it as `driver.ErrBadConn`, whatever the driver answered. A retry therefore
   never runs beside a live predecessor holding rows: either the engine already ended that
   transaction, which is why it complained, or the connection was dropped and the engine ended it
   on disconnect. No SQL engine leaves a transaction open on a failed `ROLLBACK` with a healthy
   connection.
3. **It is decided, with its reason recorded.** Agreement section 4.2 says "rollback errors on
   this path are logged, not returned, because MySQL has already rolled a deadlock victim back and
   reports so", and seam 1 asks for exactly the test in point 1. The helper's doc comment carries
   the same sentence.

The commit path is not affected and never was: a failed commit returns its error unchanged and
attempts no rollback, which `run_in_transaction_test.go:195` pins.

No code changed.

#### The `unchecked` list, six entries, all answered

1. *No real driver forced to return a rollback error, so post-failure connection state is
   unmeasured.* **Dismissed.** Per finding 4 point 2 there is no such state for the helper to
   observe: `database/sql` finishes the `Tx` and releases or discards its connection before
   `Rollback` returns. The helper's own behaviour on a rollback error is covered at the unit tier
   by the scripted driver, on all three shapes (ordinary error, deadlock, already-rolled-back).
2. *Local tiers and the four database jobs not rerun.* **Covered this session.** Unit on three
   modules and data on all four engines rerun after the fix; results in the tier section below.
3. *The forced-deadlock cases not separately rerun locally on MySQL and SQL Server.* **Covered
   this session.** All three pairs ran and deadlocked on both: MySQL `Error 1213 (40001)` and
   SQL Server `mssql: Transaction (Process ID 61) was deadlocked ...`, each followed by the
   helper's `WARN rerunning a transaction the engine aborted as a deadlock victim attempt=2` and a
   PASS. PostgreSQL cycles on pairs 2 and 3 (`SQLSTATE 40P01`) and is clean on pair 1, which the
   test records rather than requires: `engine postgres: issuance blocked behind the credential
   sweep = false`, against `true` on MySQL and SQL Server.
4. *Deleted ordering tests inventoried by symbol; obsolete bodies and mechanical mock restubs not
   reread line by line.* **Dismissed, with a spot check.** Every restub goes through one of three
   package-level helpers (`handlers`, `apihandlers`, `core/user`) plus two inline sites
   (`handler_api_clients_test.go:510`, `signing_key_rotator_test.go:45`, `user_creator_test.go:132`);
   each calls `fn(tx)` and records `bodyRan`, so a stub that silently skipped its body is not among
   them. The only `Return(err)` forms are the deliberate `expectRunInTransactionRefused` shape,
   where the body is meant not to run. Seam 9 asks no more of the conversion than this, and the
   lint is what proves it complete.
5. *No browser pass.* **Dismissed.** No user-visible behaviour moves; section 2's documentation
   sweep found one page and it stays true.
6. *No fresh end-to-end attacks against PKCE, redirect registration, token verification, cookies
   and CSRF, rate limits, secret comparison.* **Dismissed.** All outside the diff; the reviewer
   confirms it inspected the changed call paths into them.

#### Round 1 tiers, after the fix

- unit, three modules, `where.sh test --type modules`: pass, ~2 min.
- data, all four engines, `where.sh test --type data`: 1614 pass, 0 fail, 46 skip (SQLite skipping
  what needs two overlapping transactions), ~4 min.
- integration: not applicable. The round changed one file, `tests/data/deadlock_pairs_test.go`, and
  no handler, route or schema. The four-engine result for the branch is the check suite's on the
  commit this round pushes.
- Mutation: one, recorded under finding 2, `MUTATION CAUGHT`.
- `tree-hash.sh --split` after the round, tree restored and clean but for the fix:
  production `00dc174302b4` (UNCHANGED from the tree the round reviewed), unit-tests
  `eb51a3830363` (unchanged), data-tests `7e7efcaf5e7e` (was `87915b40feef`),
  integration-tests `9f21fd92c02c` (unchanged), docs `9761398fa1c4` (unchanged).

Production and docs identical before and after: this round moved test code only. Round 2 is
therefore unconditional under the budget rather than funded by it, and it is requested to read the
two fixes.

### Tree-hash staleness check, before round 2

`tree-hash.sh --verify` after `b9cb8a0a`, the round-1 fix commit:

```
current   production 00dc174302b4   unit-tests eb51a3830363   data-tests 7e7efcaf5e7e
          integration-tests 9f21fd92c02c   docs 9761398fa1c4   all 6c993fc9797e

stage 1  unit stale          data stale          integration ?       docs -
         integration-tests not recorded here, so those claims cannot be settled either way
         moved since: production, unit-tests, data-tests
stage 2  unit stale          data stale          integration stale   docs -
         moved since: production, unit-tests, data-tests
stage 3  unit stale          data stale          integration stale   docs -
         moved since: production, unit-tests, data-tests
stage 4  unit live           data stale          integration live    docs live
         moved since: data-tests
stage 5  unit live           data stale          integration ?       docs -
         integration-tests not recorded here, so those claims cannot be settled either way

5 of 5 entries record a result against a tree that has since moved.
```

What changed against the block recorded when round 1 was requested: `data-tests` moved from
`87915b40feef` to `7e7efcaf5e7e`, which is the round-1 fix, so stage 5 goes from `data live` to
`data stale` and every stage now reads stale on data. Nothing else moved. `production`
(`00dc174302b4`), `unit-tests` (`eb51a3830363`), `integration-tests` (`9f21fd92c02c`) and `docs`
(`9761398fa1c4`) are byte-identical to the tree round 1 reviewed and to the tree the check suite
judged green on `925946fb`.

Every stale data claim is answered by the rerun recorded in the round-1 tier section: the whole
data tier, all four engines, against the current tree, 1614 pass and 0 fail. So no stage's data
claim rests on a tree that no longer exists.

Check suite on `b9cb8a0a`: pending at the time round 2 was requested.

### Final gate, round 2 of 4, 2026-09-06

Reviewer: `openai-codex/gpt-5.6-sol` at effort `high`, session resumed from round 1
(`.review/reviewer.env`: `GATE=final`, `ROUND=2`, `MAX_ROUNDS=4`). Scope: the whole branch against
`origin/main`, all three axes.

**Verdict: `pass`. No findings.** Conformance, quality and security all `reviewed`.

Its summary: "Round 1's two test defects are fixed. The repaired issuance oracle catches both a
missing code and a fabricated unpersisted code, the barrier cleanup and bounds cover every wait
introduced by the residual pair tests, the rollback finding is conceded after engaging its
refutation, and the whole branch remains conformant, sound, and security-reviewed."

So finding 4 of round 1, the rollback-visibility one, was refuted rather than fixed and the
reviewer accepted the refutation on its second reading, after going to the sources: it read the
committed already-rolled-back-victim test, Go's `database/sql` rollback and connection-release
paths, and the rollback implementations of all four installed drivers. That is the one refutation
in this run that was independently re-derived rather than merely not re-raised.

What it did, from its `ran` list: re-read the contract, the round 1 account and the request;
reviewed `b9cb8a0a` independently across every changed result-state, catalog-oracle, cleanup,
release, barrier, timer and worker-channel branch; confirmed that only
`src/authserver/tests/data/deadlock_pairs_test.go` moved since round 1 and that production, unit
tests, integration tests and documentation are byte-identical; recounted the production transaction
openings, all 15 `RunInTransaction` closures, the removed acquisition-method references, the
production lock-order prose and the documentation changes; ran `check-anchors.sh` over all 24
anchors, `git diff --check`, the split tree hash and the HEAD check-suite status; inspected the
barrier's cleanup under success, early fatal exit, duplicate release, missing arrival and missing
worker completion; inspected `issuanceOut` evaluation and reset across retries against the real
`GetCodeById`; and reapplied the security reference over untrusted input, authorization, injection,
secret and log exposure, crypto and token handling, session and replay integrity, concurrency,
denial of service and rate limiting.

**One mutation, run by the reviewer, caught.** It made `mintCode` return a fabricated non-nil code
without inserting it and ran `TestDeadlockRetry_CredentialSweepAgainstIssuance` on PostgreSQL. The
real-catalog assertion added in `b9cb8a0a` failed, so the round 1 fix closed both halves of the
defect: a missing code and an invented one.

#### The six `unchecked` entries, answered

1. *The unit and four-engine data tiers were not rerun locally.* **Covered.** `where.sh ci` this
   session reports all twelve jobs green on `b9cb8a0a`, which is HEAD: Build validation, Tests on
   sqlite, mysql, mssql and postgres, Unit on authserver, adminconsole, core and setup, Lint,
   Vulnerabilities and Versions. That is the four-engine data and integration result, against the
   exact tree the verdict judged.
2. *No early `t.Fatal` was injected while a real transaction was parked.* **Dismissed.** The
   property is the data-tier harness's own robustness, not shipped behaviour: the worst case it
   guards is a package timeout in place of a clean failure message. Go has no supported way to
   assert a subtest fails without re-exec, and the experiment needs a live database, so covering it
   would mean a subprocess harness inside the tier to protect a diagnostic. Control-flow review is
   proportionate here, and the three properties it checked are the ones that matter: the release is
   registered as a `t.Cleanup`, it is idempotent, and the park and both worker result channels are
   bounded.
3. *No real driver was forced to fail `ROLLBACK`.* **Dismissed.** The helper's decision does not
   depend on which driver produced the error. It classifies what `fn` or the commit returned and
   never returns the rollback's own error, so a real driver failing rollback reaches exactly the
   path already covered: `TestRunInTransaction_AVictimTheEngineAlreadyRolledBackIsStillRerun` in
   `src/core/data/commondb/run_in_transaction_test.go` fails the rollback three times through the
   scripted driver and asserts the reruns still happen, that the deadlock is what surfaces, and
   that the rollback failure never does.
4. *Byte-identical production and mock hunks were not reread line by line.* **Covered.**
   `tree-hash.sh --split` this session: `production 00dc174302b4`, identical to the value round 1
   reviewed and to the tree the check suite judged. Only `data-tests` has moved since round 1, and
   that is the round 1 fix itself.
5. *No browser or end-to-end attack pass over PKCE, redirect, token verification, cookie, CSRF,
   rate limit or secret comparison.* **Dismissed.** None of those behaviours changed on this
   branch. The two handler files it touches, `handler_auth_issue.go` and `handler_token.go`, change
   only how the transaction is opened and which rows are acquired inside it; grepping their diff
   for `pkce|code_challenge|redirect|verifier|hmac|subtle|rate` returns a single hunk, a comment
   re-indented by the closure. CI's integration tier exercised those endpoints on four engines and
   is green.
6. *No external specification was fetched.* **Dismissed.** Section 1 records that nothing here is
   wire-observable and that no normative text bears on the change; neither round disturbed that.

#### Where the review ends

Round 2 is clean and nothing is left to raise, so `.review/final-review.md` is deleted and the
review closes at two rounds. Round 1 found two real defects, both in stage 5's forced-deadlock
pairs, both fixed in `b9cb8a0a` with no production code touched; round 2 confirmed the fixes by
mutation and passed the whole branch. No follow-up and no decision came out of either round.

---

### After the review closed

`where.sh ci` on `b9cb8a0a`, which is HEAD: **green, all twelve jobs**, including the data and
integration tiers on sqlite, mysql, postgres and mssql.
[Run 34034455618](https://github.com/leodip/goiabada/actions/runs/34034455618). That settles the
`pending` recorded in the staleness block above, and it is the four-engine result for the exact tree
round 2 judged.

PR #303, opened 2026-09-06
feat(data): a migration runner of our own, and schema rollback

### LINKED ISSUE #268: golang-migrate links a second Postgres driver, a second JWT library and five Azure modules into both binaries, to run numbered .sql files

leodip:
Closes #268.

Goiabada's schema is migrated by `golang-migrate`, and that one library brings in eight more
modules: `lib/pq`, `golang-jwt/jwt/v4` and five `Azure/go-autorest*` packages, none of which
Goiabada itself uses. This replaces it with a migration runner of our own, in `src/core/data/migrator`.

The runner is deliberately not a redesign. It writes the same `schema_migrations` row, in the same
order, under the same lock resources, and runs each file the same way the library did, so a database
at any recorded version sees no difference between its last start under the library and its first
under the runner: no step is re-run, none is skipped, and a failure leaves the same dirty marker in
the same place. What it does differently is stricter (every version-read error is an error rather
than silently "never migrated"), returns what it took (no connection is pinned for the life of the
process), or is new surface.

The new surface is schema rollback. The runner steps down as well as up, a chain-wide test walks
every migration down and back up on all four engines comparing the catalog at each version, and
`goiabada-authserver migrate to <version>` gives an operator a way to step the schema back before
installing an older release. `goiabada-authserver migrate version` prints what the binary expects
and what the database records.

**The rule this establishes:** rolling a deployment back means stepping the schema down with the
*newer* binary first, because the down files for the versions being removed only exist there, and
then installing the older release. A server refuses to start against a database a newer release
migrated, which is what makes the order safe rather than merely recommended.

**Status:** complete. All five stages have landed, the whole branch has been reviewed over three
rounds and the last came back clean, and the check suite is green on all twelve jobs at `69e84787`.
The runner replaces `golang-migrate` on all four engines, every down migration runs and restores the
catalog its up built, `goiabada-authserver migrate version` / `migrate to <version>` make a rollback
reachable without a database client, and neither server binary links the library or the eight modules
that came with it.

### Stage 1: the runner package, with nothing calling it
`b9306bbf`. New package `src/core/data/migrator`: `errors.go`, `source.go`, `engine.go`,
`migrator.go`. Additive, so the four engines still build library instances and the tree compiles
throughout.

The runner parses a set of embedded `.up.sql` / `.down.sql` files eagerly with the library's own
filename pattern, walks the versions the engine actually carries rather than counting (the four
sets have different gaps), and per step writes `(target, dirty)`, runs the file, writes
`(target, clean)`. A file runs inside a transaction on SQLite and bare on the other three, which
is what the migrations are written around. The public surface is `New`, `Head`, `Engine`,
`Version`, `Up`, `Migrate`, `Force` and `Plan`, and there is no `Close`.

The lock resource names are a byte-for-byte copy of the library's formula, salt included: during
an upgrade one replica computes the name with the library and another with this, and a different
name means both migrate the same database at once. A unit test pins the six values the library
produced for Goiabada's own database names.

Three deliberate departures:

- **Every version-table read is fail-closed.** The library's SQLite driver swallowed every read
  error and answered "never migrated", which re-runs the chain from the first migration against a
  populated database. More than one row is refused naming every row.
- **No `Close`.** A connection is taken for one operation and given back, where the library pinned
  one per pool for the life of the process. A failed unlock joins the operation's error, and that
  session is handed `driver.ErrBadConn` so it is discarded rather than lent to the next borrower
  still holding a migration lock.
- **`ErrDirty` names recovery versions from the direction that failed**, not from arithmetic on
  the marker: a down step from N marks N-1, so its "did not apply" end state is N, above the
  marker. Where the marker was read back from the table the direction is not recorded, so the
  message says so and names all three candidates rather than guessing.

Tiers: unit on all three modules, 1368 pass / 0 fail, the new package at 0.492s; `where.sh
migrations` all rules pass. Data and integration not applicable, since nothing imports the
package. Eight mutations, one per behaviour the package makes observable, every one caught.

### Stage 2: the four engines cut over, and every caller with them

`ab59370b`. Every `NewMigrator()` returns `*migrator.Migrator` and every `Migrate()` steps through
it, so no Go file outside `src/core/data/migrator` imports `github.com/golang-migrate/migrate/v4`.
The module requirement itself stays until stage 5: `go mod tidy` keeps a module while `go.sum`
still names it.

The bookkeeping is deliberately unchanged, which is the property that makes the swap safe: the same
table, the same single row rewritten in one transaction per step, the same dirty marker written
before a file and cleared after it, the same per-engine transaction wrap, and the same lock
resources. A database migrated by the previous release and then opened by this one sees no
difference.

What did change:

- **SQLite's `schema_migrations` pre-create issues `version_unique` itself.** The library's driver
  created that index on every construction. Without a replacement a fresh install would lack an
  index every deployed database has and `schema.golden` records, so the golden comparison and the
  shape test would both fail. One statement leaves all four true; a migration would have moved a
  golden and two tests for an index nobody queries by.
- **SQL Server's pre-create takes its lock through the runner's own `Engine`** rather than through
  `database.GenerateAdvisoryLockId`. That was the last non-migrator use of the library, and taking
  the resource from the same place the migration does is what stops the two diverging.
- **`migrator.StartupRefusal`** composes the "a newer release migrated this database" refusal once
  for all four engines, naming the recorded version, this binary's head, the release it is, and the
  two ways out.
- **The runner has no `Close`.** The data-tier helper drops its cleanup and the RCSI fixture closes
  the handle it opened; that fixture's zero-open-connections assertion now reads whether the runner
  gave its connection back, which is a property the library never had.
- **New `migration_lock_test.go`**: live contention on MySQL, PostgreSQL and SQL Server, holding the
  resource from the production symbol on an independent session. Nothing else in the change can tell
  a runner that takes the lock from one that never reaches for it, since the names are pinned as
  strings and every other test runs serially.
- **Folded in:** `AGENTS.md` was missing the Transactions bullet #301 added to `CLAUDE.md`. The two
  files are meant to be identical and now are.

Tiers, local: unit **1370 pass** on three modules, data **1620 pass** on mysql, postgres, mssql and
sqlite, integration **1012 pass** on sqlite, and the migration rules gate green. Six mutations, five
caught first time. The sixth survived and found a real defect in the new lock test: all three
locks are re-entrant within a session and the runner returns its connection to the pool, so the
"is the lock free now" check was handed the runner's own session and passed against a runner that
never unlocked. It opens a pool of its own now, and the mutation is caught on all three engines.

### Stage 3: the chain-wide round trip, and the downs it breaks
Landed in `55d0d281`.

A new data-tier test, `TestMigrationChain_EveryDownRestoresTheCatalog`, walks one isolated database
per engine up the whole migration chain recording the catalog at every version it carries, back
down to an unmigrated database comparing at each landing, and up again comparing at every version.
The comparison unit is the canonical text a `schema.golden` is written in, with the recorded
migration version in its header, so a step that moves the schema correctly and the version marker
wrongly fails too.

It found **twenty broken down migrations, nine of which did not run at all**. The downs from 000001
to 000023 had never been executed by any test; the 21 existing per-migration round-trip tests all
start at 000024.

| Engine | Version(s) | What was wrong |
|---|---|---|
| all four | 000003 | re-added `settings.session_authentication_key` and `session_encryption_key` NULLABLE, where 000001 declares both NOT NULL. MySQL's also used `BLOB` where 000001 uses `longblob` |
| sqlite | 000002 | rebuilt three tables declaring `INTEGER` where 000001 declares `numeric` and `int`. SQLite reports the declared type verbatim, so a rolled-back database and a fresh install of the target release disagreed |
| mysql | 000011 | **did not run**: dropped an index before the foreign key InnoDB enforces through it |
| mysql, postgres, mssql | 000031 | restored `user_sessions.level2_auth_config_has_changed` carrying a default the original does not have |
| mssql | 000004, 000006, 000009, 000010, 000013, 000016, 000017, 000018 | **none of these eight ran**: `DROP COLUMN` on a column carrying an unnamed `DEFAULT` constraint, which SQL Server refuses |
| mssql | 000007 | restored `codes.code_challenge` as `VARCHAR` where 000001 declares `NVARCHAR` |

SQLite's 000031 down was already correct and is untouched: SQLite's 000002 adds that column *with*
`DEFAULT 0`, so restoring one is right there and wrong on the other three. The existing
`TestMigration000031_OTPConfigGeneration` asserted the defect as if it were intended, and is now
engine-aware.

Two things are deliberately not compared, both explained in the test: SQLite's `schema_migrations`
below 000041, whose shape depends on whether the database was pre-created by the runner or arrived
by rolling 000041 back, and SQL Server's column collation below 000040, which no migration there
declares and which 000040's down deliberately returns to the pre-#283 value. Each has two correct
answers, and comparing either reports every version below it as broken and buries the real
findings.

**Tiers:** unit, 58 packages green. Data green on all four engines locally (sqlite 11.7s, mysql
134.2s, postgres 18.3s, mssql 84.7s), run unnarrowed on purpose because three of the four sets of
broken downs are invisible from SQLite. Integration not applicable: no down migration runs in the
up chain every integration test starts through. Four mutations against the new test, all caught,
including reverting the SQL Server unnamed-default drop.

**Not green:** `where.sh migrations` fails its snapshot rule, which asks an engine whose migrations
changed to move its `schema.golden` too. Regenerating all four against live databases rewrites them
byte-identically, because a `.down.sql` never runs in the up chain a golden records. That is
decision 14, below.

### Stage 4: `OpenDatabase`, the `migrate` subcommand, and the docs

`ed1f7079`. Rollback becomes something an operator can ask for, rather than something the code
merely permits.

```
goiabada-authserver migrate version        what this binary expects, and what the database records
goiabada-authserver migrate to <version>   step the schema there, up or down
```

Both read the usual `GOIABADA_DB_*` variables and neither starts the server. The dispatch sits
after the configuration loads and before the data-encryption key is validated: a migration touches
no encrypted value, and requiring the key would make it a precondition for repairing a database on
a deployment that has not set one.

`data.OpenDatabase` is `NewDatabase`'s engine switch moved out, returning the concrete engine
database having migrated nothing and run no startup task. `NewDatabase` is now a call to it
followed by exactly what it did before, so its contract and its 906 lines of callers do not move.
The command needs it because `NewDatabase` brings the schema to head on the way out, which would
make a step down impossible and a `migrate version` on a database behind this binary a lie.

`migrate to` refuses a target below schema version 000044, the version this release ships with.
Rolling further back would put the schema where an older release expects it while leaving that
release unable to read the data: v1.6.0 moved every TOTP secret into an encrypted column and
blanked the plaintext one, and moved the data key out of the database into the environment.
Neither is a SQL migration, so no `.down.sql` reverses either. The floor lives in the authserver,
not in the runner, which stays unrestricted so the tests can step anywhere.

It also refuses a target above the head, a dirty database, and a version this binary does not
carry, each naming what an operator needs: the recorded version, the nearest versions the binary
does carry, and the two legal end states after an interrupted migration. Exit code 2 for a
malformed invocation, 1 for a refusal, so a deployment script can tell them apart.

**Two defects found while building it, both fixed here rather than filed:**

- **The authserver unit tier never ran `./cmd`.** `run-tests.sh` ran `./internal/... ./web`, with a
  comment saying `./cmd` has no tests, true until this commit. A mutation against a deliberately
  deleted rollback-floor check came back `MUTATION SURVIVED`, and the log showed the package absent
  from the tier: this commit's eleven test functions were passing without being executed. CI's
  `Unit / authserver` job runs the same tier, so the gap was CI's too.
- **Every binary was built from `main.go` by filename**, in six places across both servers.
  `go build <file>` compiles that file alone, so `migrate.go` would not have been in any binary.
  All six now name the package, with `-buildvcs=false` to keep what ships today: building from a
  file never stamped VCS either, the release build injects version, date and commit through
  `-ldflags`, and stamping fails outright in the dev container, where git refuses the mounted
  workspace for dubious ownership. The adminconsole halves are folded in for consistency rather
  than forced by the change.

Docs: a `## Schema migrations and rollback` section on the database setup page, saying that
migration is automatic, what the two commands do, and that a rollback is back up, step down on the
current binary, then install the older release.

Tiers: unit 4624 pass, data on all four engines 2171 pass, integration on sqlite 1333 pass, no
failures. Seven mutations, all caught; two survived first and each was a real finding, one the tier
gap above and one about tests asserting an exit code against the constant that produced it.

Follow-up in `e5e515da`: CI's Lint job was red on `ed1f7079` with twenty-two `errcheck` findings, all
of them unchecked `fmt.Fprintf` and `fmt.Fprintln` writes to the writer the `migrate` subcommand
prints through. errcheck excludes those calls when they go to `os.Stderr` and not when they go to an
arbitrary `io.Writer`, and this is the first file in the repository to print through an injected one.
The twenty-two sites now go through one `outf` helper that discards the write error once, with the
reason beside it; no output text changed. Lint is green over all four modules, and the unit and data
tiers were re-run.

### Stage 5: the library and its eight leave both binaries

`30c3d3c7`. `go mod tidy` in `src/core`, `src/authserver` and `src/adminconsole`, which is all that
was left: stages 1 to 4 had already removed every `import "github.com/golang-migrate/..."` line, so
this stage changed no Go source at all. Six files, 2 insertions and 322 deletions.

Gone from all three `go.mod` files and all three `go.sum` files: `github.com/golang-migrate/migrate/v4`,
`github.com/lib/pq`, `github.com/golang-jwt/jwt/v4`, and the five `github.com/Azure/go-autorest*`
modules. `go mod why -m` over all nine names, from each of the three modules, answers "main module
does not need" for all 24 pairs.

The external modules linked into each binary, measured by listing the packages reachable from `main`
and folding them to modules:

| Binary | Before | After |
|---|---|---|
| `goiabada-authserver` | 50 | 43 |
| `goiabada-adminconsole` | 48 | 41 |

Seven rather than nine, because `github.com/Azure/go-autorest` (the base module) was a requirement
that contributed no package to either binary. Nothing else left either list.

Two lines were added rather than removed, and neither reaches a binary. `src/core` gains
`davecgh/go-spew` as an indirect requirement, because it is reachable through `go-sqlbuilder`'s own
tests and `golang-migrate`'s module graph is no longer there to supply that requirement transitively;
it is linked into neither binary. `src/adminconsole` moves `golang-jwt/jwt/v5` from its indirect block
to its direct block, correcting a stale marker: a middleware test has imported it directly for some
time, and this is the first `go mod tidy` that module has had since.

`src/cmd/goiabada-setup` keeps its own `github.com/lib/pq`. It is a separate module and that is its
PostgreSQL driver, not a consequence of the library.

Tiers: unit 4624 pass across all three modules, data 2171 pass on mysql, postgres, mssql and sqlite,
integration 1333 pass on sqlite. No failures.

## For the release notes

**This release expects schema version 000044.**

Nothing has to be done on upgrade: the auth server migrates the schema when it starts, exactly as it
did before, and a database at any recorded version moves forward without re-running or skipping a
step. What is new is what to do when you go the other way.

**Rolling a deployment back means stepping the schema down with the *newer* binary first, and only
then installing the older release.** The `.down.sql` files for the versions being removed exist only
in the newer release, so the older binary cannot undo them; and the older binary refuses to start
against a database a newer release migrated, which is what makes the order safe rather than merely
recommended.

The procedure, in full:

1. Back the database up. A rollback discards whatever the newer release wrote into structures the
   older one lacks, and that is not recoverable from the schema.
2. Read the target release's notes for the schema version it expects.
3. On the **current** (newer) binary: `goiabada-authserver migrate to <that version>`.
4. Install the target release.

`goiabada-authserver migrate version` prints the version this binary expects and the version the
database records, which is how you check both ends before and after. Both subcommands read the usual
`GOIABADA_DB_*` variables and neither starts the server.

**Rollback is supported between releases from this one onwards.** `migrate to` refuses a target below
schema version 000044 on purpose: earlier releases changed *data* in ways no SQL migration reverses.
v1.6.0 moved every TOTP secret into an encrypted column and blanked the plaintext one, and moved the
data-encryption key out of the database into the environment. Stepping the schema below that floor
would put it where an older release expects it while leaving that release unable to read the data.

The same rule is on the documentation site, under **Schema migrations and rollback** on the database
setup page.

## Tiers

Locally, on the final tree: unit on all three modules (59 packages, 0 fail), the data tier on all
four engines (0 fail), integration on sqlite (0 fail), `where.sh migrations --no-snapshot` with every
rule OK, `check-anchors.sh` 23 of 23, and a lint sweep — `gofmt`, `go vet`, `golangci-lint`,
`unparam` — reporting `0 issues.` on each module.

**The four-engine claim is the check suite's, not a local run.** `where.sh ci` at
`69e84787b0f5966b9aee5093345455d2d0df7da1` is green on all twelve jobs: `Tests / sqlite`, `mysql`,
`postgres` and `mssql` — the data *and* integration tiers on four live engines — plus `Unit / core`,
`Unit / authserver`, `Unit / adminconsole`, `Unit / setup`, `Build validation`, `Lint`,
`Vulnerabilities` and `Versions`. Local integration runs on sqlite only, by design.

Sixteen mutations were run across the branch against its new tests, every one caught, each restored
and verified by hash before its commit. The per-round tables are in the review comment.

## Deviations worth knowing

- **Twenty broken down migrations, against the agreement's assumption of two.** Nine did not run at
  all (MySQL's 000011, and eight SQL Server downs dropping columns carrying unnamed `DEFAULT`
  constraints); eleven ran and left the catalog wrong. All twenty are fixed, and the chain-wide round
  trip on four engines is what found them and what holds them.
- **A lock-acquisition test that section 5 had rejected.** The rejection assumed contention needs two
  processes; it needs two *sessions*, which one test can hold. Stage 2 gained contention tests on
  MySQL, PostgreSQL and SQL Server.
- **`where.sh migrations --no-snapshot` gated stages 4 and 5**, under decision 14, which you settled
  on this thread. The rule has since been fixed to key on `.up.sql` changes. Details in the decisions
  comment.
- **Five problems were folded into the stage that found them rather than filed**, including a unit
  tier that never ran the authserver's `./cmd` package and six build invocations that compiled
  `main.go` by filename. Listed with their reasoning in the follow-ups comment, so you can object to
  any of them.

## Review

Reviewed by `openai-codex/gpt-6-astra` at effort `xhigh` (rounds 1 and 2) and
`deepseek/deepseek-v4-pro-0813` at effort `high` (round 3), three rounds over the whole branch. No
stage was reviewed on its own: this run used end-of-run review mode, so this is the only reading the
code got. A fourth, static round ran against the plan before any code existed.

|---|---|---|---|
| 1 | 4 (2 significant, 2 minor) | 0 | all fixed, `2410892a`, with `565f44c7` as a Lint fixup |
| 2 | 2 (both significant) | 0 | all fixed, `69e84787` |
| 3 | 0 | 0 | clean |

- **significant, conformance:** `ErrDirty` told an operator to recover a half-finished migration to a
  version that does not exist. It derived the "did not apply" end state as the dirty marker minus
  one, and the marker minus one is not a version: the four sets have gaps, so an interrupted up to
  `000005` on a database last at `000002` named `000004`, which the binary then refuses on the next
  start. Fixed in `2410892a`: the error carries the source's own predecessor, and the one marker that
  records no version at all is answered explicitly.
- **significant, quality:** every caller that treats "nothing to do" as success accepted an
  operational failure with it. The runner joins a failed unlock onto the operation's error on
  purpose, because a session-scoped lock that did not come back blocks every other migrator on the
  database. At head that join carries `ErrNoChange`, and `errors.Is` found it, so all four engine
  wrappers logged "no need to migrate the database" and started the server with the lock held. Fixed
  in `2410892a`: `migrator.IsNoChange` and `migrator.IsNilVersion` are identity tests, all six
  production sites use them, and a lint holds every future caller under `src/` to the same rule.
- **minor, conformance:** eight SQL Server down migrations discovered their unnamed default
  constraints by table name, and `sys.default_constraints` spans the whole database while the
  statement they build is schema-qualified, so a same-named table in another schema lost its defaults
  during a rollback. Fixed in `2410892a`: each query is pinned with `OBJECT_ID` to the exact object
  its own `DROP COLUMN` targets.
- **minor, conformance:** the new documentation section said each server migrates on startup. Only
  the auth server opens the database. Fixed in `2410892a`.
- **significant, quality:** the lint round 1 added was defeated by an import alias — it matched the
  receiver `errors` by spelling, and `mssqldb/db.go` already imports `goerrors "errors"`, so
  reverting that one engine would have restored the defect and been reported by nothing. Fixed in
  `69e84787`: the receiver is resolved through the file's own import list.
- **significant, quality:** round 1's lock-leak repair had no failing-path test — with the session
  discard deleted, the covering test still passed. Fixed in `69e84787` with a driver-level fault
  injection that fails only `EXEC sp_releaseapplock` against real SQL Server, asserting on both DDL
  outcomes that a separate pool can take the resource afterwards.

Round 2 reported no production defect: both its findings were about guards round 1 itself added. Round
3 read the whole branch and found nothing.

Two problems were folded in from the reviewer's follow-up sections rather than filed: SQL Server's
`schema_migrations` pre-create discarded the result of releasing its lock (`2410892a`), and the dirty
refusal named the predecessor of a version the running binary does not carry (`69e84787`).

Full account, with the evidence, the mutation tables and every `unchecked` entry answered, in
[this comment](https://github.com/leodip/goiabada/pull/303#issuecomment-5571466425).

---

- Deferred decisions, and the one escalation you settled:
  [comment](https://github.com/leodip/goiabada/pull/303#issuecomment-5571416941)
- Follow-ups, fold-ins and ceilings:
  [comment](https://github.com/leodip/goiabada/pull/303#issuecomment-5571423296)
- The review account:
  [comment](https://github.com/leodip/goiabada/pull/303#issuecomment-5571466425)


leodip:
## Decision 14: this change edits twenty `.down.sql` files and no `.up.sql`, and the migration checker wants a `schema.golden` edit that does not exist. What should it do?

Stage 3 of the #268 run has landed (`55d0d281`). It added a data-tier test that walks each engine's
whole migration chain up, back down to an unmigrated database, and up again, comparing the catalog
at every version, and it found twenty broken down migrations. Nine of them did not run at all:
MySQL's 000011 dropped an index before the foreign key InnoDB enforces through it, and eight SQL
Server downs dropped columns carrying unnamed `DEFAULT` constraints, which SQL Server refuses. All
twenty are fixed and the data tier is green on sqlite, MySQL, PostgreSQL and SQL Server.

The gate that now fails is not about the code. `where.sh migrations` runs ten text rules over the
four `migrations/` directories, and the tenth, called `snapshot`, says: an engine whose migration
files changed must also change that engine's `schema.golden`. Those four files are the checked-in
record of the catalog each engine's migration chain builds, generated by
`cd src/core && go run ./cmd/schemadump`.

Stage 3 changed twenty `.down.sql` files across all four engines and no `.up.sql` file. A down
migration never runs while the chain is being applied forward, so it cannot change the catalog a
golden file records. I ran the generator in the dev container against all four live databases; it
rewrote all four files and `git status` reported no change to any of them. So there is no golden
edit that would satisfy the rule, and the rule fails:

```
FAIL snapshot             4 engine(s) have an unrefreshed schema.golden
     mysqldb: migrations changed, src/core/data/mysqldb/schema.golden did not
     postgresdb: migrations changed, src/core/data/postgresdb/schema.golden did not
     mssqldb: migrations changed, src/core/data/mssqldb/schema.golden did not
     sqlitedb: migrations changed, src/core/data/sqlitedb/schema.golden did not
```

This blocks rather than waits, because the check is scoped to the whole branch rather than to the
last commit. Now that the down edits are on the branch it will fail again at the close of every
remaining stage, so stages 4 and 5 cannot land without an answer.

### Options

**(a) Fix the rule so it keys on `.up.sql` changes rather than on any file under `migrations/`.**
A `.down.sql` cannot move a golden file, so requiring one is asking for something impossible.
Everything else the rule protects is unchanged: an engine that adds or edits an up migration still
owes a regenerated golden, which is the case #282 and #284 built it for. The checker is not a file
in this repository, so this is a change in the tooling repository and not in this pull request.

**(b) Run the gate as `where.sh migrations --no-snapshot` for the rest of this run**, recording the
byte-identical regeneration as the evidence that there is nothing to refresh. The flag already
exists. This unblocks the run today and leaves the rule able to fire the same way on any future
down-only fix.

**(c) Hand-edit a golden file so something changes.** I have not done this and would not: those
files are generated from a live database, so an edit is a claim about a catalog rather than a
record of one, and the data tier compares a freshly migrated database against them on each engine.
It would go red in CI immediately.

### Recommendation

**(a), with (b) as the interim so this run can finish.** The rule is right about the property it
protects and wrong about which files imply it. Getting it wrong in the (b)-only direction costs a
repeat of this conversation the next time a down migration is repaired, which after this run is
likely to be rare. Getting it wrong in the (c) direction breaks the one record that made #282's
five cross-engine divergences reviewable at all.

I have not touched the checker or the golden files, and I have not raised the rule's baseline. If
you would rather I do something other than the above, say which and I will apply it.

leodip:
(a), with (b) as the interim so this run can finish

leodip:
## Decision 14: settled, with one correction to what I told you

Applied your answer as written: **(a)**, with **(b)** as the interim.

**(a) is done.** `check-migrations.sh`'s snapshot rule now keys on a changed `.up.sql` inside an
engine's `migrations/` directory rather than on any file in it. An engine that changed only downs
reports `OK snapshot  only down migrations changed (<engines>); a down cannot move a golden`, so it
stays distinguishable from "no migration changed" rather than passing silently. The property the
rule protects is unchanged: an up that lands without a refreshed `schema.golden` still fails.
Committed in the tooling repository as `ai@09539a6` and synced. Nothing in this pull request moved.

**The correction.** I told you stage 3 changed twenty `.down.sql` files and no `.up.sql`. The second
half was wrong, and making the change is what found it. Against the merge base `df02d1aa` the branch
changes 18 `.down.sql` files and one `.up.sql`:

```
src/core/data/mssqldb/migrations/000031_otp_config_generation.up.sql
```

Its entire diff is inside the comment block above the
`IF OBJECT_ID('df_user_sessions_level2_auth_config_has_changed', 'D')` guard, rewritten to say the
guard now exists for a database rolled back by an *older* release rather than by this one. No
statement moved. I wrote that escalation from the intent of the stage instead of from
`git diff --name-only`.

It does not change your answer, and it does not weaken the evidence: a comment cannot move a
catalog, and `go run ./cmd/schemadump` against all four live engines still rewrote all four goldens
byte-identically. But it does mean **(a) took the failure from four engines to one rather than to
zero**, and that one is legitimate: a text checker cannot tell a comment-only up edit from a real
one, and teaching it to strip comments would buy a false negative on exactly the rule that matters.

**So (b) is in force for the rest of the run**, for `mssqldb` alone:

```
where.sh migrations                 # FAIL snapshot (1 engine), exit 1
where.sh migrations --no-snapshot   # All rules pass, exit 0
```

Stages 4 and 5 will each record the flag and this decision in their tier section.

No production code changed in settling this. The branch is still `55d0d281`, CI green on all twelve
jobs. Stage 4 (`OpenDatabase`, the `migrate` subcommand, and the docs) is next.

leodip:
## Deferred decisions

**None.** All fourteen decisions in the agreement are `Decided`, and none was resolved by taking an
assumption forward that a later reader would have to unwind. Nothing here is waiting on you.

One was escalated to you mid-run and one correction came out of settling it, so both are recorded
below rather than left in a document that does not survive the run.

### Decision 14, the only escalation, and your answer

**The question.** Stage 3 fixed twenty broken `.down.sql` files. `where.sh migrations` then failed
its *snapshot* rule, which asks an engine whose `migrations/` directory changed to change its
`schema.golden` too. A down never runs in the up chain a golden records: `go run ./cmd/schemadump`
against all four live engines rewrote all four goldens byte-identically and `git status` reported no
change, so no golden edit could satisfy the rule. The check is branch-scoped, so it would have fired
at the close of every remaining stage — stages 4 and 5 could not land without an answer.

**Your answer**, on this thread: *"(a), with (b) as the interim so this run can finish."*

**(a), done.** The snapshot rule now keys on a changed `.up.sql` rather than on any file under
`migrations/`. An engine that changed only downs reports `OK snapshot  only down migrations changed
(<engines>); a down cannot move a golden` instead of falling through to the "no migration changed"
branch, so the two stay distinguishable. The protected property is untouched: an up that lands
without a refreshed golden still fails. That change is in the tooling repository (`ai@09539a6`),
not in this pull request — the checker is not this repository's file.

**(b), in force for the rest of the run.** Stages 4 and 5 and every review round gated on
`where.sh migrations --no-snapshot`, recorded in each tier section. The other nine rules ran
unmodified. Nothing was worked around: the checker's baseline is untouched and no `schema.golden`
was hand-edited.

**A correction the fix surfaced, worth having on the record.** The escalation told you stage 3
changed twenty `.down.sql` files and no `.up.sql`. The second half was wrong. Against the merge base
the branch changes 18 `.down.sql` files *and one `.up.sql`*:
`src/core/data/mssqldb/migrations/000031_otp_config_generation.up.sql`. Its entire diff is inside the
comment block above the `IF OBJECT_ID('df_user_sessions_level2_auth_config_has_changed', 'D')` guard,
rewritten to say the guard exists for a database rolled back by an *older* release rather than by
this one. No statement moved. The escalation had been written from the intent of the stage rather
than from `git diff --name-only`, and the sharpened rule is what surfaced it.

So (a) took the failure from four engines to one, and it is right that it still fires on that one: a
text checker cannot tell a comment-only up edit from a real one, and teaching it to strip comments
would buy a false negative on the rule that matters. The catalog demonstrably did not move — the
evidence is the four-engine regeneration recorded in stage 3.

**What reversing costs.** Reverting (a) puts the snapshot rule back to keying on any file under
`migrations/`, which means any future change that fixes a down migration is blocked by a rule it
cannot satisfy, exactly as this run was. Nothing in *this* pull request depends on it: no production
file, no migration, no golden and no baseline moved in settling it, and the branch was byte-identical
before and after.

leodip:

### Follow-ups to file

**None.** This run drafted no issue. Every real problem it found next to the work met §9's fold-in
test and was fixed inside the stage that found it; they are listed below so you can object to any of
them. There is nothing here for you to file and nothing left half-done behind a TODO.

One thing was found and deliberately *not* filed against this tracker, because it is not this
repository's defect: `where.sh mutate` cannot be used on this Windows host. It compares
`git rev-parse --show-toplevel` (which answers `C:/Code/goiabada` under Git for Windows) against
`realpath` (which answers `/c/Code/goiabada/...`), so every path is rejected as "outside the
repository". The round 2 mutations were run through a hand equivalent that backs the file up,
asserts the replacement matches exactly once, runs the tier, restores from the backup and compares
by hash. That belongs to the tooling repository.

Five, each riding on a surface the stage already touched, each with nothing to settle, each the
whole of its own fix, and none of them making §2 untrue.

**1. `AGENTS.md` was missing #301's Transactions bullet** — `ab59370b` (stage 2).
`CLAUDE.md` and `AGENTS.md` are meant to be byte-identical and differed in exactly one line: the
numbered "Transactions" bullet #301 added, which landed in `CLAUDE.md` alone. Copied verbatim while
the stage was already editing both files. `diff CLAUDE.md AGENTS.md` is now empty. Filing it would
have spent a triage entry, an agreement and a run on a one-line copy.

**2. The authserver unit tier never ran `./cmd`** — `ed1f7079` (stage 4).
`src/authserver/run-tests.sh` ran `./internal/... ./web`, with a comment reading "`./cmd` has no
tests" — true until stage 4 put the `migrate` subcommand's tests there. The stage's first mutation
came back `MUTATION SURVIVED` against a deliberately deleted rollback-floor check, and the log showed
`cmd/goiabada-authserver` absent from the tier entirely: eleven new test functions were passing
without being executed. CI's `Unit / authserver` job runs the same tier, so the gap was CI's too. The
tier now runs `./internal/... ./web ./cmd/...`. Filing this instead would have meant shipping a
command whose tests nothing runs.

**3. Every binary was built from `main.go` by filename** — `ed1f7079` (stage 4).
Six invocations named `./cmd/goiabada-authserver/main.go` or its adminconsole twin:
`src/authserver/build.sh`, `src/adminconsole/build.sh`, `src/build/Dockerfile-authserver`,
`src/build/Dockerfile-adminconsole`, `src/build/build-binaries.sh` (twice) and
`src/.vscode/launch.json`. `go build <file>` compiles that file alone, so `migrate.go` would not have
been in any binary. The authserver's four broke loudly the moment the file was added, which is the
good case; the shape is the problem, because it silently omits whatever is added beside `main.go`
next time. All six now name the package, with `-buildvcs=false` to keep exactly what ships today
(building from a file never stamped VCS either, and the release build injects version, date and
commit through `-ldflags`). **The adminconsole halves are the fold-in proper** — the authserver
halves were forced by the change; leaving one binary built by file and the other by package would
have been an inconsistency shipped on purpose.

**4. SQL Server's `schema_migrations` pre-create discarded its unlock** — `2410892a` (review round 1).
`src/core/data/mssqldb/db.go:ensureSchemaMigrationsTable` is the one place outside the runner that
takes the migration lock: SQL Server has no atomic check-then-create, so #293 wrapped it in
`sp_getapplock` at `LockOwner='Session'`. It read `defer func() { _ = eng.Unlock(ctx, conn) }()` and
then handed the connection back to the pool. The reviewer fault-injected only the public
`Engine.Unlock` against a live isolated SQL Server: `NewMigrator` reported success and a different
pool then timed out taking the same resource. That is precisely the leak goal 5 of this change exists
to end, in the one place still reproducing it — inside the goal rather than beside it. The unlock
error now joins the function's result, and the session that failed to release is handed
`driver.ErrBadConn` so `database/sql` discards it instead of lending a migration lock to the next
borrower. Pinned by `TestMigrationLock_ThePreCreateGivesTheResourceBack` and, after round 2,
`...WhenTheReleaseFails`.

**5. The dirty refusal invented the predecessor of a version it does not carry** — `69e84787`
(review round 2). Raised by the reviewer under its `## Follow-ups` heading, not as a finding.
`ErrDirty.Below` is the highest version the *source* carries beneath the dirty marker, which is an
endpoint only while the marker is a version that source knows — and `currentVersion` built the error
before asking whether it was. So a newer release that dies mid-migration leaves a marker an older
binary reads by naming the nearest version *it* carries, silently skipping every version in between.
Verified against real SQLite: a set carrying 1, 2, 4, 5 fails at 5 after 000004 committed; a binary
carrying only 1 and 2 refuses the row correctly and then advises "leaving version 000002 if its
statements did not apply". The truth is 000004, and an operator who records 000002 clean has a
database that re-runs 000003 and 000004 against a schema that already has them. `ErrDirty` now
carries `Carried`, and where it is false the message says a newer release wrote the row and only
that release can name the endpoints. The refusal itself is unchanged, and asserted so.

One.

`src/core/data/migrator/engine.go:49` — the SQLite migration mutex is process-wide rather than
per database file, so two SQLite migrations in one process run one after the other even when they
touch different files. **Revisit when** a caller migrates several SQLite databases concurrently and
the serialisation shows up as wall clock; keying the mutex by resolved file path is the next shape.

leodip:
## The review account

This run went under end-of-run review mode: no stage was reviewed on its own, and the whole branch
was read once after stage 5 landed. Three rounds ran at that gate, plus one static round against the
plan before any code existed. Six findings were raised across the code rounds, all six were confirmed
against the code before anything changed, all six were fixed, and none was contested or refuted. Round
3 came back clean.

This is the only independent judgement this code received, so it is pasted here in full rather than
summarised — the verdicts themselves live in a gitignored directory that the next round overwrites.

---

## Plan gate, round 1 (`type: plan-review`), 2026-09-06

Reviewer `openai-codex/gpt-6-astra` at effort `medium`, fresh session. Scope: the plan's five stages
against the sealed agreement, all three axes. Nothing had landed — the branch was one empty commit at
`639c3d66`. Verdict `static-review-exhausted`: three findings, all quality, all significant, all
`judgement: true` at `confidence: high`. No shipping defect was demonstrated, which is what a plan
gate with no code can conclude. The reviewer executed rather than argued, in the dev container against
real SQLite and the dev PostgreSQL, each claim carrying a control and a mutant.

|---|---|---|---|---|
| 1 | quality | significant | `OpenConnections == 0` is not the connection-release guarantee | Fixed in the plan, step 5 |
| 2 | quality | significant | Nothing in the plan proves the migration lock is ever acquired | Fixed in the plan, stage 2 |
| 3 | quality | significant | Decision 5's fail-closed version read and the bookkeeping transaction are unpinned | Fixed in the plan, steps 5 and 7 |
| — | quality | (note) | `ErrDirty` would name the wrong recovery version after a failed down | Promoted to a defect; fixed in the plan, steps 1, 5 and 7 |

**1. The release assertion rejects correct pooling.** `sqlitedb/db.go` sets `SetMaxOpenConns(1)` /
`SetMaxIdleConns(1)`, so a correctly closed `*sql.Conn` returns to the pool and leaves
`OpenConnections` at 1, idle. The planned assertion fails a *correct* runner, and the only ways to
pass it are to close or reconfigure a pool the runner does not own — which decision 8 rejected. The
plan now asserts `InUse == 0` after success and failure with the pool open, plus a following
operation proving the pool is still borrowable.

**2. The lock is never observed.** Pinning the three lock names as strings does not prove the runner
acquires, holds and releases the session lock, and every chain and catalog test runs serially, so an
adapter with acquisition omitted would pass the whole plan. That loses goal 2's "same lock resources"
in the one window it exists for. Section 5 had rejected a lock test because "two processes contending
is not observable from one test"; the premise was wrong, not the conclusion — contention needs two
*sessions*, and `database_create_test.go:holdCreationLock` already does this for #293's creation lock.
Stage 2 gained contention tests on MySQL, PostgreSQL and SQL Server, with acquisition and release each
mutated away.

**3. Decision 5's change is untested.** The case list tested a failing SQL migration but never a
failing version *read* or a failing bookkeeping *write*. Decision 5 exists to stop the library
swallowing every read error and answering nil version — the value that reruns the chain from 000001
against a populated database — and the reviewer showed both a query fault and a scan fault going
undetected by the driver being replaced. Step 5 gained an unreadable `schema_migrations` making
`Version`, `Plan` and `Up` fail without executing a file, and a rejected `INSERT` at each of the two
bookkeeping writes leaving the prior row and the schema unmoved.

**The note the run promoted to a defect.** The verdict set a "dirty-recovery wording problem" aside as
a follow-up. Read against section 1 it is a defect: decision 7's two legal end states are written from
the up step, and a down step from `N` to `N-1` marks `N-1` while running `N.down.sql`, so its "did not
apply" end state is `(N, clean)` — the version *above* the marker. An operator told to force
`marker-1` would land on `N-2`, a version the schema was never at. Correcting it states the facts
decision 7 chose rather than choosing different ones, so it landed in stage 1.

Round 2 was available and not taken: the reviewer returned `static-review-exhausted` and asked to
proceed to implementation, which is right with nothing on the branch to read.

---

## Final gate, round 1 — four findings, all fixed in `2410892a`

Reviewer `openai-codex/gpt-6-astra` at effort `xhigh`, session resumed, `MAX_ROUNDS=4`. Scope: the
whole branch against its base, 70 files, +3468/−676, over the five stage commits. Verdict `findings`:
four — **two significant and two minor, no blocking finding and nothing on the security axis**. All
three axes read `reviewed`. CI was green on all twelve jobs at the reviewed HEAD.

The reviewer earned its verdict by writing its own programs rather than re-running the tiers: scratch
overlays in the dev container, a live SQL Server fixture, a fault-injected unlock adapter, and two
built binaries under smoke test. Every finding was reproduced before it was acted on.

### Finding 1 — conformance, significant. Dirty recovery derived from arithmetic. **Fixed.**

`ErrDirty.Error` computed the "did not apply" end state as the marker minus one. **Confirmed:**
`errors.go` read `notApplied = e.Version - 1` on the up branch and `formatVersion(e.Version-1)` on the
read-back branch, and neither is a version. The four sets have gaps by construction, so an interrupted
up to 000005 on a database last at 000002 told the operator to record 000004 — a number no engine
carries. `checkCarried` then refuses it on the next start, which turns a dirty database into one that
will not open. Below the first migration it named 000000, which is not the unmigrated state either,
and a dirty nil marker (a rolled-back first migration, interrupted) produced "either migration none
(never migrated) was being applied".

Fixed as the reviewer forced it: `ErrDirty` gained `Below`, the source's own predecessor, set at both
construction sites from `m.src.prev`, and the nil marker is answered explicitly as the one case whose
direction is not ambiguous, since an up step never marks a nil version. Four new subtests: a failed up
over a gap, a failed first up, a read-back over a gap, and a read-back after a real failed first down
driven end to end rather than by inserting a row.

### Finding 2 — quality, significant. A benign sentinel masking an operational failure. **Fixed.**

`errors.Is(err, ErrNoChange)` in all four engine `Migrate()` wrappers and twice in `migrateTo`.
**Confirmed:** `Migrator.run` joins a failed unlock onto the operation's error *on purpose*, and the
runner's own `TestRun_AFailedUnlockIsReportedAndTheSessionIsDiscarded` already asserted that
`m.Migrate(2)` at head answers an error satisfying `ErrorIs(err, ErrNoChange)` **and**
`ErrorIs(err, unlockErr)` together. The wrappers read the first half and dropped the second — so a
server would log "no need to migrate the database" and start while holding a session-scoped lock that
blocks every other migrator on the database, indefinitely on PostgreSQL and SQL Server. The reviewer
reproduced it with a real SQLite runner and only the unlock adapter faulted.

Fixed with identity rather than `errors.Is`, and **resolved as a family rather than an instance**:
enumerating the family found a member the finding did not name, `errors.Is(err, migrator.ErrNilVersion)`
in `migrateVersion`, where `Version()` runs through `withConn` and joins a failed connection close the
same way. Both sentinels now have an identity helper — `migrator.IsNoChange` and
`migrator.IsNilVersion` — and all six production sites use them. `ErrLocked` and the two typed errors
are deliberately outside the rule: they are failures, so asking whether one is inside a wrapped error
is exactly what `errors.Is` is for. `src/core/data/benign_sentinel_lint_test.go` holds the whole tree
to it, parsing rather than grepping because the identifiers appear in the comments that explain the
rule.

### Finding 3 — conformance, minor. Catalog discovery unrestricted by schema. **Fixed.**

The eight `sys.default_constraints` queries stage 3 added to SQL Server's downs filtered on `t.name`
alone. **Confirmed** by reading all eight: the view spans the database while the statement they build
is schema-qualified, so a same-named table in another schema lost its defaults. The reviewer
demonstrated it live — eight unrelated defaults gone during a rollback from 000018 to 000003.

Each query is now pinned with `dc.parent_object_id = OBJECT_ID(...)` to the exact object its own
`DROP COLUMN` targets, spelled as that statement spells it.
`TestMigrationDowns_DropOnlyTheirOwnDefaultConstraints` builds the companion schema no other test in
the repository has — every test database holds one schema, which is why nothing could see this — and
asserts all eighteen companion defaults survive a full rollback.

### Finding 4 — conformance, minor. Documentation attributed startup migration to both binaries. **Fixed.**

**Confirmed:** the production `data.NewDatabase` fanout is exactly one caller,
`authserver/cmd/goiabada-authserver/main.go:94`; the admin console's main says in its own comment that
it keeps no database connection. "Each server applies whatever migrations its release carries" was
false of half the product. The opening paragraph now names the auth server and says what the admin
console does instead. `database.mdx`'s new section: opening paragraph 57 words to 80, whole section
260 words, 11 sentences plus the two-row table, against the agreement's "about eight to twelve short
sentences plus one two-row table". In budget.

### Round 1's `unchecked` list, answered

- **No old-release/new-release lock-contention run**, because the retired library source was gone from
  the local caches. **Dismissed** — it cannot be run and does not need to be: the contract is the
  resource *name*, and `TestAdvisoryLockID_MatchesGolangMigrate` pins all six values against
  `probe/lock_ids.go`'s output, captured from golang-migrate v4.19.1 before it was removed.
- **No release-to-release deployment rollback**, since the shipped floor equals this release's head.
  **Dismissed**, and it is decision 13 rather than a gap: the floor is vNext's schema version by
  design, so there is no supported target below it yet.
- **Unchanged security paths not re-audited.** **Dismissed** — none is in this change's fanout, and the
  reviewer says it did read the changed startup, SQL, logging and encryption boundaries.

### Round 1 tiers, on the fixed tree

|---|---|---|---|
| unit | `where.sh test --type modules` | n/a | 4632 pass, 0 fail, ~8s |
| data | `where.sh test --type data` | sqlite, mysql, postgres, mssql | 2173 pass, 0 fail, ~7m |
| integration | `where.sh test --type integration --db sqlite` | sqlite | 1333 pass, 0 fail, ~1m |
| migrations | `where.sh migrations --no-snapshot` (decision 14) | all four | all rules pass |
| anchors | `check-anchors.sh` | n/a | 23 of 23 |

### Round 1 mutations — six, all caught

|---|---|---|---|
| 1 | `migrator/errors.go` | the carried predecessor back to marker minus one | CAUGHT, `TestErrDirty` |
| 2 | `migrator/errors.go` | `IsNoChange`'s identity test back to `errors.Is` | CAUGHT, `TestIsNoChange` |
| 3 | `migrator/errors.go` | the nil-marker branch made unreachable with `if false` | CAUGHT, `TestErrDirty` |
| 4 | `mssqldb/migrations/000017_*.down.sql` | the object-identity filter back to the table name | CAUGHT, `TestMigrationDowns_DropOnlyTheirOwnDefaultConstraints` on mssql |
| 5 | `mssqldb/db.go` | the pre-create unlock replaced with a nil error, so it never runs | CAUGHT, `TestMigrationLock_ThePreCreateGivesTheResourceBack` on mssql |
| 6 | a temporary probe file under `migrator/` | a production file containing `errors.Is` on a benign sentinel | CAUGHT, `TestNoErrorsIsOnABenignMigratorSentinel` named the file and line |

Mutation 6 is a probe file rather than a `where.sh mutate` run, and the reason is worth recording. The
first attempt did use `where.sh mutate`, reverting `sqlitedb/db.go` to `errors.Is`, and it came back
CAUGHT — *by `TestGoSourcesAreGofmted`, not by the lint*, because the rewrite disturbed the file's
formatting. A mutation caught by the wrong test proves nothing about the test it was aimed at, so it
was discarded and replaced with a gofmt-clean probe file, run through `go test -run` in the dev
container and deleted afterwards.

### Round 1's fix went red on CI's Lint job, and why nothing local could see it

`2410892a` pushed green on every tier this run has locally and CI's **Lint** job failed on it:
staticcheck `QF1002` on `defaultLiteral` in the new SQL Server fixture — an untagged `switch` comparing
the same operand in every case. Fixed in `565f44c7`, a tagged switch, no behaviour change.

Recording it because the gap is structural rather than an oversight. `where.sh test` has five tiers and
none of them is lint: golangci-lint, `go vet` and `unparam` run per module in CI's Lint job and nowhere
local. So "unit, data and integration all green" is exactly compatible with a red branch. What closes
it locally, in under a minute, for anyone doing this again: run `gofmt -l .`, `go vet ./...`,
`golangci-lint run ./...` and `unparam ./...` in each of `src/core`, `src/authserver` and
`src/adminconsole` inside the dev container.

`production` did not move between `2410892a` and `565f44c7`, so every tier result recorded for round 1
still binds the code under review; only the fixture file moved.

---

## Final gate, round 2 — two findings, both fixed in `69e84787`

Reviewer `openai-codex/gpt-6-astra` at effort `xhigh`, session resumed, round 2 of 4, completed
2026-09-07T11:48:36Z. Verdict `findings`: two, both `quality`, both `significant`, both `judgement:
true` with a populated `forced_answer` — a finding the code settles rather than a question for you.
All three axes `reviewed`.

**No production defect was reported this round.** The reviewer's own note: *"Two significant quality
findings concern regression coverage, not remaining production failures."* Both are about guards
**round 1 added**, which is worth stating plainly: the review of the shipped behaviour came back
clean, and what remained was whether round 1's repairs are pinned against being undone. Both were
confirmed against the code before anything changed; neither was refuted.

### Finding 1 — quality, significant. The new lint was defeated by an import alias. **Fixed.**

`benign_sentinel_lint_test.go:findErrorsIsOnBenignSentinels` recognised its receiver by the literal
identifier `errors`:

```go
pkg, ok := fun.X.(*ast.Ident)
if !ok || pkg.Name != "errors" {
```

The file that defeats it is already in the tree: `src/core/data/mssqldb/db.go:8` imports
`goerrors "errors"` and calls `goerrors.Is` at line 347. So reverting *that* engine's wrapper to
`goerrors.Is(err, migrator.ErrNoChange)` would have restored the exact defect round 1 fixed, added no
import, changed no semantics, and been reported by nothing. The reviewer established it with two
fixtures differing only in the alias — the `errors` control caught, the `goerrors` one answered
`findings=[]`.

Fixed by resolving the receiver through the file's own import list rather than its spelling, covering
`errors` and `github.com/pkg/errors` (whose `Is` forwards to the standard one, and which 220 files
under `src/` import unaliased, binding the same identifier). Blank and dot imports are excluded with
the reason in a comment: a dot import binds `Is` with no receiver at all, and attributing a bare call
needs type information — a tier this test deliberately stays below. The synthetic half gains all three
cases the `forced_answer` named: the `goerrors` alias, an aliased `migrator` import, and a control
calling `Is` on a package that is not an errors package at all.

**Mutation, caught:** binding an aliased import to the default name — `idents[spec.Name.Name] = true`
to `idents["errors"] = true` — reproduces the defect while still compiling. The two `mssqldb/db.go`
findings vanish, 6 expected against 4 actual, test fails. (A first attempt reverting the condition left
`errorsIdents` declared and not used and did not compile, so it proved nothing and was discarded rather
than recorded as a pass.)

### Finding 2 — quality, significant. Round 1's lock-leak repair had no failing-path test. **Fixed.**

`TestMigrationLock_ThePreCreateGivesTheResourceBack` covers a release that *succeeds*, so the branch
round 1 added — joining a failed release onto the error and destroying the session with
`conn.Raw(func(any) error { return driver.ErrBadConn })` — never executes, and the test cannot tell the
repair from its absence. **Confirmed by reading, then measured: with the discard deleted, the test
passes.**

The closing account had recorded that the failure case needed a production injection seam. The reviewer
showed it does not, with a working program: wrap the real `go-mssqldb` connector, fail only
`EXEC sp_releaseapplock`, call the unchanged `NewMigrator`. That removed the stated reason for the
omission, which is why this is fixed rather than deferred.

Fixed with `TestMigrationLock_ThePreCreateGivesTheResourceBackWhenTheReleaseFails`. Acquisition, the
catalog check and the CREATE all run against real SQL Server; only the release's answer is replaced.
Both DDL outcomes run, because they leave the function through different returns: a successful CREATE
reaches the end and the deferred unlock supplies the error, a failing CREATE is already returning one
and the unlock's must *join* rather than replace it. Each asserts the injected sentinel survives to the
caller, `InUse` and `OpenConnections` are both zero, and a **separate pool** can then take the resource
— the assertion a connection count cannot make.

One departure from the reviewer's scratch, worth recording because it inverts a detail: the scratch and
the first draft wrapped `PrepareContext` only, and an `ExecContext` override was added on the
assumption the driver would take that path. It does not compile — `*mssql.Conn` implements no
`driver.ExecerContext`, so `database/sql` prepares every statement and the prepared statement is the
only path the release can take. The override is gone and a comment records the fact, including that a
future driver gaining `ExecerContext` makes this stop intercepting and the assertions fail rather than
quietly pass. The wrapper embeds the concrete `*mssql.Conn` rather than the `driver.Conn` interface
deliberately: embedding the interface would hide `ConnBeginTx`, `NamedValueChecker` and the rest, which
the parameterised lock statements need.

**Mutation, caught, and it reproduces the finding exactly.** Replacing the discard with
`_ = driver.ErrBadConn`: the pre-existing `...ThePreCreateGivesTheResourceBack` **passes**, and the new
`...WhenTheReleaseFails` **fails on both DDL outcomes** — at `migration_lock_test.go:241` ("the session
that failed to release the lock must be destroyed, not pooled for the next borrower") and at `:247`
(a separate session could not take the lock). That contrast is the whole of finding 2.

### Round 2's `unchecked` list, four entries, each answered

- *No retired-library / new-runner live lock-contention test.* **Dismissed** — the library left both
  binaries at stage 5 by design, so this is not re-runnable and no change here could make it so. The
  property rests on decision 4's formula and `probe/lock_ids.out`, recorded while the library was still
  in the tree, plus `TestMigrationLock_*` on three engines.
- *No supported release-to-release CLI rollback exercised; the shipped floor equals the head.*
  **Dismissed** — decision 13 behaving as decided. Every down is exercised regardless, by the
  chain-wide round trip on four engines.
- *No independent four-engine catalog reconstruction this round.* **Covered** — the data tier ran on
  all four engines this session, green, golden comparison included.
- *No new endpoint-level probes for unchanged security families.* **Dismissed** — this branch adds no
  HTTP surface; `migrate to` / `migrate version` are local administrative subcommands, and no route,
  handler or template changed.

### Round 2 tiers, at `69e84787`

|---|---|---|---|
| unit | `where.sh test --type modules` | n/a | 59 packages ok, 0 fail |
| data | `where.sh test --type data` | sqlite, mysql, postgres, mssql | green, 0 fail |
| integration | `where.sh test --type integration --db sqlite` | sqlite | green, 0 fail |
| lint | `gofmt`, `go vet`, `golangci-lint`, `unparam`, all three modules | n/a | `0 issues.` each |

Integration was run although stage 4 records it as not applicable, because this round moved production
code. The lint sweep was run because of what round 1 cost. All three mutations were verified restored
by hash before the commit.

---

## Final gate, round 3 — clean, and the gate closes on it

`deepseek/deepseek-v4-pro-0813` at effort `high`, session resumed, round 3 of a maximum of 4, started
2026-09-07T13:13:54Z. Scope the whole branch against `origin/main`, all three axes.

**Verdict: clean. Zero findings.** `axis_coverage` reports conformance, quality and security each
`reviewed`. Nothing was raised, so nothing was fixed, deferred or refuted. The reviewer's own note:
*"Zero findings after reading the whole branch end to end. Prior rounds' six findings are confirmed
fixed against the code, and the one production change since round 2 (`ErrDirty.Carried`) is correct and
mutation-pinned."*

**The tree it read is the tree that is here.** The reviewer records computing `all 0b2e7c463aad`
itself, byte-identical to the block in round 3's request. The working tree is clean and HEAD is
`69e84787`, so no mutation, no fixup and no stray edit sits between what was reviewed and what will
merge.

**What was verified here rather than taken on trust.** A zero-finding verdict is the one that most
deserves an independent look, so four load-bearing claims were re-derived:

- **Goal 1, the headline.** `grep` over all three `go.mod` files for `golang-migrate`, `lib/pq`,
  `golang-jwt/jwt/v4` and `go-autorest`: absent from all three. No Go source imports the library —
  `grep -rn '"github.com/golang-migrate' src --include=*.go` returns nothing — and all 41 surviving
  mentions of the name are prose inside comments, which is what the documentation sweep sanctioned.
  `probe/linked_modules_{before,after}.out` shows authserver 50 to 43 and adminconsole 50 to 41
  external modules.
- **Citation accuracy.** `check-anchors.sh` PASS, all 23 rows resolve to exactly one line in the named
  function.
- **The migrations gate.** `where.sh migrations --no-snapshot` exits 0 with every rule OK. The reviewer
  had left this one unchecked; it costs nothing, so it was run.
- **No endpoint surface.** `git diff --name-only origin/main...HEAD` matches no `handlers/`,
  `routes.go`, `web/template/` or `.html` path at all — the premise the reviewer's security dismissal
  rests on.

### Round 3's `unchecked` list, six entries, each answered

1. **No full four-engine data or integration re-run.** *Covered* — the check suite ran exactly that at
   this HEAD and is green. CI supersedes the reviewer's substitute with a result naming a commit anyone
   can reproduce from the link.
2. **No live lock-contention test between the retired library and the new runner.** *Dismissed* — not
   runnable by construction; there is no second implementation left in the tree to contend with.
3. **No independent four-engine catalog reconstruction.** *Covered*, by the same green suite: the chain
   round trip is a data-tier test, so `Tests / mysql`, `postgres`, `mssql` and `sqlite` *are* it
   running on four live engines.
4. **No RFC fetched.** *Dismissed* — section 1 reads "Nothing. No OAuth2, OIDC or IETF document reaches
   schema migration", so there is no text to ground against and fetching one would be theatre.
5. **`where.sh migrations` not re-run.** *Covered* — run here, exit 0, every rule OK.
6. **Standing security families not re-probed at endpoint level.** *Dismissed*, on the premise checked
   above rather than asserted: the diff adds no endpoint, route, handler or template.

### Why there is no round 4

Round 3 found nothing, and a verdict that reports no finding cannot buy the next round however the tree
moves afterwards. More to the point, the test is whether anything is left to raise, and nothing is:
three rounds read the whole branch, six findings were raised and all six confirmed and fixed, none was
contested, and the one production change made since round 2 read the code was itself reviewed clean in
this round.

### The check suite, which is where the four-engine claim lives

`where.sh ci` at `69e84787b0f5966b9aee5093345455d2d0df7da1`: **green**, twelve jobs, none failing.
`Tests / sqlite`, `Tests / mysql`, `Tests / postgres` and `Tests / mssql` each pass — that is the data
and integration tiers on all four engines — plus `Unit / core`, `Unit / authserver`,
`Unit / adminconsole`, `Unit / setup`, `Build validation`, `Lint`, `Vulnerabilities` and `Versions`.

Issue #280, opened 2026-08-27
go-chi/cors wraps one Handler call whose interesting half is already ours


leodip:
`go-chi/cors` is imported by one file, `core/middleware/middleware_cors.go`, for one call.

## What we use

```go
return cors.Handler(cors.Options{
    AllowOriginFunc: func(r *http.Request, origin string) bool { ... },
    ...
})
```

The `AllowOriginFunc` body is entirely ours: the per-path logic (always allow `/.well-known/openid-configuration` and `/certs`, index lookup on the `web_origins` UNIQUE constraint from migration 000034 for `/auth/token`, `/auth/logout` and `/userinfo`), including the reasoning about preflights carrying neither the `Authorization` header nor the form body per Fetch Standard §4.8.

What the library contributes is the mechanical half: preflight detection, `Vary` headers, and writing the `Access-Control-*` response headers.

## What replaces it

About 80 lines of middleware:

- On `OPTIONS` with an `Access-Control-Request-Method` header, treat it as a preflight: run the origin check, write `Access-Control-Allow-Origin`, `-Allow-Methods`, `-Allow-Headers`, `-Max-Age`, and respond 204 without calling the next handler.
- On any other request with an `Origin` header, run the origin check and write `Access-Control-Allow-Origin` (and `-Allow-Credentials` / `-Expose-Headers` if configured) before calling next.
- Always write `Vary: Origin` on any response whose `Access-Control-Allow-Origin` depends on the request origin, plus `Vary: Access-Control-Request-Method, Access-Control-Request-Headers` on preflights.

## Priority: low, and the risk is asymmetric

This one I would leave alone unless the goal is a zero-third-party middleware layer:

- It removes **one** module, with no transitive tail. `go-chi/cors` has no dependencies.
- CORS reads simple and is not. The `Vary` handling in particular is the difference between working correctly and a cache serving one origin's `Access-Control-Allow-Origin` to another origin, which is a security bug that no test we would think to write catches, because it lives in a cache we do not control.
- The interesting, Goiabada-specific half is already ours and stays ours either way.

If it does go ahead, the deliverable is the test suite as much as the middleware: preflight and simple-request paths, allowed and denied origins, the `Vary` headers on every branch, and the `/certs` and discovery-endpoint always-allow cases.

---

### Dependency reduction

[...]

Issue #279, opened 2026-08-27
pkg/errors is archived upstream and reaches 210 files; a local errs package keeps the %+v stack traces


leodip:
`github.com/pkg/errors` has been **archived upstream since December 2021**. It is our most widely imported third-party package.

## Scope

210 files import it. Counting only within those files:

| Symbol | Calls |
|---|---|
| `errors.Wrap` | 684 |
| `errors.New` | 625 |
| `errors.WithStack` | 554 |
| `errors.Is` | 10 |
| `errors.Wrapf` | 9 |
| `errors.Errorf` | 7 |
| `errors.As` | 1 |
| `errors.Cause` | 0 |

No file imports both stdlib `errors` and `pkg/errors`, so the migration is a straight import swap per file with no aliasing to untangle.

The stack traces are **actually consumed**: 59 `%+v` format sites across the repository. Dropping them silently would be a real loss of diagnostic quality, so a replacement has to keep them.

## What replaces it

A `core/errs` package, roughly 60 lines:

```go
type withStack struct {
    err   error
    stack []uintptr   // runtime.Callers(3, pcs[:])
}

func (e *withStack) Error() string  { return e.err.Error() }
func (e *withStack) Unwrap() error  { return e.err }
func (e *withStack) Format(s fmt.State, verb rune)  // %+v prints the frames
```

with three constructors mirroring what we use:

| Today | Replacement |
|---|---|
| `errors.New(msg)` | `errs.New(msg)` (stdlib `errors.New` + stack capture) |
| `errors.Wrap(err, msg)` | `errs.Wrap(err, msg)` (`fmt.Errorf("%s: %w", msg, err)` + stack, only if none present) |
| `errors.WithStack(err)` | `errs.WithStack(err)` |
| `errors.Wrapf` / `errors.Errorf` | `errs.Wrapf` / `fmt.Errorf` |
| `errors.Is` / `errors.As` | stdlib `errors.Is` / `errors.As` |

`Unwrap` is the important part: `pkg/errors` predates `%w`, so its wrapped errors work with `errors.Is`/`As` only through its own `Cause`. A replacement built on `fmt.Errorf("%w")` is strictly better here, and `customerrors.ErrNoAuthContext`, `AuthCodeReusedError` and friends keep matching through `errors.Is` as they do now.

Capture the stack **only when none is present** in the chain, or a wrapped error accumulates one stack per wrap and `%+v` becomes unreadable. That is the single behavioural detail to get right.

[...]

Issue #278, opened 2026-08-27
google/uuid supplies random identifiers only, never UUID semantics


leodip:
`google/uuid` is imported by 101 files, but nothing in Goiabada uses UUID *semantics*: no version inspection, no namespace UUIDs, no time-ordered v7, no MAC-derived v1. We use it to produce **random opaque identifiers**.

## What we use

| Symbol | Calls |
|---|---|
| `uuid.New` | 421 |
| `uuid.Parse` | 31 |
| `uuid.NewString` | 25 |
| `uuid.MustParse` | 19 |
| `uuid.UUID` (type) | 7 |
| `uuid.Nil` | 1 |

Most of the `Parse` / `MustParse` calls are in tests. The production use is "give me a fresh random identifier and format it canonically".

## What replaces it

A `core/uuidutil` package:

```go
func New() string        // random v4, canonical 8-4-4-4-12 hex
func Parse(s string) (string, error)  // validate + normalize to lowercase
```

Generation is `crypto/rand.Read(b[:])`, then set the version and variant bits:

```go
b[6] = (b[6] & 0x0f) | 0x40   // version 4
b[8] = (b[8] & 0x3f) | 0x80   // variant RFC 4122
```

then format. Around 30 lines including parse and validation.

**Do not silently swallow the `crypto/rand` error.** `google/uuid`'s `New` panics on a failed read; `NewRandom` returns the error. Whatever the replacement does, it must not return a zero or partially-random identifier, because these values become session identifiers, authorization codes and database keys.

## Priority: low, and worth saying why

This is on the list for completeness, not because it earns its place ahead of the others:

- `google/uuid` has **zero dependencies**. Removing it takes nothing out of the module graph beyond itself.
- It is actively maintained by Google and is one of the most widely deployed Go modules in existence.
- The change touches **101 files** for no transitive benefit.

The honest case for doing it is philosophical (one fewer third-party name in `go.mod`) rather than practical. It is mechanical and compiler-checked, so the risk is low, but it should be scheduled behind every other dependency ticket. If the crusade has to stop somewhere, this is a reasonable place.

---

### Dependency reduction

**Clear replacement, low risk:** #268 golang-migrate, #269 gorilla/sessions, #270 gorilla/securecookie, #271 unknwon/paginater, #272 gofakeit (test-only).

[...]

PR #304, opened 2026-09-07
refactor(sessionstore): gorilla/sessions and gorilla/securecookie come out of the session store

### LINKED ISSUE #269: gorilla/sessions now contributes only type declarations: after #266 the store is ours, the interface should be too
Before #266, `gorilla/sessions` supplied a working cookie store. After #266 it supplies **type declarations only**. The store is `core/sessionstore.ServerSideStore`, written here; the interface it satisfies is still imported.

## What is left of the dependency

Every symbol the repository uses from `gorilla/sessions`:

| Symbol | What it is |
|---|---|
| `sessions.Store` | A three-method interface: `Get`, `New`, `Save` |
| `sessions.Session` | A struct: `ID`, `Values map[interface{}]interface{}`, `Options`, `IsNew`, plus `Name()` |
| `sessions.NewSession` | A two-line constructor |
| `sessions.Options` | A struct mirroring `http.Cookie`'s attributes |
| `sessions.GetRegistry` | One call site |

`sessions.NewCookieStore` has **zero** non-test call sites. The parts of the library that did work are gone.

## Why remove it

- `sessions.Values` is `map[interface{}]interface{}`, a pre-generics shape we inherit and then work around at every read.
- `Store.Get` is documented to return a session *and* a non-nil error together, which is a subtle contract we have to keep honouring for a library we no longer take behaviour from.
- The `Registry` mechanism caches sessions per request in a context value. We use it in one place. It is behaviour we did not ask for sitting on the request path of every handler.

## What replaces it

A `sessionstore.Session` struct and a `sessionstore.Store` interface declared in `core/sessionstore`, alongside the store that already implements them. Optionally `Values map[string]any`, since every key we use is a string constant already.

## Scope

- 69 files import `gorilla/sessions`, 51 of them only to name `*sessions.Session` or `sessions.Store` in a signature. Mechanical, and the compiler finds all of it.
- `core/sessionstore/mocks/store_mock.go` needs regenerating.
- No behaviour change, no migration, no effect on live sessions. This is a rename.

Best done together with the `securecookie` removal, which touches the same package.

---

### Dependency reduction

### LINKED ISSUE #270: securecookie encodes a blob that no longer travels in a cookie: gob + AES-CTR + HMAC where AES-GCM would do
`gorilla/securecookie` exists to make a value safe to hand to a browser. Since #266 the value it encodes is a **database row**, and the cookie carries nothing but a 64-character identifier.

## What it does for us now

In `core/sessionstore/server_side_store.go`:

| Call | Encodes |
|---|---|
| `EncodeMulti(name, id, s.Codecs...)` | The session identifier, which does go in a cookie |
| `EncodeMulti(name, session.Values, s.DataCodecs...)` | **The session blob, which goes in the database** |
| `DecodeMulti` ×2 | The inverses |
| `CodecsFromPairs` | Builds both codec sets from the key pairs |

Plus `securecookie.GenerateRandomKey` in `core/data/database_seeder.go`, which is `crypto/rand.Read` with a length.

## Why it is the wrong tool for the blob

`securecookie.Encode` is: gob encode, AES-CTR encrypt with a random IV, HMAC-SHA256 over `name|timestamp|value`, base64, base64 again. That layering is what you build when the ciphertext has to survive a `Set-Cookie` header and you need encrypt-then-MAC out of two separate primitives.

For a `BLOB` column, `crypto/cipher` AES-GCM is one primitive, authenticated by construction, with no base64 and no manual MAC composition. It is less code and a stronger construction.

The awkwardness is already visible in the current implementation: two codec sets built from the same key pairs, differing only in a `MaxLength` we have to reach through a type assertion to set, plus a `MaxAge` backstop of a year that exists solely because securecookie insists on a freshness check the `expires_at` column already performs.

## What replaces it

In `core/sessionstore`, a versioned envelope:

```
[1-byte version][12-byte nonce][AES-256-GCM ciphertext]
```

with the plaintext being `encoding/gob` (or JSON) of `session.Values`, and the AAD being the session name. Around 80 lines including the multi-key rotation logic that `CodecsFromPairs` gives us today, which must be preserved: decrypt tries each key in turn, encrypt always uses the first.

`GenerateRandomKey(n)` becomes a four-line helper over `crypto/rand`.

## Risk: this invalidates live sessions

The blob format changes, so every session in the database becomes unreadable at upgrade. Options:

1. **Accept it.** Everyone signs in again once. Simplest, and defensible for a store that is one release old.
2. **Version the envelope and read both.** Keep a securecookie decode path for one release, write only the new format, drop the old path afterwards. Costs a release of carrying the dependency, buys a silent upgrade.

Option 2 is the safer default but only if #266 has actually shipped to deployments by then; if it has not, option 1 costs nobody anything.

Best done together with the `gorilla/sessions` removal, which touches the same package.

---

### Dependency reduction

leodip:
Closes #269. Closes #270.

`github.com/gorilla/sessions` and `github.com/gorilla/securecookie` are required by all three Go
modules and named by 72 Go files, and almost none of that is behaviour Goiabada wants from a
library. 57 of the 72 files write `sessions.Store` in a signature and nothing else. The one
concrete store constructed in production is Goiabada's own `ServerSideStore`, written by #266, and
`sessions.NewCookieStore` survives only in eight tests. What is left is `securecookie`'s codec:
gob, AES-CTR, base64, HMAC-SHA256 over `name|timestamp|value`, base64 again, a length knob and an
age knob, guarding a value that since #266 is an opaque 64-character identifier.

This change makes `core/sessionstore` the owner of both halves. It declares `Session`, `Store`,
`Options` and `NewSession` itself, and it seals its own values with one AEAD in a versioned
envelope.

## The rule this establishes

**The package that hands out a session owns the type, and the package that stores a value owns how
it is sealed.** Concretely:

- `Store` is `Get` and `Save`. `New` is the store's own method, not an obligation on every mock and
  stub, and its "never a nil session beside an error" rule is the store's own contract.
- Per-request memoisation stays and moves inside `sessionstore.Get`. It is load bearing, contrary
  to #269: the JWT middleware refreshes tokens into a session obtained by a second `Get` and reads
  the result from its own, and three middlewares each ask for the session on one request.
- One AEAD, XChaCha20-Poly1305, with a random 24-byte nonce, over an envelope of
  `version ++ nonce ++ ciphertext`, base64 text, with the session name bound as associated data.
  Two keys are derived with HKDF-SHA256 from the two keys a deployment already configures, one for
  the cookie and one for the stored blob, so a cookie's ciphertext cannot open as a blob.
- Session keys become rotatable with overlap, through four new optional `_PREVIOUS` variables. The
  store seals with the current pair and opens with the current pair then the previous one.

## For the release notes

**Everyone signs in once on the upgrade.** The session cookie and the stored session blob change
format: a value sealed by the old codec does not open, the store answers a fresh session as it does
for any undecodable value, and each user signs in again. Nothing is required of an operator to make
that happen, no data is lost and nothing needs clearing by hand. This adds no invalidation the same
upgrade did not already owe — #266, which introduced the server-side store, is in no release either,
so no released deployment holds a session in the format being replaced.

**No configuration change is required.** The two existing session keys per application keep their
names, their lengths and their required status. Both are still read: the encryption key is HKDF's
secret and the authentication key its salt, so neither becomes a variable the documentation calls
required while nothing consults it.

**Session keys can now be rotated without signing anyone out**, which was not possible before. Four
new optional variables, `GOIABADA_AUTHSERVER_SESSION_AUTHENTICATION_KEY_PREVIOUS`,
`GOIABADA_AUTHSERVER_SESSION_ENCRYPTION_KEY_PREVIOUS` and the `ADMINCONSOLE` twins. To rotate: set
the new pair as the current one, move the old pair to the `_PREVIOUS` variables, restart both
applications, wait, then remove the `_PREVIOUS` variables. The store seals with the current pair and
opens with the current pair then the previous one. Each application logs one line at startup when a
previous pair is configured, which is the only place the operator can confirm it was read. Half a
pair is refused at startup naming the missing variable, rather than accepted and then quietly
opening nothing.

**One caveat on when to remove the previous pair**, because the obvious rule is wrong. For signed-in
sessions, waiting out the maximum session lifetime is enough: their deadline is
`createdAt + maxLifetime` and the row is created at the moment they authenticate. A **sign-in still
in progress** has no such bound — an unauthenticated row gets a flat 30-minute window that every
touch renews, so there is no finite wait that guarantees none is outstanding. Removing the previous
pair may therefore restart a sign-in someone is part way through, which costs them a retry and
nothing else. Documented on the environment variables page under "Rotating the session keys".

### Stage 1: `core/sessionstore` owns `Session`, `Store` and `Options`

Landed in `5321e28a`. `github.com/gorilla/sessions` is now required by no module, imported by no
file, and none of its source was copied in. The codec is untouched, so `securecookie` still seals
both values; that is stage 2's, and leaving it made this stage compile and review on its own.

`core/sessionstore` declares for itself:

- **`Options`**, the six fields the tree sets. `MaxAge` below zero stays the deletion signal the
  two logout handlers write. `Partitioned` is dropped: nothing set it, and the store would have
  had to define its meaning for a `__Host-` cookie.
- **`Session`**, with `Values` as `map[string]any`. Every key in the tree is a string constant,
  and `Values` is the stored wire type, so it changes now while the format is changing anyway
  rather than later behind a compatibility path.
- **`Store`, as `Get` and `Save`.** `New` leaves the interface: nothing outside the store called
  it and no test stubbed it, so its "never a nil session" rule becomes the store's own to keep
  rather than an obligation on every implementer. The mock and the one hand-written test store
  lose a method nobody used.
- **A flash pair written rather than copied**: `SetFlash(key, value)` and
  `TakeFlash(key) (string, bool)`, built from what the 69 call sites actually ask for. Copying
  would have been redistribution: the licence's source condition would have followed those lines
  into this repository and its binary condition into every release artifact, and both would have
  outlived the `go.mod` entry — a permanent obligation created by the change that claimed to
  remove the dependency. Storing a `string` also retires the `gob.Register([]interface{}{})` that
  the library's package `init` performed on the stored shape's behalf, so this package registers
  nothing and neither does either `main.go`.
- **The per-request cache, inside `Get`.** Not an optimisation. The JWT middleware obtains the
  session, calls `refreshToken`, which obtains it *again* through a second `Get`, writes the
  refreshed tokens into that object and saves, and the outer function then reads them back out of
  the object it is still holding. Two objects and the outer function serves pre-refresh tokens.
  The `(session, error)` pair is memoised rather than just the session, so a lookup that failed
  refuses every middleware on the request identically.

Two things landed here because this is where they became reachable or wrong:

- **`encryption.RandomKey(n) ([]byte, error)`**, and the seeder's four key generations now fail
  `Seed` instead of hex-encoding a nil slice. A CSPRNG failure used to produce a bootstrap env
  file naming a session key of no bytes at all, and a deployment that came up and ran on it.
- **Comments that named the library for behaviour this repository now owns** were corrected in the
  same commit: `i18n/middleware.go`'s note about the registry caching the decoded session, the
  store's and the regenerator's doc comments, three test comments, and `backend_memory.go`'s file
  count, recounted from three to six. `i18n/middleware.go` is the one file this change's goal does
  not mention, which is why it is named here. The ten `gorilla/csrf` mentions are a different
  issue's and were left alone.

The 8 test cookie stores became a real store over `NewMemoryBackend` through a per-package helper,
which also drops `securecookie` from 8 of its 11 test importers. The mock was regenerated from the
updated `.mockery.yaml`, never edited.

**Tiers.** Unit green in all three modules: 59 packages, 4661 cases, 0 failures, with the gofmt
guard and `go vet` clean. Integration green on sqlite: 1333 cases, 0 failures, including the five
`TestBrowserSession_*` cases. Data and the migrations gate are not applicable — no query, no
schema and no migration is touched. The four-engine data and integration verdict is the check
suite's on this commit; it was `pending` when this entry was written.

**Mutations.** Six, one per behaviour this stage makes observable. Five caught first time. The
sixth survived, and it was a real defect in the new test rather than in the code: with the
change-password GET's consumption save made unreachable, the case still passed, because it built
each hop's cookies from the previous response — and a handler that saves nothing sets no cookie,
so the reload went out with no session and found no notice in a session it had never opened. The
helper is now an accumulating cookie jar, as a browser is, and the mutation is caught. That case
would otherwise have shipped green and testing nothing.

### Stage 2: one AEAD, one envelope, and `securecookie` leaves
Landed in `c6fa08d9`.

New `core/sessionstore/codec.go`. Two 32-byte keys are derived from the pair a deployment already
configures, with HKDF-SHA256 over the encryption key as secret and the authentication key as salt,
and a purpose-naming `info` string per key. Both configured secrets therefore keep a job, startup
validation and the seeder are untouched, and every existing environment file stays valid. Two keys
rather than one is what stops a value sealed for the backend from opening as a cookie, which the
two codec sets used to give for free.

The AEAD is XChaCha20-Poly1305. Its 24-byte nonce is long enough to draw at random with no
bookkeeping, where a 96-bit nonce would bound one key to 2^32 messages, and nothing in the tree
rotates a session key, so a per-key ceiling would be a per-deployment-lifetime ceiling. The
envelope is a version byte, the nonce, then the ciphertext and tag, base64 encoded once, with the
logical session name bound in as associated data so a value sealed under one name does not open
under another. What goes away is the second base64, the separate MAC, the timestamp, and the
length and age knobs that existed only because a library insisted on them.

The store now takes a current key pair and an optional previous one, seals with the current and
opens with the current then the previous. Both binaries pass `nil` for now; stage 3 supplies the
configuration. `NewServerSideStore` returns an error, so a deployment whose keys cannot build a
cipher fails at startup rather than at its first save. The size ceiling is the store's own, in one
helper both write paths go through, applied to the sealed length. `MiddlewareCookieReset` loses
the decode branch that has been unreachable since #266 and keeps the stale-cookie sweep, and the
`Get` that installs the per-request session cache.

**A cookie or blob written by the old codec does not open, and the visitor signs in once.** No
release carries the old format, so no deployment on a release pays anything for that.

**Both halves of the goal are closed here.** After `go mod tidy` in all three modules,
`grep -rn gorilla src/*/go.mod src/*/go.sum` and `grep -rl --include=*.go 'github.com/gorilla' src`
both return nothing. Nothing gorilla wrote was copied in either: the flash pair, the per-request
registry and the four types were written from this repository's own call sites in stage 1, which
is what keeps a BSD-3-Clause source and binary obligation out of every future release artifact.

Tiers: unit 4675 pass on all three modules, integration 1333 pass on sqlite. Data and migrations
not applicable — no query and no schema changed. Six mutations against the new mechanisms (the
collapse of the two derived keys, the version check, the minimum-length check, the name binding,
the previous-pair fallback, the size ceiling), all six caught.

Two notes worth keeping. The version byte is deliberately **outside** the associated data:
binding it would make the version check unobservable from outside the package, since removing the
check would then send a restamped envelope to an AEAD that refuses it for the tag instead, and
relabelling an envelope only ever turns it into a refusal anyway. And the cross-name case is new
rather than existing: the agreement pointed at a test that asserts the opposite property, about
the `__Host-` prefix, so the associated-data binding was unpinned until this stage.

Folded in, all three the same class of comment falsified by this change: `handler_authorize.go`
said the parked authorization error rides on a context "which securecookie encrypts and HMACs";
`core/api/requests.go` said the session endpoint's `data` field is a string because it is
securecookie output; and `backend_memory.go` said six test files drive the store for real and then
enumerated six, where it is eight — the two admin console flash tests added in stage 1 were not
counted by the change that added them.

### Stage 3: the previous pair, wired and documented

Landed in `9ab174e4`. Session keys can now be rotated. The store has taken a current pair and an
optional previous one since stage 2, and both binaries passed `nil` for the previous one because
nothing configured it; this is the configuration, which is what makes the rotation reachable from
a deployment rather than only from a test.

Four new optional variables, `GOIABADA_AUTHSERVER_SESSION_AUTHENTICATION_KEY_PREVIOUS`,
`GOIABADA_AUTHSERVER_SESSION_ENCRYPTION_KEY_PREVIOUS` and the `ADMINCONSOLE` twins, validated to
the same lengths as the current pair by both startup validators and accepted as a pair. Half a
pair opens nothing, so an operator who mistypes one variable name is refused at startup naming
the missing one, rather than told a rotation is in place while every session it was meant to keep
alive is turned away.

**The decoding is two tested helpers rather than two `if`s in `func main()`, and that is the
security-relevant part of the stage.** `nil` and a zero-value `KeyPair` are not the same thing
here: `hkdf.Key` accepts an empty secret and an empty salt and returns a valid 32 byte key,
`chacha20poly1305.NewX` accepts that key, and a value seals and opens under it. A binary that
built its previous pair out of two empty configuration values would therefore hand the store a
second, permanently valid opening key derived from nothing but two `info` constants that ship in
the source, and nothing would error or look wrong. `sessionstore.DecodeKeyPair` and
`sessionstore.DecodePreviousKeyPair` decide it once with a table on them, so each binary is a
straight pass-through with no branch left to trust, and `newSealer` now refuses a pair with an
empty key as well, so the state is unreachable from both sides. That guard is not a length rule
-- lengths stay in the startup validators -- so the two cannot disagree.

The current pair goes through a helper too, rather than keeping `hex.DecodeString` with a dropped
error beside a checked one; `encoding/hex` leaves both `main.go` files as a result. Each binary
logs one line when a previous pair is configured, which is the only place an operator performing
a rotation can confirm the pair was read.

Docs: four table rows, 27 words each against the 31-word `GOIABADA_AES_ENCRYPTION_KEY_PREVIOUS`
row they are modelled on, and a "Rotating the session keys" subsection of 91 words against its
93-word sibling in the data-at-rest section.

Tiers: unit 4708 pass, 0 fail, all three modules; integration 1333 pass, 0 fail, on sqlite. Data
and the migrations gate are not applicable -- no query, no schema, no migration. Six mutations
run against the new seams and all six caught, the first of them the one that returns a zero-value
pair where `nil` belongs.

## Tiers

Named by where each result comes from, because that is what makes the claim checkable rather than
trusted.

**The check suite, on the final commit `97477248` — green, all twelve jobs:**
[run 34154417628](https://github.com/leodip/goiabada/actions/runs/34154417628). That is the
four-engine data and integration verdict, `Tests / sqlite`, `Tests / mysql`, `Tests / postgres` and
`Tests / mssql`, plus the four unit jobs, `Build validation`, `Lint`, `Vulnerabilities` and
`Versions` — the last three of which no local tier runs at all. Under this run's tier discipline the
four-engine result is reported there and nowhere else, so nothing on this branch claims it from a
local run.

**Local**, with each row saying when it was measured rather than implying all of them were run at
once:

| Tier | Command | Engines | Result | Measured |
|---|---|---|---|---|
| unit, all three modules | `where.sh test --type modules` | n/a | 59 packages, **4905 cases pass, 0 fail** | at the finish |
| anchors | `check-anchors.sh` | n/a | PASS, all 49 rows resolve to one line in the named function | at the finish |
| integration | `where.sh test --type integration --db sqlite` | sqlite | **1333 pass, 0 fail**, 73s | at review round 1 |

The integration row was not re-run at the finish and does not need to be: the `integration-tests`
tree hash is byte-identical to the value that run was recorded against, so the claim is live rather
than stale. Re-running it would have re-measured the same tree. The unit tier was re-run because a
figure going into this body is worth measuring on the tree that ships.

That unit figure counts every `--- PASS` line the run emits, subtests included, from the command
named beside it. The per-stage figures above were each measured during their own stage and are not
line-for-line comparable with it — the counts moved with what each stage added and with the tier
variant it used, so treat 4905 as the shipping tree's number and each stage's as that stage's.

**Data and the migrations gate are not applicable**, and that is a statement about the diff rather
than a tier that was skipped: no query, no schema, no `schema.sql` and no `migrations/` directory is
touched anywhere on this branch. The `data-tests` tree hash is byte-identical to every value
recorded on the branch, and the four-engine catalog comparison is what the four `Tests /` jobs above
perform.

**Mutations: 22 run across the branch, all caught** — six in stage 1 (one of which survived first
time and exposed a real defect in the new test, described in the stage 1 entry above), six in stage
2, six in stage 3, and four at final review round 1 against the codec cases that round added. Every
one was proven gone from the tree by hash before its commit, not assumed.

**The tree at the finish is byte-identical, in every category, to the tree both reviewers read:**
`production 5a633a41aade`, `unit-tests 59456d014883`, `data-tests 07da31b1288e`,
`integration-tests b2a2b8bfd505`, `docs 7d44a35bf675`. So round 1's mutation matrix is live evidence
about this code rather than a claim about a tree three commits back, which by the end of a
multi-stage run is usually not true of anything.

## Deviations worth knowing

Everything below was recorded when it happened, not reconstructed here.

- **The previous-pair decoding is two exported, tested helpers rather than two `if`s in `func
  main()`**, which is the deviation with security behind it. `nil` and a zero-value `KeyPair` are
  not the same thing: `hkdf.Key` accepts an empty secret and salt and returns a valid 32-byte key,
  `chacha20poly1305.NewX` accepts it, and a value seals and opens under it. A binary building its
  previous pair from two empty configuration values would have handed the store a second,
  permanently valid opening key derived from nothing but two `info` constants that ship in the
  source — and nothing would have errored. `DecodeKeyPair` and `DecodePreviousKeyPair` decide it once
  with a table on them, and `newSealer` refuses an empty key as well, so the state is unreachable
  from both sides.
- **`NewServerSideStore` has 15 call sites, not the 9 the plan recorded.** The plan measured them on
  `main`, before its own stage 1 turned the 8 test cookie stores into real stores — so two rows of
  the plan were inconsistent with each other by construction. Absorbed rather than escalated,
  because the change that made the number stale is the one the plan specifies. Four of the sites are
  in test files `go build ./...` never compiles, which is why every check after that point used
  `go vet ./...` per module.
- **Six test helpers panic on the constructor's new error** rather than returning it, each with a
  comment saying the panic is unreachable and why dropping the error would not be. Threading it
  would have touched 65 call sites to carry an error that cannot occur on literal keys.
- **`cookieNamer` loses `CookieName`, with `physicalName` and `deletionCookie`.** Decision 11
  deletes the branch that was their only caller, and leaving them would fail CI's Lint job, since
  there is no `.golangci` config and `unused` is in golangci-lint's default set.
- **The current key pair goes through a helper too**, rather than leaving `authKey, _ :=
  hex.DecodeString(...)` beside a checked previous decode. The dropped error's failure mode was the
  same skeleton key the new guard exists to refuse. Both `main.go` files lose `encoding/hex`.
- **Anchors were re-swept twice.** Stage 2 falsified 15 of 53 rows, all in code it deleted or
  rewrote: 11 were repointed and 4 left the table because the code they cited is gone with nothing
  to repoint them at — one of which is the goal being reached rather than a loss. All 49 remaining
  rows pass.
- **Documentation stayed inside its stated size.** The "Rotating the session keys" paragraph is 93
  words against a 93-word sibling, and the four new table rows are 27 words each against the 31-word
  row they are modelled on. Round 1's correction cost 26 words, paid for by tightening the same
  paragraph rather than by exceeding the ceiling.

## Review

Reviewed by `openai-codex/gpt-6-astra` at effort `high` (final round 1) and
`deepseek/deepseek-v4-pro-0813` at effort `high` (final round 2), **2 rounds over the whole branch.
No stage was reviewed on its own** — under `LEO_REVIEW_MODE=end` the entire diff is read once at the
end, so these two rounds are the only independent reading this code received. Two plan-gate rounds
preceded any code.

|---|---|---|---|---|
| plan | 1 | 1 | 0 | fixed in the plan, then superseded by a better fix |
| plan | 2 | 2 | 0 | both fixed in the plan |
| final | 1 | 2 | 0 | both fixed, `97477248` |
| final | 2 | 0 | 0 | `clean` |

No finding was blocking, none was refuted, and nothing was deferred. Every one was verified against
the code — or, at plan round 1, against the licence text in the module cache — before being acted
on, and every one stood.

- **significant, conformance (plan 1):** a comment naming gorilla, its copyright holder and
  "BSD-3-Clause" does not discharge that licence over two copied function bodies. Correct, and its
  cause was removed rather than annotated: decision 4 was rewritten so the flash pair is written for
  this repository instead of copied, which is why no licence obligation reaches this tree or any
  future release artifact. A green build cannot check that, which is why it is stated in the goal.
- **significant, quality (plan 2):** the planned tests for the flash pair could not see the 69
  rewritten call sites. Demonstrated by mutating scratch copies of two real handlers — reversed
  arguments, a deleted consumption save, a `||` guard turned `&&` — all three of which left the
  planned tests and the packages' existing tests green. Answered with handler-level coverage over a
  real store, and three mutations proving it can fail.
- **minor, conformance (plan 2):** the both-directions decryption assertion round 1 asked for had
  only one public entry point, so it could be met only by testing a private function or exporting an
  API for a test. Answered at the public cookie opener, which is the direction that discriminates
  two derived keys from one; no API was exported.
- **significant, quality (final 1):** eleven mutations against the shipped tests, four surviving —
  the codec's tests could not see which HKDF input was consulted, that a *successful* nonce read is
  used, or that the blob's session name is bound. Not defects in the implementation; gaps in what
  the tests could observe. Fixed in `97477248` with three cases at the public store surface, each
  proven to catch its own mutation and, for the two key subtests, to pass under the other's.
- **minor, conformance (final 1):** the rotation paragraph claimed that waiting out the maximum
  session lifetime retires the old pair. False for a sign-in in progress, whose 30-minute window
  every touch renews. Fixed in `97477248` in the docs, in the store's field comment and in a test
  message — and deliberately *not* fixed by widening the wait, because a renewing window has no
  finite bound to wait out.

Round 2 read the branch cold with a different model, re-checked round 1's own fixes and reported the
three new cases as discriminating rather than tautological, then returned `clean`. No production
code changed, so no further round is funded — and a clean verdict is a conclusion, not a stall.

Full account with all the evidence, both mutation tables, the tier figures and every `unchecked`
entry answered: [the review comment](https://github.com/leodip/goiabada/pull/304#issuecomment-5575110542).

## The three closing comments

- **[Deferred decisions](https://github.com/leodip/goiabada/pull/304#issuecomment-5575087609)** —
  there are none. All 11 decisions are `Decided`, nothing waits on anybody. The comment lists the
  six the run chose itself, with what reversing each would cost, since those are the ones no human
  picked.
- **[Follow-ups, fold-ins and ceilings](https://github.com/leodip/goiabada/pull/304#issuecomment-5575098656)**
  — one drafted issue, not filed, on released ZIPs and images shipping no third-party licence
  notices; the nine comment corrections folded into the stages that falsified them, one of them in a
  file this change's goal does not name; and no ceilings.
- **[The review account](https://github.com/leodip/goiabada/pull/304#issuecomment-5575110542)** —
  all four rounds in full.

## Status

**Complete.** All three stages landed, in `5321e28a`, `c6fa08d9` and `9ab174e4`, with the review's
two accepted fixes in `97477248`.

No gorilla module is required, imported or vendored by this repository, and none of its source is
copied in. Both halves were verified directly rather than inferred: `gorilla/(sessions|securecookie)`
across every `.go` file under `src`, `gorilla` in all three `go.mod` and all three `go.sum` files,
and the copied-source markers `The Gorilla Authors`, `CodecsFromPairs`, `NewCookieStore` and
`GenerateRandomKey` all return nothing, as does the abandoned `map[interface{}]interface{}` value
type. The second half is the one a green build cannot check, which is why it is checked by grep and
stated here.

`core/sessionstore` owns its four types, its per-request cache and its flash pair; it seals its own
values with one AEAD in a versioned envelope with the session name bound in; and the session keys
are rotatable with overlap, with the page that documents them saying how and when it is safe to
remove the previous pair. The check suite is green on the final commit across four engines, and the
whole branch has been read twice by two different models with every finding fixed.


leodip:
## Deferred decisions: none

Section 3 of the agreement holds **11 decisions, all of them `Decided`**. None is open, none was
deferred with a stated assumption, and none was left for the reviewer or for you to settle. Neither
review gate raised a decision, and neither reopened one. So the list this comment exists to carry
is empty.

That is worth a comment rather than a silence, because an empty list and an unwritten one look
identical, and only one of them means there is nothing waiting on you.

### What you chose

Five decisions were yours: both issues land in one run (1), `Values` becomes `map[string]any` (2),
live sessions are invalidated once at upgrade (7), session keys rotate with overlap through
`_PREVIOUS` variables (10), and — after plan round 1 found that a comment naming BSD-3-Clause does
not discharge it — that the flash pair be **written for this repository rather than copied**, with
no gorilla licence obligation entering the tree at all (4).

### What the run chose, and what reversing each costs

These six are the ones no human picked, so they are the ones worth your eye. All are `Decided`, all
are implemented, and each reversal cost is stated so you can weigh it without reading the agreement
— which is gitignored and does not survive this run.

| # | Decision | Reversing it costs |
|---|---|---|
| 3 | Per-request memoisation **stays**, moved inside `sessionstore.Get` | Correctness, not performance. The JWT middleware refreshes tokens into a session obtained by a second `Get` and reads the result from the object it already holds; without one object per name per request the outer function serves pre-refresh tokens. #269 called the registry "behaviour we did not ask for"; grounding contradicted the issue and the agreement records that |
| 5 | `Store` is `Get` and `Save`; `New` leaves the interface | Re-adding a method no caller outside the store invokes and no test stubs, to every mock and hand-written stub |
| 6 | `Options` is an own six-field struct | Carrying options as an `http.Cookie` instead exposes `Name`, `Value` and `Expires` to handlers and puts the `MaxAge < 0` deletion signal beside a conflicting `Expires` |
| 8 | The AEAD is **XChaCha20-Poly1305**, random 24-byte nonce | AES-256-GCM via `cipher.NewGCMWithRandomNonce` (what #270 proposed) is standard-library-only but carries a documented 2^32-messages-per-key bound. Nothing rotates a session key today, so that is a per-deployment-lifetime ceiling — about 430 days at ten million saves a day |
| 9 | Both configured keys feed **HKDF-SHA256** — encryption key as secret, authentication key as salt, one derivation per purpose | Using the encryption key alone is smaller in code but turns a variable that `native-binaries.mdx`, `production-checklist.mdx` and `environment-variables.mdx` all call **required** into one nothing reads, and edits three pages. It would also collapse the two derived keys into one, so a cookie's ciphertext could open as a blob |
| 11 | The cookie-reset middleware's dead decode branch is deleted, with its two tests | Keeping it needs the store to **refuse** a bad cookie, reversing #266's deliberate fresh-session answer. The branch has been unreachable since #266 and was reached only by tests manufacturing `securecookie.MultiError` through a mock — and that type is what this change removes |

Decision 9 is the one with a test behind it rather than only an argument: final review round 1 found
the case that was supposed to pin it could not see which HKDF input was consulted, because
`hkdf.Key` derives a valid key from either input alone. Two subtests now vary one half of the pair
at a time and each fails under its own key's removal and passes under the other's. Details in the
review comment.

leodip:

### 1. Drafted issue, not filed — filing is yours

**Duplicate search re-run at the finish, 2026-09-07**, because the tracker may have moved while the
run was working: `gh issue list --state all --search "license"` returns nothing, `"notice
attribution"` nothing, and `"third-party"` returns only the unrelated dependency-reduction family
(#271, #278, #279, #280) plus security issues matching on other words. Not a duplicate.

**One note on provenance.** This was first drafted at plan review round 1, when this change still
copied two gorilla function bodies verbatim. You then required that nothing be copied, decision 4
was rewritten, and that deleted the half of this follow-up that was specific to #269 — there is no
copied source here, so there is no source-form condition to discharge and nothing for a
graph-walking notice generator to miss. What survives is the general defect, which predates #269
and is true with or without it. **This change makes it smaller, not larger:** it removes two
requirements from the graph.

**Title:** Released ZIPs and images ship no third-party licence notices
**Labels:** `enhancement`, `dependencies`, `go`

> Goiabada is MIT. Every dependency it links into a binary keeps its own licence, and several of
> those licences attach a condition to *binary* redistribution. Today nothing in the release
> carries any of them.
>
> **What ships.** `src/build/build-binaries.sh` zips exactly two files per platform,
> `goiabada-authserver` and `goiabada-adminconsole`, and nothing else — not the dependencies'
> licences, and not the repository's own `LICENSE`. `src/build/Dockerfile-authserver` and
> `Dockerfile-adminconsole` copy one binary onto `alpine:latest` with the same result. There is no
> `NOTICE` file, no `third_party/` directory and no `vendor/` directory anywhere in the tree.
>
> **Why that is a defect.** Several licences in the graph of the three modules require, on binary
> redistribution, that their notice, conditions and disclaimer be reproduced "in the documentation
> and/or other materials provided with the distribution". The published ZIPs and images are binary
> redistributions and reproduce none of them. The question is owed to every entry in the graph, and
> each has to be answered from its own licence rather than assumed.
>
> **What closing this takes.**
>
> - Decide the form: a generated `THIRD-PARTY-NOTICES` covering all three modules' graphs, or
>   vendored `LICENSE` files. A notice listing one dependency out of dozens is worse than none,
>   because it reads as complete.
> - Ship it: add it to the ZIP in `build-binaries.sh` and `COPY` it into the final stage of both
>   Dockerfiles. Shipping is the whole point — a notice file that stays in the repository does not
>   accompany the distribution.
> - Consider shipping Goiabada's own `LICENSE` in the ZIP too, which is absent for the same reason.
> - A graph walk finds required dependencies and nothing else. If a future change ever copies
>   third-party source into the tree rather than requiring it, that copy needs its own entry and no
>   tool will find it. #269 deliberately does not: it writes its own replacements.
>
> **Anchors:** `src/build/build-binaries.sh` (the `zip -v` line in `build_platform`);
> `src/build/Dockerfile-authserver` and `-adminconsole` (`FROM alpine:latest AS final`).
>
> Found while planning #269, and out of that change's scope: the obligation predates it, covers the
> whole graph rather than any one library, and #269 removes two requirements from it.

To file it, save the quoted body above to `notice-issue.md` and run:

```bash
gh issue create --title "Released ZIPs and images ship no third-party licence notices" \
  --label enhancement --label dependencies --label go \
  --body-file notice-issue.md
```

### 2. Folded in rather than filed

Section 2 of the agreement makes correcting comments this change falsifies its own business, in the
stage that falsifies them. It names three; the stages found nine. They are listed here because you
cannot object to a fold-in you were never shown, and because one of them is in a file the change's
goal mentions nowhere — which would otherwise read as an unexplained hunk in the diff.

Each fits on the fold-in test: it rides on a surface the stage already touched, in the same commit
as the change that falsified it; there was nothing to settle, because a comment describing a
dependency the file no longer has is wrong on its face; the comment edit is the whole of the fix;
and the goal stays true as written, because the goal is what asks for it. **No code changed in any
of them.**

| Where | What it said | Commit |
|---|---|---|
| `src/core/i18n/middleware.go` | "gorilla/sessions caches the decoded session on the request via the registry" — the registry is now this repository's, inside `sessionstore.Get`. **This is the file the goal does not mention.** | `5321e28a` |
| `src/core/sessionstore/server_side_store.go` | "It implements sessions.Store" → the package's own `Store` | `5321e28a` |
| `src/core/sessionstore/regenerator.go` | "sessions.Store has no such method" → "Store has no such method" | `5321e28a` |
| three test comments | two in `server_side_store_test.go` explaining the non-nil-session rule and `GetSurvivesAStorageFailure` by naming gorilla's registry; one in `routes_ratelimiter_test.go` calling a cookie "what gorilla wrote" | `5321e28a` |
| `src/authserver/internal/handlers/handler_authorize.go` | the parked authorization error rides on a context "which securecookie encrypts and HMACs" | `c6fa08d9` |
| `src/core/api/requests.go` | the session endpoint's `data` field is a string because it is securecookie output | `c6fa08d9` |
| `src/core/sessionstore/backend_memory.go` | said six test files drive the store for real and then enumerated six, where it is eight — the two admin console flash tests added in stage 1 were not counted by the change that added them | `c6fa08d9` |

The ten remaining `gorilla` mentions in the tree are the `gorilla/csrf` ones the goal puts out of
scope by name — three dead `gorilla.csrf.Token` lookups in `web/static/*.js` and comments recording
why code is shaped as it is after #155. They belong to their own issue and were left alone.

### 3. Ceilings left by this change

**None.**

```
git diff --unified=0 origin/main...HEAD | grep -n '^+.*ceiling:'
```

returns nothing. This change added no simplification carrying a known limit, so there is nothing
here a later reader could mistake for a final design.

leodip:
## The review account

This is the code review, pasted in full rather than pointed at, because the verdicts live in a
gitignored directory that the next run overwrites. Under `LEO_REVIEW_MODE=end` **no stage was
reviewed on its own**: the whole branch was read twice, at the end, by two different models. Two
plan-gate rounds preceded any code.

| Gate | Round | Reviewer | Findings | Severity | Outcome |
|---|---|---|---|---|---|
| plan | 1 | `openai-codex/gpt-6-astra` @ high | 1 | significant, conformance | fixed in the plan — and then superseded by a better fix |
| plan | 2 | `openai-codex/gpt-6-astra` @ high | 2 | significant quality; minor conformance | both fixed in the plan |
| final | 1 | `openai-codex/gpt-6-astra` @ high | 2 | significant quality; minor conformance | both fixed in `97477248` |
| final | 2 | `deepseek/deepseek-v4-pro-0813` @ high | 0 | — | `clean` |

Nothing was deferred. Nothing was refuted. No decision was raised by a review or reopened by one.
Every finding at every gate was verified against the code or the licence text first, and every one
of them stood.

---

### Plan gate, round 1 — the licence finding

Request `269-20260907T143321Z`, `type: plan-review`, reviewing `plan.md` against the agreement.
All three axes reviewed. Verdict `findings`: one, plus four `unchecked` entries.

**Finding 1, conformance, significant, `confidence: high`, `judgement: false` with a populated
`forced_answer`.** At that point the plan copied gorilla's `Flashes` and `AddFlash` verbatim and
added a comment naming gorilla, its copyright holder and "BSD-3-Clause". The reviewer's point was
that such a comment does not satisfy that licence, which requires the copyright notice, the list of
conditions and the disclaimer to be retained with copied source.

*Verified from the licence text, not from the finding.* Read out of the container's module cache at
`$(go env GOMODCACHE)/github.com/gorilla/sessions@v1.4.0/LICENSE`: first condition,
"Redistributions of source code must retain the above copyright notice, this list of conditions and
the following disclaimer", over `Copyright (c) 2024 The Gorilla Authors. All rights reserved.`
Also confirmed that the bodies were genuine copies rather than a resemblance — `Flashes` and
`AddFlash` are unchanged character for character under `map[string]any`, because `key` is already a
`string`. The finding stood.

**Fixed, and then fixed better the same day, before any code was written.** The first fix was to
reproduce the LICENSE in full above both methods. On reading the finding you required instead that
**no gorilla licence obligation enter this repository at all**, and decision 4 was rewritten: the
flash pair is written for this repository from the 69 call sites' one shape, not copied. Nothing is
redistributed, so no condition attaches, so the attribution block was deleted rather than corrected
— and `[]interface{}`, the `gob.Register([]interface{}{})` and a panicking type assertion went with
it.

That is worth stating plainly because a green build cannot check it: **a change can delete every
`require` line and still leave a copied body in the tree under the original licence**, which is a
worse position than the dependency was, since a dependency is declared, auditable and removable
while a copied body is none of those. The finding was right and its cause is now gone.

**The half of the `forced_answer` not applied**, and why. It also asked that the notice accompany
distributed binaries and images. Real, and separate — with nothing copied, gorilla raises it no
longer, but it is unmet today for every other linked dependency. It predates this change, and this
change *reduces* the graph by two entries. A gorilla-only notice beside dozens of equally linked
dependencies would mislead rather than comply, so it was drafted as a follow-up covering the whole
graph. It is in the follow-ups comment on this PR.

**Four `unchecked` entries, all answered.** Two were findings wearing an `unchecked` label and were
promoted into stage sketches: (2) that stage 2's purpose-separation case must assert rejection at
the point of decryption rather than through the downstream fresh session — the reviewer ran the
key-collapse mutation and the fresh-session assertions stayed green in *both* directions while the
direct assertion failed as required, so written the other way that case would have passed with
decision 9 deleted; and (3) that stage 3 owes a case on each wiring branch — executed at the gate,
`hkdf.Key` with an empty secret and salt returns `err=<nil>` and a usable key, so an absent previous
pair passed as a zero-value `KeyPair` rather than `nil` would give the store a second valid opening
key derived from public constants alone. (1) and (4) were dismissed with reasons: no code exists at
a plan gate, and (4) names areas the plan touches only through imports and signatures.

---

### Plan gate, round 2 — the 69 rewritten callers

Verdict `static-review-exhausted`; conformance, quality and security all reviewed. Two findings,
both marked forced by the code rather than a choice, both fixed in the plan.

**Finding 1, quality, significant.** The change rewrites 69 flash call sites across 35 files onto
the new pair, and the plan's tests for that pair exercised only the session object and the store.
The reviewer's point was that those tests cannot see the 69 rewritten callers, and it demonstrated
the gap by mutating scratch copies of two real handlers: reversing a `SetFlash` call's two
arguments, deleting the branch that saves a consumed flash, and changing a two-key save guard from
`||` to `&&`. **All three left the planned session and store tests green, and left the affected
packages' existing tests green.**

*Confirmed independently before fixing.* Both arguments are `string`, so a reversed call compiles.
The save that persists a consumption is the caller's `if`, not the primitive's: 34 of the 35 read
sites guard it on one key and one guards it on two. And nothing existing covered it — the
`adminuserhandlers` package had no test functions at all beyond its `TestMain`, and the only admin
console tests mentioning a flash key are six lines in `rendertest/render_test.go` that pass
`savedSuccessfully: false` in as a literal template binding, observing the template and never the
handler that computes it.

*Fix.* Handler-level coverage added to stage 1, over a real session store on the in-memory backend
with the template render stubbed to capture its bindings: `HandleAccountChangePasswordPost` then
`Get` then `Get`, asserting the notice appears once and is gone on the next load; and
`HandleAdminUserDetailsGet` with neither, either and both of its two keys set, asserting they are
consumed independently. Three mutations were added to the stage to prove that coverage can fail.

**Finding 2, conformance, minor.** Round 1 had required the two-derived-keys rejection be asserted
in both directions at the point of decryption. Finding 2 was that only one of those directions has
a public entry point — the design exports a cookie opener and keeps the blob opener private, and
the agreement's test rules reject reaching into the private sealer. So the requirement as written
could be met only by testing a private function or by exporting an API for one test.

*Fix.* The rejection is asserted at the public cookie opener, which is the direction that actually
discriminates two derived keys from one on round 1's own mutation evidence. The other direction is
kept as a public outcome — the store answers a fresh session — and is claimed to show only that.
**No API was exported for a test.**

**Two bookkeeping items** fixed in passing: a plan step said "no type assertion" where the decision
says "no type assertion that can panic", and the anchor checker failed 12 of 53 rows naming stage 1
code that did not exist yet, which is expected at a plan gate.

**No round 3.** Both findings were requirements whose proof is a mutation run against real code,
which no further reading of a plan can produce.

---

### Final review, round 1 — the whole branch, `openai-codex/gpt-6-astra` @ high

Request `269-20260907T182443Z`, `gate: final`, round 1 of max 4, fresh session, started
2026-09-07T18:33:35Z. Scope was `origin/main...HEAD`, all three axes; `axis_coverage` reports
conformance, quality and security all `reviewed`. Verdict `findings`: two, both `judgement: true` at
`confidence: high`. Both verified and both stood. **No implementation security defect was
confirmed.** The reviewer recorded the tree byte-identical after its run, which `tree-hash.sh`
agreed with.

#### Finding 1, quality, significant — codec coverage: unpinned key contribution, nonce freshness, blob domain separation

*Family:* "Unpinned codec key contribution, nonce freshness, and blob domain separation". The
reviewer ran **eleven Go-overlay mutations against the existing tests: seven caught, four
survived** — `authentication-key-unused`, `encryption-key-unused`,
`fixed-nonce-after-successful-read`, `data-name-not-bound`. These were scratch mutations of the new
guarantees, **not defects in the shipped implementation**: the code was correct, the tests could not
see it.

*Verified against the code, and the claim is exact:*

- `TestServerSideStore_CookieFromOtherKeysGivesAFreshSession` keys its foreign store with `"0000…"`
  and `"1111…"`, changing **both** halves at once. `newSealer` feeds the encryption key to HKDF as
  the secret and the authentication key as the salt, and `hkdf.Key` derives a valid 32-byte key from
  either input alone — so a derivation that stopped reading one input still produces different keys
  for two pairs differing in both. The case could not see which input is consulted. That matters
  beyond tidiness: decision 9 exists precisely so the authentication key keeps a job rather than
  becoming a variable three docs pages call **required** while nothing reads it.
- The two name-binding cases both go through the *cookie* key. The blob's name is bound at two
  separate sites — `sealSessionData` passing `session.Name()` and `New` passing `name` into
  `openWithEither` — so removing it on the blob path alone changed nothing either case observes.
- The CSPRNG cases assert a *failed* nonce read is reported. Nothing asserted a *successful* read is
  used, so a nonce zeroed after the read survived.

**Fixed** in `97477248`, three cases in `server_side_store_test.go`, all at the public store surface
over the existing `fakeBackend`, with no copied codec:

- `TestServerSideStore_BothConfiguredKeysSealTheValue`, two subtests varying one half of the pair at
  a time, each requiring that **neither** the cookie **nor** the blob opens. The blob half is
  reached by transplanting a foreign blob behind a cookie the store under test did write, because
  presenting the foreign cookie stops at the cookie and would let a store that opened the blob read
  as a pass.
- `TestServerSideStore_EverySealUsesAFreshNonce`, saving an unchanged session twice and requiring
  both the cookie and the single-entry blob ciphertexts to differ while the identifier and the
  values stay equal.
- `TestServerSideStore_ABlobDoesNotOpenUnderAnotherSessionName`, transplanting a blob sealed under
  `constants.AdminConsoleSessionName` behind an auth server cookie and requiring an empty fresh
  session.

#### Finding 2, conformance, minor — the rotation paragraph's expiry claim is false for pending sign-ins

*Family:* "Rotation-retirement guidance assumes one absolute lifetime for both session phases".
`environment-variables.mdx` claimed that once the maximum session lifetime has passed, nothing
sealed under the old pair is still alive.

**Verified against the code, and false as written.** `ExpiresAt` returns `now.Add(PreAuthLifetime)`
when `authenticated` is false, consulting neither the idle timeout nor the maximum lifetime, and
`dbBackend.expiryFor` recomputes it from `now` on every `Update` and `Touch`. So an unauthenticated
row has a flat 30-minute window that **each touch renews**, independent of the configured maximum:
a sign-in somebody keeps returning to has no bound in that maximum at all. The claim holds for
authenticated sessions, whose deadline is `createdAt + maxLifetime` and whose row is created at the
moment they authenticate, because the identifier rotates at sign-in.

This overclaim originated in the agreement's own section 1 and decision 10. The agreement is sealed
and gitignored, so the correction is carried in the code and the docs rather than by editing it.

**Fixed** in `97477248`, taking the forced answer as written. The decided no-reseal-on-open design
is untouched, and the wait is **not** replaced with `max(maximum lifetime, thirty minutes)`, because
a window that every touch renews has no finite bound to wait out. The paragraph now scopes the
guarantee to signed-in sessions and says removing the pair may restart a sign-in in progress. Two
further edits in the same commit correct the same overclaim where it is durable: the store's
`previous` field comment now carries the pre-authentication nuance and why waiting longer does not
fix it, and a rotation test message that repeated "costs nothing" was narrowed to what its case
actually shows.

#### Mutations, round 1

Four, one per previously surviving member of finding 1's family, each proving the new case catches
what nothing caught before. The first three through `where.sh mutate`, which restores from a trap
and verifies by hash; the fourth by hand under a trap, because it needs two sites.

| Mutation | Site | Replacement | Result |
|---|---|---|---|
| `authentication-key-unused` | `codec.go`, `newSealer` | `pair.AuthenticationKey = nil` before both `hkdf.Key` calls | **CAUGHT** by `BothConfiguredKeysSealTheValue/only_the_authentication_key_differs` |
| `encryption-key-unused` | `codec.go`, `newSealer` | `pair.EncryptionKey = nil` before both `hkdf.Key` calls | **CAUGHT** by `BothConfiguredKeysSealTheValue/only_the_encryption_key_differs` |
| `fixed-nonce-after-successful-read` | `codec.go`, `seal` | `nonce = make([]byte, nonceBytes)` after the successful `io.ReadFull` | **CAUGHT** by `EverySealUsesAFreshNonce` |
| `data-name-not-bound` | `server_side_store.go`, `sealSessionData` and `New` | both blob AAD sites pass the empty string instead of the session name | **CAUGHT** by `ABlobDoesNotOpenUnderAnotherSessionName`, and by nothing else in the package |

Two properties of that table are worth keeping. The two key subtests **discriminate**: each fails
under its own key's removal and passes under the other's, which is what establishes that each HKDF
input is independently observable rather than that the pair as a whole is. And `data-name-not-bound`
was caught by exactly one test with the rest of the package still green, because binding the empty
string on both sides preserves the round trip — the clean demonstration that the new case pins the
binding itself rather than the ability to reload a session.

**Every mutation proven gone before the commit.** `production` returned to `d5ad22195cda` after each
of the four, checked with `tree-hash.sh --split` rather than assumed, with `git status --porcelain`
showing only the intended test file between them.

#### Documentation words, round 1

| Page, section | Stated size | Before | After |
|---|---|---|---|
| `environment-variables.mdx`, "Rotating the session keys" | no longer than its 93-word sibling | 91 | 93 |

At the ceiling, not over it. The correction plus the new caveat cost 26 words, so the paragraph was
tightened elsewhere to pay for them: the opening sentence drops two explicitly named variables for
"each application's current pair" and "the four current variables", which loses nothing because all
eight appear in tables on the same page. No deviation to record.

#### Tiers after the round 1 fixes

| Tier | Engines | Result |
|---|---|---|
| unit, all modules | n/a | PASS, no failures |
| integration | sqlite | **1333 pass, 0 fail**, 73s |
| anchors | n/a | PASS, all 49 rows resolve |
| docs / gofmt | n/a | PASS / clean |

#### The `unchecked` array, four entries, each answered

1. **No interactive browser pass, no live two-binary key-rotation restart.** *Dismissed, with the
   reason already in the agreement.* The integration harness reads config once at start and cannot
   restart the server under a second pair, which is why seam 1 owns the rotation table at the store.
   The reviewer substituted real-store handler tests, HTTP integration and store-level rotation
   probes — the same substitution the agreement made deliberately.
2. **Expiry reproduction on SQLite only; no four-engine catalog comparison.** *Covered.* No schema
   changed and no migration was touched; `data-tests` is byte-identical to its recorded value; CI
   was green on all twelve jobs. The behaviour the finding turns on is in `ExpiresAt`, pure Go over
   a `time.Time` with no engine-specific path.
3. **Unchanged security surface not re-audited** — refresh-token and OTP atomicity, revocation
   fanout, PKCE, redirect and audience validation, permission inventories, injection, rate limits,
   enumeration timing, audit delivery. *Dismissed.* Outside the goal, untouched by this branch, and
   the diff is the boundary. The consumers that did move are covered by the unit and sqlite
   integration tiers.
4. **No cryptographic-library implementation audit, no long-running concurrency or fuzz campaign.**
   *Dismissed.* The primitives are `crypto/hkdf` from the standard library and
   `x/crypto/chacha20poly1305`, both already in the module graph and neither vendored nor modified
   here; auditing them is upstream's. What this change owes is that it calls them correctly, and the
   reviewer's stated execution covered exactly that.

No entry needed filing, so round 1 added no follow-up.

---

### Final review, round 2 — read cold by a different model, `deepseek/deepseek-v4-pro-0813` @ high

Request `269-20260907T191058Z`, round 2 of max 4, fresh session, started 2026-09-07T19:31:08Z. Same
scope, all three axes `reviewed`.

**Verdict: `clean`. Zero findings.** Nothing to fix, nothing to refute, nothing deferred.

**A different model from round 1, which is the point of the round.** Round 1's two accepted fixes in
`97477248` became part of what needed reading — new test code and a corrected docs paragraph that
nothing independent had yet seen.

**What the clean verdict rests on**, recorded rather than summarised, because a clean verdict is
only worth its stated execution. The reviewer read the full `origin/main...HEAD` diff across
`session.go`, `codec.go`, `server_side_store.go`, `regenerator.go`, `middleware_cookie_reset.go`,
`config.go`, both `main.go` files, `database_seeder.go`, `encryption/random.go`, the store, codec and
session test files, both flash handler tests, `config_test.go`, the docs diff and the mock diff. It
ran `check-anchors.sh` (PASS, 49 rows), `tree-hash.sh`, `where.sh ci 97477248`,
`where.sh test --type core` (full core unit suite PASS), and targeted `go test` runs over the codec,
memoisation and flash cases in `./sessionstore/...` plus the flash tests in `accounthandlers` and
`adminuserhandlers`, all passing.

It also re-read round 1's two fixes specifically and reports the three new cases as
**discriminating rather than tautological**, and the rotation paragraph as now correctly scoped to
signed-in sessions. That independent check is what round 2 existed to obtain.

**Verified first-hand at ingest, not taken on the reviewer's word.** The change's goal is that
nothing of gorilla survives, and that is one grep rather than an inference:
`gorilla/(sessions|securecookie)` across `--include=*.go` in `src`, `gorilla` in every `go.mod` and
`go.sum`, the copied-source markers `The Gorilla Authors`, `CodecsFromPairs`, `NewCookieStore` and
`GenerateRandomKey`, and the abandoned `map[interface{}]interface{}` value type. **All five return
nothing** — no import, no module requirement, no copied source, no legacy type.

**Tree integrity, and why round 1's mutation proofs are still live.** At ingest, on a clean tree at
`97477248`:

```
production 5a633a41aade   unit-tests 59456d014883   data-tests 07da31b1288e
integration-tests b2a2b8bfd505   docs 7d44a35bf675   all 6a13eafdd3b3
```

Byte-identical to what round 1 recorded after its fixes and to what the reviewer reported for its
own run. Two things follow. No mutation was left behind in the tree by either round, **checked by
hash rather than assumed** — a real hazard, since a session that dies between writing a mutation and
restoring it leaves a corrupted production file that looks like an ordinary dirty tree. And round 1's
four-mutation matrix, the evidence this record leans on hardest, was run against exactly this
`production` value.

**No mutations were run in round 2**: zero findings, no new code, and the matrix establishing the
codec's boundary is still live against this precise tree. Inventing a mutation at a clean gate would
add no evidence the existing table does not already carry.

**Three `unchecked` entries, each answered.** (1) No browser pass or live rotation restart —
*dismissed* on the ground the agreement settled at seal time, and worth recording rather than waving
through: this is the second independent reviewer to raise the gap and then answer it with the same
substitution, unprompted, which says the trade is legible from outside the run rather than a hole
the agreement talked itself past. (2) No local data tier or four-engine schema comparison —
*covered, by the strongest available evidence*: no migration directory or `schema.sql` was touched,
`data-tests` is byte-identical at `07da31b1288e`, and the four-engine comparison the entry asks for
is exactly what `Tests / sqlite|mysql|postgres|mssql` perform, all four green at `97477248`. (3)
Unchanged security surface not re-audited — *dismissed*, outside the goal and untouched.

**Why the review closes here.** No production code changed, which under the budget funds no further
round — and none is wanted. **A clean verdict is a conclusion, not a stall.** The stall rule exists
to stop rounds that churn while finding nothing worth fixing; this reviewer found nothing at all and
said so, having read the branch cold with a different model and re-checked round 1's own output.

---

### The four-engine result

Round 1 closed leaving one evidentiary item open: the data and integration tiers on mysql, postgres
and mssql were the check suite's to report, and at the time it had not run. It has.

**Green at `97477248837e9e795a1b4949daba102c53c4008b`, all twelve jobs:**
https://github.com/leodip/goiabada/actions/runs/34154417628

`Tests / sqlite`, `Tests / mysql`, `Tests / postgres`, `Tests / mssql`, `Unit / core`,
`Unit / authserver`, `Unit / adminconsole`, `Unit / setup`, `Build validation`, `Lint`,
`Vulnerabilities`, `Versions`.

Nothing on this branch is claimed green on evidence that has not concluded, or measured against a
tree that no longer exists.

Issue #136, opened 2026-08-03
Flaky: TestSearchUsersPaginated collides with accumulated users in the never-reset shared databases


leodip:
## Symptom

`TestSearchUsersPaginated` (`src/authserver/tests/data/user_test.go`) fails intermittently on the
server-backed engines:

```
user_test.go:185: Expected 1 search result, got 2
user_test.go:188: Expected total of 1, got 2
```

Seen on MySQL during a full `./run-tests.sh` run. The same test then passed twice on re-run with
identical code, and passed in isolation. It is not caused by any recent change; it is a latent
condition that gets more likely with every test run.

## Mechanism

Three facts combine. None is a bug on its own.

**1. The test asserts a globally unique match.** It creates five users and searches for
`users[0].Username`, requiring exactly one result:

```go
users[i] = createTestUser(t)   // x5
searchResults, total, err := database.SearchUsersPaginated(nil, users[0].Username, 1, 10)
if len(searchResults) != 1 { ... }
if total != 1 { ... }
```

**2. The search is a substring match across six columns.** `commondb.SearchUsersPaginated` builds:

```go
selectBuilder.Or(
    selectBuilder.Like("subject",     "%"+query+"%"),
    selectBuilder.Like("username",    "%"+query+"%"),
    selectBuilder.Like("given_name",  "%"+query+"%"),
    selectBuilder.Like("middle_name", "%"+query+"%"),
    selectBuilder.Like("family_name", "%"+query+"%"),
    selectBuilder.Like("email",       "%"+query+"%"),
)
```

So the assertion holds only if the freshly generated username appears nowhere else in the table, in
any of six columns, as a substring.

**3. The shared databases are never reset, and the generated username space is small.**
`createTestUser` uses `gofakeit.Username()`, which emits short values like `Fay5454` or
`Parker8278`: a name from a fixed pool plus a four-digit number. `run-tests.sh` deletes only the
SQLite files (`rm -f /tmp/goiabada*.db*`) and contains no `DROP DATABASE` or `CREATE DATABASE`, so
the MySQL, PostgreSQL and SQL Server databases accumulate users across every run, forever.

## Evidence

Measured directly against the running dev containers:

| Engine | Users in `goiabada_data` | Distinct usernames | Duplicates already present |
|---|---|---|---|
| MySQL | 2781 | 2779 | `Parker8278`, `Weber9487` |
| PostgreSQL | 2420 | 2419 | one |

[...]

PR #305, opened 2026-09-07
refactor(adminconsole): replace unknwon/paginater with internal/pagination

### LINKED ISSUE #271: unknwon/paginater is unmaintained since March 2020 and supplies five fields to one template partial
`github.com/unknwon/paginater` has one release, `v0.0.0-20200328080006`, and no commits since **March 2020**. It supplies five fields to one template partial.

## The entire surface we use

Five call sites, all identical:

```go
p := paginater.New(total, pageSize, pageInt, 5)
```

- `adminconsole/internal/handlers/adminuserhandlers/handler_admin_users.go`
- `adminconsole/internal/handlers/admingrouphandlers/handler_admin_group_members.go`
- `adminconsole/internal/handlers/adminresourcehandlers/handler_admin_resource_groups_with_permission.go`
- `adminconsole/internal/handlers/adminresourcehandlers/handler_admin_resource_users_with_permission.go`
- `adminconsole/internal/handlers/adminsettingshandlers/handler_admin_settings_audit_logs.go`

`adminconsole/web/template/partials/paginator.html` consumes exactly:

| Field | Used for |
|---|---|
| `HasPrevious`, `Previous` | The back arrow, enabled or disabled |
| `HasNext`, `Next` | The forward arrow, enabled or disabled |
| `Pages` | The number buttons |
| `.IsCurrent` on a page | The active button |
| `.Num` on a page | The label, with `-1` meaning "render an ellipsis" |

Nothing else in either module touches the library.

## What replaces it

An `adminconsole/internal/pagination` package with the same shape, so the partial does not change:

```go
type Page struct {
    Num       int  // -1 means ellipsis
    IsCurrent bool
}

type Paginator struct {
    HasPrevious bool
    Previous    int
    HasNext     bool
    Next        int
    Pages       []Page
}

func New(total, pageSize, current, numPages int) *Paginator
```

The only real logic is the sliding window that decides which page numbers to show and where the `-1` ellipsis sentinels go. Around 60 lines, and it is the kind of arithmetic that deserves a table test anyway, which we do not have today because the behaviour belongs to a third party.

## Why bother

It is not big and it is not heavy. It is **unmaintained**, and it is the one dependency in the tree where the replacement is unambiguous, self-contained and covered by a handful of assertions. Low risk, and it takes an abandoned module off the supply chain.

---

### Dependency reduction

leodip:
Closes #271.

`github.com/unknwon/paginater` is unmaintained (one pseudo-version, dated 2020) and is required by
`src/adminconsole/go.mod` alone, where five admin console handlers use it to compute the page bar
that `partials/paginator.html` renders. Its own test file drags four more modules into the graph
(`gopherjs`, `jtolds/gls`, `smartystreets/assertions`, `smartystreets/goconvey`) that serve nothing
else.

This change replaces it with `src/adminconsole/internal/pagination`, a small package written for
this repository from a recorded table of the library's behaviour rather than ported from its
source, so the tree carries no third-party code and no licence notice. The partial does not change:
Go templates resolve a field and a niladic method alike, so the new type exposes plain fields under
the names the partial already reads.

**One visible rule changes.** The library showed "..." even when it hid exactly one page, and the
dots took the slot that page could have used: 6 pages on page 1 rendered `1 2 3 4 5 ...`, with page
6 reachable only through the forward arrow. Now a single hidden page is shown by number, at either
end, and the dots only ever stand for two or more pages.

**Scope grew after the seal.** The two follow-ups this run had left unfiled — the five lists
disagreeing about `?page=`, and the overflowing page that panicked the groups-with-permission page
— were folded into this branch at the user's request rather than filed as issues. They are stages 3
and 4 below. Chasing the second one out of the admin console led into `core/data`, where the same
overflow was returning the wrong rows on three of the four engines; that is stage 4, and it is the
only part of this PR outside `src/adminconsole`.

## Status

**Complete, ready for review.** Four stages landed. The first two, and the review's fix commit, were
reviewed in two rounds over the whole branch and the check suite is green on `c80ef2e3` — all twelve
checks, including the data and integration tiers on mysql, postgres, mssql and sqlite
([run](https://github.com/leodip/goiabada/actions/runs/34175829416)).

Stages 3 and 4 (`c7ee0ef8`, `6243567b`) are the two folded-in follow-ups. They have not been through
the reviewer; they were verified locally against all four engines and the check suite is running on
`6243567b` ([run](https://github.com/leodip/goiabada/actions/runs/34217154801)).

### Stage 1: `adminconsole/internal/pagination`, and its table

`f799ea7b`. A new package, `src/adminconsole/internal/pagination`, and the table that holds it to
the behaviour it replaces.

`New(total, pageSize, current, numPages)` returns a `*Paginator` carrying `HasPrevious`,
`Previous`, `HasNext`, `Next` and `Pages []Page`, with `Page{Num, IsCurrent}` and `Num == -1`
meaning an ellipsis. They are plain fields where the library exposed niladic methods; a Go template
resolves the two alike, which is why `web/template/partials/paginator.html` does not change. It was
written from the library's recorded behaviour, not ported from its source, so nothing here carries
third-party code or an Apache-2.0 notice.

Behaviour is parity with the library, including the clamps on an out-of-range `?page=`, except for
the ellipsis rule described above.

The test embeds all 363 recorded rows verbatim: every page count from 0 to 12 with every current
page from before the first to two past the last, the exact multiples, the three clamps, the narrow
windows at `n` 0 to 4, and the audit log's page size of 20. Each row is compared as a whole
formatted line, so a failure prints the recorded bar beside the one produced. 73 of those rows are
the ellipsis rule reversing the library's answer deliberately, and a second test pins the row
counts so a trimmed table cannot pass quietly.

Tiers: unit green (1411 pass), data green on mysql, postgres, mssql and sqlite (1627 pass).
Integration is not applicable at this stage, since nothing imports the package yet. Four mutations
were run against the table and all four were caught: the single-page arm at the leading end, the
trailing-edge window slide, the current-page clamp, and the `HasNext` boundary.

One note worth carrying: the first data run aborted on postgres in `TestSearchUsersPaginated`,
which is the known flake #136 in the never-reset shared test databases. It cannot be this stage's,
as nothing imports the new package. The re-run was green on all four engines.

### Stage 2: the swap, the tidy, the render case

`ed4991ca`. The five call sites moved onto the new package, the module left the graph, and one real
page now renders through the unchanged partial.

Each of the five handlers, in `adminuserhandlers`, `admingrouphandlers`, `adminresourcehandlers`
(two) and `adminsettingshandlers`, changed one import line and one call: `paginater.New(...)` became
`pagination.New(...)` with all four arguments untouched. Nothing else in those functions moved and
`bind["paginator"]` keeps its key, so the seven `{{template "paginator"}}` includes across five
pages, the partial itself and `addUrlParam` are all byte-identical.

`go mod tidy` then removed the `require` line and 15 `go.sum` lines: the hash pairs for
`paginater` and the four test-only modules it dragged in, plus five stale `/go.mod`-only hashes for
2019-era `golang.org/x/*` pseudo-versions that were in the graph only because the library's own
`go.mod` pinned them. Every module those five names still needs keeps its current version.
`grep -rn unknwon src/` is empty, which is this change's first goal. `src/core` and
`src/authserver`'s module files did not move.

`TestRender_AdminUsersPaginator`, the tenth case in `internal/rendertest`, is the template hop: it
renders `/admin_users.html` through the real renderer, the real embedded partials and pt-BR over a
real `adminuserhandlers.PageResult` and `pagination.New(73, 10, 4, 5)`. That is 8 pages on page 4,
so the bar reads `1 2 3 [4] 5 6 ...` and exercises both ends at once. It asserts the page 1 link
that the library would have replaced with dots, exactly one ellipsis, `?page=3` and `?page=5` twice
each because each is both a number and an arrow target, page 4 as the active button with no link of
its own, and that `page=-1` appears nowhere: the sentinel must never reach a URL.

Tiers: unit green (1412 pass, the one new case), data green on mysql, postgres, mssql and sqlite
(1627 pass), integration green on sqlite (1012 pass). The three remaining engines' data and
integration runs are the check suite's, at `ed4991ca`. Three mutations were run against the new
render case and all three were caught: the trailing ellipsis arm, the page-1 arm that carries the
ellipsis rule, and the back arrow's target.

Two deviations from the plan, both recorded in full in the run's log. The predicted `go.sum`
removal was 10 lines and the actual was 15, for the reason above, which is the same removal
reaching further rather than an unaccounted module. And integration, which the stage's early sketch
called not applicable, was run: that answer belonged to stage 1, when nothing imported the package,
and these five handlers are on live routes.

The `?page=` parsing was deliberately left alone. Two of the five handlers clamp bad input to 1 and
three answer 500, which the agreement named as out of scope: it is a behaviour with a choice in it,
and this change does not touch a line of it.

### The review's fix commit

`c80ef2e3`. Three files, all from round 1 of the whole-branch review: `pagination.go` computes the
page count as quotient-then-remainder, the way the library did, so a `total` near `math.MaxInt` no
longer overflows the ceiling expression into a one-page bar; `pagination_test.go` gains
`TestNew_HugeTotalDoesNotOverflow` and reorders its formatter's arms into the partial's order, so a
sentinel wrongly marked current can no longer print as honest dots; and `render_test.go` drops its
one `github.com/google/uuid` import, which had made the module untidy, rather than promoting that
module in `go.mod`. All 363 table rows are unchanged by both fixes. Details and the mutation
evidence are in the review comment below.

### Stage 3: one answer for `?page=` in the admin console

`c7ee0ef8`. The first of the two follow-ups. The five paginated lists read `?page=` three different
ways — the user list and the audit log viewer clamped anything unusable to page 1, while group
members and both "with permission" lists answered 500 to `page=abc` and to `page=0` — and none of
the five clamped a page past the end, so `?page=99` over three pages rendered an empty table under a
bar highlighting page 3. The bar clamped; the query did not.

`pagination.ParsePage` and `pagination.ClampPage` are now the one answer and all five call them.
Unusable input is page 1 everywhere. A page past the end is the last page: the handler learns the
total only by asking, so it clamps on the way back and asks again, once, and the rows and the bar
then name the same page. `ClampPage` counts pages through the same `pageCount` as `New`, factored
out of it here, so the page the handler settles on and the page the bar highlights cannot drift
apart. A sixth reader turned up while doing it — the users-with-permission *add* page, which only
carries the value into a return link and answered 500 to a page the list beside it rendered — and it
parses the same way now.

The second follow-up, the panic, lives here too:
`handler_admin_resource_groups_with_permission.go` holds the only page offset the console computes
for itself, validated as `>= 1` alone, so `?page=9223372036854775807` wrapped `(pageInt-1)*pageSize`
negative; the two guards below it bounded only from above and never saw it, and
`allGroups[start:end]` panicked before anything rendered. The clamp now bounds that offset and the
guards stay as the bound on the slice itself.

Three decisions were the user's, taken before any code moved: clamp bad input to 1 rather than
refuse it, clamp past-the-end in the handler with a second query rather than redirect, and put the
shared parse in `internal/pagination` beside the bar it has to agree with.

1,589 lines of tests across six files: tables for both functions, a sweep holding the clamp and the
bar to the same page for every total to 200 at both page sizes, the offset arithmetic, then all six
handlers asserted on what they ask the API for, what they settle on, and whether the rows, the page
number and the bar agree. The panic is its own sweep, over the whole overflowing range against six
list sizes. Eight mutations, all eight caught — including the original code restored, which fails
with the reported `panic: runtime error: slice bounds out of range [:-10]`.

### Stage 4: the same overflow in `core/data`, which was worse than reported

`6243567b`. Stage 3's bound stops a nonsense page leaving the console, but the same page reaches the
auth server admin API directly, and there it becomes a SQL `OFFSET`. Running the mutation against
all four engines showed the negative offset does **not** simply fail:

| Path | Engines | What actually happened |
|---|---|---|
| Six reads via `sqlbuilder` | sqlite, mysql, postgres | **No error — the first page's rows came back**, labelled as the page asked for. `SelectBuilder.Offset` drops the clause when its argument is negative ([select.go:316](https://github.com/huandu/go-sqlbuilder/blob/v1.43.0/select.go#L316)), so the query ran with no offset at all. |
| The SQL Server audit log override | mssql | Refused outright — *"The offset specified in a OFFSET clause may not be negative"* — which the API turns into a 500. That one formats the number into the SQL text and so has no such guard. |

So the common case was not a crash but silently wrong data, and the two halves fail so differently
that fixing the one you notice leaves the other in place. `commondb.PageOffset` is now the one place
the offset is computed: it normalises both inputs and saturates the product at the largest multiple
of `pageSize` an int can hold, which is past the end of any table that will exist, so the read
returns the empty page every other page past the end returns.

**Seven** call sites, not six — the seventh is the `mssqldb` audit log override, found by grepping
for the SQL text rather than for `.Offset(`. `TestNoHandRolledPageOffset` now holds `core/data` to
the rule by parsing it, following the tree's existing lint-test idiom (real tree, plus a synthetic
tree proving the checker matches the shape and not the spelling).

The half that cannot be tested in Go is whether a nine-quintillion-row `OFFSET` is a number the
engines accept, so `TestPaginatedReads_AnOverflowingPageIsAnEmptyPage` runs all six readers in the
data tier on each of the four. It asserts no error **and** no rows: an error-only assertion passes on
three engines while they hand back page 1, and a rows-only assertion misses the engine that refuses.
Removing the saturation fails the data tier on all four.

The console's own bound stays. It should not send a page it knows is nonsense, and the two guards sit
on opposite sides of an HTTP call.

One thing deliberately left: the API still echoes the requested page back — `"page": 9223372036854775807`
beside an empty list and a correct `total`. That is truthful about the request, and the console
clamps for display, so the contract is unchanged.

## Verification

**Local, in the dev container:** unit and data on all four engines at both stages, integration on
sqlite at stage 2, and `where.sh test --type adminconsole` again after the review's fix commit.
`gofmt -l src/` clean, `GOPROXY=off go mod tidy -diff` exits 0, `check-anchors.sh` resolves all 11
anchors in the agreement's section 0.

**The check suite's, and only its:** the data and integration tiers on mysql, postgres and mssql,
plus Lint, Vulnerabilities, Versions and Build validation. Green on all twelve at `c80ef2e3`,
[run 34175829416](https://github.com/leodip/goiabada/actions/runs/34175829416).

**Stages 3 and 4, local:** the three module tiers, the full data tier on mysql, postgres, mssql and
sqlite, integration on sqlite, `gofmt -l src/` clean, `go vet` clean, `golangci-lint run` 0 issues on
all three modules, `GOPROXY=off go mod tidy -diff` exits 0. The check suite is running on
`6243567b`.

**Mutation testing:** seven mutations run by the run (four at stage 1, three at stage 2), all
caught, plus seven run independently by the reviewer, one of which survived and became finding 3.
Stage 3 added eight more, all caught. Stage 4 added one — the saturation removed — run against all
four engines and caught on all four.

**Deviations from the agreement:** two, both mechanical and both recorded above — the `go.sum`
removal reaching 15 lines rather than the predicted 10, and integration being run at stage 2 where
an early sketch had called it not applicable. No design decision changed during the run and no
decision was deferred.

### Review

round at the plan gate. No stage was reviewed on its own (`LEO_REVIEW_MODE=end`).

|---|---|---|---|
| plan 1 | 0 | 0 | clean |
| 1 | 3 | 0 | all confirmed, all fixed, `c80ef2e3` |
| 2 | 0 | 0 | clean |

- **significant, quality:** the table test's formatter checked `Num == -1` before `IsCurrent`, while
  the partial checks `IsCurrent` first, so a sentinel wrongly marked current would have matched the
  table and rendered as an active `-1` button. Proved by a mutation that survived. Fixed in
  `c80ef2e3`: the formatter's three arms are now the partial's three renderings, in the same order.
- **minor, conformance:** `(total + pageSize - 1) / pageSize` overflows near `math.MaxInt`, where the
  library's divide-then-increment did not — an undeclared second departure from the contract. Fixed
  in `c80ef2e3`, with a test that the old expression fails.
- **nit, quality:** a new test import left the module untidy. Fixed in `c80ef2e3` by removing the
  import rather than promoting the module, since taking modules out of this graph is the point.

Full account, with every mutation and its result, in
[this comment](https://github.com/leodip/goiabada/pull/305#issuecomment-5577672519).

## The three closing comments

- [Deferred decisions](https://github.com/leodip/goiabada/pull/305#issuecomment-5577653348) — none;
  all four of the agreement's decisions were settled at the seal.
- [Follow-ups](https://github.com/leodip/goiabada/pull/305#issuecomment-5577656561) — two, both
  about `?page=` handling. **Superseded:** neither was filed as an issue. At the user's request both
  were folded into this branch instead, as stages 3 and 4 above, and the second turned out to reach
  further than the report described. No `ceiling:` marker was added.
- [The review account](https://github.com/leodip/goiabada/pull/305#issuecomment-5577672519) — every
  round, every finding, and what answered it.

## For the release notes

The admin console's page bar no longer collapses a single hidden page into "...": where exactly one
page falls outside the run of numbers, at either end, that page's number is shown in the slot the
dots would have taken, so the ellipsis now always stands for two or more pages.

The five paginated admin console lists now agree about the `page` in a URL. A value that is not a
usable page number shows the first page rather than, on three of the five, an error screen; a page
past the end shows the last page with its rows, rather than an empty table under a page bar
highlighting a full one.

Fixed: a `page` large enough to overflow could crash the resource groups-with-permission page, and,
through the admin API, could make any paginated endpoint return the first page of results labelled
as the page that was requested — or fail outright on SQL Server. Every paginated read now treats a
page past the end as an empty page.


leodip:
## Deferred decisions

**None.** No decision was deferred, and none was left to the reader to settle after the fact.

All four decisions in section 3 of the agreement were made and closed at the seal, with the user in
the loop for the first one and the run raising the other three:

| # | Question | Answer taken |
|---|---|---|
| 1 | Where the replacement lives and what shape it has | `src/adminconsole/internal/pagination`, a `Paginator` with five exported fields, raised by the user |
| 2 | Whether the window size stays a parameter of `New` | It stays: `New(total, pageSize, current, numPages)`, callers pass 5 |
| 3 | Exact parity with the library, or fix a quirk while the file is open | Parity, with one departure: dots never stand for a single page |
| 4 | Reimplement from the recorded table, or port the library's code under its Apache-2.0 notice | Reimplement, so the tree carries no third-party code and no notice |

Neither stage nor either review round raised a new one, and nothing was escalated to the PR thread
during the run. Decision 3 is the only one with a user-visible consequence, and reversing it costs
one `case` arm at each end of `New` plus the 73 rows in the table test that record it — the rows are
marked, so the reversal is mechanical rather than a re-derivation.

leodip:
## Follow-ups

Two, both left unfiled for the user to file or discard. Neither was folded in. The duplicate search
was re-run at the closing pass, after the work: `gh issue list --state all` for `"invalid page"`,
`"paginater OR paginator OR pagination"` and `"page offset OR pagination panic OR slice bounds"`
return nothing about the admin console's page arithmetic — the paginator search returns only #271
and its ten sibling dependency issues.

### 1. The five paginated admin lists parse `?page=` three different ways, and none clamps a page past the end

Found during grounding, before any code moved. Users and the audit log viewer turn a malformed or
non-positive `page` into 1; group members and both "with permission" lists answer 500 to `page=abc`
and to `page=0`. All five keep an out-of-range page for the query, so `?page=99` over 3 pages
renders an empty list under a bar highlighting page 3 — the bar clamps, the handler does not.
Verified at all five `pagination.New` call sites.

Out of scope for #271 because the fix is a behaviour with a choice in it (clamp or refuse) spread
across five handlers, and section 2 named it out of scope in as many words.

```bash
gh issue create \
  --title "Five admin console lists parse ?page= three different ways, and none clamps a page past the end" \
  --body "Five admin console lists take a \`page\` query parameter and disagree about bad values: two clamp anything unparseable or below 1 to page 1, three answer 500. None clamps a page beyond the last, so \`?page=99\` shows an empty list under a page bar that highlights the last page.

Pick one behaviour (clamping to \`[1, last]\` is the friendlier one) and apply it in all five handlers, with a table test on the shared parse.

The five: \`handler_admin_users.go\`, \`handler_admin_group_members.go\`, \`handler_admin_resource_groups_with_permission.go\`, \`handler_admin_resource_users_with_permission.go\`, \`handler_admin_settings_audit_logs.go\`.

Found while running #271, which replaced the paginator library underneath these handlers without touching a line of their \`?page=\` parsing."
```

### 2. Bug: a `?page=` large enough to overflow panics the resource groups page

Raised by the reviewer at the final gate. In
`handler_admin_resource_groups_with_permission.go:133`, the branch taken when a resource has no
permissions paginates in the handler: `start := (pageInt - 1) * pageSize`. `pageInt` is validated
only as `>= 1`, so `?page=9223372036854775807` makes that product wrap negative. The two guards
below it only bound from above, so `allGroups[start:end]` runs with negative bounds and panics
before anything renders. It needs an authenticated administrator holding `authserver:manage`, and
is reachable only on a resource with no permissions defined — so the impact is a failed request,
not a bypass.

This is the only client-side page offset in the admin console:
`grep -rn 'pageInt - 1' src/adminconsole/ --include=*.go` returns this one line.

Not folded in, although #271 had that file open and changed one line in it. Section 2 put the
`?page=` parsing out of scope explicitly; a bare `if start < 0` would render an empty list under a
bar highlighting the last page, which is the quirk decision 3 chose to leave; and closing it
properly means deciding one answer for all five handlers, which is follow-up 1 and a stage of its
own.

```bash
gh issue create \
  --title "Admin console: an overflowing ?page= panics the groups-with-permission page" \
  --label bug --label go \
  --body "\`GET /admin/resources/{id}/groups-with-permission?page=9223372036854775807\` panics when the resource has no permissions.

\`handler_admin_resource_groups_with_permission.go\` validates \`page\` only as \`>= 1\`, then computes \`start := (pageInt - 1) * pageSize\`, which overflows to a negative offset; the following bounds checks only clamp from above, so the handler slices \`allGroups\` with negative bounds. It needs an authenticated administrator, so the impact is a failed request rather than a bypass.

The wider problem is that the five paginated admin pages disagree about invalid \`page\` input: two clamp to 1 and three return 500. Decide one answer, an upper bound included, and apply it to all five, then bound or remove the hand-rolled slice arithmetic here. Note that clamping alone leaves the existing quirk where a page past the last renders an empty list under a bar highlighting the last page, so the two are worth deciding together.

Found by the review of #271."
```

## Folded in rather than filed

**None.** Nothing found beside the work was folded into this branch. The three findings the review
raised were defects in this change's own code and were fixed as part of it (`c80ef2e3`), not
fold-ins; they are in the review comment.

## Ceilings left by this change

**None.** `git diff --unified=0 origin/main...HEAD | grep '^+.*ceiling:'` matches nothing.

One boundary is deliberately left and is recorded in the code's own comment rather than as a
ceiling, because it is parity and not a limit of the implementation: `HasNext` is
`total > current*pageSize`, and that product wraps on the last page of a `total` at `math.MaxInt`.
It is character-for-character what the library did, and correcting it would be a second undeclared
departure from the table this change is held to.

leodip:
## The review account

`LEO_REVIEW_MODE=end`: no stage was reviewed on its own. The whole branch was read once after both
stages landed, in two rounds, plus one review of the plan before any code was written. This is that
account in full — the verdicts live in a gitignored directory that the next run overwrites, so this
comment is where they survive.

Reviewer throughout: `openai-codex/gpt-5.6-sol` at effort `xhigh`, fresh session per gate.

|---|---|---|---|---|
| plan 1 | plan | 0 | 0 | clean |
| 1 | final | 3 (1 significant, 1 minor, 1 nit) | 0 | all confirmed, all fixed, `c80ef2e3` |
| 2 | final | 0 | 0 | clean |

---

### Plan gate, round 1 — clean

Request `271-20260907T232153Z`. Conformance, quality and security all `reviewed`, zero findings.

The reviewer did not take the plan's word for anything. It re-executed `probe/expected` in the
container and diffed it against `expected.out`, ran fourteen branch mutations against a scratch
implementation of section 4's shape, tidied the module graph in a scratch copy of the swap, and
rendered `admin_users.html` over the scratch type through the real renderer — once green, once with
`IsCurrent` neutralised.

`unchecked` held two entries, both dismissed: the stage-gate artifacts (table test, mutation
records, tier hashes, release-note sentence) are assigned to stages 1 and 2 by the plan, and a
browser pass cannot observe more than the HTML assertion does, because the partial and every page
including it are unchanged. No plan change, no decision raised.

### Before round 1: tree-hash verification

`tree-hash.sh docs/issue-271-paginater/ --verify` at `ed4991ca`, clean worktree, before the first
reading of this code:

```
current   production e432bf91f09b   unit-tests b3478b7c0dde   data-tests 07da31b1288e
          integration-tests b2a2b8bfd505   docs 7d44a35bf675   all 6719e9b64cc8

stage 1  unit stale          data stale          integration stale   docs live
stage 2  unit live           data live           integration live    docs live
```

Stage 1's tiers and its four mutations were recorded against a tree whose `production` and
`unit-tests` categories moved afterwards — which is exactly what stage 2 did to them: it edited the
five call sites and added `TestRender_AdminUsersPaginator`. `pagination.go` itself was untouched by
stage 2. Independently of the hashes, CI was green at `ed4991ca` on all twelve checks including the
data and integration tiers on mysql, postgres, mssql and sqlite, so both stages' tier claims are
confirmed at the branch tip by something no session wrote.

---

### Final gate, round 1 — three findings, all confirmed, all fixed

Request `271-20260908T003834Z`, round 1 of a budget of 4. All three axes `reviewed`.

The reviewer did not re-run what the run had already recorded. It wrote its own programs: a
**12,525,000-case** comparison of the new package against the actual library source with decision
3's rule applied, a **100,000-case** property probe, and **seven** `where.sh` mutations of its own.
Six it reported caught; the seventh is finding 3.

#### Finding 1 — the page count overflows where the library's did not

`conformance`, **minor**, confidence high. `New` computed
`totalPages := (total + pageSize - 1) / pageSize`, and that addition wraps once `total` is within
`pageSize` of `math.MaxInt`. The negative count is then caught by the `totalPages < 1` clamp and
becomes 1, so `New(MaxInt-8, 10, 1, 5)` returned `[1]` where the library returns `[1 2 3 4 5 ...]`.

**Confirmed and fixed.** Verified against the library's own source rather than the reviewer's
account of it, in the module cache at `paginater@v0.0.0-20200328080006-042474bd0eae/paginater.go:84-92`:
`TotalPages()` divides first and adds one on a non-zero remainder, so it has no addition to
overflow. Goal 3 and decision 3 make the library's table the contract with exactly one departure,
and this was an undeclared second one.

`New` now computes the count the same way, quotient then remainder. For every `total >= 0` the two
expressions are equal, so all 363 recorded rows are unchanged. `TestNew_HugeTotalDoesNotOverflow`
pins it at both page sizes the handlers pass, on page 1 and on the clamped last page.

```
where.sh mutate src/adminconsole/internal/pagination/pagination.go \
  --replace 'totalPages := total / pageSize\n\tif total%pageSize != 0 {\n\t\ttotalPages++\n\t}' \
  --with    'totalPages := (total + pageSize - 1) / pageSize' \
  --type adminconsole --run TestNew_HugeTotalDoesNotOverflow
MUTATION CAUGHT
```

That mutation leaves all 363 table rows passing, which is the point: the table could not reach this
and the new case is what closes it.

**One boundary is deliberately left, and it is parity.** `HasNext` is `total > current*pageSize`,
and that product still wraps on the last page of a `total` at `math.MaxInt` itself. It is
character-for-character what the library did (`paginater.go:59`); decision 3 keeps every quirk but
the ellipsis, and correcting it would be a second undeclared departure. Recorded in the test's own
comment. The reviewer reached the same conclusion independently and listed it under
`siblings_checked` rather than raising it.

#### Finding 2 — the module is no longer tidy after a new direct import

`quality`, **nit**, confidence high. Stage 2's `TestRender_AdminUsersPaginator` imported
`github.com/google/uuid` for one `uuid.New()`, while `go.mod:25` still carried that module
`// indirect`, so `GOPROXY=off go mod tidy -diff` exited 1 and wanted to promote it.

**Confirmed and fixed, the other way round.** The reviewer offered both remedies; promoting the
module was the wrong one here. Goal 1 is about taking modules *out* of `src/adminconsole`'s graph,
and that test file was the only Go file in the whole module importing `uuid`
(`grep -rln github.com/google/uuid src/adminconsole/` gave `go.mod`, `go.sum`, `render_test.go`).
The `Subject` field only has to render into a `<td>`, so it is left at its zero UUID and the import
is gone. `go.mod` is byte-identical to what stage 2 committed, and `GOPROXY=off go mod tidy -diff`
exits 0.

#### Finding 3 — the table's formatter hid a `Page` state the template can see

`quality`, **significant**, confidence high. `format` in the test tested `Num == -1` before
`IsCurrent`, while `partials/paginator.html:21-27` tests `.IsCurrent` first. A sentinel carrying
`IsCurrent: true` therefore printed as `...` and matched the table, while the partial would have
drawn it as an active `-1` button. The reviewer proved it rather than asserting it: mutating only
the *leading* sentinel to `Page{Num: -1, IsCurrent: true}` reported `MUTATION SURVIVED`, while the
same mutation on the trailing sentinel reported `MUTATION CAUGHT`, because the render fixture has a
trailing sentinel and no leading one.

**Confirmed and fixed.** Verified both halves by reading: the partial's `{{if .IsCurrent}}` at line
21 precedes its `{{if eq .Num -1}}` at line 24. `format`'s arms are now in the partial's order,
`IsCurrent` first, so a sentinel wrongly marked current prints as `[-1]` and no recorded row
matches it. All 363 rows still pass.

```
--replace 'case first > 2:\n\t\tp.Pages = append(p.Pages, Page{Num: -1})'
--with    'case first > 2:\n\t\tp.Pages = append(p.Pages, Page{Num: -1, IsCurrent: true})'
--type adminconsole --run 'TestNew_|TestRender_AdminUsersPaginator'
MUTATION CAUGHT          (it was MUTATION SURVIVED when the reviewer ran it)
```

The trailing sentinel was mutated the same way and stayed caught, so the family closes at both
ends. And it now closes by construction rather than by enumeration: `format`'s three arms are the
partial's three renderings, arm for arm and in the same order — active button, dots, link. Every
`Page` value the template can distinguish prints differently in the table, so there is no fourth
member to find.

#### The two `unchecked` entries

- **"The pull request body's required release-note sentence was not inspected because it is outside
  the repository."** *Covered, and it was a real gap.* Section 2's "Documentation owed" names one
  place: this body, under `## For the release notes`, one sentence for the ellipsis rule. That
  heading was not there. The rule was described in the opening prose, so nothing on the PR was
  wrong, but the one deliverable section 2 asked for was missing under the name it asked for. Added
  in this round, one sentence, the rest of the body untouched.
- **"No interactive browser layout pass was run."** *Dismissed.* Section 5 rejected a browser pass
  when the seams were chosen and the reason holds: `partials/paginator.html` and its CSS classes are
  byte-identical on this branch, and the maximum number of entries in the bar is unchanged at 7.
  Decision 3 can only replace a `...` with a page number in a slot the ellipsis already occupied, so
  no arrangement of the bar is wider than one that already shipped.

#### Tiers after the fix

`where.sh test --type adminconsole` in the dev container: green, including all 363 table rows,
`TestNew_HugeTotalDoesNotOverflow` and `TestRender_AdminUsersPaginator`. `gofmt -l src/` clean.
`GOPROXY=off go mod tidy -diff` exits 0. Data and integration were not re-run locally and did not
need to be: the round touched one pure function and two test files — no handler, route, query or
schema. Tree after the round (`tree-hash.sh --split`):

```
production 83732d1f3897   unit-tests 906c675af301   data-tests 07da31b1288e
integration-tests b2a2b8bfd505   docs 7d44a35bf675
```

`data-tests` and `integration-tests` are unchanged from the values round 1 was reviewed at, so stage
2's data and integration claims still bind. `production` moved by finding 1's fix, and every
mutation above was run against that tree; `git status --porcelain` listed exactly the three intended
files after the last mutation, so no mutation left residue.

**One correction to the round 1 record**, made when it was noticed rather than left standing: the
`unit-tests` hash first written down, `f2d8c828cb93`, was taken one edit before the commit — a
comment in `TestNew_HugeTotalDoesNotOverflow` was reworded afterwards. The committed value at
`c80ef2e3` is `906c675af301`, as above. `production 83732d1f3897` is unchanged and correct, and that
is the category all three mutations were run against, so no mutation evidence is affected.

Fix commit `c80ef2e3`, three files: `internal/pagination/pagination.go` (finding 1),
`internal/pagination/pagination_test.go` (findings 1 and 3), `internal/rendertest/render_test.go`
(finding 2). `go.mod` and `go.sum` did not move. `check-anchors.sh` passes: all 11 rows in section 0
still resolve to exactly one line in the named function.

---

### Final gate, round 2 — clean

Request `271-20260908T011209Z`, resumed session, round 2 of max 4. `axis_coverage` reports
conformance, quality and security all `reviewed`; `unchecked` is **empty**, which asserts full
coverage. Verdict `clean`, zero findings.

Round 2 was not a formality. Round 1's fix changed `New`'s page-count arithmetic — production code
written *after* the only reading this change had had — and the same commit reordered the arms of the
helper every one of the 363 table rows is compared through. Neither had been read by anything but
the session that wrote them.

A clean verdict is only worth its stated execution. The reviewer reports having run:

- `review-diff.sh origin/main` and `git show c80ef2e3`, with complete reads of the repaired
  production and test code
- `check-anchors.sh` and `tree-hash.sh` on the agreement
- `where.sh ci HEAD` — green on all twelve checks at `c80ef2e3`
- `where.sh test --type adminconsole --run 'TestNew_|TestRender_AdminUsersPaginator'`
- a fresh scratch predecessor-parity comparison at ordinary values and at the largest 41 totals for
  each production page size
- a fresh scratch invariant grid and 100,000 deterministic random pagination cases
- mutations restoring the old overflowing ceiling expression (caught), disabling the remainder
  increment (caught), and marking the leading and trailing sentinels current (both caught)
- `GOPROXY=off go mod tidy -diff` and module-graph residue checks
- expected-table identity, caller and include recounts, documentation and catalog sweeps,
  route-authorization review, security-sensitive diff search, gofmt and diff checks
- a final tree-hash and clean tracked-worktree verification

Reviewer's note: *"Clean. All three round 1 findings are correctly repaired, their mutation families
are closed, HEAD CI is green, and no conformance, quality, or security defect remains in the
reviewed branch."*

Tree at ingest:

```
production 83732d1f3897   unit-tests 906c675af301   data-tests 07da31b1288e
integration-tests b2a2b8bfd505   docs 7d44a35bf675   all 59c5b7441374
```

Round 2 was recorded by the driver rather than by a session: the verdict met the definition of clean
exactly — no findings, every axis reviewed, nothing unchecked — so no session was spent to say so,
and nothing in the tree was touched.

leodip:
## The two follow-ups are folded in, not filed

Superseding the [follow-ups comment](https://github.com/leodip/goiabada/pull/305#issuecomment-5577656561)
above, which said neither was folded in and gave the `gh issue create` for each. At the user's
request both were fixed on this branch instead. No issue was filed for either. The PR description
now carries them as stages 3 and 4.

Two commits, on top of the reviewed `c80ef2e3`:

| Commit | What |
|---|---|
| `c7ee0ef8` | Follow-up 1 and the panic of follow-up 2, in `src/adminconsole` |
| `6243567b` | The same overflow one module over, in `src/core/data` |

### `c7ee0ef8` — one answer for `?page=`

Three decisions were settled with the user before any code moved: **clamp** bad input to page 1
rather than refuse it; clamp past-the-end **in the handler**, with a second query, rather than
redirect; and put the shared parse in `internal/pagination`, beside the bar it has to agree with.

`ParsePage` and `ClampPage` are now the one answer and all five lists call them. `ClampPage` counts
pages through the same `pageCount` as `New` — factored out of it here — so the page a handler
settles on and the page the bar highlights cannot drift apart. A **sixth** reader turned up while
doing it: the users-with-permission *add* page, which only carries the value into a return link but
answered 500 to a page the list beside it rendered happily. It parses the same way now.

The panic is fixed by the same clamp: it bounds the offset before
`allGroups[start:end]`, and the two existing guards stay as the bound on the slice itself.

Eight mutations, all eight caught — including the original code restored, which fails with the
reported `panic: runtime error: slice bounds out of range [:-10]`.

### `6243567b` — and the second follow-up reached further than the report

The report scoped the overflow to the console's own slice. But the same page number reaches the
admin API, where it becomes a SQL `OFFSET`, and running the mutation against all four engines showed
a negative offset does **not** simply fail:

| Path | Engines | What actually happened |
|---|---|---|
| Six reads via `sqlbuilder` | sqlite, mysql, postgres | **No error — the first page's rows came back**, labelled as the page asked for. `SelectBuilder.Offset` drops the clause when its argument is negative ([select.go:316](https://github.com/huandu/go-sqlbuilder/blob/v1.43.0/select.go#L316)), so the query ran with no offset at all. |
| The SQL Server audit log override | mssql | Refused outright — *"The offset specified in a OFFSET clause may not be negative"* — a 500. It formats the number into the SQL text, so it has no such guard. |

So the common case was not a crash but **silently wrong data**, and the two halves fail so
differently that fixing the one you notice leaves the other in place. That is why the rule is now a
parsed test (`TestNoHandRolledPageOffset`) rather than a comment, and why the data-tier test asserts
no error **and** no rows: an error-only assertion passes on three engines while they hand back
page 1, and a rows-only assertion misses the engine that refuses.

`commondb.PageOffset` saturates the product at the largest multiple of `pageSize` an int can hold —
past the end of any table that will exist — so every read returns the empty page any other
past-the-end page returns. **Seven** call sites, not six: the seventh is the `mssqldb` override,
found by grepping for the SQL text rather than for `.Offset(`.

### Verification

Three module tiers; the full data tier on **mysql, postgres, mssql and sqlite**; integration on
sqlite; `gofmt` clean; `go vet` clean; `golangci-lint run` 0 issues on all three modules;
`GOPROXY=off go mod tidy -diff` exits 0. Check suite running on `6243567b`
([run 34217154801](https://github.com/leodip/goiabada/actions/runs/34217154801)).

### Two things worth a reviewer's eye

- **Stages 3 and 4 have not been through the reviewer.** The two rounds recorded above covered the
  branch as it stood at `c80ef2e3`.
- **The PR title still says `refactor(`,** and the branch now carries two bug fixes, one of them a
  panic. Worth retitling before merge if that feeds the changelog.

### Deliberately left alone

The admin API still echoes the requested page back: `"page": 9223372036854775807` beside an empty
list and a correct `total`. That is truthful about the request, and the console clamps for display,
so the contract is unchanged.

leodip:
## Both review findings fixed — `3881fea3`

Both confirmed, both real, neither reachable in production. Thanks — the second one in particular would have quietly rotted.

### 1. `maxPage` was a page tight

Confirmed by exact arithmetic at `maxApiPageSize = 200`:

```
maxPage (old)                = 46116860184273879
offset for page ...880: exact=9223372036854775800  int=9223372036854775800  fits
offset for page ...881: exact=9223372036854776000  int=-9223372036854775616  wraps
```

So `MaxInt/pageSize + 1` is the last good page, and the bound is now that. More to the point, you identified *why* it went unnoticed: `TestMaxPage_IsTheLargestSafeBound` checked `page*pageSize` rather than `(page-1)*pageSize` — off by exactly one page, so it agreed with a bound one page too tight while claiming to pin the largest.

Every assertion in that test is now about the offset, computed in `big.Int`, and the upper half is stated as a live claim rather than a decoration:

```go
// One page further does not fit, which is what makes this bound the largest
// rather than merely a safe one. If this assertion ever fails, the bound can
// be raised.
assert.NotEqual(t, offsetOf(maxPage+1).String(), strconv.Itoa(maxPage*maxApiPageSize), ...)
```

**Mutation:** restoring the old bound now fails the test. The old version of the test passed with it — which is the finding.

### 2. The offset lint read one operand

Right, and worse than cosmetic: a lint rule that a keystroke defeats is not a rule. `offsetText` now examines both operands via a shared `minusOne` helper, and renders whichever way the expression is written, so a failure still points at the line rather than the file.

The synthetic tree gained the commuted form on *both* sides of the rule — rejected where it is an offset, accepted where the subtraction is not `- 1`:

```go
// Rejected
write("core/data/sqlitedb/user.go", `... return pageSize * (page - 1) ...`)
// Accepted, both ways round, so widening the checker did not widen what it matches
return (a-2)*b, a*b, (a+1)*b, b*(a-2), b*a, b*(a+1)
```

Plus an assertion that the commuted one renders as `pageSize * (page - 1)` rather than being silently normalised.

**Mutation:** reverting the matcher to the left operand fails the synthetic test.

### Verification

Three module tiers, `gofmt`, `go vet`, `golangci-lint` (0 issues) and `go mod tidy -diff` on all three modules. The data and integration tiers were not re-run locally: the only production change is one constant in the console, which no database test reaches — CI covers them, and it was green on `6243567b`.

### On the local run

For the record, in case it is useful: Go is in the dev container rather than on the host. `docker exec goiabada-devcontainer-1 bash -lc 'cd /workspaces/goiabada/src/... && go test ./...'` works against the same working tree, and `src/authserver/run-tests.sh --type <tier> --db <engine>` drives the data and integration tiers.

Issue #56, opened 2026-02-01
Add configurable setting to control inclusion of OIDC scope claims in ID tokens


leodip:
## Problem

Goiabada currently includes all scope-based claims (email, profile, address, phone) directly in ID tokens. While this is common practice among major OIDC providers (Auth0, Microsoft, Keycloak), the OIDC conformance test suite warns that these claims should only be available via the `/userinfo` endpoint for strict spec compliance.

**OIDC Conformance Suite Warning:**
```
WARNING OIDCC-5.4: Unexpectedly found email in id_token. 
The conformance suite did not request the 'email' claim is returned in the 
id_token and hence did not expect the server to include it.
```

Per [OIDC Core 5.4](https://openid.net/specs/openid-connect-core-1_0.html#ScopeClaims), scope claims (email, profile, etc.) **MAY** be in ID tokens but **SHOULD** be available from `/userinfo`. The conformance suite expects minimal ID tokens by default.

## Proposed Solution

Add a configurable setting `IncludeOpenIDConnectClaimsInIdToken` that controls whether scope claims are included in ID tokens:

- **Enabled (default: true)**: Current behavior - claims included in ID tokens (matches Auth0/Microsoft/Keycloak)
- **Disabled**: Minimal ID tokens - scope claims only via `/userinfo` (strict OIDC conformance)

Should support both global default and per-client override (similar to existing `IncludeOpenIDConnectClaimsInAccessToken`).

## Why Default to Enabled?

1. Industry standard - all major OIDC providers do this by default
2. Backward compatible - won't break existing deployments  
3. Better developer experience - most applications expect user info in ID tokens
4. OIDC spec compliant - spec says claims "MAY" be in ID tokens

Users requiring strict conformance can disable globally or per-client.

## Implementation Approach

- Add boolean setting to `Settings` model with global default
- Add three-state setting to `Client` model for per-client override ("default"/"on"/"off")
- Modify token generation logic to conditionally include scope claims
- Update Admin Console UI with explanatory tooltips about OIDC conformance implications
- Database migrations for all supported databases
- Update tests

**Note:** Custom Goiabada scopes (`groups`, `attributes`) should always be included when requested.

PR #306, opened 2026-09-08
test: retire gofakeit for an in-tree core/testutil/fake package

### LINKED ISSUE #136: Flaky: TestSearchUsersPaginated collides with accumulated users in the never-reset shared databases

### LINKED ISSUE #272: gofakeit is a test-only dependency in 109 files for six functions, 962 of the calls being LetterN
`github.com/brianvoe/gofakeit/v6` is imported by **109 test files** across all three modules. Counting the calls shows what we actually want from it:

| Function | Calls |
|---|---|
| `gofakeit.LetterN` | **962** |
| `gofakeit.Password` | 177 |
| `gofakeit.Email` | 111 |
| `gofakeit.URL` | 91 |
| `gofakeit.UUID` | 47 |
| everything else combined (`FirstName`, `LastName`, `Name`, `Username`, `LoremIpsumSentence`, `IPv4`, `Bool`, `CountryAbr`, `Zip`, `State`, `StreetNumber`, `Phone`, `Number`, `MiddleName`, `Date`) | ~80 |

962 of roughly 1470 calls are `LetterN`, which returns N random letters. We are pulling in a fake-data generator with a very large API surface, a bundled corpus of names, addresses, companies, car models, animals and beer brands, to mostly produce random strings.

## Why it matters even though it is test-only

It never reaches a binary, so this is not about size. It is about the supply chain of CI: a test-only dependency runs with the same privileges as the test suite, on every push, on every contributor's machine. Anything we can retire from that set for 50 lines of code is worth retiring.

There is also a correctness angle. `gofakeit.LetterN` is not seeded per test and draws from a package-global source; the values are not reproducible from a failure message, so a flake that depends on a generated value cannot be replayed.

## What replaces it

A `core/testutil` (or a small internal `faker` package) with the handful of generators that are actually used:

```go
func LetterN(n uint) string
func Password(lower, upper, numeric, special, space bool, n int) string
func Email() string
func URL() string
func UUID() string
func Name() string   // and the two or three name variants
```

`crypto/rand` or a `math/rand/v2` source over a fixed alphabet covers all of it. Around 50 to 80 lines. Giving the source an explicit seed that a failing test can print would be a genuine improvement over what we have.

## Scope

- 109 files, all `_test.go`.
- Purely mechanical: the signatures can match gofakeit's so the call sites only change their import.
- `Password` has the widest signature and is the one worth reading gofakeit's implementation for, to keep generating values that satisfy the password validators.

---

### Dependency reduction

leodip:
Closes #272. Closes #136.

## What this is for

`github.com/brianvoe/gofakeit/v6` is a direct requirement of `core` and `authserver`, reaching
`adminconsole` through `core`, and it is used by nothing but `_test.go` files: 114 of them, 1508
calls, to draw random letters, passwords, emails, redirect URIs, UUIDs and a handful of profile
fields. Every generated value is opaque to the test that draws it. Only a validator, a column width
or a parser ever constrained one, and the tree already has a `crypto/rand` letters generator in
`core/stringutil`.

This change replaces the library with a small package, `src/core/testutil/fake`, standard library
plus `core/stringutil` and `google/uuid`, whose functions satisfy the consumers by construction:
`LetterN` is letters only, `Password(n)` always carries all four character classes so a test
raising the policy to `High` cannot flake, `Email` and `URL` are shapes the validators accept.

It also closes #136. `TestSearchUsersPaginated` asserted that a fresh `gofakeit.Username()` matched
exactly one row in a MySQL table that is never reset, and that generator's value space is about
4.7 million: at the row count #136 measured, a duplicate was expected. `fake.Username()` is
`"user"` plus 12 random letters, 52^12 values, so the collision is unreachable.

## The rule it establishes

Test fixtures draw random values from `core/testutil/fake`. No test-only faker library comes
back; the compiler guards it once the two `go.mod` files stop requiring one.

## Status

Complete, and ready for review. Both stages have landed and the whole change has been reviewed.
`grep -rl gofakeit src` returns nothing, no `go.mod` requires the module and no `go.sum` lists it.
The check suite is green at `9a8a8b87` on all twelve jobs, which is where the four-engine data and
integration result lives.

### Stage 1: the `fake` package and its unit table

Landed in `e71fd694`. `src/core/testutil/fake` holds the whole surface the agreement's design lists:
`LetterN`, `DigitN`, `Password`, `Email`, `URL`, `UUID`, `Username`, the three name generators plus
`Name`, `IPv4Address`, `Bool`, `Number`, `Date` and `Sentence`. `LetterN` and `DigitN` delegate to
`core/stringutil`, so the tree gains no second letters generator; every other draw goes through one
private `intn` over `crypto/rand.Int`, so the package carries no global, no seed and no mutex.

Three properties are load-bearing rather than incidental. `Password(n)` guarantees a lowercase
letter, an uppercase letter, a digit and a special character at every length, because the
integration tests run under the seeded `low` policy that requires none of them and a test raised to
`high` would otherwise flake. `Username()` is `"user"` plus 12 lowercase letters, which is the fix
for #136: a few thousand surnames collide inside a test database of a few thousand rows and 26^12
does not. `Date()` is whole seconds, so the same value round-trips unchanged through MySQL
`datetime`, SQL Server `datetime2` and SQLite.

Tiers: unit green in the dev container, 61 packages ok, 0 fail. Data and integration are not
applicable — nothing outside the package's own test imports it, so no query, route or handler can
reach it. Four mutations run and all four caught: `Password`'s per-class draws removed, `Username`'s
entropy shrunk to a 676-value space at the same 16 characters, `Bool` pinned to `true`, and a
fragment appended to `URL`'s literal.

Two departures from the plan, both recorded in full in the run log: `strconv` joined the import list
the plan named, since rendering an IPv4 octet as decimal needs it and the plan listed neither it nor
`fmt`; and the `Username` mutation was padded back to 16 characters so that only the
10000-draw distinctness assertion could catch it, rather than the length assertion catching it
first.

### Stage 2: rewrite the 114 call-site files, drop the requirement, one line in CLAUDE.md

Landed in `f3b41f20`. 120 files changed: the 114 test files, two `go.mod` files, three `go.sum`
files and `CLAUDE.md`.

The 22 calls with no counterpart in `fake` went first, by hand, before any `sed` could bury them.
All 22 sit in the three profile fixtures that write a whole `models.User` through the database and
read it back, so each became a prefixed letter run sized against the column it lands in, read from
`mysqldb/schema.golden`: `"City" + fake.LetterN(6)` for `address_locality`, `"tz" + fake.LetterN(6)`
for `zone_info`, `strings.ToUpper(fake.LetterN(2))` for the two country columns, and so on.
`compareUsers` reads every one of those columns back, so a value a column truncated would fail the
data tier on the engine that truncated it. `Gender` is picked from `enums.GenderFemale` and
`enums.GenderMale` rather than generated, since the column is 16 characters and the value is echoed
back as an OIDC claim.

The remaining 1486 calls moved by four mechanical shapes, each safe because the corpus was uniform
rather than merely usually uniform: the import line had one spelling in all 114 files, all 178
`Password` calls carried the identical five booleans, and both `Numerify` calls were `"##"`.
`LoremIpsumSentence` became `Sentence`; `Numerify`, `Phone`, `Zip` and `StreetNumber` became
`DigitN` at the length each column wants. `gofmt -w` then reordered the import blocks. Only after
the last import was gone did `go mod tidy` run, in `core`, `authserver` and `adminconsole`; it
removed the gofakeit lines and nothing else, 8 deletions across the five files.

No assertion changed anywhere.

Tiers, run against the tree **before** the rewrite as well as after, which is what makes the claim
a comparison rather than a green suite: unit 1453 ok 0 fail, data green on mysql, postgres, mssql
and sqlite at 1631 ok 0 fail, integration on sqlite 1012 ok 0 fail. All three counts are identical
before and after, so the rewrite added, removed and skipped no test. No mutation was owed: this
stage adds no test, and stage 1's table already pins every generator these call sites now stand on.

Three deviations, recorded in full in the run log. Two of the agreement's code anchors cited the
`gofakeit` requirement line, which no longer exists in either `go.mod`, so both now anchor on the
file's `module` line with a note saying why. `Gender` needed a statement rather than the expression
the plan sketched, since Go cannot branch inside a composite literal. And the plan's two counts for
the `createTestUserOn` fixture, 24 and 8, count different things — every non-`LetterN` call against
the hand-rewrite subset — which is a wording point rather than a discrepancy in the work.

### After the review: the CSPRNG guard

Landed in `9a8a8b87`, answering final review round 1. `stringutil`'s letters and digits helpers
return `""` when the system CSPRNG is unavailable, a contract their production callers were written
against. Inherited by `fake`, it meant a broken CSPRNG turned `Username()` into the constant `"user"`
for every call and the suite still went green — the exact collision shape this change exists to
close. A `mustDraw` helper now panics unless a draw is the length asked for, and both wrappers route
through it. Two cases pin it by swapping `crypto/rand.Reader` for a reader that only errors.
`stringutil` itself was left alone: its `""` is load-bearing for two production call sites and
changing it is a question this change has no mandate for. The same round prompted one fixture value
to become the constant `"America/New_York"`; the regression claim behind it was refuted, and the
demonstration is in the review comment.

## How it was verified

| Tier | Where it ran | Coverage | Result |
|---|---|---|---|
| unit | locally, dev container, at `9a8a8b87` | all three modules | 5397 pass, 0 fail |
| data | locally, dev container, at `9a8a8b87` | sqlite, mysql, postgres, mssql | 2208 pass, 0 fail |
| integration | locally, dev container, at `9a8a8b87` | sqlite | 1333 pass, 0 fail |
| all of the above, plus lint, vulnerabilities, versions and build | **the check suite**, at `9a8a8b87` | four engines in parallel | [green, 12 of 12](https://github.com/leodip/goiabada/actions/runs/34235792672) |

The four-engine **integration** result exists only in the check suite; locally only sqlite was run.
Six mutations across the run, all six caught. Test counts before and after the rewrite are identical
except for the two entropy cases the review added, so nothing was dropped, skipped or renamed.

### Review

Reviewed by `openai-codex/gpt-5.6-sol` at effort `xhigh`, 3 rounds: 1 at the plan gate and 2 over
the whole branch. No stage was reviewed on its own — this run used `LEO_REVIEW_MODE=end`.

|---|---|---|---|---|
| plan | 1 | 2 | 0 | both fixed in the plan before any code was written |
| final | 1 | 2 | 0 | 1 fixed in `9a8a8b87`, 1 refuted with the improvement taken anyway |
| final | 2 | 0 | 0 | clean, all three axes reviewed |

- **significant, quality:** `LetterN` and `DigitN` suppressed a CSPRNG failure, so a broken CSPRNG
  made `Username()` a constant and the suite went green. Fixed in `9a8a8b87`: a `mustDraw` guard
  that panics on a short draw, plus two regression cases; both guard-removal mutations caught.
- **significant, conformance:** the OIDC fixture's `zoneinfo` was claimed to be a new regression.
  Refuted — `gofakeit.TimeZoneFull()` drew 108 Windows display strings, never IANA identifiers, so
  the old value failed the same validator. The improvement was taken anyway in `9a8a8b87`:
  `"America/New_York"`, matching the already-constant `Locale` beside it.
- **significant, quality (plan):** the table had no `Bool` case, so a constant implementation would
  pass. Fixed in the plan; the constant-`true` mutation was later run and caught.
- **significant, conformance (plan):** the `URL` case asserted `url.Parse` and a host where the
  authorization endpoint applies `urlutil.IsAbsoluteRedirectURI`, which also refuses a fragment.
  Fixed in the plan; the fragment mutation was later run and caught.

Full account, with the evidence and both `unchecked` dispositions, in
[this comment](https://github.com/leodip/goiabada/pull/306#issuecomment-5586702888).

## The record

- [Deferred decisions](https://github.com/leodip/goiabada/pull/306#issuecomment-5586669379) — none;
  all four decisions were settled with the user before the agreement was sealed.
- [Follow-ups, fold-ins and ceilings](https://github.com/leodip/goiabada/pull/306#issuecomment-5586674584)
  — one follow-up to file (OIDC `phone_number` is not E.164), two fold-ins, no ceilings.
- [Review account](https://github.com/leodip/goiabada/pull/306#issuecomment-5586702888) — all three
  rounds, what each found and what answered it.

Nothing here asks an operator to do anything on upgrade: the change is confined to `_test.go` files,
one new package no binary imports, and the `go.mod`/`go.sum` lines that required a test-only module.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -267,6 +267,10 @@ Three test types:
 
 **Best way to run all tests**: `./run-tests.sh` inside the dev container (from `src/authserver/`).
 
+**Test fixtures**: tests draw random values from `core/testutil/fake`, a `_test.go`-only package
+over `crypto/rand` that replaced a third-party faker in #272. Reach for it rather than adding a
+dependency the next time a test needs a random string.
+
 **gofmt guard**: every module's unit tier runs `TestGoSourcesAreGofmted`, which holds every Go
 file under `src/` to gofmt's formatting through `core/testutil.AssertGofmted`. The walk is
 repository-wide from each tier because `cmd/goiabada-setup` has no tier of its own. CI's Lint job
--- a/src/adminconsole/go.sum
+++ b/src/adminconsole/go.sum
@@ -14,8 +14,6 @@ github.com/AzureAD/microsoft-authentication-library-for-go v1.6.0 h1:XRzhVemXdgv
 github.com/AzureAD/microsoft-authentication-library-for-go v1.6.0/go.mod h1:HKpQxkWaGLJ+D/5H8QRpyQXA1eKjxkFlOMwck5+33Jk=
 github.com/BurntSushi/toml v1.6.0 h1:dRaEfpa2VI55EwlIW72hMRHdWouJeRF7TPYhI+AUQjk=
 github.com/BurntSushi/toml v1.6.0/go.mod h1:ukJfTF/6rtPPRCnwkur4qwRxa8vTRFBF0uk2lLoLwho=
-github.com/brianvoe/gofakeit/v6 v6.28.0 h1:Xib46XXuQfmlLS2EXRuJpqcw8St6qSZz75OUo0tgAW4=
-github.com/brianvoe/gofakeit/v6 v6.28.0/go.mod h1:Xj58BMSnFqcn/fAQeSK+/PLtC5kSb7FJIq4JyGa8vEs=
 github.com/davecgh/go-spew v1.1.0/go.mod h1:J7Y8YcW2NihsgmVo/mv3lAwl/skON4iLHjSsI+c5H38=
 github.com/davecgh/go-spew v1.1.1/go.mod h1:J7Y8YcW2NihsgmVo/mv3lAwl/skON4iLHjSsI+c5H38=
 github.com/davecgh/go-spew v1.1.2-0.20180830191138-d8f796af33cc h1:U9qPSI2PIWSS1VwoXQT9A3Wy9MM3WgvqSxFWenqJduM=
--- a/src/authserver/go.mod
+++ b/src/authserver/go.mod
@@ -4,7 +4,6 @@ go 1.27.0
 
 require (
 	github.com/PuerkitoBio/goquery v1.12.0
-	github.com/brianvoe/gofakeit/v6 v6.28.0
 	github.com/go-chi/chi/v5 v5.3.2
 	github.com/go-jose/go-jose/v4 v4.1.4
 	github.com/golang-jwt/jwt/v5 v5.3.1
--- a/src/authserver/go.sum
+++ b/src/authserver/go.sum
@@ -21,8 +21,6 @@ github.com/andybalholm/cascadia v1.3.3/go.mod h1:xNd9bqTn98Ln4DwST8/nG+H0yuB8Hmg
 github.com/boombuler/barcode v1.0.1-0.20190219062509-6c824513bacc/go.mod h1:paBWMcWSl3LHKBqUq+rly7CNSldXjb2rDl3JlRe0mD8=
 github.com/boombuler/barcode v1.1.0 h1:ChaYjBR63fr4LFyGn8E8nt7dBSt3MiU3zMOZqFvVkHo=
 github.com/boombuler/barcode v1.1.0/go.mod h1:paBWMcWSl3LHKBqUq+rly7CNSldXjb2rDl3JlRe0mD8=
-github.com/brianvoe/gofakeit/v6 v6.28.0 h1:Xib46XXuQfmlLS2EXRuJpqcw8St6qSZz75OUo0tgAW4=
-github.com/brianvoe/gofakeit/v6 v6.28.0/go.mod h1:Xj58BMSnFqcn/fAQeSK+/PLtC5kSb7FJIq4JyGa8vEs=
 github.com/creack/pty v1.1.9/go.mod h1:oKZEueFk5CKHvIhNR5MUki03XCEU+Q6VDXinZuGJ33E=
 github.com/davecgh/go-spew v1.1.0/go.mod h1:J7Y8YcW2NihsgmVo/mv3lAwl/skON4iLHjSsI+c5H38=
 github.com/davecgh/go-spew v1.1.1/go.mod h1:J7Y8YcW2NihsgmVo/mv3lAwl/skON4iLHjSsI+c5H38=
--- a/src/authserver/tests/data/cascade_delete_test.go
+++ b/src/authserver/tests/data/cascade_delete_test.go
@@ -5,8 +5,8 @@ import (
 	"testing"
 	"time"
 
-	"github.com/brianvoe/gofakeit/v6"
 	"github.com/leodip/goiabada/core/models"
+	"github.com/leodip/goiabada/core/testutil/fake"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
@@ -34,8 +34,8 @@ func createROPCRefreshToken(t *testing.T, userId, clientId int64) *models.Refres
 	refreshToken := &models.RefreshToken{
 		UserId:            sql.NullInt64{Int64: userId, Valid: true},
 		ClientId:          sql.NullInt64{Int64: clientId, Valid: true},
-		RefreshTokenJti:   gofakeit.UUID(),
-		SessionIdentifier: gofakeit.UUID(),
+		RefreshTokenJti:   fake.UUID(),
+		SessionIdentifier: fake.UUID(),
 		RefreshTokenType:  "Offline",
 		Scope:             "openid profile offline_access",
 		IssuedAt:          sql.NullTime{Time: now, Valid: true},
@@ -55,8 +55,8 @@ func createCodeLinkedRefreshToken(t *testing.T, codeId int64) *models.RefreshTok
 	now := time.Now().UTC().Truncate(time.Microsecond)
 	refreshToken := &models.RefreshToken{
 		CodeId:            sql.NullInt64{Int64: codeId, Valid: true},
-		RefreshTokenJti:   gofakeit.UUID(),
-		SessionIdentifier: gofakeit.UUID(),
+		RefreshTokenJti:   fake.UUID(),
+		SessionIdentifier: fake.UUID(),
 		RefreshTokenType:  "Bearer",
 		Scope:             "openid profile",
 		IssuedAt:          sql.NullTime{Time: now, Valid: true},
--- a/src/authserver/tests/data/client_permission_test.go
+++ b/src/authserver/tests/data/client_permission_test.go
@@ -5,13 +5,13 @@ import (
 	"testing"
 	"time"
 
-	"github.com/brianvoe/gofakeit/v6"
 	"github.com/leodip/goiabada/core/models"
+	"github.com/leodip/goiabada/core/testutil/fake"
 )
 
 func TestCreateClientPermission(t *testing.T) {
 	client := &models.Client{
-		ClientIdentifier: "test_client_" + gofakeit.LetterN(6),
+		ClientIdentifier: "test_client_" + fake.LetterN(6),
 		Description:      "Test Client",
 	}
 	err := database.CreateClient(nil, client)
@@ -21,7 +21,7 @@ func TestCreateClientPermission(t *testing.T) {
 
 	// Create a resource for testing
 	resource := &models.Resource{
-		ResourceIdentifier: "test_resource_" + gofakeit.LetterN(6),
+		ResourceIdentifier: "test_resource_" + fake.LetterN(6),
 		Description:        "Test Resource",
 	}
 	err = database.CreateResource(nil, resource)
@@ -31,7 +31,7 @@ func TestCreateClientPermission(t *testing.T) {
 
 	// Create a permission for testing
 	permission := &models.Permission{
-		PermissionIdentifier: "test_permission_" + gofakeit.LetterN(6),
+		PermissionIdentifier: "test_permission_" + fake.LetterN(6),
 		Description:          "Test Permission",
 		ResourceId:           resource.Id,
 	}
@@ -98,7 +98,7 @@ func TestCreateClientPermission(t *testing.T) {
 func TestUpdateClientPermission(t *testing.T) {
 	// Create a client for testing
 	client := &models.Client{
-		ClientIdentifier: "test_client_" + gofakeit.LetterN(6),
+		ClientIdentifier: "test_client_" + fake.LetterN(6),
 		Description:      "Test Client",
 	}
 	err := database.CreateClient(nil, client)
@@ -108,7 +108,7 @@ func TestUpdateClientPermission(t *testing.T) {
 
 	// Create another client for updating
 	newClient := &models.Client{
-		ClientIdentifier: "new_test_client_" + gofakeit.LetterN(6),
+		ClientIdentifier: "new_test_client_" + fake.LetterN(6),
 		Description:      "New Test Client",
 	}
 	err = database.CreateClient(nil, newClient)
@@ -118,7 +118,7 @@ func TestUpdateClientPermission(t *testing.T) {
 
 	// Create a resource for testing
 	resource := &models.Resource{
-		ResourceIdentifier: "test_resource_" + gofakeit.LetterN(6),
+		ResourceIdentifier: "test_resource_" + fake.LetterN(6),
 		Description:        "Test Resource",
 	}
 	err = database.CreateResource(nil, resource)
@@ -128,7 +128,7 @@ func TestUpdateClientPermission(t *testing.T) {
 
 	// Create a permission for testing
 	permission := &models.Permission{
-		PermissionIdentifier: "test_permission_" + gofakeit.LetterN(6),
+		PermissionIdentifier: "test_permission_" + fake.LetterN(6),
 		Description:          "Test Permission",
 		ResourceId:           resource.Id,
 	}
@@ -139,7 +139,7 @@ func TestUpdateClientPermission(t *testing.T) {
 
 	// Create another permission for updating
 	newPermission := &models.Permission{
-		PermissionIdentifier: "new_test_permission_" + gofakeit.LetterN(6),
+		PermissionIdentifier: "new_test_permission_" + fake.LetterN(6),
 		Description:          "New Test Permission",
 		ResourceId:           resource.Id,
 	}
@@ -195,7 +195,7 @@ func TestUpdateClientPermission(t *testing.T) {
 func TestGetClientPermissionById(t *testing.T) {
 	// Create a client for testing
 	client := &models.Client{
-		ClientIdentifier: "test_client_" + gofakeit.LetterN(6),
+		ClientIdentifier: "test_client_" + fake.LetterN(6),
 		Description:      "Test Client",
 	}
 	err := database.CreateClient(nil, client)
@@ -205,7 +205,7 @@ func TestGetClientPermissionById(t *testing.T) {
 
 	// Create a resource for testing
 	resource := &models.Resource{
-		ResourceIdentifier: "test_resource_" + gofakeit.LetterN(6),
+		ResourceIdentifier: "test_resource_" + fake.LetterN(6),
 		Description:        "Test Resource",
 	}
 	err = database.CreateResource(nil, resource)
@@ -215,7 +215,7 @@ func TestGetClientPermissionById(t *testing.T) {
 
 	// Create a permission for testing
 	permission := &models.Permission{
-		PermissionIdentifier: "test_permission_" + gofakeit.LetterN(6),
+		PermissionIdentifier: "test_permission_" + fake.LetterN(6),
 		Description:          "Test Permission",
 		ResourceId:           resource.Id,
 	}
@@ -274,7 +274,7 @@ func TestGetClientPermissionById(t *testing.T) {
 func TestGetClientPermissionByClientIdAndPermissionId(t *testing.T) {
 	// Create a client for testing
 	client := &models.Client{
-		ClientIdentifier: "test_client_" + gofakeit.LetterN(6),
+		ClientIdentifier: "test_client_" + fake.LetterN(6),
 		Description:      "Test Client",
 	}
 	err := database.CreateClient(nil, client)
@@ -284,7 +284,7 @@ func TestGetClientPermissionByClientIdAndPermissionId(t *testing.T) {
 
 	// Create a resource for testing
 	resource := &models.Resource{
-		ResourceIdentifier: "test_resource_" + gofakeit.LetterN(6),
+		ResourceIdentifier: "test_resource_" + fake.LetterN(6),
 		Description:        "Test Resource",
 	}
 	err = database.CreateResource(nil, resource)
@@ -294,7 +294,7 @@ func TestGetClientPermissionByClientIdAndPermissionId(t *testing.T) {
 
 	// Create a permission for testing
 	permission := &models.Permission{
-		PermissionIdentifier: "test_permission_" + gofakeit.LetterN(6),
+		PermissionIdentifier: "test_permission_" + fake.LetterN(6),
 		Description:          "Test Permission",
 		ResourceId:           resource.Id,
 	}
@@ -344,7 +344,7 @@ func TestGetClientPermissionByClientIdAndPermissionId(t *testing.T) {
 func TestGetClientPermissionsByClientId(t *testing.T) {
 	// Create a client for testing
 	client := &models.Client{
-		ClientIdentifier: "test_client_" + gofakeit.LetterN(6),
+		ClientIdentifier: "test_client_" + fake.LetterN(6),
 		Description:      "Test Client",
 	}
 	err := database.CreateClient(nil, client)
@@ -354,7 +354,7 @@ func TestGetClientPermissionsByClientId(t *testing.T) {
 
 	// Create a resource for testing
 	resource := &models.Resource{
-		ResourceIdentifier: "test_resource_" + gofakeit.LetterN(6),
+		ResourceIdentifier: "test_resource_" + fake.LetterN(6),
 		Description:        "Test Resource",
 	}
 	err = database.CreateResource(nil, resource)
@@ -366,7 +366,7 @@ func TestGetClientPermissionsByClientId(t *testing.T) {
 	permissions := make([]*models.Permission, 3)
 	for i := 0; i < 3; i++ {
 		permissions[i] = &models.Permission{
-			PermissionIdentifier: "test_permission_" + gofakeit.LetterN(6),
+			PermissionIdentifier: "test_permission_" + fake.LetterN(6),
 			Description:          "Test Permission " + strconv.Itoa(i+1),
 			ResourceId:           resource.Id,
 		}
@@ -415,7 +415,7 @@ func TestGetClientPermissionsByClientId(t *testing.T) {
 func TestDeleteClientPermission(t *testing.T) {
 	// Create a client for testing
 	client := &models.Client{
-		ClientIdentifier: "test_client_" + gofakeit.LetterN(6),
+		ClientIdentifier: "test_client_" + fake.LetterN(6),
 		Description:      "Test Client",
 	}
 	err := database.CreateClient(nil, client)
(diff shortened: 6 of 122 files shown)


leodip:
## Deferred decisions: none

The run deferred nothing. The agreement's four decisions were all settled with the user before it
was sealed, and no stage raised a fifth. There is no assumption in this branch waiting on an answer,
and nothing here is cheaper to reverse now than later.

For the record, since this is the comment a reader opens first, the four settled decisions and what
reversing each would cost today:

1. **A new `src/core/testutil/fake` package, name `fake`.** Reversing means moving 16 functions into
   `core/testutil` proper and re-running one `sed` over 114 files. Cheap, but it would put plain
   generators beside two assertion helpers that take a `*testing.T`, and `schemadump/golden.go`
   imports that package from non-test code.
2. **`crypto/rand`, no printable seed.** Reversing means a global, an environment variable and a
   print in 21 `TestMain` functions. A seed only replays a value if the whole package re-runs in
   source order, which `-run TestX` does not do; `testify` already prints the generated value it
   compared.
3. **`Password(n int)`, class-complete at every length.** Reversing to the six-boolean signature is
   one `sed`, and would restore five parameters that were identical at all 178 call sites.
4. **Fold #136 in.** It is a property of `fake.Username()` rather than a stage, so reversing means
   nothing to undo in code — only reopening #136 against a collision that is now unreachable.

The one open question this run *found* is not a decision it deferred; it is a defect in production
code the change never enters, drafted as a follow-up in the next comment.

leodip:

### Follow-ups to file: 1

**1. Build OIDC `phone_number` claims from the complete international number.** `bug`, `go`.

Found by the final review of this run, round 1. `src/core/oauth/token_issuer.go:686` emits
`user.PhoneNumber` verbatim and never reads `user.PhoneNumberCountryCallingCode`, though the account
and admin phone handlers (`handler_api_account_phone.go`, `handler_api_users_phone.go`) store the two
parts separately on `models.User`. OpenID Connect Core 1.0 section 5.1 requires `phone_number` to be
in E.164 format whenever `phone_number_verified` is true, and line 687 sets that claim from
`user.PhoneNumberVerified`, so a verified non-local number goes out in a form the specification
forbids.

Not folded in: it touches production issuer behaviour, a layer this test-only change never enters, so
nothing in this branch's tiers would catch a mistake there. It is also not a one-liner — the calling
code and the national number have to be combined and canonicalised, which raises a question about
rows already stored without a calling code, and it needs unit coverage at the issuer plus integration
coverage for a verified cross-country number. That is a stage of its own.

Duplicate search re-run at the close of the run: `gh issue list --state all --search "phone_number"`,
`"E.164"` and `"phone claim"` return nothing open or closed on this. The nearest is #56, closed,
which added the setting controlling whether OIDC scope claims appear in ID tokens at all, not how one
is formatted.

```bash
gh issue create --title 'Build OIDC phone_number claims from the complete international number' \
  --label bug --label go --body 'TokenIssuer.addOpenIdConnectClaimsFromUser emits the `phone_number` claim as the stored national number alone and ignores `users.phone_number_country_callingcode`, which the phone handlers write as a separate column. OpenID Connect Core 1.0 section 5.1 says `phone_number` MUST be in E.164 format when `phone_number_verified` is true, which the issuer sets from the stored flag, so a verified international number goes out non-conformant and a relying party cannot dial or match it. Combine the calling code and the national number into an E.164 string at the issuer, decide what to emit for rows stored without a calling code, and cover it with a unit test at the issuer and an integration test that reads the claim back for a verified cross-country number.

Found by the final review of #272 (PR #306), which did not fold it in: it is production issuer behaviour, a layer that test-only change never entered.'
```

### Folded in rather than filed: 2

- **#136, the username collision in the never-reset test databases** — `f3b41f20`, and settled as
  decision 5 of the agreement before the run began. `TestSearchUsersPaginated` asserted a fresh
  `gofakeit.Username()` matched exactly one row across six `LIKE` columns of a MySQL table three
  server-backed engines never reset, and that generator's space is about 4.7 million values against a
  table that already held duplicates. `fake.Username()` is `"user"` plus 12 random letters, 52^12, so
  the birthday expectation at the 2781 rows #136 measured falls from 0.82 to about 2e-15. It fit
  because it is a property of a function this change had to write anyway, and there was nothing left
  to settle once the generator's shape was decided. This PR closes #136. What it does **not** do is
  #136's option 3, a per-run database reset, which stays out of scope as a harness question.
- **The OIDC claims fixture's `zoneinfo`** — `9a8a8b87`, arising from final review round 1. The
  agreement's §4.2 prescribed `"tz" + fake.LetterN(6)`; the fixture now writes the constant
  `"America/New_York"`. The reviewer's regression claim was refuted — `gofakeit.TimeZoneFull()` drew
  108 Windows display strings, not IANA identifiers, so the old value was equally invalid — but the
  observation under it was worth taking: `Locale: "en-US"` on the line below was already a validated
  constant, and a full-profile fixture the server's own `ProfileValidator` would refuse is a weaker
  thing for the claim assertions to run against. One line, one anchor re-swept, nothing added to the
  `fake` package. Detail and the demonstration are in the review comment.

### Ceilings left by this change: none

`git diff --unified=0 main...HEAD | grep '^+.*ceiling:'` returns no hits. The branch adds one
non-test file, `src/core/testutil/fake/fake.go`, and it makes no bounded assumption that a future
change would need to revisit.

leodip:
## Review account

Three rounds in total: one at the plan gate, two over the whole change. All by
`openai-codex/gpt-5.6-sol` at effort `xhigh`. Stage gates were not reviewed individually —
`LEO_REVIEW_MODE=end`, so the code got one reading as a whole rather than two partial ones.

|---|---|---|---|---|---|
| plan | 1 | 2 | 0 | 2 | both fixed in `plan.md` before any code was written |
| final | 1 | 2 | 0 | 2 | 1 fixed in `9a8a8b87`, 1 refuted with the improvement taken anyway |
| final | 2 | 0 | 0 | 0 | clean, all three axes reviewed |

---

### Plan gate, round 1 — request `272-20260908T114451Z`

Verdict `static-review-exhausted`, two findings, all three axes reviewed. `unchecked` held only what
a plan gate cannot reach: code-level mutations and tier results, which stage 1 and the end-of-run
review cover, and a live database or browser pass, which nothing in this change needs.

1. **Quality, significant: the test table had no case for `Bool`,** so a constant implementation
   would pass it. Confirmed and fixed in the plan: a both-values case over 1000 draws, and a
   constant-`true` mutation added to step 3. That mutation was later run and caught.
2. **Conformance, significant: the `URL` case checked the wrong oracle.** It asserted `url.Parse`,
   `IsAbs` and a host, while the authorization endpoint applies `urlutil.IsAbsoluteRedirectURI`,
   which also refuses a fragment. The reviewer's scratch test showed
   `https://client.example/callback#fragment` passing the planned oracle and failing the endpoint's.
   Confirmed and fixed in the plan: the case now asserts the endpoint's predicate, and step 3 mutates
   the literal with a fragment. That mutation was later run and caught.

Nothing refuted, nothing deferred. Two bookkeeping notes fixed in passing: the package comment must
not name gofakeit or goal 1's grep never empties, and decision 3's `Password` guarantee is
class-completeness rather than acceptance under `High` at 8 characters, whose minimum is 10.

---

### Final gate, round 1 — request `272-20260908T131515Z`

Verdict `findings`: 2, both `significant`, both `confidence: high`, both `judgement: true` with a
populated `forced_answer` — findings the code settles rather than choices to escalate. All three axes
reviewed. Both disposed of at `9a8a8b87`.

#### Finding 1, quality: `LetterN` and `DigitN` suppressed a CSPRNG failure — **confirmed, fixed**

The claim: the package promises every generator either produces its documented shape or stops the
test; `intn` and `uuid` hold that line, and the two `stringutil` delegates did not.

Confirmed by reading the delegate. `stringutil.randomStringFromAlphabet`
(`src/core/stringutil/stringutil.go:48`) is four lines and the middle two are
`if err != nil { return "" }`; its own comment says it "preserves the package's historical contract
of returning `""` when the system CSPRNG is unavailable", which is a contract its production callers
were written against and not one `fake` should inherit. So a broken CSPRNG gave `LetterN` an empty
string, `Username()` became the constant `"user"` for every call, and the suite went green — the
exact shape of the #136 collision this package was written to close. The reviewer's `why_exhaustive`
holds: the package has four direct entropy paths, `intn` and `uuid` already panic, and every other
generator composes one of the four.

Fixed as the `forced_answer` directed, keeping the decided string-returning API: a `mustDraw` helper
panics unless the returned draw is the length asked for, and both wrappers route through it. Two
cases pin it, `TestLetterN_PanicsOnEntropyFailure` and `TestDigitN_PanicsOnEntropyFailure`, which
swap `crypto/rand.Reader` for a reader that only errors and restore it from a `defer`. No test in the
tree runs concurrently with another, so swapping the package variable for the length of one case
races with nothing; the constraint is stated in a comment above `withBrokenEntropy`.

Both mutations `MUTATION CAUGHT`:

```
where.sh mutate src/core/testutil/fake/fake.go
  --replace 'return mustDraw(stringutil.GenerateRandomLetterString(int(n)), n, "LetterN")'
  --with    'return stringutil.GenerateRandomLetterString(int(n))'
  --type modules --run TestLetterN_PanicsOnEntropyFailure          CAUGHT
where.sh mutate src/core/testutil/fake/fake.go
  --replace 'return mustDraw(stringutil.GenerateRandomNumberString(int(n)), n, "DigitN")'
  --with    'return stringutil.GenerateRandomNumberString(int(n))'
  --type modules --run TestDigitN_PanicsOnEntropyFailure           CAUGHT
```

`stringutil` itself was left alone. Its `""` is load-bearing for two production call sites that have
never handled an error return, and changing it is a production behaviour question this change has no
mandate for.

#### Finding 2, conformance: the OIDC fixture's `zoneinfo` — **regression claim refuted, improvement taken**

The claim: the hand rewrite replaced `gofakeit.TimeZoneFull()` with `"tz" + fake.LetterN(6)`, so the
fixture "**now** always carries an invalid zoneinfo value" that `ProfileValidator` rejects.

The word "now" is wrong, and here is the demonstration. `ProfileValidator`
(`src/core/validators/profile_validator.go:203`) accepts only a `Zone` present in `timezones.Get()`,
which is 423 IANA identifiers. `gofakeit.TimeZoneFull()` did not draw from IANA identifiers: read out
of the module cache at `gofakeit/v6@v6.28.0/data/datetime.go:8`, its `"full"` corpus is 108 Windows
display strings — `"(UTC-12:00) International Date Line West"`,
`"(UTC-08:00) Pacific Time (US & Canada)"`, `"(UTC+13:00) Samoa"`. Not one is in `timezones.Get()`;
`"(UTC-08:00) Pacific Time (US & Canada)"` returns `present=false`, against `present=true` for
`America/New_York` and `America/Sao_Paulo`. The library's IANA generator is a different function,
`TimeZoneRegion()`, which this fixture never called. So the fixture carried a zoneinfo the validator
rejects both before this change and after it, the rewrite preserved the behaviour exactly as the
agreement's §4.2 intended, and this is not a regression the change introduced.

The observation underneath it was still worth taking on its own merit: nothing forces a generated
value here, the fixture writes through `database.CreateUser` with no validator in the path, and
`Locale: "en-US"` immediately below was already a validated constant. `ZoneInfo` is now
`"America/New_York"`. Nothing asserts on the value; the only `zoneinfo` assertion in the file
(`token_oidc_claims_test.go:88`) checks the claim is absent when the setting is off. This is a named
deviation from the sealed plan — §4.2 step 5 and the §0 `fixture/claims` anchor both specified
`"tz" + fake.LetterN(6)` — and the anchor was re-swept, with `check-anchors.sh` passing 23 of 23. The
reviewer's second instruction, to keep the `fake` package free of a zone generator, is honoured:
nothing was added to it.

#### `unchecked` at round 1, both entries answered

- *No interactive browser pass or duplicate local four-engine run.* **Dismissed.** The branch changes
  no route, template, query, migration or schema, and CI covered all four engines at `f3b41f20`.
- *Untouched runtime implementations for fail-open storage, token and session transitions, replay,
  revocation, client binding, redirect handling and concurrency not re-exercised.* **Dismissed.** The
  reviewer checked the dependency graph of all three main packages and found the `fake` package
  unreachable from every one, which is the compiler-backed guard the design relies on. Goal 2 asks
  for a package no binary imports, and that is what was verified.

#### Tiers re-run after the fixes, at `9a8a8b87`

| Tier | Command | Coverage | Result |
|---|---|---|---|
| unit | `where.sh test --type modules` | all three modules | 5397 pass, 0 fail, ~40s |
| data | `where.sh test --type data` | sqlite, mysql, postgres, mssql | 2208 pass, 0 fail, ~30s |
| integration | `where.sh test --type integration --db sqlite` | sqlite | 1333 pass, 0 fail, ~5m |

Unit is +2 on the pre-fix tree, which is the two new entropy cases and nothing else; no test was
removed or renamed. The four `TestToken_IdToken_OIDCClaims_*` cases that run the edited fixture all
pass. Tree hashes after the fixes, with the mutations proven gone: `production b80f35d93d09`,
`unit-tests 30bf34f56dae`, `data-tests 6483c3303341`, `integration-tests 85350a2f2c09`,
`docs 5cb6d7031d45`, `all c0a6f81783ec`.

---

### Final gate, round 2 — request `272-20260908T140513Z`

Session resumed, started 2026-09-08T14:07:08Z, completed 14:16:51Z. **Verdict `clean`. Zero
findings.** All three axes read `reviewed`: conformance, quality, security. Clean rather than merely
empty, because `unchecked` held nothing the gate depended on and the verdict stated its scope. Tree
at ingest, unmoved since the request was written:

```
production b80f35d93d09   unit-tests 30bf34f56dae   data-tests 6483c3303341
integration-tests 85350a2f2c09   docs 5cb6d7031d45   all c0a6f81783ec
```

**What it ran**, which is what gives the absence of findings a stated scope. It read the whole
`f3b41f20..9a8a8b87` repair delta and the branch against `origin/main`; re-derived the inventory by
exact mechanical normalization over all 114 rewritten files and recounted 1508 former `gofakeit`
calls against 1507 current `fake` calls, the difference being round 1's fixed `"America/New_York"`;
confirmed HEAD carries the original 2894 test declarations plus the package's 16 cases and the two
new entropy cases, so nothing was dropped in the rewrite; read `mustDraw` and both regression tests
and re-ran the recorded guard-removal mutations at this tree; ran `where.sh test --type modules` and
saw both entropy cases execute and pass; ran the letter, digit and entropy cases 100 times in order
and 100 times shuffled, to show the swapped `rand.Reader` does not leak between cases; passed
`America/New_York` with `en-US` through the real `validators.ProfileValidator` in a scratch
assertion; ran `tree-hash.sh`, `check-anchors.sh`, `git diff --check`, the documentation sweep and
the three main-package dependency checks; and read `where.sh ci` at `9a8a8b87`.

Round 1's refutation was re-examined and accepted on its own evidence. No follow-ups raised; the
E.164 `phone_number` defect from round 1 stands as drafted in the previous comment.

#### `unchecked` at round 2, both entries dismissed with a reason

1. **No `go test -race` result for the two entropy cases.** Dismissed, with the premise verified
   rather than taken on trust. The container genuinely cannot run it: `go env CGO_ENABLED` is `0` and
   neither `gcc`, `cc` nor `clang` is on the path. More to the point there is nothing to instrument.
   `grep -rn 't\.Parallel()' src --include=*.go` with comment lines filtered out returns **no hits at
   all** — the two files that match, `fake_test.go` and
   `core/data/commondb/email_backfill_failclosed_test.go`, each match only inside a comment
   explaining why they do not call it, the latter's real call having come out at `ffff26d0`. No test
   in the tree runs concurrently with another, Go gives each package's tests their own process, and
   `withBrokenEntropy`'s swap of the `crypto/rand.Reader` package variable is confined to one test
   binary running one case at a time. A race detector reports races between goroutines; with no
   second goroutine there is nothing to report.
2. **Untouched runtime security families not dynamically re-exercised** — storage-error gates, token
   and session atomicity, replay, revocation, client binding, exact identifier matching, redirect
   handling. Dismissed on two independent grounds. *Reachability:* the only non-test Go file this
   branch adds is `src/core/testutil/fake/fake.go`, and
   `grep -rln 'testutil/fake' src --include=*.go | grep -v '_test.go$'` returns nothing, so no served
   binary's graph contains it and the new panic path cannot be reached from any of those families.
   *Exercise:* they were re-run regardless, by the check suite — the four `Tests /` jobs are the data
   and integration tiers on all four engines, green at HEAD.

---

### The four-engine result

`where.sh ci` at `9a8a8b87b13a37850af7eaf5240078f075e2583d`: **green, 12 of 12 jobs** —
<https://github.com/leodip/goiabada/actions/runs/34235792672>. That is where the four-engine data and
integration claim for this branch lives, and it is the strongest evidence in this record because it
names its own commit and no session can write it: `Tests / sqlite`, `Tests / mysql`,
`Tests / postgres`, `Tests / mssql`, alongside `Unit / core`, `Unit / authserver`,
`Unit / adminconsole`, `Unit / setup`, `Build validation`, `Lint`, `Vulnerabilities` and `Versions`.

Issue #230, opened 2026-08-17
Six validators' catalog messages are never asserted, only their codes


leodip:
`src/core/validators` has seven validators returning `*i18n.LocalizedError`: authorize, address,
email, identifier, password, phone and profile. Their tests assert `locErr.Code` and nothing else,
for example in `email_validator_test.go` and `address_validator_test.go`. Nothing in the package
asserts the sentence a user actually reads.

The reason is mechanical rather than a choice anyone made. Until #213 the package's `TestMain`
(`src/core/validators/test_main_test.go`) did not call `i18n.LoadBundle`, and `EnglishFallback()`
returns the message id when `defaultBundle` is nil, which is i18n's deliberate visible-miss policy.
Any assertion on rendered text would therefore have compared `validator.email.required` against
"Please enter an email address." and failed, so nobody wrote one. #213 added the `LoadBundle` call,
because the seven authorize messages are asserted on their sentences: keeping those sentences
byte-identical through the move into the catalog is exactly what proves the refusal page still says
what it always said.

So the mechanism is now in place for the other six and unused by them.

**Why it is worth closing.** A code assertion proves a validator picked the right branch. It cannot
see a catalog entry that is wrong, empty of meaning, reworded by accident, or mapped to the wrong
condition, and `TestCatalog_ParityEnPtBR` and `TestCatalog_NoEmptyValues` cannot see those either:
parity proves a key exists in both locales and non-emptiness proves it is not blank. Nothing proves
`validator.password.too_short` is about a password being too short. The concrete failure is a
catalog edit that swaps two entries in a group: every test stays green and every affected form tells
the user to fix the wrong field.

**What closing it takes.** Each of the six test files has a table with an `expectedCode` column. Add
an `expectedMessage` column and one `assert.Equal(t, tt.expectedMessage, locErr.EnglishFallback())`
per table, mirroring what `authorize_validator_test.go` does now. Roughly six tables. It needs no
production change: `TestMain` already loads the bundle since #213.

Worth doing per group rather than all at once, and worth a mutation per group afterwards (reword one
English catalog value, confirm the group's table catches it), because that mutation is the only
thing that distinguishes an assertion that reads the catalog from one that reads a duplicate of it.

**Anchor:** `src/core/validators/test_main_test.go`, the `i18n.LoadBundle` call in `TestMain`.

leodip:
Folded into #344.

[...]

PR #307, opened 2026-09-08
refactor(core): retire go-i18n, the catalogs are a flat map and a text/template

### LINKED ISSUE #273: go-i18n's plural machinery is unused: zero plural forms across 3044 catalog lines, and the rest is a map lookup
`go-i18n` exists to solve pluralization across CLDR plural categories. We have **zero plural forms**: `grep -c '^\[' core/i18n/catalogs/active.en.toml` returns 0, across 1562 lines of English and 1482 of Brazilian Portuguese. Every entry is a flat `"key" = "value"`.

## What we actually use

| Symbol | Call sites | What it does for us |
|---|---|---|
| `i18n.NewBundle` | 2 | Holds the message maps |
| `i18n.NewLocalizer` | 3 | Picks a locale, with fallback |
| `i18n.LocalizeConfig` | 3 | `MessageID` and `TemplateData` only |
| `i18n.Bundle` / `i18n.Localizer` | type refs | |

`core/i18n/i18n.go:T` is the whole story: build a `LocalizeConfig{MessageID: key}`, optionally attach `TemplateData` from `args[0].(map[string]any)`, call `Localize`, fall back to the key on error.

That is a two-level map lookup and a `text/template` execution.

## What it costs

`go.yaml.in/yaml/v3` enters the module graph solely through `go-i18n`. `BurntSushi/toml` stays either way and is worth keeping: TOML is meaningfully better than JSON for translators (comments, no escaping traps in message text), and it has zero dependencies of its own.

## What replaces it

In `core/i18n`:

- `map[string]map[string]string`, locale tag to key to message, populated by the existing TOML loaders in `raw.go` and `overrides.go`.
- `text/template` for the `TemplateData` substitution, parsed once at load and cached per message. Only messages containing `{{` need a template at all.
- `language.NewMatcher` from `golang.org/x/text/language`, which we keep regardless for the CLDR display names in `reference.go` and `locale_label.go`.
- English fallback and the visible-miss behaviour (return the key) exactly as today.

Roughly 150 lines. **The public surface does not change**: `T(ctx, key, args...)`, `Localizer(ctx)`, `LoadBundle`, `SupportedTags`, `RefineLocalizerWith*` keep their signatures, so the 28 `i18n.T` call sites, every template, and `core/i18n/error.go`'s `LocalizedError` are untouched.

## The one real objection

A flat map cannot express plural rules. If a future locale needs them (Polish has four categories, Russian four, Arabic six), we would be reimplementing CLDR plural selection or coming back to go-i18n.

Worth deciding deliberately rather than by accident. Two mitigations if we proceed:

1. English and Portuguese both have a two-form rule, so nothing today is at risk.
2. Keep the loader tolerant of a `[key]` table with `one`/`other` so adding plural support later is additive rather than a format migration.

`core/i18n/catalog_hygiene_test.go` should grow an assertion that no catalog contains a plural table, so the day someone adds one it fails loudly instead of silently rendering the key.

---

### Dependency reduction

leodip:
Closes #273.

## What this is for

`core/i18n` renders its message catalogs through `github.com/nicksnyder/go-i18n/v2`, of which it
uses one thing: the lookup of a flat `"key" = "value"` TOML file by a locale matched from
`Accept-Language`, with Go `text/template` placeholders. No plural table exists in either catalog
and no caller sets a plural count, so the library's plural machinery is dead weight, and the
package already parses every catalog a second time into a raw map for the JS bootstrap.

This change drops the dependency. The catalogs stay exactly as they are: a flat map parsed once with
`BurntSushi/toml`, locales matched with `golang.org/x/text/language` the same way go-i18n did, and
`{{.name}}` values compiled once at startup with `text/template`. Every public signature in the
package is unchanged, so no caller, template or JS file moves.

## The rule it establishes

A catalog value is a plain string. A value that is anything else, a `[table]` in particular, is
refused at startup with the file name and key, so a plural form can never render silently as one
form on a page. An empty value means "use English". And a key the active locale lacks renders the
English text, as `concepts/localization.mdx` has always said, on every path: page strings, the JS
bootstrap, and localized errors alike.

## Status

**Complete.** Both stages landed, the final review is clean, and the check suite is green on the
head commit. Ready for review.

### Stage 1: `core/i18n` renders its catalogs with a flat map and `text/template`

Landed in `df672c5b`.

One parser reads every catalog file once into `map[string]string`; `LoadBundle` merges per key with
the later file winning, builds a `language.NewMatcher` over the bundle's tags with English at index
0, and compiles every `{{`-carrying value into a `*template.Template` that is immutable for the
process lifetime, so rendering needs no lock and no cache. The parallel raw catalogs and their
second parse of every file are gone, as are `Bundle.English()` and `DefaultBundle()`, neither of
which had a caller outside the package. `go-i18n` is out of `src/core/go.mod`.

Locale resolution is go-i18n's algorithm kept verbatim — `language.ParseAcceptLanguage` per
preference, then `matcher.Match` — and `TestLocalizerFor_MatchesAsGoI18nDid` holds all 17 rows of a
probe table that ran the old library and the new matcher side by side over the same inputs,
including the ones nothing else observes: a later range winning, a `q=0` range excluded, quality
reordering the list, unparseable input falling to English, and the `ui_locales` shape whose first
tag is unsupported.

Two defects inside the rewritten functions were fixed rather than re-implemented:

* **`T` rendered the key** for a key the matched locale lacks, while `LocalizedError.Localize`,
  `Raw` and the docs all promised English. The English hop now happens on lookup, before any
  failure can turn it into the key. It was invisible on a stock install only because the hygiene
  test forces `en` and `pt-BR` to have identical key sets — it is exactly the path a self-hoster
  shipping one override catalog takes.
* **`Raw` and `T` could disagree on the locale.** `Raw` indexed by the request's primary tag string
  while `T` went through the matcher, so `Accept-Language: pt`, `pt-PT` or `fr-FR,pt;q=0.8` gave a
  pt-BR page with English JS strings. `Raw` now reads the same translator, falling back to the
  request tag only when none is attached (the `EmailContext` path).

Unit tier green on all three modules, 5408 pass / 0 fail across 61 packages. Six mutations run
against the new cases — the English hop, `Raw`'s locale source, the empty-value delete, the matcher
over all parsed tags, and `Args` in both error renderings — all caught. Data and integration are not
applicable: no query, handler, route or schema is touched.

### Stage 2: go-i18n leaves both binaries' module graphs

Landed in `9fa733b3`.

`go mod tidy` in `src/core` (a no-op, already tidy from stage 1), then `src/authserver` and
`src/adminconsole`, which is the order their manifests require: each binary module resolves its
indirect set from core's imports. That dropped `github.com/nicksnyder/go-i18n/v2 v2.6.1 // indirect`
and its two `go.sum` lines from both binaries. `go mod graph` now has no edge into go-i18n in any of
the three modules, and the retirement reaches the shipped binaries and not just the manifests:
`go list -deps ./cmd/...` names 0 `nicksnyder` packages, of 396 for `authserver` and 377 for
`adminconsole`.

**`go.yaml.in/yaml/v3` does not leave, and the issue text expects it to.** go-i18n was not its only
requirer: `stretchr/testify v1.12.1` pulls it through `testify/assert/yaml` in all three modules, so
the `// indirect` line stays in every `go.mod` while the module leaves the *production* import graph
(0 packages in `go list -deps ./cmd/...` for both binaries). Anyone reading this change as "yaml is
gone" will go looking for a line that is still there.

Unit tier green on all three modules, 5408 pass / 0 fail, 34s. Seam 5's 19 middleware cases and the
45 `core/i18n` tests are the guard that pruning the manifests pruned nothing the code still resolves.
Data and integration are not applicable: only `go.mod` and `go.sum` changed, so neither tier can
reach anything this stage touched. No new test, so no mutation to run.

### Follow-up commit: `31a96d7b`

The final review's two findings, both about test coverage rather than shipped behaviour. No
production code changed; see the Review section below.

## Testing

|---|---|---|
| Unit, all three modules + setup | locally in the dev container at each stage, and the check suite at every commit | green, 5408 pass / 0 fail across 61 packages |
| Data, four engines (sqlite, mysql, postgres, mssql) | **the check suite only** | green at `31a96d7b` |
| Integration, four engines | **the check suite only** | green at `31a96d7b` |
| Lint, Vulnerabilities, Versions, Build validation | **the check suite only** | green at `31a96d7b` |

The data and integration tiers were not run locally at any point, deliberately: this change touches
no query, handler, route, migration or schema, so neither tier can reach anything it changed. The
four-engine result is the check suite's, and it is green on the head commit —
[run 34275670015](https://github.com/leodip/goiabada/actions/runs/34275670015), all twelve jobs.

Mutation testing stood in for the coverage the absent tiers would not have provided anyway: ten
mutations across the two stages and the review round, every one caught.

None in substance. Two decisions were added during the run rather than at the interview, both at the
plan gate and both settled against the code without escalating:

* **Decision 8** — the planned `type Localizer` and the retained `func Localizer` are a Go
  redeclaration. The getter keeps its name (as section 6 had already fixed) and the struct is
  `Translator`. Invisible outside the package.
* **Decision 9** — an empty value in a later catalog file *deletes* the key rather than being
  skipped at parse, which is what decision 7's own worked case requires.

No stage was narrowed, nothing in section 2's goal went unbuilt, and no scope was added beyond it.

## For the release notes

**Only relevant to a self-hoster using `GOIABADA_I18N_OVERRIDES_DIR`.** A stock install is
unaffected: the shipped `en` and `pt-BR` catalogs are held to identical key sets by a hygiene test,
so none of the three changes below can be observed without an override catalog.

* **A non-string value in a catalog now stops the server.** A `[table]` — the shape a go-i18n plural
  form took — fails `LoadBundle` at startup with the offending file name and key, so the process
  does not start. Previously go-i18n accepted it and could render one plural form for every count.
  If you carry a plural table in an override catalog, flatten it to a single `"key" = "value"` line
  before upgrading, or the server will refuse to boot. There is no plural support to move it to;
  that is the point of the change.
* **A key missing from your override catalog now renders the English text, not the key name.** This
  is what `concepts/localization.mdx` has always documented, and it is now true on every path. If
  you were relying on seeing the raw key to spot untranslated strings, that signal is gone — diff
  your catalog against `active.en.toml` instead.
* **An empty value in an override catalog now means "use English".** Previously it left whatever the
  embedded catalog held for that locale, so a blank line in an override of a shipped locale kept the
  shipped translation. Now it removes the key and falls through to English.

Nothing else changes for an operator: no configuration, no schema, no endpoint, no wire behaviour.

## Review

Reviewed by `openai-codex/gpt-5.6-sol` at effort `xhigh`. Three rounds: one at the plan gate, two at
the final gate over the whole branch. No stage was reviewed on its own — `LEO_REVIEW_MODE=end`, so
the final gate is the only reading the code got.

|---|---|---|---|---|
| 1 | plan | 4 | 1 | all confirmed, all resolved in the plan before any code was written |
| 1 | final | 2 | 0 | both confirmed, both fixed, `31a96d7b` |
| 2 | final | 0 | 0 | clean, every axis reviewed, `unchecked` empty |

- **blocking, conformance (plan):** `func Localizer` and `type Localizer` cannot coexist in one Go
  package namespace. Resolved as decision 8 before implementation: the getter keeps its name, the
  struct is `Translator`.
- **significant, conformance (plan):** an empty override value would have been discarded at parse,
  leaving the embedded translation instead of falling through to English. Resolved as decision 9.
- **significant, conformance (plan):** the locale parity table was not pinned — a first-range
  matcher passed every planned case. Resolved by a table-driven test over all 17 probe rows.
- **significant, quality (plan):** parameterized `LocalizedError` rendering was not owned by its
  seam; two mutations dropping template data survived. Resolved by `Args` cases on all three paths.
- **significant, quality (final):** the template *execution*-error branch of `renderOrMiss` had no
  regression test — changing it to return the raw value left the suite green. Fixed in `31a96d7b`:
  `TestT_TemplatedValueThatFailsToExecuteRendersTheKey`.
- **significant, quality (final):** every pre-`LoadBundle` rendering fallback could regress
  unnoticed, because `TestMain` loads the bundle before any test runs. Fixed in `31a96d7b`:
  `TestRenderingBeforeLoadBundle_ResolvesEveryKeyToItself`.

Neither final-gate finding was a defect in shipped behaviour: the reviewer confirmed the code
returns the right answer in both cases, so `31a96d7b` changed no production code. Nothing was
refuted in any round.

Full account, with the mutation tables and the evidence,
[in this comment](https://github.com/leodip/goiabada/pull/307#issuecomment-5591652340).

## Closing comments

* [Deferred decisions](https://github.com/leodip/goiabada/pull/307#issuecomment-5591631377) — none;
  nothing is waiting on you.
* [Follow-ups, fold-ins and ceilings](https://github.com/leodip/goiabada/pull/307#issuecomment-5591635198)
  — no issues to file, two defects folded in, no ceilings left behind.
* [Review account](https://github.com/leodip/goiabada/pull/307#issuecomment-5591652340) — all three
  rounds in full.

### DIFF
--- a/src/adminconsole/go.mod
+++ b/src/adminconsole/go.mod
@@ -35,7 +35,6 @@ require (
 	github.com/microsoft/go-mssqldb v1.10.0 // indirect
 	github.com/mileusna/useragent v1.3.5 // indirect
 	github.com/ncruces/go-strftime v1.0.0 // indirect
-	github.com/nicksnyder/go-i18n/v2 v2.6.1 // indirect
 	github.com/remyoudompheng/bigfft v0.0.0-20230129092748-24d4a6f8daec // indirect
 	github.com/shopspring/decimal v1.4.0 // indirect
 	github.com/stretchr/objx v0.5.3 // indirect
--- a/src/adminconsole/go.sum
+++ b/src/adminconsole/go.sum
@@ -73,8 +73,6 @@ github.com/mileusna/useragent v1.3.5 h1:SJM5NzBmh/hO+4LGeATKpaEX9+b4vcGg2qXGLiNG
 github.com/mileusna/useragent v1.3.5/go.mod h1:3d8TOmwL/5I8pJjyVDteHtgDGcefrFUX4ccGOMKNYYc=
 github.com/ncruces/go-strftime v1.0.0 h1:HMFp8mLCTPp341M/ZnA4qaf7ZlsbTc+miZjCLOFAw7w=
 github.com/ncruces/go-strftime v1.0.0/go.mod h1:Fwc5htZGVVkseilnfgOVb9mKy6w1naJmn9CehxcKcls=
-github.com/nicksnyder/go-i18n/v2 v2.6.1 h1:JDEJraFsQE17Dut9HFDHzCoAWGEQJom5s0TRd17NIEQ=
-github.com/nicksnyder/go-i18n/v2 v2.6.1/go.mod h1:Vee0/9RD3Quc/NmwEjzzD7VTZ+Ir7QbXocrkhOzmUKA=
 github.com/pkg/browser v0.0.0-20240102092130-5ac0b6a4141c h1:+mdjkGKdHQG3305AYmdv1U2eRNDiU2ErMBj1gwrq8eQ=
 github.com/pkg/browser v0.0.0-20240102092130-5ac0b6a4141c/go.mod h1:7rwL4CYBLnjLxUqIJNnCWiEdr3bn6IUYi15bNlnbCCU=
 github.com/pkg/errors v0.9.1 h1:FEBLx1zS214owpjy7qsBeixbURkuhQAwrK5UwLGTwt4=
--- a/src/authserver/go.mod
+++ b/src/authserver/go.mod
@@ -41,7 +41,6 @@ require (
 	github.com/mattn/go-isatty v0.0.24 // indirect
 	github.com/mileusna/useragent v1.3.5 // indirect
 	github.com/ncruces/go-strftime v1.0.0 // indirect
-	github.com/nicksnyder/go-i18n/v2 v2.6.1 // indirect
 	github.com/remyoudompheng/bigfft v0.0.0-20230129092748-24d4a6f8daec // indirect
 	github.com/rogpeppe/go-internal v1.16.0 // indirect
 	github.com/shopspring/decimal v1.4.0 // indirect
--- a/src/authserver/go.sum
+++ b/src/authserver/go.sum
@@ -86,8 +86,6 @@ github.com/mileusna/useragent v1.3.5 h1:SJM5NzBmh/hO+4LGeATKpaEX9+b4vcGg2qXGLiNG
 github.com/mileusna/useragent v1.3.5/go.mod h1:3d8TOmwL/5I8pJjyVDteHtgDGcefrFUX4ccGOMKNYYc=
 github.com/ncruces/go-strftime v1.0.0 h1:HMFp8mLCTPp341M/ZnA4qaf7ZlsbTc+miZjCLOFAw7w=
 github.com/ncruces/go-strftime v1.0.0/go.mod h1:Fwc5htZGVVkseilnfgOVb9mKy6w1naJmn9CehxcKcls=
-github.com/nicksnyder/go-i18n/v2 v2.6.1 h1:JDEJraFsQE17Dut9HFDHzCoAWGEQJom5s0TRd17NIEQ=
-github.com/nicksnyder/go-i18n/v2 v2.6.1/go.mod h1:Vee0/9RD3Quc/NmwEjzzD7VTZ+Ir7QbXocrkhOzmUKA=
 github.com/pkg/browser v0.0.0-20240102092130-5ac0b6a4141c h1:+mdjkGKdHQG3305AYmdv1U2eRNDiU2ErMBj1gwrq8eQ=
 github.com/pkg/browser v0.0.0-20240102092130-5ac0b6a4141c/go.mod h1:7rwL4CYBLnjLxUqIJNnCWiEdr3bn6IUYi15bNlnbCCU=
 github.com/pkg/errors v0.9.1 h1:FEBLx1zS214owpjy7qsBeixbURkuhQAwrK5UwLGTwt4=
--- a/src/core/go.mod
+++ b/src/core/go.mod
@@ -15,7 +15,6 @@ require (
 	github.com/jackc/pgx/v5 v5.10.0
 	github.com/microsoft/go-mssqldb v1.10.0
 	github.com/mileusna/useragent v1.3.5
-	github.com/nicksnyder/go-i18n/v2 v2.6.1
 	github.com/pkg/errors v0.9.1
 	github.com/pquerna/otp v1.5.0
 	github.com/stretchr/testify v1.12.1
--- a/src/core/go.sum
+++ b/src/core/go.sum
@@ -76,8 +76,6 @@ github.com/mileusna/useragent v1.3.5 h1:SJM5NzBmh/hO+4LGeATKpaEX9+b4vcGg2qXGLiNG
 github.com/mileusna/useragent v1.3.5/go.mod h1:3d8TOmwL/5I8pJjyVDteHtgDGcefrFUX4ccGOMKNYYc=
 github.com/ncruces/go-strftime v1.0.0 h1:HMFp8mLCTPp341M/ZnA4qaf7ZlsbTc+miZjCLOFAw7w=
 github.com/ncruces/go-strftime v1.0.0/go.mod h1:Fwc5htZGVVkseilnfgOVb9mKy6w1naJmn9CehxcKcls=
-github.com/nicksnyder/go-i18n/v2 v2.6.1 h1:JDEJraFsQE17Dut9HFDHzCoAWGEQJom5s0TRd17NIEQ=
-github.com/nicksnyder/go-i18n/v2 v2.6.1/go.mod h1:Vee0/9RD3Quc/NmwEjzzD7VTZ+Ir7QbXocrkhOzmUKA=
 github.com/pkg/browser v0.0.0-20240102092130-5ac0b6a4141c h1:+mdjkGKdHQG3305AYmdv1U2eRNDiU2ErMBj1gwrq8eQ=
 github.com/pkg/browser v0.0.0-20240102092130-5ac0b6a4141c/go.mod h1:7rwL4CYBLnjLxUqIJNnCWiEdr3bn6IUYi15bNlnbCCU=
 github.com/pkg/errors v0.9.1 h1:FEBLx1zS214owpjy7qsBeixbURkuhQAwrK5UwLGTwt4=
--- a/src/core/i18n/catalog_hygiene_test.go
+++ b/src/core/i18n/catalog_hygiene_test.go
@@ -3,32 +3,48 @@ package i18n
 import (
 	"testing"
 
-	"github.com/BurntSushi/toml"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
 
-// loadCatalogFlat parses an embedded catalog TOML into a flat key->value map.
-// The catalogs use quoted dotted keys (e.g. "auth.pwd.title"), which TOML
-// treats as literal single keys, so a flat map[string]string is correct.
+// loadCatalogFlat parses an embedded catalog TOML into a flat key->value map,
+// through the loader's own parser, so the assertions below hold the embedded
+// catalogs to exactly the rules LoadBundle enforces at startup. The catalogs
+// use quoted dotted keys (e.g. "auth.pwd.title"), which TOML treats as literal
+// single keys, so a flat map[string]string is correct.
 func loadCatalogFlat(t *testing.T, name string) map[string]string {
 	t.Helper()
 	data, err := embeddedCatalogs.ReadFile("catalogs/" + name)
 	require.NoError(t, err, "reading %s", name)
-	m := map[string]string{}
-	require.NoError(t, toml.Unmarshal(data, &m), "parsing %s", name)
+	_, m, err := parseCatalog(name, data)
+	require.NoError(t, err, "parsing %s", name)
 	require.NotEmpty(t, m, "%s parsed to zero keys", name)
 	return m
 }
 
-// TestCatalog_NoEmptyValues guards against the "empty_prefix" class of bug:
-// go-i18n drops a message whose value is "" and then reports it identically to
-// a missing key, so T() falls back to the visible-miss policy and renders the
-// raw key onto the page. An intentionally-empty slot can therefore never work.
+// TestParseCatalog_RefusesATableValue pins decision 3 of #273: a [table]
+// section, which is how go-i18n spelled plural forms, fails the load naming
+// the file and the key rather than being tolerated into one form rendered for
+// every count.
+func TestParseCatalog_RefusesATableValue(t *testing.T) {
+	const table = `[greeting]
+other = "hi"
+`
+	_, _, err := parseCatalog("active.xx.toml", []byte(table))
+	require.Error(t, err)
+	assert.Contains(t, err.Error(), "active.xx.toml")
+	assert.Contains(t, err.Error(), `"greeting"`)
+}
+
+// TestCatalog_NoEmptyValues guards against the "empty_prefix" class of bug: an
+// empty value means "no translation here", so the loader removes the key from
+// that locale and T() renders English. In the English catalog itself that
+// leaves nothing to render but the key. An intentionally-empty slot can
+// therefore never work.
 func TestCatalog_NoEmptyValues(t *testing.T) {
 	for _, name := range []string{"active.en.toml", "active.pt-BR.toml"} {
 		for k, v := range loadCatalogFlat(t, name) {
-			assert.NotEmptyf(t, v, "%s: key %q has an empty value; go-i18n treats it as missing and leaks the key in the UI", name, k)
+			assert.NotEmptyf(t, v, "%s: key %q has an empty value; the loader treats it as absent, so the key renders English or leaks into the UI", name, k)
 		}
 	}
 }
--- a/src/core/i18n/catalogs/active.en.toml
+++ b/src/core/i18n/catalogs/active.en.toml
@@ -3,8 +3,8 @@
 # Every key referenced by code or templates must have an entry here.
 # Other locales are translations of this file.
 #
-# Format: simple "key" = "value" (go-i18n parses dotted keys via TOML
-# string keys). Use [section]\nother = "..." for plural forms when needed.
+# Format: simple "key" = "value". Values are plain strings; {{.name}}
+# placeholders render with Go text/template. A [table] is refused at startup.
 
 # --- Common shared strings ---------------------------------------------------
 
--- a/src/core/i18n/error.go
+++ b/src/core/i18n/error.go
@@ -2,8 +2,6 @@ package i18n
 
 import (
 	"context"
-
-	"github.com/nicksnyder/go-i18n/v2/i18n"
 )
 
 // LocalizedError carries a stable error code and template arguments. The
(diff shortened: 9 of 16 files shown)


leodip:
## Deferred decisions

**None.** Nothing in this run proceeded on an assumption that is still waiting on you.

All nine decisions in the agreement's section 3 are `Decided`. Seven were settled at the interview,
before any code was written. Two were raised by the run itself, at the plan gate, and settled there
rather than escalated, because neither had an option you could have observed:

- **Decision 8 — `func Localizer` and `type Localizer` cannot coexist.** Go has one package
  namespace for types and functions, so the planned struct and the retained getter were a
  redeclaration; the plan reviewer type-checked the pair and got `Localizer redeclared in this
  block`. Decision 6 had already fixed that the getter keeps its name, so the only free choice was
  what to call the struct: it is `Translator`. Nothing outside `core/i18n` calls the getter or names
  the type, so the choice is invisible past the package boundary either way.
- **Decision 9 — an empty value in a later file deletes the key, it is not skipped.** Decision 7
  had already said a blank line in an override file means "use English", and the plan's own worked
  case pins it. The plan's first parser kept the embedded translation instead, which failed that
  case. `parseCatalog` now keeps empty strings and the per-key merge deletes the key.

Both are recorded in full in `docs/issue-273-retire-go-i18n/agreement.md` section 3, with what was
rejected and why.

leodip:

### Follow-ups to file

**None.** Neither stage found a real defect belonging to a different change, so there is nothing
here for you to file.

That is not for want of finding things. This work found two defects — but both were found at spec
time, put through the four fold-in tests, and fixed here, because both lived inside functions the
change was rewriting anyway. Drafting either would have meant filing an issue against code that no
longer exists.

Duplicate search re-run at the finish, in case the tracker moved while the run was working:
`gh issue list --state all --search "i18n locale translation fallback"` returns nothing, open or
closed. #230 remains the only open i18n-adjacent issue and it asserts validator catalog messages,
which this change does not touch.

Both were sealed into the agreement as decisions before the first commit, so they are part of the
goal rather than scope the run took on quietly. Both are in the pull request body too.

1. **`T` rendered the key, not English, for a key the matched locale lacks.** Agreement decision 2.
   Landed in `df672c5b`. `LocalizedError.Localize`, `Raw` and `concepts/localization.mdx` all
   promised English; `T` turned go-i18n's fallback-with-error into the key. It rides on `T`, which
   was rewritten regardless; the fix has one shape (matched locale, then English, then the key); it
   is the whole fix; and section 2 already claimed the English fallback. Zero extra production
   lines. It was invisible on a stock install only because the hygiene test forces `en` and `pt-BR`
   to have identical key sets — it is exactly the path a self-hoster shipping one override catalog
   takes.
2. **`Raw` and `T` could disagree on the locale.** Agreement decision 5. Landed in `df672c5b`.
   `Raw` indexed the parallel raw catalogs by the request's primary tag string while `T` went
   through the matcher, so `Accept-Language: pt`, `pt-PT` or `fr-FR,pt;q=0.8` rendered a pt-BR page
   with English JS bootstrap strings. The flat map replaces those raw catalogs, so `Raw` was
   rewritten regardless; one shape, whole fix, section 2 unchanged. A few lines and one test case.

Both are pinned by tests that fail against the old behaviour.

**None.** `git diff --unified=0 main...HEAD | grep '^+.*ceiling:'` returns nothing.

The change ships no shortcut with a known limit: the template compile is eager and covers every
`{{`-carrying value, the matcher is go-i18n's algorithm kept verbatim, and the flat map is the
catalog format's actual shape rather than a simplification of it. The one deliberate limit — no
plural support — is not a ceiling but the point of the issue, and it is enforced loudly rather than
silently: a `[table]` in a catalog fails `LoadBundle` with the file name and the key, so the process
does not start.

leodip:
## Review account

Under `LEO_REVIEW_MODE=end` no stage was reviewed on its own, so the final gate is the only reading
this code got. `.review/` is gitignored and overwritten at the next gate, so this comment is where
the verdicts survive.

Three rounds ran in total: one at the plan gate, before any code existed, and two at the final gate
over the whole branch.

---

### Plan gate, round 1 — 4 findings, all resolved in the plan

Reviewer `openai-codex/gpt-5.6-sol`, effort `xhigh`, fresh session, request
`273-20260908T185053Z`, `type: plan-review`. Verdict **findings**, four. All three axes reviewed;
`unchecked` held only the three things a plan review cannot reach without code (the implementation
itself, stage 2's tidied module graphs, a dynamic test of the renderer), each dismissed to the final
review.

1. **Empty values discarded before they can override an embedded translation** — conformance,
   significant. **Confirmed**, resolved as agreement decision 9: `parseCatalog` keeps empty strings
   and the per-key merge deletes the key. *Evidence:* the reviewer's scratch model of the planned
   parser failed the plan's own worked case (an override `active.pt-BR.toml` carrying
   `"auth.pwd.title" = ""` must render `Login`, and the planned parser left the embedded `Entrar`),
   and decision 7 had already said a blank override line means English.
2. **The locale parity table was not pinned** — conformance, significant. **Confirmed**, resolved:
   the stage gained a table-driven case over all 17 rows of the probe table. *Evidence:* a
   first-range matcher — a plausible wrong implementation — passed every case the plan had planned
   and got `fr-FR,pt;q=0.8` wrong.
3. **Parameterized `LocalizedError` rendering not owned by its seam** — quality, significant.
   **Confirmed**, resolved: the stage gained `Args` cases on `EnglishFallback`, `Error` and a pt-BR
   `Localize`. *Evidence:* two reviewer mutations dropping template data survived the core unit
   tier.
4. **`func Localizer` and `type Localizer` cannot coexist** — conformance, **blocking**, marked
   needs-human. **Confirmed**, resolved without escalating as decision 8: the getter keeps its name,
   the struct is `Translator`. *Evidence:* the reviewer type-checked the pair and got `Localizer
   redeclared in this block`. Not escalated because nothing outside the package calls one or names
   the other, so no option was observable, and decision 6 had already fixed which name survives.

No production code moved: nothing was built yet. Stage 1 owed the final review a mutation per
resolved finding.

---

### Final gate, round 1 — 2 findings, both confirmed, both fixed in `31a96d7b`

Reviewer `openai-codex/gpt-5.6-sol`, effort `xhigh`, fresh session, request
`273-20260908T200646Z`, round 1 of max 4. Scope: the whole branch against `origin/main`, all three
axes, all reported `reviewed`. Verdict **findings**, two, both axis `quality`, both severity
`significant`, `confidence: high`, `judgement: false` — settled by the code rather than choices to
escalate. No follow-ups raised.

Both findings are about the same thing: behaviour stage 1 implemented and documented that the suite
could not tell had changed. **Neither is a defect in shipped behaviour** — the reviewer confirmed
the current code returns the right answer in both cases with fresh external assertions — so this
round changed no production code. `tree-hash.sh --split` gives `production ff6e3e305019` before and
after, identical to the value the review read.

**Finding 1 — the template execution-error outcome has no regression test. Confirmed, fixed in
`31a96d7b`.** `Translator.renderOrMiss` has five outcome branches: key missing, value that failed to
parse, plain string, template executed, template execution failed. The reviewer showed the fifth was
mutable with the suite green: in a scratch copy it changed the `Execute` error arm from
`return key, true` to `return m.raw, true` and `go test ./i18n` still passed. Verified here by
reading the arm and re-running the mutation, which now reports `MUTATION CAUGHT`. The fix is
`TestT_TemplatedValueThatFailsToExecuteRendersTheKey`: an override whose value is `{{.x.Y}}`,
rendered with `map[string]any{"x": 1}`, so the field lookup fails at execution rather than at parse,
and `T` must return `auth.pwd.title`. That is the branch separating an execution failure from
missing data, which `missingkey=default` renders as `<no value>` without erroring.

**Finding 2 — every pre-`LoadBundle` rendering fallback can regress with the suite green.
Confirmed, fixed in `31a96d7b`.** The package `TestMain` calls `LoadBundle` before any test in
`core/i18n` runs, so no test ever reached a `defaultBundle == nil` arm. The reviewer changed three
at once in a scratch copy — `Localizer` to return `nil`, `Raw`'s and `EnglishFallback`'s to return
`""` — and the whole package suite passed. Verified by reading the three arms and mutating each
separately, all now caught. The fix is `TestRenderingBeforeLoadBundle_ResolvesEveryKeyToItself`,
which saves `defaultBundle`, nils it under `t.Cleanup`, and asserts the four public rendering
surfaces (`T`, `Raw`, `LocalizedError.EnglishFallback`, `LocalizedError.Localize`) plus a non-nil
`Localizer`, which `T` dereferences without checking. Note `lookup` has a nil-receiver guard, so
`Localizer` returning nil would *not* panic — `T` would still render the key — which is exactly why
that mutation needs the explicit `require.NotNil`, and why the gap was invisible.

**Bookkeeping, fixed in passing in the same commit.** The reviewer noted that
`TestLocalizerFor_MatchesAsGoI18nDid`'s comment named a probe file that does not exist in a fresh
clone (the agreement directory is gitignored). The parity table is self-contained in the test, so
the comment now says the rows were derived by running go-i18n's localizer and this one side by side,
without naming the file.

**Mutations run this round**, all through `where.sh mutate --type core`, which restores from a trap
and compares by hash:

| # | File | Replacement | Caught by | Result |
|---|---|---|---|---|
| 1 | `i18n.go` | `Execute` error arm `return key, true` → `return m.raw, true` | `TestT_TemplatedValueThatFailsToExecuteRendersTheKey` | CAUGHT |
| 2 | `i18n.go` | `Localizer` pre-load arm → `return nil` | `TestRenderingBeforeLoadBundle_ResolvesEveryKeyToItself` | CAUGHT |
| 3 | `raw.go` | `Raw` pre-load arm `return key` → `return ""` | `TestRenderingBeforeLoadBundle_ResolvesEveryKeyToItself` | CAUGHT |
| 4 | `error.go` | `EnglishFallback` pre-load arm `return e.Code` → `return ""` | `TestRenderingBeforeLoadBundle_ResolvesEveryKeyToItself` | CAUGHT |

`git status --porcelain` after the last one listed `src/core/i18n/i18n_test.go` alone, and
`production` was unchanged at `ff6e3e305019`, so no mutation survived on disk. `unit-tests` moved
`f122f8eea3f5` → `f29f474da99d`, which is the one test file.

**Tiers.** Core unit tier green, run through `where.sh test --type core` inside the dev container,
whole tier not narrowed, ~90s; both new cases appear as `--- PASS`. No data or integration tier
applies: this round changed no production code, no handler, no route, no schema.

**`unchecked`, three entries, all answered.**

1. *No interactive browser pass inspected the final `window.i18n` script.* **Dismissed.** The
   bootstrap's escaping is untouched — it still calls `i18n.Raw` through the same
   `SetEscapeHTML(true)` encoder — and what `Raw` returns for a given locale is pinned at seam 3
   through real `MiddlewareLocale`, including the `{{param}}` placeholders surviving verbatim. A
   browser would observe the strings the middleware test already asserts.
2. *No local data or integration tier was rerun.* **Covered.** CI ran both tiers on all four engines
   at `9fa733b3`, green. `core/i18n` has no database, route, handler, migration or schema.
3. *The security classes (PKCE, tokens, sessions, authorization, rate limiting, one-shot claims,
   revocation, oracles, audit) were not dynamically exercised.* **Dismissed.** The reviewer checked
   their absence from the fanout statically and reported the security axis reviewed; the change's
   whole surface is catalog loading and string rendering. Nothing in the diff reaches a credential
   comparison, a storage predicate or an audit path.

---

### Final gate, round 2 — clean, 0 findings

Reviewer `openai-codex/gpt-5.6-sol`, effort `xhigh`, resumed session, request
`273-20260908T203639Z`, round 2 of max 4. `axis_coverage` reports conformance, quality and security
all `reviewed`; `unchecked` is **empty**, which asserts full coverage. Verdict **clean**: nothing to
fix, nothing to refute, nothing deferred. Nothing in the tree was touched.

A clean verdict is only worth its stated execution. The reviewer reports having run:
`review-diff.sh 9fa733b3` for the complete round 1 repair commit; `agreement.sh digest` and this
closing account; `tree-hash.sh --split` and `check-anchors.sh`; `git diff --check`, changed-file
fanout, production-hash, status and diff-checksum checks; `rg` checks for `t.Parallel`,
`defaultBundle` assignments, cleanup ordering, the new test assertions and security-surface
identifiers; and `where.sh ci HEAD`, rechecked until all twelve jobs were green at `31a96d7b`.

> **Reviewer's note:** Clean. The two round 1 coverage findings are correctly pinned by focused
> tests and four caught mutations. Production is unchanged, and all CI jobs are green at HEAD.

Tree at ingest: `production ff6e3e305019`, `unit-tests f29f474da99d`, `data-tests 6483c3303341`,
`integration-tests 85350a2f2c09`, `docs 5cb6d7031d45`, `all 060a15f1834f`.

**Nothing was refuted in any round.** Every finding across all three rounds was confirmed and fixed.

Issue #36, opened 2025-11-29
Store AES encryption key in environment variable instead of database


leodip:
Problem

  The AES-256 encryption key used to encrypt sensitive data (client secrets, SMTP passwords, etc.) is currently generated during database
  seeding and stored in plaintext in the settings table (aes_encryption_key column).

  This creates a security concern: if the database is compromised, an attacker obtains both the encrypted data AND the key to decrypt it,
  rendering the encryption ineffective.

  Current flow:
  Database compromise → attacker gets encrypted client secrets + AES key → full decryption possible

  Current Implementation

  The key is generated in database_seeder.go and stored in the database:

  // database_seeder.go:63
  encryptionKey := securecookie.GenerateRandomKey(32)

  // database_seeder.go:357
  settings := &models.Settings{
      AESEncryptionKey: encryptionKey,  // stored in DB
  }

  Meanwhile, goiabada-setup already generates other secrets externally (session keys, OAuth client secret) and passes them via environment
  variables — but the AES key is not handled this way.

  Proposed Solution

  Align AES encryption key handling with the existing pattern used for session keys:

  1. Add to goiabada-setup — Generate GOIABADA_AES_ENCRYPTION_KEY alongside other secrets
  2. Add to Config struct — Include AESEncryptionKey field in goiabada-setup/main.go
  3. Update output templates — Add the env var to docker-compose, Kubernetes manifests, and .env files
  4. Add to core config — Read GOIABADA_AES_ENCRYPTION_KEY in src/core/config/config.go
  5. Modify database seeder — Accept key from config instead of generating it
  6. Remove from database — Drop aes_encryption_key column from settings table

  Changes Required

  src/cmd/goiabada-setup/main.go:
  // Add to credential generation (around line 601)
  aesEncryptionKey := generateHexKey(32)
  printSuccess("AES encryption key generated")

  // Add to Config struct
  type Config struct {
      // ... existing fields ...
      AESEncryptionKey string
  }

  Output templates (docker-compose, k8s, .env):
  GOIABADA_AES_ENCRYPTION_KEY: ${config.AESEncryptionKey}

  src/core/config/config.go:
  AESEncryptionKey string `env:"GOIABADA_AES_ENCRYPTION_KEY"`

  src/core/data/database_seeder.go:
  - Remove encryptionKey := securecookie.GenerateRandomKey(32)
  - Accept key from config/parameter instead

  Migration for Existing Installations

  Add startup validation with three scenarios:

[...]

PR #308, opened 2026-09-08
refactor(core): retire go-simple-mail, the sender speaks net/smtp

### LINKED ISSUE #274: go-simple-mail is 40 lines of SMTP plus a DKIM signer we never call
`xhit/go-simple-mail/v2` is used in exactly one file, `core/communication/email_sender.go`, for about 40 lines of work. It brings `toorop/go-dkim` with it, which we never call.

## What we use

```go
server := mail.NewSMTPClient()          // host, port, username, password
server.Encryption = mail.EncryptionSSLTLS | STARTTLS | None
smtpClient, err := server.Connect()
email := mail.NewMSG().SetFrom(...).AddTo(...).SetSubject(...)
email.SetBody(mail.TextHTML, input.HtmlBody)
email.Send(smtpClient)
```

No attachments, no CC/BCC, no multipart alternative, no DKIM signing, no connection pooling, no templates.

## What replaces it

`net/smtp` + `crypto/tls` + `mime`:

| Today | Replacement |
|---|---|
| `EncryptionSSLTLS` (implicit TLS, port 465) | `tls.Dial("tcp", addr, cfg)` then `smtp.NewClient(conn, host)` |
| `EncryptionSTARTTLS` (port 587) | `smtp.Dial(addr)` then `client.StartTLS(cfg)` |
| `EncryptionNone` | `smtp.Dial(addr)` |
| `server.Username` / `server.Password` | `smtp.PlainAuth("", user, pass, host)` |
| `NewMSG().SetFrom/AddTo/SetSubject` | Write the headers directly |
| `SetBody(mail.TextHTML, body)` | `MIME-Version: 1.0` + `Content-Type: text/html; charset=UTF-8` |
| `Send` | `client.Mail`, `client.Rcpt`, `client.Data`, write, close, `client.Quit` |

Roughly 120 lines including error handling.

## Two things to get right

**1. Encoded-word headers (RFC 2047).** `settings.SMTPFromName` is operator-editable and the subject lines come from the localized catalogs, so both can contain non-ASCII. `mime.QEncoding.Encode("UTF-8", s)` handles it; forgetting it means accented names and Portuguese subjects arrive as mojibake. go-simple-mail does this for us today, so it is a real regression risk rather than a theoretical one.

**2. `net/smtp` is frozen.** It accepts no new features, ever. That is fine for what we do (it already supports STARTTLS, PLAIN auth and the full command sequence) but it means if we ever want OAuth2 SMTP auth (XOAUTH2, which Gmail and Microsoft 365 increasingly require over app passwords) we implement `smtp.Auth` ourselves. Worth knowing before committing, since that is a plausible future request rather than an exotic one.

## Test coverage

There is no test for `EmailSender` today because it needs an SMTP server. A replacement should come with one: `net/smtp` talks to a `net.Listener`, so a fake SMTP server in-process that asserts on the wire bytes is straightforward and would cover all three encryption modes plus the header encoding.

---

### Dependency reduction

leodip:
Closes #274.

## What this is for

`src/core/communication/email_sender.go` sent mail through `github.com/xhit/go-simple-mail/v2`, a
library that dragged `github.com/toorop/go-dkim` into all three module graphs for a DKIM signer the
server never calls. The sender is about 40 lines of SMTP, and the standard library (`net/smtp`,
`crypto/tls`, `net/mail`, `mime`, `mime/quotedprintable`) covers all of it. This change rewrites
the sender on the standard library and drops both modules.

The rewrite also fixes what the library did wrong on the wire:

- With encryption `none` and credentials, the password went out as cleartext `AUTH PLAIN`. Now
  PLAIN and LOGIN are offered only over TLS or to a localhost address; otherwise the send fails with
  an error naming the encryption setting.
- With encryption `starttls` against a server not offering STARTTLS, the library fell back to
  cleartext. Now the send fails and says the server did not offer STARTTLS.
- With credentials configured and no `AUTH` advertised, the library silently sent unauthenticated.
  Now the send fails and says the server offers no authentication.
- A from-name containing a comma, a quote or angle brackets failed every send, because the name was
  concatenated unquoted and parsed. The `From` header is now built with `net/mail.Address`, which
  quotes or encodes the name.
- Every message now carries a `Message-ID`.

The rule this establishes: the password leaves the process only on a TLS connection or to a
localhost address, a connection the operator asked to be STARTTLS is TLS or the send fails, and
every header value is either RFC 2047 encoded or parsed by `net/mail`, so no setting or catalog
string can add a header.

The `EmailSender` interface, its mock, its six callers and the settings surface are unchanged,
with one addition: the settings API now bounds the SMTP password at 256 characters, the one SMTP
field that had no bound (see *Stage 3*). Out of scope: XOAUTH2 (fails today, stays failing),
CC/BCC/attachments, DKIM, a plain-text alternative part.

## Status

**Complete.** All three stages landed. Stages 1 and 2 are traced and had the single code review
over two rounds with all ten findings resolved and none deferred, green on `d98ffda9` across all
twelve jobs ([run](https://github.com/leodip/goiabada/actions/runs/34295379281)). Stage 3 was
folded in afterwards at the author's request and had a separate review, three findings, all fixed
in `822f1399`. The check suite is green on `cb266e43`
([run](https://github.com/leodip/goiabada/actions/runs/34296776770)) and running on the head
`822f1399` ([run](https://github.com/leodip/goiabada/actions/runs/34297753687)).

### Stage 1: the sender on `net/smtp`, the fake server and its table

`e4a10031`. `email_sender.go` rewritten on `net/smtp`, `crypto/tls`, `net/mail`, `mime` and
`mime/quotedprintable` (335 lines, from 78), plus `smtp_login_auth.go` for the AUTH LOGIN
mechanism the standard library does not ship and smtp.office365.com is the only password mechanism
of. `src/core/go.mod` and `go.sum` lost go-simple-mail and go-dkim; the other two modules' indirect
lines are stage 2's, because a stale indirect requirement still builds and core has to be tidied
first.

Tests: `smtp_fake_test.go` is an in-process SMTP server that generates its certificate at test
time, records every client line and the DATA block, and speaks STARTTLS, implicit TLS, AUTH PLAIN,
AUTH LOGIN and AUTH CRAM-MD5. 36 rows over it in 12 test functions cover the three encryption
modes, mechanism choice in client preference rather than server order, the four fail-closed
refusals, an unknown authority and a wrong hostname on both TLS paths, the message down to header
order, RFC 2047 encoding, 78-byte folding, the `Message-ID`, dot-stuffing and two CRLF-injection
attempts, and the conversation deadline. `TestSendEmail` against the dev container's mailpit is
untouched and stays as the one real delivery test.

Verification: unit tier green on all three modules, 1480 pass 0 fail, 37s. Seven mutations against
the new gates, every one caught — including the two the plan review asked for specifically, one
narrowing the cleartext-password gate to PLAIN alone so the LOGIN row has to catch it unaided, and
one adding `InsecureSkipVerify: true`, which all four certificate rows catch. Data and integration
tiers not applicable: no query, migration, handler, route or schema moved.

Two commits correct this stage's work: `8e58e72b` and `d98ffda9`, from the final review's two
rounds.

### Stage 2: the module graphs and the mailpit end-to-end assertions

`fa2d71f0`. `go mod tidy` in `src/authserver` and then `src/adminconsole`, in that order because
core had to be tidied first. Each module lost its two `// indirect` requirements and seven `go.sum`
lines: go-simple-mail, go-dkim, and go-dkim's own test dependency `go-test/deep`, which left with it
and which the plan had not predicted because it was never a build-graph node.
`grep -rn "go-simple-mail\|go-dkim" src/*/go.mod src/*/go.sum` is now empty across all three
modules, which is the issue's goal, and all three build and vet clean, which is the proof nothing
else reached the library.

Tests: `TestSendEmail`, the one real delivery test against the dev container's mailpit, now sends
with from-name `Acme, Inc.` and reads back through mailpit's API that `From.Name` is `Acme, Inc.`
and that the `Message-ID` is present and rooted in the from address's domain. Both properties were
already pinned against the in-process fake at stage 1; the point here is that a real SMTP
implementation and a real MIME parser agree, and that the from-name case is the one that failed
outright before this change. Reading them back needed `testutil.AssertEmailSent` to return the
message it had already fetched, whose final check became `require` rather than `assert` in the same
edit, since continuing past a miss would now assert on a zero value.

Verification: unit tier green on all three modules, 1480 pass 0 fail. Data and integration tiers
not applicable — the only files that changed are a test, a test helper and four module files. Two
mutations, both caught, each failing `TestSendEmail` itself: the from-name dropped from the `From`
header, and the `Message-ID` header not written.

### Stage 3: the SMTP password bound, folded in after the run

`cb266e43` and `822f1399`. `HandleAPISettingsEmailPut` bounded `SMTPHost` at 120 and
`SMTPUsername`, `SMTPFromName` and `SMTPFromEmail` at 60, and `SMTPPassword` not at all. It now
refuses past 256 bytes with the same `VALIDATION_ERROR` shape as its neighbours, message
`SMTP password must be at most 256 bytes.` 256 is an application bound: no PLAIN or LOGIN provider
credential is known to exceed it, and it must not shrink to the neighbours' 60 because SendGrid's
SMTP password is its 69-character API key. The admin console form has no `maxlength`, so the API is
the single point of validation. This is the one handler file the branch touches.

`822f1399` also moved every check that needs no network (encryption value, username, password,
from-name, from-email length and format) ahead of the 3s TCP connectivity probe, so a bad field
paired with an unreachable host is answered with the field error rather than the connect error.

Tests: a refusal row at 257 in `TestAPISettingsEmailPut_ValidationErrors` asserting code and
message, a second refusal row sending 257 bytes to an unreachable host to pin
validation-before-dial, and `TestAPISettingsEmailPut_PasswordAtBoundIsAccepted`, which stores 256
bytes and decrypts the stored value back to the input. Run on sqlite in the dev container, all ten
`TestAPISettingsEmail*` tests green, plus the `apihandlers` unit package. One mutation, the bound
lifted to 1e6, caught by the refusal row. The tiers on the other three engines ran in the check
suite only.

This was the run's one drafted follow-up; it is folded in here at the author's request rather than
filed. Its review found three things, all fixed in `822f1399`: the probe ran before the pure
validation, the message said "less than 256 characters" while 256 was accepted and `len` counts
bytes, and the Azure Communication Services rationale was wrong (its SMTP password is a single
Entra client secret).

## Verification, across the whole run

|---|---|---|
| Unit, all three modules | locally, at stages 1 and 2 and both review rounds | green; 5467 tests, 0 failing at the last run |
| Unit, `apihandlers` package, and the ten `TestAPISettingsEmail*` integration tests on sqlite | locally, stage 3 | green |
| Unit on core, authserver, adminconsole, setup | check suite, `d98ffda9` and `cb266e43` | green, 4 jobs |
| Data, on sqlite, mysql, postgres and mssql | locally on all four at the final gate, and the check suite | green; 2208 tests locally, and 4 green CI jobs |
| Integration, sqlite | locally at the final gate | green, 1333 tests |
| Integration, four engines | check suite, `d98ffda9` and `cb266e43` | green, inside the same four `Tests / <engine>` jobs |
| Lint, Vulnerabilities, Versions, Build validation | check suite, `d98ffda9` and `cb266e43` | green |
| `check-anchors.sh` | locally, at every gate | all 27 rows resolve to one line in the named function |
| Mutation | locally, 20 mutations over the run (7 at stage 1, 2 at stage 2, 7 at review round 1, 3 at round 2, 1 at stage 3) | every one caught |

The data and integration tiers on mysql, postgres and mssql exist only in the check suite; the
local runs above are the sqlite and four-engine data runs the dev container can produce. Stages 1
and 2 touch no query, migration, handler, route or schema, so those tiers were run for assurance
rather than because a surface moved. Stage 3 touches one handler, `HandleAPISettingsEmailPut`, whose
integration tests are the ones run locally above and in the check suite on all four engines.

Six, all recorded in the stage logs and none of them changing what §2 asked for.

1. **`EmailSender` has four unexported fields, not §4's three.** A `rand io.Reader` was added so the
   Message-ID's failure branch is testable; Go documents `crypto/rand.Read` as never failing, so
   the field exists for the test and not for production.
2. **`smtp_plain_auth.go` is a file §4 did not plan.** Written at the final review's round 1,
   finding 5: `smtp.PlainAuth` always sends the credentials inline, which can overrun RFC 5321's
   512-octet command line.
3. **`writeFoldedHeader` refuses rather than emitting an over-long line.** The judgement at round 1,
   finding 3; §4 specified the 78-byte SHOULD and did not say what to do when there is no legal
   fold point.
4. **`where.sh mutate --run` has no effect on a module tier**, so per-row mutation attribution is
   read out of each run's failed-test list rather than from narrowed runs.
5. **Three `go.sum` stanzas left each module, not two.** `go-test/deep` was go-dkim's own test
   dependency, never a build-graph node, which is why the module-graph probe did not name it.
6. **`AssertEmailSent`'s final check became `require.True`.** It returns a message now, so
   continuing past a miss would assert on a zero value and report a second, misleading failure.

## Review

Stages 1 and 2 were reviewed by `openai-codex/gpt-5.6-sol` at effort `xhigh`, 2 rounds over the
whole branch as it stood at `d98ffda9`. No stage was reviewed on its own (`LEO_REVIEW_MODE=end`). A
separate plan-gate round ran before any code existed and is included below for completeness. Stage
3 landed after that review and had its own round, recorded at the end of this section.

|---|---|---|---|
| plan gate | 6 | 2 | all resolved in the plan before stage 1 |
| 1 (code) | 7 | 0 | all fixed, `8e58e72b` |
| 2 (code) | 3 | 0 | all fixed, `d98ffda9`, tests only |
| stage 3 | 3 | 0 | all fixed, `822f1399` |

- **blocking, security (plan gate):** the cleartext-password refusal row used PLAIN only, which
  `net/smtp.PlainAuth` refuses on its own, so an ungated `loginAuth` would have passed the table.
  Answered in the plan: one gate for both mechanisms, a LOGIN-only refusal row, and a mutation
  narrowing the gate to PLAIN that the row must catch.
- **blocking, security (plan gate):** no row would have detected TLS certificate verification being
  disabled. Answered in the plan: an untrusted pool, four unknown-authority and wrong-host rows
  across both TLS paths, and an `InsecureSkipVerify` mutation per path.
- **significant, conformance (round 1, finding 5):** `AUTH PLAIN <base64>` can exceed RFC 5321
  §4.5.3.1.4's 512-octet command line for a password an admin can save today, which RFC 4954 §4
  says a client MUST NOT do. Fixed in `8e58e72b`: `smtp_plain_auth.go` falls back to the
  challenge form, supplying the credentials after the server's empty `334`.
- **minor, conformance (round 1, finding 3):** an unbroken 1000-byte subject would emit a header
  line past RFC 5322 §2.1.1's 998-byte MUST, with no legal fold point. Fixed in `8e58e72b`: the
  send is refused before the dial, with a named error, and both sides of the boundary are pinned.
- The other eight code findings are mutation-proven coverage gaps: the relay's verdict on the
  completed message, LOGIN and the PLAIN challenge form over a protected connection, mechanism
  second place, the dial timeout, `loginAuth`'s positional fallback, PLAIN's repeated-challenge
  guard, and the 512-octet transition sampled at the transition. Each has a mutation that survived
  before and is caught now.
- The plan gate's other four were defects in the plan itself, all corrected before stage 1: a
  Message-ID built from letters through a helper that returns `""` on a random failure, a fold that
  still left a 79-byte first line, a row pinning Q encoding where the standard library picks B, and
  a multi-mechanism row written in preference order so it could not detect server order winning.

[this comment](https://github.com/leodip/goiabada/pull/308#issuecomment-5594002628).

## The closing comments

- [Deferred decisions](https://github.com/leodip/goiabada/pull/308#issuecomment-5593973393) — none
  deferred; the eight agreement decisions and the one judgement taken inside the run.
- [Follow-ups, fold-ins and ceilings](https://github.com/leodip/goiabada/pull/308#issuecomment-5593979236)
  — one drafted issue, since folded in as stage 3; one fold-in from the seal; no ceilings.
- [The review account](https://github.com/leodip/goiabada/pull/308#issuecomment-5594002628) — all
  three rounds in full.

## For the release notes

**An SMTP configuration that worked before can now refuse to send, on purpose.** Three cases, each
failing with an error naming the setting to change, visible at the admin console's *Send test email*
button:

- **Encryption `None` with a username and password, against anything but a localhost relay.** The
  password used to go out in cleartext. It now refuses: set the encryption to `STARTTLS` or
  `SSL/TLS`. `None` remains valid without credentials, and remains valid with credentials to a relay
  on `localhost`, `127.0.0.1` or `::1` — mailpit and same-host relays are unaffected.
- **Encryption `STARTTLS` against a server that does not offer STARTTLS.** It used to continue in
  cleartext. It now refuses: use the right port, or set the encryption to `None` if the server
  genuinely has no TLS.
- **A username configured against a server advertising no `AUTH`.** It used to send unauthenticated.
  It now refuses: that is usually the wrong host or port, or a STARTTLS-only server reached as
  `None`.

What will not work either way: XOAUTH2. It failed before this change and still fails — it is not a
regression and is tracked separately.

What starts working: a from-name containing a comma, a double quote or angle brackets. `Acme, Inc.`
was accepted by the settings form and then failed every send; it now sends correctly. Every message
also carries a `Message-ID` now, which gives you a value to search relay logs for.


leodip:
## Deferred decisions

**None.** The run deferred nothing and raised nothing.

All eight decisions in the agreement's §3 were settled with you at the seal, before any code was
written, and none of them reopened during the run:

| # | Question | Settled as |
|---|---|---|
| 1 | Plaintext password mechanisms over an unencrypted connection | Fail closed, Go's rule: PLAIN and LOGIN only over TLS or to a loopback host |
| 2 | STARTTLS requested but not advertised | Fail, naming the mode; no silent downgrade |
| 3 | Which mechanisms, in which order | PLAIN, LOGIN, CRAM-MD5, client preference; CRAM-MD5 usable over cleartext since it sends no password |
| 4 | Credentials configured, server advertises no AUTH | Error, as `net/smtp.SendMail` does |
| 5 | Add a `Message-ID` header | Yes, `<random hex>@<from domain>` |
| 6 | From-name with a comma, quote or angle brackets fails every send | Folded in and fixed here |
| 7 | Timeouts | 10s dial, one 30s deadline covering EHLO through QUIT |
| 8 | How the tests reach the TLS paths | In-process fake SMTP server with a certificate generated at test time; mailpit stays as the one real delivery test |

One judgement was taken **inside** the run rather than deferred, and it is worth your eye because it
chose a refusal over a tolerated non-conformance. Final review round 1, finding 3: a subject with no
whitespace in it long enough to make the header line exceed RFC 5322 §2.1.1's 998-byte MUST has no
legal fold point, so the choice was to emit an over-long line or refuse the send. **Refused**, before
the dial, with a named error, matching what decisions 1, 2 and 4 already do in this file. It is not
reachable from any caller — every subject is a catalog string and the longest is bounded to about 82
bytes by `appNameMaxLength = 30` — so nothing an operator can configure today reaches it.

**Cost to reverse:** one branch in `writeFoldedHeader` and one row of
`TestSendEmail_SubjectHardLineLimit`. Contained, and reversible without touching anything else.

leodip:

### Drafted, not filed

One, and it was then folded into this PR at the author's request as `cb266e43`, corrected by
`822f1399` after its own review (stage 3 in the description): the bound is 256 bytes, checked before
the connectivity probe, with refusal rows at 257 and an acceptance row at 256. The text
below is the draft as it stood; the `gh issue create` command is no longer needed.

1. **The SMTP password is the one settings field with no length bound on input.**
   `enhancement`, `go`. Found during the final review, round 1 (finding 5). Status: **Folded in, `cb266e43` + `822f1399`**

   `HandleAPISettingsEmailPut` bounds `SMTPHost` at 120 and `SMTPUsername`, `SMTPFromName` and
   `SMTPFromEmail` at 60 each, and bounds `SMTPPassword` not at all — it goes straight from
   `len(req.SMTPPassword) > 0` into `encryption.EncryptData` and the
   `settings.SMTPPasswordEncrypted` column. Verified at
   `src/authserver/internal/handlers/apihandlers/handler_api_settings_email.go:123`, `:150`, `:158`
   and `:175`.

   The wire-level consequence is already fixed on this branch, which is why this is an enhancement
   rather than a bug: an oversized password used to make `AUTH PLAIN <base64>` exceed RFC 5321
   §4.5.3.1.4's 512-octet command line and fail every send, and `smtp_plain_auth.go` now falls back
   to the challenge form in exactly that case. What is left is the missing bound itself, which
   belongs to the settings handler and not to the sender.

   Out of scope here because it fails fold-in test 1: it rides on an authserver API handler that
   this change's diff does not touch at all (twelve files, five Go files under
   `src/core/communication`, six module files and one test helper), so folding it in would put a new
   package and a new tier in front of the reviewer for a bound the sender does not need.

   Searched `gh issue list --state all --search "SMTP password length"` and
   `--search "smtp settings validation"`: nothing, open or closed. (#36, on the AES key, is the only
   near hit and is unrelated.)

   **Body:** Every SMTP settings field validated by `HandleAPISettingsEmailPut` carries a maximum
   length except the password, which is written to `settings.SMTPPasswordEncrypted` at whatever size
   it arrives. Pick a bound in keeping with the neighbours and refuse past it with the same
   `VALIDATION_ERROR` shape the other fields use. Note that #274 already removed the send-time
   consequence: an oversized password no longer breaks SMTP AUTH, since `smtp_plain_auth.go` sends
   the credentials after the server's `334` challenge when the inline form would overrun the command
   line.

   ```bash
   gh issue create --title "Bound the SMTP password length in the settings API" \
     --body "Every SMTP settings field validated by HandleAPISettingsEmailPut carries a maximum length except the password, which is written to settings.SMTPPasswordEncrypted at whatever size it arrives. See handler_api_settings_email.go:123, :150, :158 and :175. Pick a bound in keeping with the neighbours and refuse past it with the same VALIDATION_ERROR shape the other fields use. Note that #274 already removed the send-time consequence: an oversized password no longer breaks SMTP AUTH, since smtp_plain_auth.go sends the credentials after the server's 334 challenge when the inline form would overrun the command line."
   ```

1. **A from-name containing a comma, a quote or angle brackets failed every send.** `e4a10031`
   (the fix), `fa2d71f0` (the end-to-end assertion through mailpit). Folded in at the seal as
   decision 6, so it is named in the agreement's §2 goal rather than smuggled.

   It fit because it is one line in the same function the change was rewriting anyway:
   `sender/from` concatenated the name unquoted and handed the result to a parser, which failed
   with `mail: missing '@' or angle-addr`, and `net/mail.Address{Name, Address}.String()` — which
   the rewrite had to use regardless — quotes or RFC 2047 encodes it correctly. Nothing to weigh,
   the whole of the fix, and §2 stayed true. The settings API accepts `Acme, Inc.` today
   (`caller/from-name-length` bounds the length and nothing else), so an operator with that
   from-name could not send mail at all.

**None.** `git diff --unified=0 origin/main...HEAD | grep '^+.*ceiling:'` returns nothing: this
change ships no simplification with a known limit. The three constants that look like limits are
protocol values, not ceilings — `smtpCommandLineLimit = 512` (RFC 5321 §4.5.3.1.4),
`maxHeaderLineHardBytes = 998` and the 78-byte fold target (RFC 5322 §2.1.1) — and none of them is
ours to revisit.

leodip:
## The review account

This is `log/closing.md` section 10 in full: every round, what the reviewer was, what it found, and
what happened to each finding. Under `LEO_REVIEW_MODE=end` the two final rounds are the only reading
the code got, so this comment is the whole of the review record — the verdict files live in a
gitignored `.review/` that the next round overwrites.

---

### Plan gate, round 1, 2026-09-08

Reviewer `openai-codex/gpt-5.6-sol`, effort xhigh, fresh session, request `274-20260908T214343Z`,
`type: plan-review`. Verdict `static-review-exhausted`, six findings, no follow-ups. `unchecked`
held four entries, all dismissed in `log/plan.md` as not bearing on a plan gate (no code existed;
the interface, callers and admin console are unchanged; live relays were rejected by decision 8).
Nothing outside `docs/` moved: the round changed the plan, and the plan gate closes on this verdict.

1. **LOGIN not covered by the non-loopback cleartext refusal row** (security, blocking). Confirmed:
   the only refusal row used PLAIN, which `net/smtp.PlainAuth` refuses on its own, so an ungated
   `loginAuth` would pass the table. Fixed in the plan: one gate for both mechanisms, a LOGIN-only
   refusal row, and a mutation narrowing the gate to PLAIN that the new row must catch.
2. **No row detects disabled TLS certificate verification** (security, blocking). Confirmed: every
   TLS row used a trusted, matching certificate. Fixed in the plan: the fake certificate loses its
   `localhost` SAN, an untrusted pool is added, four failure rows (unknown authority and wrong host,
   on STARTTLS and implicit TLS) require no `AUTH` and no `MAIL`, and an `InsecureSkipVerify`
   mutation per path must be caught.
3. **Message-ID used letters, not decision 5's hex, through a helper that returns `""` on a random
   failure** (conformance, significant). Confirmed at `stringutil.go:47`. Fixed in the plan: 16 bytes
   through an injectable reader, hex, error before dialing; the hex regex, a uniqueness row and an
   `iotest.ErrReader` row. Qualified: Go 1.27 documents `crypto/rand.Read` as never failing, so the
   reader field exists to make the branch testable, a recorded deviation from §4's three fields.
4. **The subject folding left a 79-byte first line** (conformance, significant). Confirmed by the
   reviewer's scratch test and by arithmetic. Fixed in the plan: a greedy 78-byte folder that folds
   after `Subject:` when the first token does not fit, applied to ASCII subjects too, with a long
   ASCII row added.
5. **The from-name CRLF row pinned Q encoding where the standard library picks B** (quality,
   significant). Confirmed at `probe/stdlib_probe.out` line 14. Fixed in the plan: the row asserts
   no `Bcc:` line, no raw CRLF in the value, and a round-trip decode, whichever encoding is chosen.
6. **The multi-mechanism row was in preference order** (conformance, significant). Confirmed. Fixed
   in the plan: a `CRAM-MD5 LOGIN PLAIN` row still picking PLAIN, and a server-order mutation.

---

### Final code review, round 1: scope and the tree it was read against

Request `274-20260908T230541Z`, gate `final`, round 1. Scope is the whole branch against
`origin/main` (11 files, +1554/-58), since no stage was reviewed on its own under
`LEO_REVIEW_MODE=end`. The check suite was green on `fa2d71f0` across all twelve jobs, including the
data and integration tiers on sqlite, mysql, postgres and mssql — the four-engine result no local
run can produce.

`tree-hash.sh --verify` at the time of writing:

```
current   production ddd8be116b31   unit-tests cdb6a7aae255   data-tests 6483c3303341
          integration-tests 85350a2f2c09   docs 5cb6d7031d45   all 6502dabccca8

stage 2  unit live           data ?              integration ?       docs live
```

Stage 1's unit result was taken on a tree whose production and unit-test files have since moved —
stage 2 edited `src/core/testutil/mailpitdata.go` and extended `email_sender_test.go` — which is
expected rather than suspicious, and CI's green on `fa2d71f0` re-establishes the unit tier at the
current tree anyway. Neither stage recorded a data or integration hash, correctly: this change
touches no handler, route or schema, so those tiers were not applicable locally, and CI ran them
regardless.

### Final code review, round 1: verdict

Reviewer `openai-codex/gpt-5.6-sol`, effort `xhigh`, fresh session, 2026-09-08T23:07:02Z. Verdict
`findings`: **7 findings**, all three axes reviewed. Its own note: "No security defect was found in
the plaintext credential gate, STARTTLS handling, TLS verification, injection defenses, or changed
secret flows."

Two of the seven are conformance defects in code this change wrote; five are mutation-proven
coverage gaps. All seven were resolved rather than deferred — each is a few lines on a surface the
change already owns. Nothing was filed as a follow-up. Fixed in `8e58e72b`.

**1 — quality, significant, CONFIRMED. The final DATA acceptance response is not held by any test.**
`email_sender.go` checks `w.Close()`, the only place `net/smtp` surfaces the relay's verdict on the
completed message, and no test made that verdict a refusal. The reviewer's mutation (`_ = w.Close()`)
left the whole suite green. Fixed: `fakeSMTP.rejectAfterData` answers the DATA terminator with
`550 5.7.1`, and `TestSendEmail_FinalDataRejection` requires the error and asserts the message did
arrive, so the refusal is the relay's verdict and not a transport failure.

**2 — quality, significant, CONFIRMED. LOGIN was only ever exercised over cleartext loopback.**
The mechanism table runs every row over `none` to 127.0.0.1, so the suite's only LOGIN exchange was
the one decision 1's loopback exception permits. A regression refusing LOGIN on every protected mode
left the table green. That is backwards from why decision 3 keeps LOGIN at all:
`probe/ehlo-public.out` has smtp.office365.com advertising `AUTH LOGIN XOAUTH2` and nothing else
*after* STARTTLS. Fixed: `TestSendEmail_LoginOverProtectedConnection` reproduces that shape —
STARTTLS to `os.Hostname()`, `AUTH LOGIN XOAUTH2` advertised only post-handshake — and asserts
delivery, the decoded credentials, and that the AUTH line follows STARTTLS.

**3 — conformance, minor, CONFIRMED, and the one the reviewer left as a judgement.** An unbroken
ASCII subject is returned unchanged by `mime.QEncoding.Encode` and has no space to fold at, so
`writeFoldedHeader` would emit a 1001-byte line for a 1000-byte subject, past RFC 5322 §2.1.1's
998-byte MUST. Not reachable from any caller: every subject is a catalog string and the longest, the
new-user one carrying the app name, is bounded by `appNameMaxLength = 30` to about 82 bytes.

The judgement was between emitting the over-long line and refusing the send. **Refusing**, because
a word with no whitespace in it has no legal fold point — the choice is a conforming error or a
non-conforming line — and because the alternative is a rejection arriving mid-DATA with a message
the admin cannot act on. It matches what the rest of the file does: decisions 1, 2 and 4 all fail
closed with a named error, and this one fails before the dial, alongside the invalid-recipient
refusal. `maxHeaderLineHardBytes = 998` is now a constant beside the 78-byte SHOULD,
`writeFoldedHeader` returns an error, and `TestSendEmail_SubjectHardLineLimit` pins both sides:
997 bytes folded and delivered, 998 refused with nothing connected.

**4 — quality, significant, CONFIRMED. Decision 3's second-place ordering was never pinned.** The
table proved PLAIN beats both and that each mechanism works alone, but nothing separated LOGIN from
CRAM-MD5: swapping them in `authenticate` left every row green. Fixed with one row,
`AUTH CRAM-MD5 LOGIN` — worst-first, like the existing reverse row. The recorded password is what
decides it, since CRAM-MD5 never sends one.

**5 — conformance, significant, CONFIRMED, and the substantive defect of the round.**
`smtp.PlainAuth` always writes the credentials inline as `AUTH PLAIN <base64>`. RFC 4954 §4: "If use
of the initial response argument would cause the AUTH command to exceed this length, the client MUST
NOT use the initial response parameter". That length is RFC 5321 §4.5.3.1.4's 512 octets. Nothing
bounds the SMTP password on the way in, so a password an admin can save today makes every send fail
against a server that enforces the limit.

Fixed in `src/core/communication/smtp_plain_auth.go` (new): `plainInitialResponseFits` measures the
command line, and where it does not fit, `plainAuth` sends the bare `AUTH PLAIN` and supplies the
credentials after the server's empty `334` challenge. That continuation line is exempt by the same
section: both ends "MUST be able to handle the maximum encoded size of challenges and responses ...
independent of any line length limitations". The inline path stays for every password that fits, so
the ordinary send pays no extra round trip. `fakeSMTP.maxCommandLine` now enforces 512, and a test
delivers with a 400-byte password and asserts the AUTH line carried no payload.

**6 — quality, significant, CONFIRMED. Only half of decision 7 was tested.** The conversation
deadline had a test; the 10-second dial timeout had none, and removing `Timeout` from the
`net.Dialer` left the suite green. Fixed: `TestSendEmail_DialTimeout` asserts
`defaultDialTimeout == 10s` and drives the `dialTimeout` seam with a duration already in the past,
which `net.Dialer` turns into an immediate `i/o timeout` against a listening fake — deterministic,
and microseconds rather than a wait.

**7 — quality, significant, CONFIRMED. `loginAuth`'s answer-by-position fallback was untested.**
The fake only ever sent the conventional `Username:` and `Password:` prompts, so deleting the
fallback changed nothing the suite could see. It exists because no RFC specifies LOGIN —
draft-murchison-sasl-login expired — so the prompt wording is convention, not syntax. Fixed:
`fakeSMTP.loginChallenges` is configurable and
`TestSendEmail_LoginAnswersUnconventionalChallenges` drives `first value:` / `second value:`.

#### Mutations, all seven re-run against the fixed suite

Each is the reviewer's own mutation, replayed with `where.sh mutate --type modules`. Each was
reported as surviving; each is now **CAUGHT**.

| Finding | File | Replacement | Result |
|---|---|---|---|
| 1 | `email_sender.go` | the checked `w.Close()` block to `_ = w.Close()` | CAUGHT |
| 2 | `email_sender.go` | `case "LOGIN"` refuses every mode but `none` | CAUGHT |
| 3 | `email_sender.go` | `if 1+len(token) > maxHeaderLineHardBytes {` to `if false {` | CAUGHT |
| 4 | `email_sender.go` | preference slice to `{"PLAIN", "CRAM-MD5", "LOGIN"}` | CAUGHT |
| 5 | `email_sender.go` | `if plainInitialResponseFits(...)` to `if true {` | CAUGHT |
| 6 | `email_sender.go` | `&net.Dialer{Timeout: dialTimeout}` to `&net.Dialer{}` | CAUGHT |
| 7 | `smtp_login_auth.go` | `switch a.answered {` to `switch -1 {` | CAUGHT |

`git status` and a grep for every mutant string confirm none survived in the tree; the tiers below
ran after that check.

#### Tiers, at `production 8c6fc76aadde  unit-tests b3c52b98fede  all de12df3a0581`

- `where.sh test --type modules` — 5463 passing, 0 failing, ~30s.
- `where.sh test --type data` — 2208 passing, 0 failing, ~40s, on mysql, postgres, mssql and sqlite.
- `where.sh test --type integration --db sqlite` — 1333 passing, 0 failing, ~50s. Run despite the
  change touching no handler, route or schema, because production code moved at the final gate.
- `check-anchors.sh` — all 27 rows resolve to exactly one line, in the named function.

#### The verdict's `unchecked`, answered

- **No live external relay interoperability pass.** *Dismissed.* No Gmail or Office 365 credentials
  exist for this project and the run cannot acquire any. The two shapes that matter are recorded in
  `probe/ehlo-public.out` and both are now covered against the fake: Office 365's post-STARTTLS
  `AUTH LOGIN XOAUTH2` by finding 2's new test, Gmail's `LOGIN PLAIN XOAUTH2` by the mechanism
  table's PLAIN-wins rows.
- **Not every SMTP command rejection or network drop was injected.** *Covered in part, dismissed for
  the rest.* The final DATA response was the one with no negative test and a surviving mutation, and
  finding 1 closed it. MAIL, RCPT, DATA initiation, the body write and QUIT are each an
  `if err != nil { return errors.Wrap(err, "unable to send SMTP message") }` immediately after their
  call, on the same three-line shape now pinned at DATA.
- **Post-dial request-context cancellation was not tested.** *Dismissed.* Decision 7 chose one
  `SetDeadline` covering EHLO through QUIT on purpose. The context is honoured for the dial, through
  `DialContext`; after that the conversation is bounded by the deadline rather than by the request,
  which is the decided design.
- **No interactive admin-console browser pass.** *Dismissed.* `git diff origin/main...HEAD --stat`
  shows no handler file in the change at all; the send error reaches the admin through the same
  `SEND_FAILED` message as before, and the wording of the new refusals is asserted at the unit seam.
- **Unrelated security classes not rerun.** *Dismissed*, and correctly reasoned: the diff is confined
  to `src/core/communication` and the three module files.

---

### Final code review, round 2: verdict, and why it is the last round

Reviewer `openai-codex/gpt-5.6-sol`, effort `xhigh`, resumed session, 2026-09-09T00:09:27Z, request
`274-20260909T000603Z`. Verdict `findings`: **3 findings**, all three axes reviewed. Its own note:
"Round 1 production repairs are correct in the inspected paths, and no security or current
implementation defect was found. Three mutation-proven gaps remain in the new challenge-form PLAIN
coverage: repeated challenges, the exact 512-octet transition, and protected non-loopback
operation."

All three are coverage gaps in `smtp_plain_auth.go`, which round 1 wrote and nothing had yet read.
All three were confirmed against the code, fixed, and each has a mutation proving the new row can
fail. None was refuted and none deferred. Fixed in `d98ffda9`.

**This is the last round.** The budget funds a further round only where the previous one changed
production code, and this one did not: `tree-hash.sh --split` reports `production 8c6fc76aadde`
before and after — the value round 2 was reviewed at — so every mutation is proven gone and the only
files that moved are `email_sender_test.go` and `smtp_fake_test.go`. A gate whose last round only
hardened tests has had its answer; a round 3 would be reviewing this round's tests.

**1 — quality, significant, CONFIRMED. `plainAuth`'s single-response guard is unreachable from the
suite.** PLAIN carries exactly one response, so `plainAuth.Next` refuses a second challenge rather
than putting another copy of the password on the wire. No test could produce a second challenge, so
the reviewer's mutation of `if a.answered` to `if false` left every `TestSendEmail` test green.
Fixed: `fakeSMTP.plainExtraChallenge` issues a second empty `334` after it has already taken the
response, and `TestSendEmail_PlainRepeatedChallenge` requires the send to fail naming
`unexpected server challenge`, and counts the recorded lines that decode to the password. The count
is the pin, not the failure: this fake refuses the login either way, so a client that answered twice
would also see the send fail — it would just have handed over the password a second time first.
Mutation `if a.answered {` → `if false {`: **caught** (the count reaches 2 and the error becomes the
server's `535`).

**2 — quality, significant, CONFIRMED. The 512-octet boundary was sampled 400 bytes away from
itself.** The round 1 test used a 400-byte password, well past the transition, so it proved the two
shapes exist and not where the switch happens. Allowing one extra octet in
`plainInitialResponseFits` left the whole suite green even though the first 513-octet command is
then written inline and answered `500` by a server that enforces RFC 5321 §4.5.3.1.4. Fixed: renamed
to `TestSendEmail_PlainInitialResponseBoundary` and made a two-row table one password octet apart.
The lengths are derived in the test from the limit and base64's 4-for-3 expansion — 499 octets for
the response, so 372 raw, less RFC 4616's two NUL separators and the username — deliberately not
read back from `plainInitialResponseFits`, which is the thing under test. The fake's
`maxCommandLine` is the independent oracle, so the rows break in both directions: too permissive and
the inline row overruns and is refused, too strict and the fitting row loses its inline response.
Mutation `<= smtpCommandLineLimit` → `<= smtpCommandLineLimit+1`: **caught**.

**3 — quality, significant, CONFIRMED. The challenge form was only ever exercised over cleartext
loopback.** The same family as round 1's finding 2, one layer down: round 1 split PLAIN into two
implementations, and every protected row in the suite runs the *inline* one. The challenge form was
reachable only through decision 1's loopback exception, so a regression refusing it on STARTTLS and
implicit TLS — the ordinary deployment shape, and the one RFC 4954 §4 requires to interoperate —
left the suite green. Fixed: `TestSendEmail_PlainChallengeFormOverProtectedConnection`, STARTTLS to
`os.Hostname()` with `AUTH PLAIN` advertised only after the handshake, asserting the bare
`AUTH PLAIN` command, that it follows STARTTLS, the decoded credentials and delivery. The boundary
table keeps the loopback-clear challenge-form coverage, so both transport classes now have a row
rather than the one the reviewer offered to trade for. Mutation: refuse `&plainAuth` whenever
`smtpEnc != enums.SMTPEncryptionNone`: **caught**.

#### What round 2's `unchecked` held, and the answer to each

- *No live Gmail or Office 365 relay.* **Dismissed.** Decision 8 settled this at the seal; the
  substitute is `probe/ehlo-public.out`, which captured what those two relays actually advertise,
  and decision 3 and round 1's finding 2 are both built from it.
- *The broad unit, four-engine data and sqlite integration tiers were not rerun.* **Covered.** The
  unit tier was rerun this round on all three modules: 5467 passed, 0 failed, ~40s.
  `tree-hash.sh --split` has `data-tests 6483c3303341` and `integration-tests 85350a2f2c09` unmoved
  from round 1's record and `production` unmoved, so those results still describe this tree.
- *The repeated-challenge case was tested against the state machine rather than a full exchange.*
  **Covered** — that is finding 1, and it is now an end-to-end row.
- *No interactive admin-console browser pass.* **Dismissed**, on the diff rather than on the
  reviewer's word: twelve files, five Go under `src/core/communication`, six `go.mod`/`go.sum`, and
  `src/core/testutil/mailpitdata.go`. No handler, route, template or migration.
- *Unrelated security classes not re-reviewed.* **Dismissed.** None is on this change's surface.

#### Verification for round 2

Unit tier, all three modules, 5467 passed, 0 failed. `gofmt -l` and `go vet` clean on the package.
`check-anchors.sh`: all 27 rows resolve. Three mutations, one per finding, all **caught**.
`tree-hash.sh --split` after the mutations: `production 8c6fc76aadde`, equal to the value round 2 was
reviewed at, so nothing was left behind.

Issue #120, opened 2026-07-26
Hardening: output encoding is incidental rather than designed — 23 innerHTML sinks in the admin console, plus the marshal template.JS helper


leodip:
## Summary

The admin console templates assign server-supplied data to `innerHTML` in 23 places. `innerHTML` parses its input as HTML, so any value reaching one of those sinks is markup rather than text, and an `onload` or `onerror` attribute inside it executes with the viewing administrator's session.

One of these sinks is confirmed reachable by an unauthenticated caller, and it plus its web origins twin are being fixed in #105. The other 21 are currently blocked by input validation on their write paths rather than by output encoding. That is a fragile position: nothing in the templates records the dependency, so relaxing any of those validators later silently creates an XSS, and the inventory has never been reviewed as a whole.

Severity: **Low** as it stands, because the one verified vector is handled in #105 and the rest are protected upstream. Filing it because the protection is incidental rather than designed, and because each remaining site needs its data source checked individually instead of assumed safe.

No CSP mitigates any of this. The app sets only `Content-Security-Policy: frame-ancestors 'none'` (`src/core/middleware/middleware_security_headers.go:31`) with no `script-src` directive, so inline event handlers execute normally.

## The confirmed case, for context

`admin_clients_redirect_uris.html:15` writes each stored redirect URI into a JavaScript array, and `refreshRedirectURIsTable()` assigns it at `:89` with `cell1.innerHTML = uri`. Go's contextual escaping stops the value breaking out of the JS string literal, but `innerHTML` then re-parses those characters as HTML.

Verified in a browser against a page replicating the template's exact output: `<svg onload>` arriving through that sink fires. `<script>` does not, which is presumably why the sink has looked safe. Combined with #105's prefix-match bug, an anonymous `POST /connect/register` could store the payload, and an administrator would trigger it by viewing that client's redirect URIs, which is what `site/src/content/docs/concepts/clients.mdx:215` advises them to do.

#105 fixes both the input side and this sink plus its web origins twin. This issue covers everything else.

## Inventory

83 `innerHTML` assignments across 18 templates. 50 are localised string literals with no data and are not at issue. The remaining 33 break down as follows.

**Group A: data assigned directly, should be `textContent` (20 sites)**

| Sites | Value | Current protection |
|---|---|---|
| `admin_clients_redirect_uris.html:89`, `admin_clients_web_origins.html:89` | `uri` | **none, verified reachable.** Fixed in #105 |
[...]

PR #313, opened 2026-09-09
feat(authserver): refuse < and > in plain-text fields and drop htmlsanitizer

### LINKED ISSUE #275: htmlsanitizer silently rewrites plain-text profile fields that html/template already escapes
`sym01/htmlsanitizer` is an HTML sanitizer. Every field we point it at is **plain text**.

## Where it is applied

`core/inputsanitizer/input_sanitizer.go` wraps `htmlsanitizer.SanitizeString` and is called on:

| Handler | Fields |
|---|---|
| `handler_api_account_profile.go` | `Username`, `GivenName`, `MiddleName`, `FamilyName`, `Nickname` |
| `handler_api_account_address.go` | `AddressLine1`, `AddressLine2`, `AddressLocality`, `AddressRegion`, `AddressPostalCode`, `AddressCountry` |
| `handler_api_groups.go` | `GroupIdentifier`, `Description` |
| `handler_api_users_attributes.go`, `handler_api_group_attributes.go` | attribute `Value` |
| `handler_api_settings_email.go` | `SMTPFromName` |

None of these is HTML. None is ever rendered as HTML. `html/template` escapes all of them on output.

## Why this is the wrong shape

A sanitizer's job is to take HTML that must stay HTML and remove the dangerous parts. Applied to a plain-text field it does something different and worse: it **silently rewrites what the user typed**.

A user whose name contains `<` gets it stripped with no error and no indication. A group description mentioning `a < b` loses characters. The user finds out later, if ever. That is a worse outcome than a validation error, and it is not buying us anything `html/template` is not already providing.

The failure mode we are guarding against is a future path that renders one of these unescaped, or ships it through a non-HTML channel. A validator covers that case at least as well, and does not corrupt data in the common case.

## What replaces it

A validator in `core/validators` that **rejects** input containing `<`, `>` or an HTML entity prefix, returning a localized error through the existing `i18n.NewLocalizedError` machinery, wired into the same handlers.

Stdlib only. Roughly 30 lines plus catalog entries for the error messages.

## This is a behaviour change, decide it first

| | Today (strip) | Proposed (reject) |
|---|---|---|
| User types `O'Brien <the third>` | Silently saved as `O'Brien ` | 400 with "this field cannot contain < or >" |
| Existing rows containing stripped values | Already stripped, unaffected | Unaffected |
| API clients sending markup | Succeed, get different data back than they sent | Get a clear error |

My view is that reject is correct: silently mutating someone's own name is worse UX than telling them no, and an API that returns different data than it accepted is a bug in its own right. But it is a product call, not a mechanical swap, and it should be settled before any code moves.

Whichever way it goes, `strings.TrimSpace` at these call sites stays: several handlers already pair it with `Sanitize`.

---

### Dependency reduction

leodip:
Closes #275.

## What this is for

Every plain-text field the admin and account APIs accept, from client descriptions to address lines
to the SMTP from name, was passed through `github.com/sym01/htmlsanitizer` on its way to the row.
None of those fields is ever rendered as HTML: the auth server escapes everything through
`html/template`. The library did not strip markup so much as rewrite text: `O'Brien <the third>`
became `O'Brien `, `x > y` was stored as `x &gt; y` and displayed that way, and `<b>` or ``
passed through untouched because they were on its allowlist.

## The rule

A value containing `<` or `>` is refused with a 400 and a localized message naming the field.
Nothing that is accepted is ever different from what was sent, beyond the whitespace trimming the
handlers already applied. Ampersands, entities, quotes and non-ASCII text pass, so `Tom & Jerry`
and `AT&T` are stored verbatim. Fields whose existing validators already exclude the characters
(names, usernames, emails, phones, identifiers, countries) are unchanged, and a unit case per
validator now pins that they do. Dynamic client registration applies the same rule to `client_name`
with `invalid_client_metadata`, so every write path to the client description agrees. The
dependency leaves all three modules.

### Stage 1: the validator, its seven codes, the address checks and the pins
Landed in `1bf173ec`.

`validators.ValidateNoAngleBrackets(value, code)` and `validators.ContainsAngleBrackets(s)` in a new
stdlib-only file. The caller supplies the error code, so the banner the API's `error_description`
feeds names the field that was refused; seven codes ship with their English and pt-BR catalog
entries and their rows in `error_codes.md`.

`ValidateAddress` now refuses `<` and `>` in address line 1, line 2, locality, region and postal
code. Both address endpoints inherit that without a handler change, since each validates before it
sanitizes, so `PUT /api/v1/account/address` and `PUT /api/v1/admin/users/{id}/address` already
answer 400 with `validator.address.angle_brackets`.

Two refused rows per existing validator (name, username, nickname, email, phone, identifier,
country), one holding `<` and one holding `>`. Those regexes already excluded both characters, so
the sanitizer was a no-op behind them; once it goes they are the only guard, and these rows are what
makes removing it safe.

Tiers: unit green across the three modules (5516 subtests, 0 fail); integration green on sqlite
(1338 subtests, 0 fail); data and migrations not applicable, no schema or data-layer code. Nine
mutations run, every one caught.

One thing worth knowing: the plan asked for a mutation isolating the country pin on `>`, which is
not possible — `U<` and `U>` are both two characters, so both are refused by the exact alpha-2 set
lookup rather than by any character class. The mutation removes that lookup's refusal instead, and
fails the two new pins together with the two pre-existing rows sharing the guard.

### Stage 2: the auth server API handlers stop sanitizing
Landed in `0c4bb1b8`.

All 49 `Sanitize` calls under `authserver/internal/handlers/apihandlers/` are gone. Thirteen
independently removable `ValidateNoAngleBrackets` checks took their place, one per wiring point
rather than one per file: client create and update description and display name, group create and
update description, resource create and update description, user attribute create and update value,
group attribute create and update value, settings general app name and issuer, settings email from
name. Each is refused with 400 and its own catalog key. At the sites whose validator already
excluded the characters the call was simply deleted, since stage 1's pins now cover them.
`TrimSpace` stayed exactly where it was, and attribute values are still untrimmed.

Two behaviour changes worth naming. The permission description's refusal moves from a literal
English string under `VALIDATION_ERROR` to the catalog key
`handler.admin_resource_permissions.description_html_not_allowed`; the sentence on the wire is
unchanged, because the catalog string and the deleted literal are the same text. And the permission
identifier now reaches `ValidateIdentifier` as it was sent: the handler used to sanitize it first,
so `valid<b` was truncated to `valid` and the permission was created under a name the caller never
asked for. It is refused now, and a row in the endpoint's table pins that.

Handler constructors lost their `inputSanitizer` parameter, `routes.go` stops constructing one, and
the unused `InputSanitizer` interface in the auth server's `interfaces.go` is deleted. The one
sentence in `openapi.yaml` that mentioned sanitizing now says trimming.

Tiers: unit green across the three modules (1491 tests, 0 fail); integration green on sqlite (1367
tests, 0 fail), with a refused case per check and a verbatim-accept case per endpoint sending `&`
and quotes and asserting the body returns them byte for byte. Data not applicable, no query, model
or migration touched.

Four mutations, all caught: neutralising the shared validator, reproducing the sanitizer's old
identifier truncation at the permissions handler, re-entity-encoding `&` on the group description's
path to the row, and removing one single endpoint's check while every other stays.

Eighteen rows of the agreement's anchor table pointed at code this stage deleted or rewrote. Three
were removed because nothing stands where they pointed; the other fifteen were repointed, with the
reasoning in the stage log.

### Stage 3: DCR, the admin console pre-check, and the package goes
Landed in `744ab66b`.

Dynamic client registration writes `client_name` into `clients.description`, the same column the
admin API refuses angle brackets in, so a self-registered client could hold markup every other
write path to that column rejects. `validateDCRRequest` refuses `<` and `>` beside its length
check, answered as `invalid_client_metadata` per RFC 7591 section 3.2.2. The description is English
like the handler's other refusals, since DCR errors are protocol responses and are not localized.

The admin console's pre-save check for a permission used to detect markup by sanitizing the
description and seeing whether the value changed. That disagreed with the API on everything the
sanitizer's allowlist passed through: `<b>x</b>` survived unchanged, so the pre-check said valid and
the save then refused it. It asks the same validator now, returning the same message it returned
before, and the permission identifier reaches `ValidateIdentifier` raw for the same reason —
sanitizing it first turned `valid<b` into `valid` and reported a name the user had not typed as
available.

With those gone, `core/inputsanitizer` and its 41 cases are deleted, the `InputSanitizer` interface
with them, and `go mod tidy` in the three modules drops the requirement.

The consent-escaping test can no longer register a `<script>` name, because registration refuses it.
It registers `Acme & "Sons"` instead and asserts the same property: stored byte for byte, reaching
the page as `Acme &amp; &#34;Sons&#34;`, checked against the raw response bytes because goquery
decodes entities. The admin console's validate-permission endpoint had no test at all; it has one
now, running the real `HttpHelper` and the real `IdentifierValidator` rather than mocks.

Tiers: unit green across the three modules (5483 ok, 0 fail); integration green on sqlite (1370 ok,
0 fail). Data and migrations not applicable, no query, schema or migration touched.

Four mutations, all caught: neutralising the pre-check's refusal, reproducing the sanitizer's old
truncation of the permission identifier, neutralising the DCR refusal, and emptying the client name
the consent page renders. One note for whoever runs mutations at the integration tier: an `if false`
replacement that leaves an import unused makes the package fail to build, and `where.sh mutate`
reports that as `MUTATION CAUGHT` rather than `MUTATION DID NOT COMPILE`. That first attempt was
discarded and redone with a replacement that compiles.

CI was red on `744ab66b` in exactly one job, `Lint`: two `staticcheck` SA1012 violations in the new
validate-permission test, which built the messages it asserts against with `Localize(nil)`. `Lint`
is CI-only in this workflow, so no local tier saw it, though `golangci-lint` is in the dev container
and reproduced it in seconds. `c264a932` passes `context.Background()` instead, which is how the
other tests here build theirs and carries no language, so the expected English text is unchanged: no
assertion moved and no production file moved. The engine jobs that had concluded on `744ab66b` —
sqlite, mysql and postgres — were green; the four-engine claim now rests on `c264a932`.

## Verification

**The check suite is the four-engine result**, green on all twelve jobs at `f4cf1f38`
([run](https://github.com/leodip/goiabada/actions/runs/34309886319)): `Build validation`, `Lint`,
`Vulnerabilities`, `Versions`, `Unit / core`, `Unit / authserver`, `Unit / adminconsole`,
`Unit / setup`, and `Tests / sqlite`, `Tests / mysql`, `Tests / postgres`, `Tests / mssql` — each of
those four running the data tier and then the integration tier.

**Locally**, each stage ran the unit tier across the three modules and the integration tier on
sqlite, plus the mutations recorded in the stage sections above (nine in stage 1, four in stage 2,
four in stage 3, every one caught) and three more at the final gate. The data tier is not applicable
to this branch on its own terms — no query, model, schema or migration file is touched — and it ran
on four engines in CI regardless. `Lint`, `Vulnerabilities` and `Versions` have no local tier, which
is how stage 3's `staticcheck` failure reached CI.

### Deviations

- **The country validator's `>` pin cannot be isolated by mutation.** `U<` and `U>` are both refused
  by the exact ISO alpha-2 set lookup rather than by any character class, so the stage 1 mutation
  removes that lookup's refusal instead and fails the two new pins together with the two
  pre-existing rows sharing the guard.
- **Section 5's seam 3 said update endpoints get no case of their own**, "since the check is the
  same call at the same layer". True of the check, not of the assignment — the final review's second
  finding showed the attribute update handlers assign the value independently, so the two verbatim
  tests now cover both wiring points. This widens the seam rather than narrowing it.
- **CI was red once**, on `744ab66b`, in `Lint` alone: two `staticcheck` SA1012 violations in the new
  admin console test. `c264a932` fixes them and moves no production file.
- **`where.sh mutate` misreports a non-compiling mutation** at the integration tier as
  `MUTATION CAUGHT` rather than `MUTATION DID NOT COMPILE` when the replacement leaves an import
  unused. One stage 3 attempt hit this, was discarded, and was redone with a replacement that
  compiles.

### Review

Reviewed by `openai-codex/gpt-5.6-sol` at effort `xhigh`, one round at the plan gate and 2 rounds
over the whole branch at the final gate. No stage was reviewed on its own.

|---|---|---|---|---|
| Plan | 1 | 3 | 0 | all confirmed, fixed in the plan before any code |
| Final | 1 | 2 | 0 | both confirmed and fixed, `f4cf1f38` |
| Final | 2 | 0 | 0 | clean |

- **significant, quality (plan):** the endpoint matrix was drawn per file rather than per wiring
  point, and `<b>` alone cannot pin both characters. Fixed in the plan: one case per wiring point,
  and two rows per validator, one per character.
- **minor, conformance (plan):** the permission handlers sanitized the identifier *before*
  validating it, so `valid<b` was silently accepted as `valid`. Falsified section 1's no-op premise.
  Fixed: the raw trimmed identifier is validated, and it is now refused.
- **significant, quality (final 1):** the consent-escaping test had been rewritten around
  `Acme & "Sons"`, which does not exercise the two characters the renderer exists to neutralise,
  while decision 7 leaves pre-existing rows holding complete tags. Fixed in `f4cf1f38`: a second
  test plants `<b>Existing</b>` in the row and asserts the raw response bytes carry the escaped form.
- **significant, conformance (final 1):** the attribute *update* handlers assign the value at their
  own site, and only create had an accepted-value case, so a mutation rewriting ampersands on update
  survived the whole tier. Fixed in `f4cf1f38`: both verbatim tests now PUT a second accepted value
  and read it back from the response and the database.

No finding was refuted, deferred or escalated. Full account, with the mutation tables and the
`unchecked` answers, in [this comment](https://github.com/leodip/goiabada/pull/313#issuecomment-5595804255).

### The three closing comments

- [Deferred decisions](https://github.com/leodip/goiabada/pull/313#issuecomment-5595790194) — none;
  every decision was settled before the run and none moved.
- [Follow-ups, fold-ins and ceilings](https://github.com/leodip/goiabada/pull/313#issuecomment-5595792638)
  — no issues to file, two fold-ins reported, no ceiling markers.
- [The review account](https://github.com/leodip/goiabada/pull/313#issuecomment-5595804255) — every
  round, every finding, and the evidence behind each answer.

## For the release notes

**Plain-text fields now refuse `<` and `>` instead of silently rewriting them.** A request that
would previously have been accepted with the characters stripped is answered with 400 and a
localized message naming the field. This affects client and group and resource descriptions, the
client display name, user and group attribute values, the five address fields, the application name,
the issuer, and the SMTP from name. Dynamic client registration refuses the same characters in
`client_name` with `invalid_client_metadata`.

**What to do about it.** Nothing on upgrade — there is no migration and no configuration. Callers of
the admin and account APIs that send these characters need to stop; they were never getting stored
what they sent. Values containing `&`, entities, quotes and non-ASCII text are unaffected and are
now stored byte for byte.

**What will not work.** Existing rows are left exactly as they are. A row that holds `&gt;` — which
is how the old sanitizer stored a `>` a user typed — keeps displaying as `&gt;` and must be edited
by hand if that is wrong; nothing can tell it apart from an `&gt;` somebody typed deliberately. A
row that holds markup the old allowlist passed through, such as `<b>`, stays in place, but
re-saving it unchanged from an edit form is now refused until the markup is removed. Both are
deliberate: cleaning them up would mean a fifth site deciding what a user's text meant.

## Status

**Complete.** All three stages landed, the whole change was reviewed over two rounds ending clean,
and the check suite is green on all twelve jobs at `f4cf1f38`. `github.com/sym01/htmlsanitizer` is
gone from all three `go.mod` and `go.sum` files, `core/inputsanitizer` no longer exists, and
`grep -rn '\.Sanitize(' --include=*.go src` reads zero.


leodip:
## Deferred decisions

**None.** Nothing in this run was postponed on an assumption.

All eight decisions in section 3 of the agreement were settled before the run started, and none of
them moved while it was building: stage 3's log entry records "No decision moved and nothing was
folded in", and no stage escalated a question to this thread. There is therefore no assumption here
for you to ratify or reverse.

Two of those eight are worth knowing about anyway, because they decided something the issue did not
ask for, and both are reported as fold-ins in the follow-ups comment rather than here:

- **Decision 6**, dynamic client registration refuses `<` and `>` in `client_name`, against the
  recommendation to leave that path alone.
- **Decision 7**, no migration and no cleanup of rows the old sanitizer already rewrote. A stored
  `<b>` value resubmitted unchanged from an edit form is now refused until the user removes it. This
  is the one user-visible consequence of the change that nothing in the code announces, so it is
  written up under `For the release notes` in the PR body.

leodip:

### Issues to file

**None.** No stage found a defect outside its own scope. The plan gate drafted no follow-ups,
stage 2's entry records "None found. Nothing in this stage's surface raised a defect outside its
scope", and stages 1 and 3 raised none. There is no `gh issue create` command to run here.

The one adjacent thing this change touches without fixing is already filed and still open:
[#120](https://github.com/leodip/goiabada/issues/120), the inventory of admin console `innerHTML`
sinks whose only protection is input validation on the write path. This branch tightens what #120
depends on rather than loosening it — the sanitizer it removes let `<b>`, `<a href>` and ``
through by design, and a validator refusing `<` and `>` outright is strictly stronger at every one
of those sinks. No sink code moved here, and #120 keeps its own scope: `textContent` and the lint
that would enforce it.

Both were absorbed at seal time, as decisions in the agreement, and both are reported here because
they widen the change beyond what the issue asked for and you have not seen them go by:

- **The admin console pre-check disagreed with the API on `<b>x</b>`.** `744ab66b`. The pre-save
  validate-permission endpoint detected markup by sanitizing the description and seeing whether the
  value changed, so everything on the library's allowlist passed the pre-check and was then refused
  on save. It rides on a file this change already had to edit, because the sanitizer *was* that
  handler's mechanism, and replacing the compare with the shared validator is the whole fix.
- **Dynamic client registration wrote the same column unchecked.** `744ab66b`. `client_name` lands
  in `clients.description` with only a length check, so after this change one write path would have
  accepted what every other refuses. `validateDCRRequest` now refuses `<` and `>` as
  `invalid_client_metadata`, per RFC 7591 section 3.2.2.

Nothing was folded in *during* the run. Stage 3's log entry: "No decision moved and nothing was
folded in."

**none.** `git diff --unified=0 main...HEAD | grep '^+.*ceiling:'` returns nothing — this branch
added no `ceiling:` marker.

leodip:
## The review account

Reviewed by `openai-codex/gpt-5.6-sol` at effort `xhigh`. One round at the plan gate and two over
the whole branch at the final gate; under `LEO_REVIEW_MODE=end` no stage was reviewed on its own, so
these three rounds are the entire independent reading this code got. `.review/` is gitignored and
overwritten at the next gate, which is why the account is pasted here rather than pointed at.

---

### Plan gate, round 1 — `275-20260909T014347Z`, fresh session

Verdict `static-review-exhausted`, three findings, all axes reviewed. `unchecked` held no tier or
file the gate depended on: nothing was built yet, no browser pass, untouched security paths, and the
probes were read rather than re-run.

1. **Endpoint matrix per file rather than per wiring point** (quality, significant). Confirmed:
   create and update handlers each carry their own check, client create and settings general carry
   two apiece, and the address endpoints change in stage 1. Fixed in the plan — address HTTP cases
   moved into stage 1, and stage 2 enumerates one refused case per check plus an `error_code`
   assertion on the permission description.
2. **`<b>` alone cannot pin both characters** (quality, significant). Confirmed: for phone, email
   and country the row is refused before the regex runs. Fixed in the plan — two plausible rows per
   validator, one per character, and a `>`-side mutation per validator.
3. **Permission identifiers are sanitized before validation** (conformance, minor). Confirmed by the
   reviewer's scratch test against the real library and by reading both handlers: `valid<b` was
   accepted as `valid`. Section 1's no-op premise was false at those two sites. Resolved under
   decision 1 without escalation — the raw trimmed identifier is validated and `valid<b` is refused
   with the identifier's existing code, which is stricter and changes no interface. One test row in
   stage 2 (API) and one in stage 3 (admin pre-check) pin it.

No refutations. No follow-ups drafted.

---

### Final gate, round 1 — `275-20260909T033151Z`, fresh session

Reviewed the whole branch against `origin/main` at `c264a932`. All three axes read `reviewed`.
Verdict `findings`, two findings, both `judgement: false` with an empty `forced_answer`, so both
were resolved rather than escalated. **Neither touched production code**; the reviewer's own note is
that "the production implementation is sound in the reviewed scope". Both fixes landed in
`f4cf1f38`.

**Finding 1 — quality, significant.** `dcr_consent_test.go:TestDCR_Consent_SelfAssertedNameIsEscaped`.
Decision 6 stopped that test registering a `<script>` name, so stage 3 rewrote it around
`Acme & "Sons"`; the reviewer's point is that ampersands and quotes are not the two characters the
renderer exists to neutralise, and decision 7 migrates nothing, so a client registered before the
reject still holds a complete tag in `clients.description`, which `consentClientName`
(`client_display.go:71`) returns for a DCR client and the consent template renders. Its scratch
mutation trusted the name after escaping only `&` and `"`, and the existing test still passed.

*Confirmed and fixed.* `TestDCR_Consent_RetainedMarkupNameIsEscaped` registers a client with an
accepted name, writes `<b>Existing</b>` into the row with `database.UpdateClient` (the only way to
hold it now — every write path refuses the characters), drives the consent flow, and asserts the raw
bytes carry `&lt;b&gt;Existing&lt;/b&gt;` and not the tag. That the fixture has to be planted by
hand is the point: it is the row shape decision 7 leaves behind.

**Finding 2 — conformance, significant.** The two attribute verbatim tests.
`handler_api_users_attributes.go:279` and `handler_api_group_attributes.go:292` assign the update
value independently of the create sites, and only create had an accepted-value case. The reject
cases return before the assignment and the update success cases send `"inactive"`, which has nothing
in it to rewrite, so the reviewer's scratch mutation rewriting ampersands in both update handlers
survived the whole tier. Every other endpoint pair on the branch has a verbatim case on both wiring
points; these two were the omission. Section 5's seam 3 says update endpoints get no case of their
own "since the check is the same call at the same layer", which is true of the check and not of the
assignment — and it is the assignment that carries the property decision 1 asked for.

*Confirmed and fixed.* Both `..._AmpersandsAndQuotesStoredVerbatim` tests now PUT a second accepted
value onto the row they created and read it back from the response and from the database.

**Evidence.** `where.sh test --type integration --db sqlite --run 'TestDCR_Consent|TestAPIUserAttribute_|TestAPIGroupAttribute_'`
— 8 tests, 8 pass, 0 fail, ~30s. Three mutations, each **caught**, each also showing which sibling
case survives it, which is the family boundary:

| File | Replacement | Result |
|---|---|---|
| `internal/handlers/handler_consent.go` | `"clientName": clientName` to the same wrapped in two `strings.ReplaceAll` calls dropping `<` and `>` | caught by `RetainedMarkupNameIsEscaped`; `SelfAssertedNameIsEscaped` passes, which is exactly the gap the finding named |
| `apihandlers/handler_api_users_attributes.go` | `attribute.Value = req.Value` to `= req.Key` | caught by the verbatim test; `AngleBracketsRejected` passes both subtests |
| `apihandlers/handler_api_group_attributes.go` | `attribute.Value = updateReq.Value` to `= updateReq.Key` | same, on the group twin |

`tree-hash.sh --split` after the last mutation: `production fbcb12968fad`, byte-identical to the
value the round was reviewed at, so no mutation survives in the tree. `integration-tests` moved from
`eb790fbe66c1` to `029224d85959` — these three files. `gofmt -l src/` is empty.

**`unchecked`, all three answered.** *No interactive browser pass* — dismissed; the change is a
validator on write paths, and the one rendering property it bears on is the consent page's escaping,
now pinned at the HTTP seam against raw bytes, which is what a browser would observe. *Preexisting
admin console `innerHTML` sinks* — dismissed as already tracked under #120, named in section 1 and
out of scope in section 2. *No concurrency, revocation, token validation, PKCE, session lifetime,
constant-time or rate limit experiment* — covered by the diff: it touches no file under `oauth/`, no
session or token handler and no rate limiter, and the only route wiring that moved is
`routes/adminconsole` dropping the sanitizer argument.

---

### Final gate, round 2 — `275-20260909T041048Z`, resumed session

Reviewed the whole branch with round 1's fixes on it. All three axes read `reviewed`. Verdict
**clean**, zero findings, so nothing was fixed, deferred or refuted.

**Why the zero is worth something.** A clean verdict is only as wide as what it executed, and this
one did not re-run the suites it had already been handed green. It wrote its own programs: three
scratch SQLite integration runs, each re-installing by hand the exact defect round 1 had found, to
check the repair actually catches it.

| Reviewer's scratch mutation | Result it reports |
|---|---|
| consent renderer trusts the name after escaping only `&` and `"` | accepted-name test passes, `RetainedMarkupNameIsEscaped` fails |
| group attribute update assignment stores the key | rejection test passes, verbatim test fails on response **and** storage |
| user attribute update assignment stores the key | same, on the user twin |

That is the family boundary round 1's own mutation table drew, re-established independently: in each
pair the sibling case still passes, so the new case carries the property rather than duplicating
cover that already existed. Its note: round 1's two findings are "correctly and exhaustively
repaired", production is unchanged, and the focused mutations are caught.

**Both of its factual premises were verified rather than taken.** `where.sh ci` at HEAD reports green
on all twelve jobs at `f4cf1f38`, the four-engine data and integration tiers included.
`git diff --stat c264a932 f4cf1f38` is three `_test.go` files and nothing else, so the production
tree round 2 read is byte-identical to the one round 1 read; `tree-hash.sh --split` agrees,
`production fbcb12968fad` at both. That is what licenses the round to reason from round 1's
production reading instead of redoing it.

**`unchecked`, all three answered** on the same grounds as round 1, with the diff premise verified
above: round 2's diff is three integration test files and `production` did not move, so there is no
new production behaviour for a dynamic probe to reach.

**Gate closed, no round 3.** Nothing was left to raise. The budget agrees independently: rounds 1
and 2 are unconditional, and past that a round is funded only when the previous one changed
production code — round 2 reported no findings and changed no byte of it. No contested finding, no
deferred finding, no open decision.

---

### What a green suite does not carry

Recorded at the final gate and unchanged since. The three stages ran their mutations against trees
that have moved (`tree-hash.sh --verify` reads three of three stage entries stale, and stage 1
recorded only an `all` hash so nothing inside it can be told apart). Every *tier* claim those stages
made locally has since been re-established by CI at `f4cf1f38` on four engines rather than sqlite
alone, so the stale hashes cost nothing to believe there. The mutation evidence is the exception: a
mutation proves a test *can fail*, CI does not run mutations, and that property is not carried
forward by a green suite. The final gate's own mutations, tabled above, are current — they were run
at `production fbcb12968fad`, which is HEAD.

PR #314, opened 2026-09-09
feat(core): replace httprate with an in-tree sliding-window limiter and IP canonicaliser

### LINKED ISSUE #276: httprate links xxh3 and cpuid for a sliding-window counter and an IPv6 truncation
`go-chi/httprate` links two extra modules, `zeebo/xxh3` and `klauspost/cpuid/v2`, for a sliding-window counter and an IP truncation.

## What we use

`core/middleware/middleware_ratelimiter.go`:

| Symbol | Purpose |
|---|---|
| `httprate.NewRateLimiter(limit, window, opts...)` | Two limiters: the request limiter and a 1-per-window `auditGate` |
| `httprate.WithResponseHeaders(ResponseHeaders{...})` | The `X-RateLimit-*` headers |
| `httprate.WithLimitCounter` | Test seam only, see `erroringLimitCounter` in the test file |
| `httprate.CanonicalizeIP` | IPv6 to /64, IPv4 unchanged |

## What replaces it

A `core/ratelimit` package:

- **Counter.** Two-bucket sliding window: `previousCount * overlapFraction + currentCount` against the limit, where the overlap fraction is how far into the current window we are. This is the same approximation httprate uses, and it is about 40 lines.
- **Storage.** `map[string]*bucket` behind a `sync.Mutex`, or a sharded map if contention shows up in practice.
- **`CanonicalizeIP`.** Parse with `net/netip`, mask IPv6 to /64, return IPv4 unchanged. About 10 lines.
- **Response headers.** `X-RateLimit-Limit`, `X-RateLimit-Remaining`, `X-RateLimit-Reset`, and `Retry-After` on a 429, matching what the current tests assert.
- **The `LimitCounter` seam.** Keep it as an interface so `erroringLimitCounter` and the existing failure-path tests carry over unchanged.

Roughly 120 lines.

## The part that needs care

**Eviction.** `map[clientIP]*bucket` with no eviction is a memory exhaustion vector: the rate-limited endpoints are the unauthenticated ones (`/auth/pwd`, ROPC, forgot-password, DCR, self-registration), so the key space is attacker-controlled. httprate evicts on a timer. The replacement needs the same: a background sweep dropping buckets whose window has fully elapsed, or eviction on write past a size threshold. This should be tested explicitly, because it is the failure that only appears under attack.

Secondary: this middleware sits on the credential-check path, and `middleware_ratelimiter_test.go` is 1400+ lines. All of it must keep passing, including the audit-gate pairing (`WithLimitHandler` taking an `http.HandlerFunc`) documented at the top of the current file.

## Value

3 modules retired (`httprate`, `xxh3`, `cpuid/v2`) for ~120 lines. Good trade on paper, but the code is security-relevant, so it earns a careful review rather than a quick merge.

---

### Dependency reduction

leodip:
Closes #276.

## What this is for

`core/middleware`'s rate limiter is built on `go-chi/httprate`, which brings `zeebo/xxh3` and
`klauspost/cpuid/v2` into all three modules for a two-bucket sliding-window counter and an IPv6
/64 truncation. This change moves that counter and that canonicaliser into a package this
repository owns, `src/core/ratelimit`, and retires the three modules from every `go.mod` and
`go.sum`.

## The rule it establishes

The admit and refuse sequence at every key is unchanged: the same rate,
`prev * (window - elapsed) / window + curr`, the same predicate, `round(rate) + 1 > limit`, and
eviction on the same event, the first write after a window rolls. Memory stays bounded by the keys
written in the last two windows, and a test proves it. The limiter's API is key-only: no
`ResponseWriter` or `Request` reaches the counter, so the throwaway writer the middleware used to
carry goes. Store errors fail closed.

Two things ride along because they land in code this change rewrites anyway:

- The audit gate becomes a fixed-window gate over its limiter's own window anchor, so the guarantee
  is exactly one audit event per key per window rather than "at most two per window length in the
  worst phase". It is deliberately not a sliding limiter of budget 1: one with a single event
  behind it goes on refusing through the whole first half of the next window, so a rejection
  twenty seconds past the boundary would never be audited. The plan review caught that before any
  code was written.
- An IPv4-mapped IPv6 address (`::ffff:203.0.113.7`) keys as its IPv4 address instead of
  collapsing to `::` with every other mapped address.

No budget, key, tier, reject shape or header changes. `Retry-After` stays, the four
`X-RateLimit-*` headers stay absent.

## Status

**Complete.** Both stages landed, the review is closed and the check suite is green on the final
commit. `go-chi/httprate`, `zeebo/xxh3` and `klauspost/cpuid/v2` are gone from all three `go.mod`
and `go.sum` files, and the middleware runs on `core/ratelimit`.

Five review rounds ran — one against the plan before any code existed, four over the whole change,
the budget's limit — and every one of them found something real. The fourth is the only one that
reported coverage on all three axes; the three before it were partial or cut off, which is why the
budget was spent to the end. It raised one minor conformance finding, now fixed, and conceded the
one thing round 3 had contested.

One question is left open for you, posted as [a comment on this PR](https://github.com/leodip/goiabada/pull/314#issuecomment-5607738394):
an account identifier longer than 320 octets is keyed by its SHA-256 digest, which means two such
identifiers colliding would share a rate-limit bucket. Every alternative is closed by one of the
findings below, and the residual sits below one this change already accepts on purpose, but it is a
choice rather than a forced answer and it is yours.

### Stage 1: the `ratelimit` package, tested with a fake clock
Landed in `f6c1867f`.

`src/core/ratelimit` is new and nothing calls it yet, so this stage changes nothing observable.
It carries the `Store` interface and the in-process two-map store, the `Limiter` (`New`,
`WithStore`, `Allow`, `Rate`, `Add`), the fixed-window `Gate`, and `CanonicalizeIP`, with 11
tests.

Three deliberate departures from the library, each pinned by a test: keys are stored exact rather
than hashed, so no two clients share a bucket by collision; the gate is fixed-window for the
reason above; and an IPv4-mapped address is unmapped before its family is decided.

Storing exact keys costs something the library did not pay, and this is where it is paid. A map
entry keeps the whole backing array of the string it holds alive, and a rate-limit key is a
substring of a parsed form body whose backing array also holds the submitted password, so the
store clones a key on insertion. The counts sit behind pointers because Go's map assignment
overwrites the stored key with the caller's string — the runtime does it deliberately, "so it can
be garbage collected" — which would have undone that clone on every hit after the first. The test
written for the clone caught it.

Tiers: unit green, 5510 ok 0 fail (`--type modules`, all three modules), `golangci-lint` and
`unparam` clean on the package. Data and integration are not applicable: no query, no schema, no
handler, and no caller at all until stage 2.

Nine mutations, one per behaviour the new tests claim to make observable, all caught. At the time
of this stage the race detector could not be run — the dev container answered `-race requires cgo`
with `CGO_ENABLED=0` and no C toolchain, and CI's unit job was equally without it — so the
concurrency case was backed by the mutation that runs check-and-charge outside the lock. That gap
was closed on this branch afterwards, at f9a773eb: see **Race detector** under Verification.

### Stage 2: the middleware moves to `ratelimit` and httprate, xxh3 and cpuid go
Landed in `925b2a9e`.

The switchover and the retirement. `tier.rl` is a `*ratelimit.Limiter` and `tier.auditGate` a
`*ratelimit.Gate` taken from it, which is what makes them share a phase; `tripped` calls `Allow`,
`Reserve` calls `Rate`, `Release` calls `Add`, `reportTrip` calls `First`, and
`clientIPRateLimitKey` calls `ratelimit.CanonicalizeIP`. The blanked-header option and the
throwaway `ResponseWriter` are deleted, and with them the `*http.Request` four functions carried
only to feed it. `grep -c -e httprate -e xxh3 -e cpuid` returns 0 on all six module files.

Watch the polarity at `tripped`: `OnLimit` answered true when a request was *over* budget and
`Allow` answers true when it is *within* budget, so the guard is inverted from the one it replaced.
Getting that wrong compiles, passes vet, and fails open on every every-request tier. A mutation
pins it.

**Release note.** Three modules retired. Budgets, keys, tiers, reject classes and the 429 shape are
unchanged, and the 20 middleware tests that pin every budget exactly on both sides pass with a
single edit — the fail-closed stub shrinks from httprate's four-method `LimitCounter` to the
two-method `ratelimit.Store`. Two changes are visible to an operator, stated once under **For the
release notes** below.

Tiers: unit green, 5510 ok 0 fail — the same count as stage 1, since no case was added, removed or
relaxed. `golangci-lint` and `unparam` clean. Data and integration not applicable: no query,
migration or schema, and the integration suite runs with the limiter disabled (#219 section 5).
Three mutations against the rewired call sites, all caught.

Section 0's anchor table was re-swept: thirteen rows re-pointed, and five retired because the code
they cited — the header option, the throwaway writer and the three `httprate v0.16.0` module lines
— no longer exists, which is the goal.

## What the review changed

Both stages were built without a per-stage review, so the whole change is read at the end. Four
rounds ran, the budget's limit, and each found something real. Three landed on the same surface —
the account rate-limit key — and the fourth on the rationale written to justify it. The full
account, including the places a fix departed from what was recommended and why, is in the
agreement's closing log.

### Round 1, in `4f2d91e6`

- **The account key was unbounded** (security). `accountRateLimitKey` returned the whole submitted
  form value, off an unauthenticated form with no body cap, and the store retains a key for two
  windows. Storing exact keys is what stops two accounts sharing a bucket, and this is what it
  costs: the identifier now has to be bounded on the way in.
- **`Gate.First`'s atomicity was not observable** (quality). It is a compound read-and-record and
  every case for it was sequential, so the lock could be removed with the suite still green.

### Round 2, in `82d753b7`

- **The bound aliased accounts it assumed could not exist** (security). Round 1 bounded the key by
  folding everything past 320 octets into one shared bucket, on the premise that nothing longer
  could name an account. It can: `ValidateEmailAddress` checks an address's shape without any
  length check, self-registration and the setup program both use it, and `users.email` is `TEXT`
  on sqlite. (The other three engines cap it at `varchar(64)`; one engine is enough.) An account
  past the threshold therefore shared a bucket with every unrelated oversized submission, and five
  of them spent its whole password-reset budget for the window.

  Fixed by digesting the overlong tail rather than folding it, so no two identifiers collide at
  any length on any engine. The threshold stops being a security boundary and becomes a legibility
  one — an address stays readable in the warning line and the audit event; only something longer
  than any address arrives as hex. (This round also claimed both branches of the key were
  injective. Round 3 showed that was false and why it cannot be true; see below.)
- **The new `Gate` atomicity test was probabilistic** (quality). It asserted an outcome, so it
  needed the scheduler to interleave two callers: against the exact mutation it was written for it
  passed four runs in six. Replaced by one that pins the cause — park the first caller inside the
  clock read `First` performs under its lock, then require `TryLock` to fail from outside. Six of
  six.

### Round 3, in `60d157bd`

This verdict was salvaged: the reviewer was cut off before it reported, so it recorded no coverage
on any axis and its two findings are a floor rather than its last word. Both were confirmed, one
of them only in part.

- **The digest key aliased the exact key** (security). A digest key is `<sha256>` and sixty-four
  hex characters — seventy-two octets, far inside the 320-octet bound — so submitting one back as
  an ordinary identifier was keyed exactly and landed in the bucket of the long identifier it
  names, with no SHA-256 collision involved. And the digest is not a secret: it is written to the
  warning line and to the audit event, so a reader of either could spend a long account's budget
  without ever knowing the identifier behind it. The prefix is now reserved — any submission
  spelled with it is digested too — so the exact branch cannot emit a key carrying it and the two
  branches are disjoint.

  Round 2's claim that both branches were injective was also false, and unachievably so: a bounded
  key over an unbounded set of identifiers cannot be injective, and the key is bounded because
  retaining a 10 MiB form value for two windows was round 1's finding. The comments now say what
  actually holds. The reviewer's wider reading — that hashing at all contradicts the agreement's
  decision 5 — is refuted in the closing log: that decision governs the store's key representation,
  which still holds exact strings and hashes nothing.
- **The `Gate` atomicity test pinned the entrance, not the section** (quality). Round 2's fix
  parked its caller inside the clock read, which `First` performs one statement after taking the
  lock, so it proved the lock is held on the way in and nothing about the read and the record that
  follow. An unlock moved below the clock read, or below the store read, left every Gate case green
  in a hundred consecutive runs. The park moves to the far end instead — inside the store's write,
  the last thing `First` does under the lock — which catches all three truncations. This is the one
  place the change touches production for a testing reason: `Gate`'s store field becomes an
  unexported interface so the package's own tests can substitute one that parks.

### Round 4, in `29378118`

The last round the budget allows, and the only one to report coverage on all three axes. It
confirmed the fixes above and their coverage, re-ran the mutations rounds 2 and 3 added, and raised
one minor finding — against a comment rather than the code.

- **The digest's rationale rested on a false analogy** (conformance). The comment justifying the
  SHA-256 key argued that collision resistance "is already what the token endpoint rests on for PKCE
  S256". It is not. RFC 7636 section 4.6 compares `BASE64URL-ENCODE(SHA256(ASCII(code_verifier)))`
  against the challenge already associated with the authorization code, so passing it with any other
  verifier is a second preimage of a fixed target, not a collision between two values an attacker
  picks. The comment now states the residual on its own terms, keeping the two costs apart: some two
  identifiers sharing a bucket is a generic collision at around 2^128, and landing a submission in
  one chosen account's bucket is a second preimage at around 2^256. The width argument the paragraph
  exists for is untouched — a 64-bit hash collides at around 2^32, which is what made the retired
  library's shared buckets constructible.

  The same conflation was in the open deferred decision below, which is what you read to settle it,
  so it is corrected in the closing log's round-4 entry. **The assumption there does not change**:
  the argument removed was comparing the residual to a different property and was doing no work, and
  the two that carry the decision — 2^128 is out of reach, and every alternative reopens an earlier
  confirmed finding — still stand.

- **Round 3's contested finding was conceded**, on the boundary rather than on assertion: decision 5
  governs the store's key representation inside `ratelimit`, which still holds exact strings and
  hashes nothing, while the digest is in `core/middleware` and bounds what the caller hands the
  limiter.

**Nothing about the release note changes.** All four rounds touched keying, test strength and the
reasoning written around them, not budgets, keys, tiers, reject classes or the 429 shape.

### Review

Reviewed by `openai-codex/gpt-5.6-sol` at effort `xhigh`, 5 rounds — 1 against the plan and 4 over
the whole branch. `LEO_REVIEW_MODE=end`, so no stage was reviewed on its own.

|---|---|---|---|
| plan | 5 | 2 | all fixed in the plan, before any code |
| final 1 | 2 | 1 | all fixed, `4f2d91e6` |
| final 2 | 2 | 0 | all fixed, `82d753b7` |
| final 3 | 2 | 1 | 1 fixed, 1 fixed in part and refuted in part, `60d157bd` |
| final 4 | 1 | 0 | fixed, `29378118` |

- **blocking, security (plan):** a sliding audit sibling of budget 1 would suppress the first
  rejection through half of the next window. Fixed in the plan: the gate became a fixed-window
  `Gate` over the parent's anchor.
- **blocking, security (plan):** string keys retain the parsed form buffer and the password in it.
  Fixed in the plan and pinned in `f6c1867f`: `memStore.add` clones the key.
- **blocking, security (final 1):** `accountRateLimitKey` returned the whole submitted form value
  off an unauthenticated form with no body cap, and the store retains a key for two windows. Fixed
  in `4f2d91e6`: the identifier is bounded on the way in.
- **significant, quality (final 1):** `Gate.First`'s atomicity was unobservable — the lock could be
  removed with the suite green. Fixed in `4f2d91e6`, then twice more as the pin was found too
  narrow (`82d753b7`, `60d157bd`).
- **significant, security (final 2):** round 1's bound folded every identifier past 320 octets into
  one shared bucket, on a premise sqlite falsifies. Fixed in `82d753b7`: the overlong tail is
  digested, so no two identifiers share a bucket at any length on any engine.
- **significant, quality (final 2):** the new atomicity test was probabilistic, 4 runs in 6. Fixed
  in `82d753b7` by pinning the cause rather than the outcome.
- **blocking, conformance (final 3):** the digest key was itself submittable, so it aliased the
  exact branch with no collision involved — and the digest is written to the warning line and the
  audit event. Fixed in `60d157bd`: the prefix is reserved, so the branches are disjoint.
- **minor, conformance (final 4):** the digest's rationale rested on a false PKCE S256 analogy.
  Fixed in `29378118`, in the comment and in the closing log's deferred decision.

Full account, with the evidence, the refutations and the places a fix departed from what was
recommended, in [this comment](https://github.com/leodip/goiabada/pull/314#issuecomment-5607738994).

## For the release notes

Nothing to do on upgrade. No configuration, schema or migration changes, and no budget, key, tier,
reject class or header changes. Two behaviours are observable to an operator:

- **An IPv4 client behind a dual-stack proxy stops sharing one bucket.** An address the proxy
  reports as `::ffff:a.b.c.d` used to collapse to `::`, the same bucket as loopback and as every
  other such client, so those clients spent the per-IP budgets collectively. It now keys as its
  ordinary IPv4 address. Deployments behind a dual-stack proxy will see per-IP limits stop firing
  early for unrelated users.
- **The rate-limit audit event fires once per window, exactly.** It used to fire on a gate phased
  independently of its limiter, giving "at most one per key per window and at most two per window
  length in the worst phase". `AuditRateLimitExceeded` volume may drop slightly around window
  boundaries; the guarantee is now exact.

Counters remain per process, so a deployment on N replicas still multiplies every budget by N and
resets them on deploy. That is unchanged and is already documented.

## How this was verified

| Tier | Where | Result |
|---|---|---|
| unit, all three modules | local, every stage and every review round | green; 5524 pass, 0 fail at the last run that added a case |
| gofmt guard, repository-wide | local, inside the unit tier | green |
| `golangci-lint`, `unparam` | local, on the new package and the rewired middleware | clean |
| data, four engines | local at rounds 1–3 | 2208 pass, 0 fail |
| integration, sqlite | local at round 3 | 1371 pass, 0 fail |
| **the check suite, all twelve jobs** | **GitHub, at `29378118`** | **green** |

The check suite is what covers what no local run here can: the data tier on mysql, postgres, mssql
and sqlite together, the integration tier, Build validation, Lint, Vulnerabilities and Versions. It
is green on the final commit, which is the precondition for this PR leaving draft.

Beyond the tiers, every behaviour the new tests claim to make observable was pinned by a mutation:
nine at stage 1, three at stage 2, and one per fix at each review round, all caught. Where a
mutation survived — the first two attempts at the `Gate` atomicity pin — that is recorded in the
review account rather than quietly re-run. `check-anchors.sh` passes: all 20 rows of the
agreement's section 0 resolve to exactly one line in the named function.

**Race detector.** Until f9a773eb the detector had not run at any tier, and could not here: the dev
container answered `-race requires cgo` with `CGO_ENABLED=0` and no C toolchain, and CI had no race
job. That was a standing property of the repository rather than a gap in this change, and it was
drafted as the one follow-up. It is closed on this branch instead. `run-tests.sh` takes `--race`,
which runs the three module tiers under `go test -race` with `CGO_ENABLED=1` for that command
alone; the dev container image gained gcc and a `CGO_ENABLED=0` pin so ordinary builds stay as they
were; and check.yml gained `Unit / race`, one job with the three tiers as steps on the same golang
image as the other unit jobs, which ships gcc. Both runs are green with no race reported: locally
in the rebuilt container, 61 packages ok across the three modules, and in CI at
[Unit / race](https://github.com/leodip/goiabada/actions/runs/34407350807/job/102653446975), 4m29s
against about 1m30s for each plain unit job. The plain jobs stay, since the detector changes timing.
The data and integration tiers do not run under it; that narrower gap is what remains of the
follow-up.

## Deviations

Five, all recorded with their evidence in the review account:

1. **Round 1's fix did not take its forced answer.** The reviewer asked for the bound at "the
   repository's 60-byte valid email limit". Only two validators carry that check; self-registration
   validates through `ValidateEmailAddress`, which has none. Bounding at 60 would have folded real
   accounts into a shared bucket. The bound is 320 octets instead.
2. **Round 1's recommendation to assert the bounded key reaches the real store was refused.** That
   is the side channel the agreement's section 5 rejects by name. The property is pinned at two
   seams instead — the function's own table, and the fold observed through HTTP.
3. **Round 2's fix did not take its forced answer.** The reviewer asked for a length invariant
   enforced across every creation, update, activation and seed path, plus remediation of existing
   sqlite rows. That is a behaviour change outside section 2 as frozen, and it defends the aliasing
   rather than removing it. The key was made disjoint instead.
4. **Round 3's framing was refuted**, with the demonstration: decision 5 governs the store's key
   representation inside `ratelimit`, which still holds exact strings and hashes nothing, and
   injectivity cannot be restored to a bounded key over an unbounded set of identifiers. The
   concrete half of the same finding was confirmed and fixed. The reviewer conceded this in round 4.
5. **Half of round 1's second finding was filed rather than fixed** — the detector run, for the
   reason above.

One bookkeeping deviation with no effect on the code: stages 1 and 2 each recorded a single `all`
tree hash rather than `tree-hash.sh --split`'s categories, so `--verify` at the final gate could
only report that the whole tree had moved. Reconciled by hand before round 1 and recorded in the
review account; every round since used `--split`.

## The closing comments

- [Deferred decisions](https://github.com/leodip/goiabada/pull/314#issuecomment-5607738394) — the one open question, and it is yours.
- [Follow-ups, fold-ins and ceilings](https://github.com/leodip/goiabada/pull/314#issuecomment-5607738714) — one drafted issue, since closed on this branch at f9a773eb rather than filed, nothing folded in, one ceiling.
- [The review account](https://github.com/leodip/goiabada/pull/314#issuecomment-5607738994) — all five rounds in full.


leodip:
## Deferred decisions

One question the run met, judged safe to proceed past on a stated assumption, and left for you.
It is the only one: the agreement's eight sealed decisions are all `Decided` and none was reopened.

---

## Deferred decision, raised at the final gate by round 3

Belongs to section 8.

9. **Should an account identifier longer than 320 octets be keyed by a SHA-256 digest, accepting
   that two such identifiers colliding would share a rate-limit bucket?** Raised by round 3's
   finding 2 at the final gate. Status: **Open**

   **Assumed:** yes, it is keyed by the digest, which is what is committed. Per `decisions.md`
   section 1b this is real but not load-bearing, and the three numbers are:

   1. *The residual of each option.* The unbounded exact key retains up to 10 MiB per accepted
      request for two windows (round 1's confirmed finding). The shared sentinel aliases at zero
      cost, constructible by anyone (round 2's confirmed finding). A 64-bit hash collides at about
      2^32 trial identifiers, which is a laptop. The 256-bit digest had one constructible alias,
      the cross-branch one, closed in this round and pinned by two subtests and a CAUGHT mutation;
      what is left needs a SHA-256 collision at about 2^128. No new probe was committed for this:
      the executed demonstration is the test suite itself, since the alias is exactly the thing the
      new subtests fail without, and the remaining number is analytic rather than measurable.
   2. *A residual this change already accepts on purpose.* Counters are per process, so a
      deployment on N replicas multiplies every budget by N and resets them on deploy. That is
      certain rather than improbable, it is stated in the reference documentation, and section 2
      puts shared counters out of scope.
   3. *The rate the underlying primitive fails anyway.* This server already treats SHA-256 as
      load-bearing for authentication: `core/oauth/oauth.go:9` hashes the PKCE code verifier and
      the token endpoint accepts an authorization code on digest equality. A SHA-256 collision
      there is an authentication bypass; here it is two accounts sharing a rate-limit bucket.

   The qualifier in 1b is met. Rarity counts only where an attacker cannot steer it, and the one
   steerable path -- spelling a submission as a digest key -- is closed rather than argued away.
   Steering what remains means constructing a SHA-256 collision, and anyone who can do that has the
   token endpoint.

   **Cost to reverse:** the alternatives are not cheap and none is contained. Enforcing a length
   invariant on account identifiers reaches the validators, self-registration, the setup program,
   the seeder and existing sqlite rows, refuses identifiers the server accepts today, and is
   outside section 2 as frozen. Keeping the digest but widening it buys nothing. Reversing to any
   bounded exact key reopens round 2's finding.


---

**Correction carried from round 4.** The third number above — "the rate the underlying primitive
fails anyway" — argued that SHA-256 is load-bearing for PKCE S256 and that a collision there would
be an authentication bypass. That is wrong: RFC 7636 section 4.6 compares the hash of the submitted
verifier against the challenge already stored with the code, so passing it means a *second preimage*
of a fixed target, not a collision. The token endpoint does not fall at the same moment this bucket
does, and that argument should be read as removed rather than weighed.

**The assumption does not change.** The argument removed compared the residual to a different
property and was doing no work. The two that carry the decision stand: a generic collision costs
around 2^128 and choosing whose bucket costs around 2^256, while every alternative reopens round 1's
or round 2's confirmed security finding or reaches outside section 2 as frozen.

leodip:

### Drafted follow-ups, not filed

One. Re-searched at the finish, over every state: `gh issue list --state all --search "race detector"`
and `--search "go test -race"` and `--search "detector"` return nothing about a detector tier — the
hits are #301, #128, #133, #127 and #199, all about concurrency in the data and session paths rather
than about how tests are run. Not tracked, open or closed.

**Update, same day:** closed on this branch at f9a773eb rather than filed. See **Closed here** at
the end of section 9. The draft below is kept as written for the record.

---

## Section 9, follow-up: no test tier runs the race detector, and the dev container cannot

Drafted at the plan review, 2026-09-09, from finding 2 of round 1. ~~Not filed.~~ Closed on this
branch, see **Closed here** below.

**Title:** No test tier runs `go test -race`, and the dev container cannot (`CGO_ENABLED=0`, no C toolchain)

**Labels:** testing, ci, go

**Anchor:** `src/authserver/run-tests.sh`, the three module legs (`go test -v -count=1` at the
`internal`, `core` and `adminconsole` invocations); `.github/workflows/check.yml`, the `unit-core`
job, which delegates to the core leg with `--type core --no-build`.

**What the code does today.** Every unit invocation is plain `go test -v -count=1`, locally and in
CI. Inside the dev container, `go test -race -count=1 ./constants/` from `src/core` answers
`go: -race requires cgo; enable cgo by setting CGO_ENABLED=1`; `go env CGO_ENABLED` is `0` and
`gcc` is absent, so the detector cannot be enabled there without installing a toolchain. The host
has no Go.

**Why it is a defect.** The repository has code whose correctness is a locking argument: the
limiter in `src/core/ratelimit` (#276) serialises check-and-charge under one mutex and has a
1,000-goroutine case asserting an exact budget; `failureTier` in `core/middleware` keeps an
in-flight count under its own mutex; the session and token paths run concurrent transactions on
purpose (#301). None of it has ever run under the detector, so a data race that does not happen to
corrupt a count in a given run is invisible to every tier. The #276 concurrency case is pinned by a
mutation that runs the body unlocked, which the runtime's concurrent-map-write check catches, but a
mutation proves one line and the detector proves the package.

**What closing it takes.** One CI job on the ubuntu runner, which has gcc: `CGO_ENABLED=1 go test
-race -count=1 ./...` in `src/core` at least, and ideally the three modules, as a fourth unit job
beside `unit-core`, `unit-internal` and `unit-adminconsole`. Optionally `gcc` in the dev container
image so `run-tests.sh` can take a `--race` flag and developers can run it before pushing. Nothing
in the production build changes: cgo is needed only for the detector's runtime.


**File it with:**

```bash
gh issue create \
  --title 'No test tier runs `go test -race`, and the dev container cannot (CGO_ENABLED=0, no C toolchain)' \
  --label testing --label ci --label go \
  --body 'Every unit invocation in `src/authserver/run-tests.sh` and in `.github/workflows/check.yml` is plain `go test -v -count=1`, locally and in CI. Inside the dev container `go test -race` answers `go: -race requires cgo; enable cgo by setting CGO_ENABLED=1`; `go env CGO_ENABLED` is `0` and `gcc` is absent, so the detector cannot be enabled there. The host has no Go.

This matters because the repository has code whose correctness is a locking argument: the limiter in `src/core/ratelimit` (#276) serialises check-and-charge under one mutex and has a 1,000-goroutine case asserting an exact budget; `failureTier` in `core/middleware` keeps an in-flight count under its own mutex; the session and token paths run concurrent transactions on purpose (#301). None of it has ever run under the detector, so a data race that does not happen to corrupt a count in a given run is invisible to every tier. #276 pinned its concurrency cases with mutations that the runtime'"'"'s concurrent-map-write check catches, but a mutation proves one line and the detector proves the package.

Closing it takes one CI job on the ubuntu runner, which has gcc: `CGO_ENABLED=1 go test -race -count=1 ./...` in `src/core` at least, ideally all three modules, as a fourth unit job beside `unit-core`, `unit-internal` and `unit-adminconsole`. Optionally `gcc` in the dev container image so `run-tests.sh` can take a `--race` flag. Nothing in the production build changes: cgo is needed only for the detector'"'"'s runtime.

Drafted as a follow-up to #276.'
```

**Why it was not folded in.** It fails fold-in test 1: it rides on no surface this change touches —
`run-tests.sh` and `.github/workflows/check.yml` are not in the diff, and adding a CI job is a
different layer at a different seam from a counter package. It would need a stage of its own to add
the job, prove it green on the runner, and decide whether the container image grows a C toolchain.

**Closed here.** That stage of its own is f9a773eb, on this branch after the review closed. Four
pieces:

1. `run-tests.sh` takes `-R`/`--race`. The three module legs share one `go test` invocation, which
   under the flag becomes `env CGO_ENABLED=1 go test -race -v -count=1`, so cgo is on for that
   command alone. Refused with exit 2 for `--type data` and `--type integration`, which it does not
   cover: integration runs the binary `build.sh` produced, and data would need the detector per
   database. Summary rows read `Core (race)` and so on.
2. `check.yml` gains `unit-race`, `Unit / race`: one job, the three tiers as steps, on the same
   golang bookworm image as the other unit jobs. The "ubuntu runner, which has gcc" above was one
   step further than needed: the golang image ships gcc too, and Go turns cgo on wherever it finds a
   compiler, so the job installs nothing. Mailpit is declared for the core leg's email test. The
   plain unit jobs stay, since the detector slows a run several times over and changes its timing.
   `release.yml` reuses `check.yml`, so the release matrix gains the job with no edit.
3. The dev container Dockerfile installs `gcc` and `libc6-dev`, and `devcontainer.json` pins
   `CGO_ENABLED=0` in `containerEnv`. The pin is what keeps the compiler's arrival from changing
   anything else: without it every `go build` in the container would silently switch to cgo, unlike
   the release binaries, which `build-binaries.sh` builds with `CGO_ENABLED=0` exported.
4. `CLAUDE.md` and `AGENTS.md` describe the flag and the job under Testing.

Both first runs are green with no race reported. Locally, in the rebuilt container, `gcc` is
present, `go env CGO_ENABLED` reads `0`, the exact command quoted above now passes, and
`./run-tests.sh --type modules --race --no-build` reports 61 packages ok across the three modules,
the 1,000-goroutine budget case, the `Gate` concurrent-first case and the two concurrent
double-spend cases at the token endpoint among them. In CI,
[Unit / race](https://github.com/leodip/goiabada/actions/runs/34407350807/job/102653446975) took
4m29s against about 1m30s for each plain unit job. What remains of the follow-up is narrower than
the draft: the data tier under the detector, per database, where the concurrent-transaction cases
of #301 would be the beneficiaries.

### Folded in

**Nothing was folded in during the run.** Every fix in the diff traces to either a sealed stage or a
confirmed review finding; there is no hunk the agreement and the review account do not explain.

Two things were folded in at *spec* time and are named in the agreement's §2 goal, so they are not
run-time fold-ins and are recorded here only so the list is complete: decision 6, the audit gate
phase-aligned with its limiter, and decision 7, IPv4-mapped addresses keyed as their IPv4 address.
Both ride on code the retirement rewrites anyway.

One `ceiling:` marker added on this branch.

- `src/core/ratelimit/limiter.go:43` — one mutex serialises every key of this limiter, so a tier
  admits one check-and-charge at a time per process. That is httprate's own shape, and the limited
  routes are login forms and a token grant rather than a hot path. **Revisit when** a profile shows
  contention here contributing measurably to the latency of a limited route; sharding the store by
  key is the next shape.

leodip:
## The review account

`LEO_REVIEW_MODE=end`, so no stage was reviewed on its own and this is the whole of the code review
for this branch. Five rounds in total: one against the plan before any code existed, and four at the
final gate over the whole change. Pasted in full rather than linked, because `.review/` is
gitignored and overwritten by the next gate — this comment is the only surviving record of what was
found, what was refuted, and what was deliberately left.

| Round | Reviewer | Findings | Blocking | Outcome |
|---|---|---|---|---|
| plan | `openai-codex/gpt-5.6-sol` xhigh | 5 | 2 | all fixed in the plan, before code |
| final 1 | same, fresh | 2 | 1 | all fixed, `4f2d91e6` |
| final 2 | same, resumed | 2 | 0 | all fixed, `82d753b7` |
| final 3 | same, resumed (salvaged) | 2 | 1 | 1 fixed, 1 fixed in part and refuted in part, `60d157bd` |
| final 4 | same, resumed | 1 | 0 | fixed, `29378118` |

Round 4 is the only one that reported coverage on all three axes; rounds 1 to 3 recorded partial or
salvaged coverage, which is why the budget was spent to four.

---

## Section 10, plan review round 1

Request `276-20260909T154235Z`, `type: plan-review`, round 1 of 2. Reviewer
`openai-codex/gpt-5.6-sol`, effort xhigh, fresh session. Verdict `findings`, five; `unchecked`
held only what a plan review cannot reach (no code, stage 2 a sketch, no integration or engine
run, all dismissed with reasons in `log/plan.md`). Full account there; per finding:

1. Sliding audit sibling suppresses the first rejection in half the next window. Security,
   blocking. **Confirmed and fixed in the plan**: the gate becomes a fixed-window `Gate` over the
   parent's anchor (`First(key)`), reading only the current-window count. Arithmetic: rate `1-f`,
   `round(1-f)+1 > 1` for all `f <= 0.5`. Lands in stage 1 with a boundary test and two mutations.
2. Concurrency case not run under `-race`. Quality, significant. **Confirmed; half fixed, half
   filed**: the serialization mutation is in stage 1; the detector run is impossible here
   (`-race requires cgo`, `CGO_ENABLED=0`, no gcc, no host Go) and is the section 9 follow-up
   above. Deviating from the forced answer on that half is on that evidence.
3. Release note claimed no observable change. Conformance, minor. **Confirmed and fixed in the
   plan**: stage 2's note names the mapped-address bucket and the audit timing.
4. `Limiter.mu` ceiling without a marker. Conformance, minor. **Confirmed and fixed in the plan**:
   `ceiling:` comment with a profiling trigger specified in stage 1 step 2.
5. String keys retain the parsed form and its password. Security, blocking. **Confirmed and fixed
   in the plan**: `memStore.Add` clones a new key; pinned with `unsafe.StringData` and mutated.
   Verified against `accountRateLimitKey` (`middleware_ratelimiter.go:891`) and `url.ParseQuery`'s
   no-escape fast path.

Bookkeeping: the zone-stripping rationale corrected in step 3. Nothing refuted. No production code
moved (none exists); the round is recorded as productive on the two §4 amendments it caused.



## Section 10, stage 1 gate: no review

`LEO_REVIEW_MODE=end`, so stage 1 was not reviewed on its own; the whole change goes to one review
after stage 2 lands. What checked this code before it was committed: the unit tier on all three
modules (5510 ok, 0 fail), `golangci-lint` and `unparam` clean on the new package, and nine
mutations, one per behaviour the new tests claim to make observable, every one caught. Full account
in `log/stage-1.md`.

One thing from that stage is worth carrying to whoever reviews the whole change, because it is the
kind of defect a reader will not see by reading the diff. `memStore` holds its counts as
`map[string]*int` rather than `map[string]int`. That is not a taste choice: Go's map assignment
overwrites the stored key with the caller's string — deliberately, "so it can be garbage collected"
— so `m[key]++` would have put the caller's uncloned substring back into the map on every hit after
the first and undone the clone that keeps a submitted password from being retained for two windows.
Simplifying `*int` back to `int` silently restores the retention. Both the struct field and
`memStore.add` carry the reason in comments; the retention test is what caught it.


## Section 10, final gate: freshness of the record before round 1

`tree-hash.sh docs/issue-276-ratelimit/ --verify`, run before the round 1 request was written, at
HEAD `925b2a9e` with a clean working tree:

```
current   production 34cba1c5def2   unit-tests e35ca55bf300   data-tests 6483c3303341
          integration-tests 029224d85959   docs acc66fc3990a   all a6a796f804c8

stage 1  only `all` recorded, d1a6f328354d, now a6a796f804c8: moved, and which tier cannot be told apart
stage 2  only `all` recorded, 34cba1c5def2, now a6a796f804c8: moved, and which tier cannot be told apart

2 of 2 entries record a result against a tree that has since moved.
```

Both entries recorded a single hash rather than the `--split` categories, so `--verify` can only say
the whole tree moved. What that costs, and what it does not:

- **Stage 1's tiers are stale in the strict sense and sound in fact.** Its recorded `all`,
  `d1a6f328354d`, is the tree before stage 2 landed. Stage 2 changed the middleware, its test, one
  constant comment and the six module files; it did not touch `src/core/ratelimit/`, which is what
  stage 1's nine mutations were run against. The re-run that settles it is free: the unit tier is 8
  seconds and CI has run it at HEAD.
- **Stage 2's hash is stale only because it is labelled `all`.** The value it recorded,
  `34cba1c5def2`, is byte for byte today's `production` category, so no production file has moved
  since that stage was verified. The mutations recorded in `log/stage-2.md` still hold against the
  tree the reviewer is reading.
- **What actually moved between stage 2's record and now is the agreement's own logs**, which
  `tree-hash.sh` excludes from every category and therefore cannot explain the difference; the
  remaining delta is stage 2's own commit landing after its hash was taken.

Independent of any of that, `where.sh ci` reports green at `925b2a9e` across all twelve jobs,
including the data and integration tiers on sqlite, mysql, postgres and mssql, which no local run in
this issue covered.

## 10. Review round 1, gate final -- request 276-20260909T165732Z

**Reviewer:** pi openai-codex/gpt-5.6-sol, effort xhigh, fresh session. Started
2026-09-09T16:58:59Z, completed 17:24:48Z.

**Verdict:** `findings`, 2 of them. All three axes read `reviewed`. `unchecked` held three
entries, answered below. This was the branch's only reading, per `LEO_REVIEW_MODE=end`.

### Finding 1 -- oversized account identifiers became long-lived exact map keys

Axis security, severity **blocking**, confidence high, `judgement: true` with a populated
`forced_answer`, so the code settles it and no escalation is owed.

**Claim.** `accountRateLimitKey` returned the whole submitted form value, and `memStore.add`
clones and retains it for two windows. httprate hashed every key to a `uint64` and so retained
8 bytes whatever arrived; exact string keys (decision 5) made the retained size the submitted
size. The forms are unauthenticated and set no body cap, so net/http's 10 MiB form limit was
the only ceiling, and `LimitForgotPwd` admits 20 per client block per window.

**Confirmed, and fixed** in `4f2d91e6`.

Verified before working it, rather than taken on the reviewer's word:

- `middleware_ratelimiter.go:803` reads `r.FormValue("email")` straight into the key, and
  `:525` and `:948` do the same for `LimitPwd` and `LimitROPC`.
- `grep -rn "MaxBytesReader" src --include=*.go` finds five call sites, all on the profile
  picture, client logo, session and OTP API handlers. None of the three limited forms.
- `ratelimit/store.go:add` clones the key into `s.curr`, and its own comment already states
  the contrast with httprate's `uint64`.

Decision 5 is not what this contradicts. It costed exact keys at "roughly 60 to 100 bytes per
entry", which is true of every identifier that can name an account and silently assumed the
identifier was bounded. Bounding it is what makes that costing correct, so this is a premise
decision 5 did not state rather than a disagreement with it.

**Where the fix departs from the forced answer, and why.** The reviewer asked for the bound at
"the repository's 60-byte valid email limit". Sixty is wrong, and reachably so: only
`ValidateEmailUpdate` and `ValidateEmailChange` carry the 60-octet check, while
`handler_account_register.go:94` validates self-registration through `ValidateEmailAddress`
alone, which has no length check at all, and `users.email` is `TEXT` on sqlite and postgres.
So a real account can hold an email past 60 octets, and folding it into the shared bucket would
let a stranger's garbage spend that account's login and reset budget -- reintroducing at the
tail exactly the cross-principal interference decision 5 rejected hashing to avoid.

The bound is instead the longest an address can be, 320 octets, fetched from the text rather
than recalled: RFC 5321 section 4.5.3.1.1, "The maximum total length of a user name or other
local-part is 64 octets", and 4.5.3.1.2, "The maximum total length of a domain name or number
is 255 octets", plus the `@`. Nothing past it can name an account on any engine, so the shared
bucket can never hold one. Retention falls from 10 MiB to 320 octets per entry, a factor of
about 32,000, and the same bound now applies to the key `reportTrip` writes into the warning
line and the audit event.

The reviewer also asked the new middleware case to "assert the bounded key reaches the real
ratelimit store". That is the side channel section 5 rejected by name -- an assertion on
`tier`'s store from the middleware, into a package that owns the assertion itself. The property
is pinned at two seams instead: `TestAccountRateLimitKey_BoundsTheIdentifier` is the function's
own table, including a 10 MiB input, and `TestLimitForgotPwd_PerEmailAndPerIP` gains
"distinct oversized addresses share one bucket", which observes the fold through HTTP and also
shows a real address surviving the flood.

**Mutations.** The length check replaced with `if false {`: MUTATION CAUGHT against
`TestAccountRateLimitKey_BoundsTheIdentifier`, and MUTATION CAUGHT again against
`TestLimitForgotPwd_PerEmailAndPerIP`, so the seam case pins it independently of the unit table.

### Finding 2 -- Gate concurrency was not observable to the suite

Axis quality, severity **significant**, confidence high, `judgement: false`, so no human call
is owed.

**Claim.** `Gate.First` is a compound read-and-record but every Gate case is sequential, and
the reviewer's scratch mutant -- lock, then unlock immediately, before get/check/add -- passed
the whole `ratelimit` suite.

**Confirmed, and fixed** in `4f2d91e6`, with a correction the reviewer's own recipe did not
survive. Its mutation was reproduced exactly, through `where.sh mutate` on
`src/core/ratelimit/gate.go`, replacing the deferred unlock with an immediate one.

The first version of the new case followed the reviewer's recommendation literally -- 1,000
goroutines released onto one fresh Gate key, asserting exactly one `true`, mirroring
`TestLimiter_ConcurrentAllowChargesExactlyTheBudget`. It reported **MUTATION SURVIVED**. The
reason is that `First`'s body is a map read and a map write, a few nanoseconds during which the
scheduler has to land a second caller, and after the first `add` every later caller reads a
count of 1 and correctly answers false. So the race window is not the test's duration, it is
one caller's get-to-add interval, and 1,000 goroutines did not hit it.

The committed case widens that interval with the package's own unexported clock, which is what
decision 3 put there: `First` calls `now()` with the lock held, so a clock that pauses a
millisecond makes a correct Gate admit one caller at a time and an unlocked one admit all 32 at
once. The pause is not what the case asserts, so a slower machine only makes it more certain.
Re-run: **MUTATION CAUGHT**, the runtime aborting the binary with `fatal error: concurrent map
writes` at `gate_test.go:100`.

That the reviewer's own recommendation did not hold is worth recording. Its `mutations_run`
entry proved the mutant survives the *existing* tests, which is a true claim about the gap, and
it did not establish that the test it recommended would close it.

- *"The race detector was not run ... CGO_ENABLED=0 and no C toolchain."* **Dismissed.** True of
  the container and of CI, which has no race job either, so it is a standing property of this
  repository rather than a gap in this gate. Go's runtime detects concurrent map access without
  cgo, and that is precisely what caught the Gate mutation above.
- *"Data and integration tiers were not rerun locally."* **Covered.** Data was re-run this
  session on all four engines after the fix, 2208 passing and 0 failing. Integration stays not
  applicable for the reason section 5 gives: the suite runs with the limiter disabled (#219), so
  nothing in this change is reachable through an endpoint.
- *"The pull request body was not fetched, so its release-note wording was not independently
  checked."* **Covered.** The body is the run's to write and is rewritten at the finish; round 2
  is told the branch is one commit further on so it can read it.

### Tiers, after the fix

|---|---|---|---|
| unit | `where.sh test --type modules` | n/a | 5521 pass, 0 fail |
| data | `where.sh test --type data` | mysql, postgres, mssql, sqlite | 2208 pass, 0 fail |
| integration | not run | -- | not applicable, limiter disabled in the suite (#219) |

`check-anchors.sh`: PASS, all 20 rows resolve to exactly one line in the named function.

Tree at the fix, `tree-hash.sh --split`: production `4f7c5ee4b820`, unit-tests `cdde886b39d9`,
data-tests `6483c3303341`, integration-tests `029224d85959`, docs `acc66fc3990a`, all
`b570b0e6c828`. `data-tests`, `integration-tests` and `docs` are unmoved from what the review
read; `production` and `unit-tests` moved with `4f2d91e6` and are what round 2 is asked to read.

The four-engine data and integration result for `4f2d91e6` is the check suite's to report.

**Round 2 is warranted and funded:** round 1 changed production code, and the change it made is
a security fix on an unauthenticated keying path that no reviewer has yet read.


---

## Round 2, at the final gate

`pi openai-codex/gpt-5.6-sol`, effort `xhigh`, resumed session, request
`276-20260909T174738Z`. Verdict `findings`, two, both `significant` and both `confidence: high`.
Both confirmed and both fixed in `82d753b7`. Nothing refuted, nothing deferred.

### Finding 1, security: the oversized sentinel aliases identifiers the application accepts

**Claim.** Round 1's fix folded every identifier past 320 octets into one shared bucket, on the
premise that nothing longer could name an account. The premise is false, so a real account past
the threshold shares its limiter key with every unrelated oversized submission and five of them
spend its whole password-reset budget.

**Confirmed.** Verified from the code rather than from the reviewer's report:

- `validators/email_validator.go:27` `ValidateEmailAddress` is a shape regex plus two dot rules.
  There is no length check anywhere in it, so a 400-octet address passes.
- `accounthandlers/handler_account_register.go:65` normalizes with `TrimSpace`+`ToLower`, exactly
  as `accountRateLimitKey` does, validates at :94 and creates the user at :242 or the
  pre-registration at :189. No length check on either arm.
- `sqlitedb/schema.golden:332` declares `users.email` as `TEXT`.

**The reviewer also corrected round 1's own record, and it was right to.** Round 1's entry said
60 was rejected as a bound because "`users.email` is `TEXT` on sqlite and postgres". Postgres is
`character varying(64)` (`postgresdb/schema.golden:353`); so are mysql and mssql. Sqlite is the
one engine that persists an overlong address. That does not change the disposition -- one engine
is enough -- but the claim as written was wrong and is corrected here.

**Fixed, and not the way the recommendation asked.** The `forced_answer` was to make the premise
an enforced invariant: reject overlong identifiers on every creation, update, activation and seed
path, and detect and remediate existing sqlite rows before they can be keyed. That was not taken.
It is a behaviour change across the validators, the setup program, the seeder and probably a
migration, none of which section 2 scopes, and it would refuse account identifiers this server
accepts today. It also treats the aliasing as a fact to be defended rather than removed.

The defect is that the key stopped being injective. So the fix restores injectivity instead:
an identifier past the threshold is keyed by `"<sha256>" + hex(sha256(normalized))` rather than by
a shared sentinel. That keeps the property round 1 was actually buying -- the store retains a key
for two windows and reads it off a form with no body cap, so 10 MiB must not be retained -- while
no two identifiers ever share a bucket, at any length, on any engine, with no premise about what
can name an account. `maxAccountIdentifierLen` drops from a security boundary to a legibility
threshold: an address still appears verbatim in the warning line and the audit event, and only
something longer than any address arrives as hex.

Two smaller things fell out of it. The length is now judged after `ToLower` rather than before,
because the digest needs the normalized form anyway, and judging before would have let an overlong
account buy a bucket per spelling -- the #219 defect this function exists to prevent. And the
reviewer's point that the RFC 5321 arithmetic was incomplete (4.5.3.1.3 caps a complete path at
256, so the local-part and domain maxima cannot simply be added to establish a mailbox maximum) is
answered by the threshold no longer needing to be a maximum of anything; the comment no longer
claims it is.

**This does not reopen decision 5.** That decision is about the store's key representation --
exact strings rather than `hash/fnv` to a `uint64` -- and it stands: the store still holds exact
strings, and the reason given, that a 64-bit hash leaves a rare untestable case where an innocent
client is refused for a stranger's attempts, is precisely the argument against the sentinel. A
256-bit digest of an identifier no address can be does not reintroduce it.

**Evidence the fix is pinned**, `where.sh mutate`, `--type modules`:

| Mutation | Result |
|---|---|
| `sha256.Sum256([]byte(normalized))` to `sha256.Sum256(nil)`, restoring the shared bucket exactly | CAUGHT |
| `if len(normalized) > maxAccountIdentifierLen {` to `if false {`, so a 10 MiB value is retained whole | CAUGHT |
| `strings.ToLower(strings.TrimSpace(identifier))` to `strings.TrimSpace(identifier)` | CAUGHT |

`TestAccountRateLimitKey_BoundsTheIdentifier` now pins the digest's length and prefix, that one
octet past the threshold is digested, that two distinct oversized identifiers keep distinct
buckets, and that an oversized identifier normalizes before it is digested. At the HTTP seam,
`TestLimitForgotPwd_PerEmailAndPerIP` gained "an oversized address keeps its own bucket": six
distinct oversized addresses all reach the handler, one address repeated spends exactly its own
budget of five and is refused on the sixth, and a seventh oversized address is untouched by it.
Round 1's "distinct oversized addresses share one bucket" case asserted the defect, so it is gone.

### Finding 2, quality: the Gate atomicity test is probabilistic

**Claim.** Round 1's `TestGate_ConcurrentFirstReportsExactlyOnce` does not deterministically catch
the mutation it was written for. The reviewer ran the exact mutant -- `First`'s unlock moved above
its clock read -- 100 times and it passed 58.

**Confirmed, and measured here.** Six independent `where.sh mutate` runs of that mutant against
that test: CAUGHT, CAUGHT, SURVIVED, CAUGHT, SURVIVED, CAUGHT. Four of six. The reviewer's finding
stands and round 1's `MUTATION CAUGHT` was a single sample of a coin.

**Why it is probabilistic.** The case asserts an outcome -- exactly one of 32 callers is told it
is first -- and an outcome assertion needs the scheduler to actually interleave two callers inside
the unlocked window. Widening that window with the fake clock makes the interleaving likely, not
certain.

**Fixed** by adding `TestGate_FirstHoldsItsLockAcrossTheWholeReadAndRecord`, which pins the cause
instead of the outcome and takes the scheduler out of it. `First` calls `now()` after taking the
lock, so parking the first caller inside `now()` means the mutex must be held for as long as it is
parked; the test then calls `TryLock` from outside and requires it to fail. Against the same
mutant, six of six runs CAUGHT. The outcome case is kept beside it, with its comment rewritten to
say what it is -- a probabilistic witness, four in six -- and to point at the deterministic one.

### `unchecked`, all four answered

- *"The race detector could not run: `CGO_ENABLED=0` and no C toolchain."* **Dismissed**, for the
  reason round 1 gave: it is a standing property of the container and of CI, which has no race
  job, rather than a gap at this gate. Go's runtime detects concurrent map access without cgo,
  which is how round 1's mutant aborted.
- *"No browser or full integration flow was run; the integration harness disables rate limiting."*
  **Dismissed.** Section 5 settles this: the suite runs with the limiter disabled (#219), so
  nothing here is reachable through an endpoint. The changed behaviour is pinned at the middleware
  HTTP seam instead, which is the seam section 5 names.
- *"No live sqlite self-registration was run."* **Covered**, by reading the three links in the
  chain rather than by running it: the validator has no length check, the handler creates on the
  validated value with none of its own, and the sqlite catalog stores `TEXT`. The fix makes the
  question moot in any case -- it no longer matters how long an accepted identifier is.
- *"The PR body was not fetched, so release-note wording was not independently checked."*
  **Covered.** The body is updated in this pass, with this round's account, so round 3 can read it.

### Tiers, after the fix

|---|---|---|---|
| unit | `where.sh test --type modules` | n/a | 5522 pass, 0 fail |
| data | `where.sh test --type data` | mysql 553, postgres 556, mssql 568, sqlite 531 | 2208 pass, 0 fail |
| integration | not run | -- | not applicable, limiter disabled in the suite (#219) |

`check-anchors.sh`: PASS, all 20 rows resolve to exactly one line in the named function.

Tree before the fix, and the value every mutation above was run against: production
`4f7c5ee4b820`, matching what the round 2 request recorded, so nothing moved under the review.
Tree at `82d753b7`: production `152344aa2c04`, unit-tests `b9f4c657c4db`, data-tests
`6483c3303341`, integration-tests `029224d85959`, docs `acc66fc3990a`, all `ca79d2b7f426`.
`data-tests`, `integration-tests` and `docs` are unmoved; `production` and `unit-tests` are what
round 3 is asked to read.

The four-engine data and integration result for `82d753b7` is the check suite's to report.

**Round 3 is warranted and funded:** round 2 changed production code, and what it changed is a
security fix on the same unauthenticated keying path, taking a different shape from the one the
reviewer recommended. No reviewer has read it.


## Round 3, at the final gate

`pi openai-codex/gpt-5.6-sol`, effort `xhigh`, resumed session, request
`276-20260909T183630Z`. Verdict `findings`, two: one `significant` on quality and one `blocking`
on conformance, both `confidence: high`.

**The verdict is salvaged, and that governs how much it settles.** The reviewer was cut off before
it reported, so `.review/response.md` does not exist, `summary.ran` is empty, all three axes read
`skipped`, and the finding list was recovered from `findings.jsonl` as the reviewer appended to it.
Its own note says to treat the list as a floor rather than a last word. So this round bought two
findings and **no reading**: nothing here says the axes were covered.

Finding 1 confirmed and fixed. Finding 2 confirmed in its concrete half and fixed, refuted in its
framing half, with the residual recorded as a deferred decision below.

### Finding 1, quality: the deterministic Gate test pins the entrance, not the section

**Claim.** `TestGate_FirstHoldsItsLockAcrossTheWholeReadAndRecord`, added by round 2, proves
`Gate.mu` is held only while `now()` is blocked, not across the store read and the record its name
claims. Moving the unlock below `now()`, or below `store.get` and above `store.add`, left both Gate
concurrency cases green in 100 of 100 runs each.

**Confirmed, from the code.** `gate.go:First` calls `g.now()` one statement after taking the lock,
and the test parked its caller inside that call. `TryLock` therefore reported on the entrance to the
critical section and on nothing after it: any unlock below `now()` has not yet executed while the
caller is parked there, so the mutex is held and the assertion passes. Both truncations the reviewer
names are real defects -- two callers read a count of zero and both are told they are first, which
turns the audit guarantee into "at least one event per key per window" -- and neither was
observable.

**This family was called closed one round ago, by this run.** Round 2's entry above says the new
test "pins the cause instead of the outcome and takes the scheduler out of it", on the evidence of
six of six CAUGHT against one mutant. That evidence was true and the conclusion drawn from it was
too wide: one mutant at one position is not a critical section. The reviewer's `why_exhaustive` is
the correction -- three mutations at every meaningful boundary of a compound operation, rather than
one -- and it is right.

**Fixed** by moving the park to the far end of the section instead of adding another test. The
store read and the record are reachable only through the store, so `Gate.store` becomes an
unexported `gateStore` interface over `memStore`'s existing unexported `get` and `add`; the test
substitutes a `parkingStore` that suspends its first caller inside `add`, which is the last thing
`First` does under the lock. An unlock anywhere above it is then visible to `TryLock`, and the
test's name is true for the first time. Production is unaffected: `Gate()` is the only constructor
and it builds a `memStore`, and the interface is unexported, so nothing outside the package can
substitute.

**Evidence, one mutation per family member**, `where.sh mutate --type modules --run
TestGate_FirstHoldsItsLockAcrossTheWholeReadAndRecord`. Each inserts `g.mu.Unlock()` plus a
`defer g.mu.Lock()` at one boundary, which keeps the function balanced while releasing the lock for
the rest of the section:

| Mutation, in `src/core/ratelimit/gate.go` | Round 2's test | This round's |
|---|---|---|
| unlock above `now := g.now()` | CAUGHT (6 of 6, round 2) | CAUGHT |
| unlock below `now := g.now()`, above the window math and the read | SURVIVED (100 of 100) | CAUGHT |
| unlock below `curr, _ := g.store.get(...)`, above the record | SURVIVED (100 of 100) | CAUGHT |

The two SURVIVED columns are the reviewer's measurement, not re-run here; the CAUGHT column is this
session's, one run each. `TestGate_ConcurrentFirstReportsExactlyOnce` is kept beside it as the
probabilistic witness it already admits to being, with its comment corrected to say it misses the
two later truncations entirely rather than merely being flaky against the first.

### Finding 2, conformance: the digest branch is not injective, and it aliases the exact branch

**Claim.** `accountRateLimitKey`'s overlong branch hashes with SHA-256, contrary to decision 5's
exact string keys and its rejection of a stdlib hash; hex-encoding a digest neither preserves the
identifier nor removes collisions, so round 2's claim that both branches are injective is false.
Concretely: submitting the 72-octet literal `<sha256>` plus the digest lands in the same bucket as
the overlong identifier it names, with no SHA-256 collision required.

**The concrete alias is confirmed and fixed.** It is exactly constructible: a digest key is eight
octets of prefix and sixty-four hex characters, seventy-two in all, far inside the 320-octet bound,
so before this round it took the exact branch and keyed as itself. Round 2's comment argued the
alias "buys nothing, because producing a given account's digest key means knowing that account's
identifier". That reasoning is wrong, and it is worth saying why rather than just deleting it:
`reportTrip` writes the key to the warning line and to the audit event, so a reader of either
learns a digest key **without** learning the identifier behind it, and could then spend that
account's budget by submitting the key straight back.

Fixed by reserving the prefix. `accountRateLimitKey` now digests any submission whose normalized
form starts with `oversizedAccountKeyPrefix`, however short, so the exact branch can never emit a
key carrying the prefix and the two branches are disjoint. One condition, no interface change,
nothing at runtime beyond a `strings.HasPrefix` on a short string: it only narrows what is keyed
exactly, which is the remedy `decisions.md` section 1 says not to stop a run over.

**The false claim is confirmed and corrected.** A bounded key over an unbounded set of identifiers
cannot be injective -- pigeonhole, and round 1's confirmed finding is what makes the key bounded --
so round 2's "both branches are injective" was not merely imprecise, it was unachievable. The
comments now say what actually holds: the exact branch is injective, the branches are disjoint, and
two accounts can be put in one bucket only by producing a SHA-256 collision. `maxAccountIdentifierLen`
and `oversizedAccountKeyPrefix` carry the corrected wording.

**The framing is refuted, with the demonstration.**

- *"Contrary to decision 5."* Decision 5 decides the **store's** key representation:
  `map[string]int` holding the string itself rather than `hash/fnv` to a `uint64`. That is intact
  and untouched. `src/core/ratelimit/store.go` still declares `curr, prev map[string]*int` and
  `add` still stores `strings.Clone(key)`; no key is hashed anywhere inside `ratelimit`.
  `accountRateLimitKey` lives in `core/middleware` and derives what the caller hands the limiter,
  which decision 5 does not address, and which only became a question when round 1 established
  that the derived key must be bounded.
- *"Injectivity is the property to restore."* It cannot be restored, per the pigeonhole argument
  above, so the real question is which non-injective bounded map. Every alternative is closed by a
  demonstration rather than by preference: the unbounded exact key is round 1's confirmed security
  finding; the shared sentinel is round 2's confirmed security finding; a 64-bit hash collides
  constructibly, which is decision 5's own stated objection and the sentinel again in a different
  shape; and enforcing a length invariant across every creation, update and seed path is this
  reviewer's own round 2 recommendation, which it accepted as out of section 2's frozen scope. The
  256-bit digest with disjoint namespaces is what is left standing.

The residual that remains -- a SHA-256 collision would share a bucket -- is real, is not closed by
this change, and is recorded as a deferred decision below with its three numbers.

**Evidence the fix is pinned**, `where.sh mutate --type modules --run
TestAccountRateLimitKey_BoundsTheIdentifier`:

| Mutation, in `src/core/middleware/middleware_ratelimiter.go` | Result |
|---|---|
| `strings.HasPrefix(normalized, oversizedAccountKeyPrefix)` to `false`, restoring the alias exactly | CAUGHT |

Two subtests were added: "a submission spelled as a digest key cannot reach a digested bucket",
which digests a real overlong identifier, checks the resulting key is short enough to be submitted
back, and requires that submitting it back does not key as itself; and "the exact branch never
emits a key carrying the digest prefix", over the bare prefix, a prefixed address, and the prefix
in upper case with surrounding whitespace, which also pins that normalization happens before the
prefix is tested.

### `unchecked`, the one entry

- *"the review was cut off before it reported, so what it covered is not recorded."* **Covered by
  asking again**, and that is the whole reason round 4 exists rather than the finish. Nothing in
  this verdict claims an axis was read: `axis_coverage` is `skipped` three times and `ran` is
  empty. Per `reviewer.md` section 4 a partial review is a reason to extend the scope and ask
  again rather than to advance, and this is as partial as a verdict gets. What this session covered
  itself is below -- three tiers, four mutations, the anchors -- and none of that is a substitute
  for an independent reading of the diff.

### Tiers, after the fix

|---|---|---|---|
| unit | `where.sh test --type modules` | n/a | 5524 pass, 0 fail |
| data | `where.sh test --type data` | mysql, postgres, mssql, sqlite | 2208 pass, 0 fail |
| integration | `where.sh test --type integration --db sqlite` | sqlite | 1371 pass, 0 fail |

Integration was run this round although round 2 recorded it as not applicable, and round 2's reason
still stands: the suite runs with the limiter disabled (#219 section 5), so nothing changed here is
reachable through an endpoint. It is evidence that the change breaks nothing else, not evidence
about the limiter. The limiter's own behaviour is pinned at the middleware HTTP seam, which is the
seam section 5 names.

`check-anchors.sh`: PASS, all 20 rows resolve to exactly one line in the named function.

Tree the review read: production `152344aa2c04`, matching what the round 3 request recorded, so
nothing moved under the review. Tree after this round's fixes, and the value all four mutations
above were run against: production `f3f5dac89a99`, unit-tests `4ac88594cb99`, data-tests
`6483c3303341`, integration-tests `029224d85959`, docs `acc66fc3990a`, all `1c7aae87f55f`.
`tree-hash.sh --split` was re-read after the last mutation and `git status` shows only the four
intended files modified, so every mutation is proven gone rather than assumed gone. `data-tests`,
`integration-tests` and `docs` are unmoved from round 2; `production` and `unit-tests` are what
round 4 is asked to read.

The four-engine data and integration result for this round's commit is the check suite's to report.

**Round 4 is warranted and funded on both counts:** round 3 changed production code, in
`ratelimit/gate.go` and `middleware/middleware_ratelimiter.go`, and the round that prompted it
never reported a reading of any axis. It is also the last round `max_rounds` allows.


---

## Round 4, at the final gate

`openai-codex/gpt-5.6-sol` at `xhigh`, resumed session, request `276-20260909T191618Z`. Verdict
`findings`, one finding, and the first round of the four to report coverage on all three axes:
rounds 1 to 3 each recorded partial or salvaged coverage, so this is the reading the code was owed.
Round 4 of 4, the last funded round.

The reviewer also conceded round 3's contested half in writing: "The contested store-key finding is
conceded after checking the package boundary." That closes the disagreement about decision 5 the way
the round-3 entry argued it, on the boundary rather than on assertion, and it needed no further
round.

### Finding 1, conformance, minor. Confirmed and fixed in `29378118`

**The claim.** The rationale on `accountRateLimitKey` justified the digest's residual by asserting
that SHA-256's collision resistance "is already what the token endpoint rests on for PKCE S256,
where breaking it is an authentication bypass rather than a shared rate-limit bucket". PKCE S256
does not rest on collision resistance, so the comment leaned on a false protocol analogy.

**Verified rather than taken.** `GeneratePKCECodeChallenge` at `src/core/oauth/oauth.go:9` hashes a
verifier and the token endpoint compares the result with the challenge stored against the
authorization code. RFC 7636 section 4.6, fetched from the publisher rather than recalled, gives the
equation the server evaluates:

> If the "code_challenge_method" from Section 4.3 was "S256", the received "code_verifier" is
> hashed by SHA-256, base64url-encoded, and then compared to the "code_challenge", i.e.:
> BASE64URL-ENCODE(SHA256(ASCII(code_verifier))) == code_challenge

The challenge on the right is "previously associated" with the code, so satisfying that equation
with anything but the original verifier is a second preimage of a fixed target, around 2^256. A
generic collision between two values an attacker picks freely does not reach it, and is worthless to
whoever constructed the pair, since they already hold a verifier that matches. The finding stands.

**The fix.** The analogy is gone and the residual is stated on its own terms, with the two costs
kept apart: making some two identifiers share a bucket is a generic SHA-256 collision at around
2^128, while making a submission land in one chosen account's bucket -- the case that would matter,
since a colliding pair of the attacker's own identifiers spends only their own budget -- is a second
preimage at around 2^256. The width argument the paragraph exists for survives intact: a 64-bit hash
collides at around 2^32, constructibly, which is the retired library's shared-bucket defect in a
different shape.

**Family, resolved rather than the instance.** The family is a security rationale conflating
collision resistance with preimage or second-preimage resistance. `grep -rn -i "preimage|collision
resist"` over `src` now returns one line, this comment's; `grep -rn "collision"` over non-test Go
returns twenty-one, and the only two making a cryptographic-strength argument are
`handler_account_activate.go:121` and the store's "no two clients can share a bucket by collision"
at `ratelimit/store.go:100`. Neither conflates the properties: the activation comment describes an
exact comparison left load-bearing behind a hash-index lookup, where a spurious candidate really
does require a digest collision, and the store line is about exact string keys, where no collision
exists at all. The reviewer's `why_exhaustive` checked source comments only, so the two records that
outlive `.review/` were checked here as well: the live PR body carries no such claim, and
`closing.md` does, which the next paragraph settles.

### The same conflation in the record, corrected

Section 9's open deferred decision -- whether an overlong identifier should be keyed by a SHA-256
digest -- argues its third number, the rate the underlying primitive fails anyway, exactly the way
the comment did: "This server already treats SHA-256 as load-bearing for authentication ... A
SHA-256 collision there is an authentication bypass; here it is two accounts sharing a rate-limit
bucket", closing with "anyone who can do that has the token endpoint". That is wrong for the reason
above, and it is the paragraph the user reads to settle the decision, so it is corrected here rather
than left standing. The log is append-only, so this entry is the correction.

Read the third number as: SHA-256 is load-bearing for authentication in this server, but on preimage
resistance against a stored challenge, not on collision resistance, so the token endpoint does not
fall at the same moment this bucket does. The comparison to make instead is with the residual's own
numbers, and they do not need the analogy. Nothing steerable is left after round 3 closed the
cross-branch alias; what remains costs around 2^128 for any shared bucket at all and around 2^256 to
choose whose, against a 64-bit hash's 2^32, which is a laptop.

**The assumption does not change.** Keeping the digest is still what is committed and still what the
run proceeded on: the correction removes an argument that was doing no work, since it compared the
residual to a different property, and leaves the two that were: 2^128 is out of reach, and section
9's cost-to-reverse is unchanged -- every alternative reopens round 1's or round 2's confirmed
finding, or reaches outside section 2 as frozen. The decision stays `Open` and stays the user's.

### `unchecked`, both answered

- *"The PR body was not fetched ... so release-note wording was not independently verified."*
  **Covered.** The live body was fetched this session with `gh pr view 314 --json body` and read for
  the family above; it does not repeat the PKCE claim, and its round-3 account matches what shipped.
  It is a record rather than a deliverable, and the finish reads it once more before the PR goes
  ready.
- *"No browser or local full integration flow was run because rate limiting is disabled in that
  harness."* **Dismissed.** That is by design and predates this change: #219 section 5 disables the
  limiter in the integration suite, which is why section 5 of this agreement puts all three seams at
  the unit tier and says nothing here is observable in integration. A browser pass would exercise the
  disabled path.

### Tiers, after the fix

`where.sh test --type modules`, all three modules, green, about 30s, including the repository-wide
gofmt guard. The change is a comment, so it owes no mutation: no test was added or altered and no
assertion depends on it. Data and integration are unaffected for the same reason.

```
production 649b800a6bc8   unit-tests 4ac88594cb99   data-tests 6483c3303341
integration-tests 029224d85959   docs acc66fc3990a   all fa4191f7f43a
```

`unit-tests` is unchanged from the tree round 4 reviewed, `4ac88594cb99`, so every mutation this run
recorded as CAUGHT still holds without a re-run: the reviewer's ten-of-ten Gate mutants and its
`accountRateLimitKey` mutants were run against exactly this test tree. `production` moved by the
comment alone, `29378118`, pushed to `issue-276-ratelimit`.

### Where the review ends

Nothing is left to raise: the one finding is fixed, its family is closed in the code and in both
records that outlive `.review/`, and the contested round-3 half was conceded by the reviewer. Round
4 was the last funded round, so the final review closes here and `.review/final-review.md` is
deleted, which lets the finish run.

PR #315, opened 2026-09-09
refactor(core): parse the dir + A256GCM id_token_hint JWE with the stdlib, retiring go-jose

### LINKED ISSUE #277: go-jose parses one scheme for one function: dir + A256GCM, pinned to a single-element allowlist
`go-jose/go-jose/v4` is imported by one non-test file, `core/encryption/jwe.go`, for one function on one scheme.

## What we use

```go
var (
    idTokenHintKeyAlgorithms     = []jose.KeyAlgorithm{jose.DIRECT}
    idTokenHintContentEncryption = []jose.ContentEncryption{jose.A256GCM}
)

object, err := jose.ParseEncryptedCompact(compactJWE, idTokenHintKeyAlgorithms, idTokenHintContentEncryption)
plaintext, err := object.Decrypt(DeriveIDTokenHintKey(clientSecret))
```

`jose.NewEncrypter` / `jose.Recipient` appear only in `jwe_test.go` and `handler_account_logout_test.go`, constructing fixtures.

We do not parse JWS with it (that is `golang-jwt/v5`), we do not use JWK, and we accept exactly **one** key-management algorithm and **one** content encryption.

## Why hand-rolling is defensible here, specifically

The standing advice is never to write your own JOSE, and the reason is algorithm agility: the parser has to decide, from attacker-supplied header fields, which of a dozen algorithms to run, and that decision is where the classic attacks live (`alg: none`, RSA-to-HMAC confusion, key-type substitution).

**We have already deleted that decision.** The allowlist has one element on each axis. What remains after pinning is:

1. Split the compact serialization on `.` into 5 parts. Reject anything that is not 5 parts.
2. Base64url-decode (raw, no padding) each part.
3. Parse the protected header JSON, assert `alg == "dir"` and `enc == "A256GCM"`, reject anything else.
4. For `dir`, part 2 (the encrypted key) **must be empty**. Reject if not.
5. `aes.NewCipher(key)` + `cipher.NewGCM`, then `Open(nil, iv, ciphertext||tag, aad)`.

Roughly 60 lines, auditable in a single sitting.

## The two details that must be exact

**AAD is the protected header as received, byte for byte.** Per RFC 7516 §5.1 step 14, the Additional Authenticated Data is the ASCII bytes of the *encoded* protected header, not a re-serialization of the parsed JSON. Decoding the header to a struct and re-encoding it will produce a different byte string and every decryption will fail. Keep the original substring.

**The GCM tag is part 5, appended to the ciphertext.** Go's `cipher.AEAD.Open` expects `ciphertext || tag`; JWE carries them separately. Concatenate before calling `Open`, and reject a tag that is not 16 bytes.

Also: the IV must be 12 bytes for A256GCM, the key 32 bytes (guaranteed by `DeriveIDTokenHintKey` returning a SHA-256 sum), and unknown critical header parameters (`crit`) should be rejected.

## Scope

- `core/encryption/jwe.go`: replace `ParseEncryptedCompact` + `Decrypt`, keep `DeriveIDTokenHintKey` and both exported signatures unchanged.
- `jwe_test.go` and `handler_account_logout_test.go` build JWEs with `jose.NewEncrypter`. Either keep go-jose as a test-only dependency for fixture construction, or write the ~30-line encrypt half too and test it round-trip. **Writing the encrypt half is better**: it removes the module entirely and gives the tests a way to construct deliberately malformed JWEs (wrong `alg`, non-empty encrypted key, tampered AAD) that go-jose would refuse to produce.

Negative tests for each rejection above are the deliverable, not an afterthought.

---

### Dependency reduction

leodip:
Closes #277.

## What this is for

The logout endpoint accepts an encrypted `id_token_hint`, a JWE in one fixed scheme: `alg` `dir`,
`enc` `A256GCM`, key SHA-256 of the client secret. `core/encryption` decrypted it through
`go-jose/go-jose/v4`, a general JOSE library, of which the two calls here used one parser mode and
one cipher. This change replaces those two calls with a stdlib parser for exactly that scheme, adds
the matching encryptor as the reference implementation of what the docs describe, and retires
go-jose from all three `go.mod` and `go.sum` files.

## The rule it establishes

The parser accepts exactly one shape: five canonical base64url segments whose protected header is a
JSON object naming `dir` and `A256GCM` and nothing the recipient does not understand, and every
departure from that shape is refused before the cipher runs, each with its own error in the server
log. Nothing in the header is dispatched on; the algorithms are asserted. The key derivation and
the exported signatures are unchanged, so every hint a client following the docs produces still
decrypts to the same plaintext.

Against the library it replaces, the parser is stricter in four places: a non-empty encrypted key
with `dir`, a `zip` parameter, line breaks inside or after the compact form, and invalid UTF-8 in
the protected header. Each is a spec requirement the library did not enforce, and each is pinned by
a test that would pass under the library. A repeated header name is refused too, which the library
also refused; there the strictness is against `encoding/json`, which silently keeps the last copy.

(The sealed agreement listed the repeated header name in place of the invalid UTF-8 as the fourth
tightening. Stage 1's probe rows settled it the other way; `log/stage-1.md` records the correction.)

## Status

**Complete.** Both stages landed, the whole change was reviewed twice over the branch, the check
suite is green on the final commit, and this pull request is out of draft and ready for review.

go-jose is out of the repository: `grep -c go-jose` over all six `go.mod` and `go.sum` files returns
0, and no Go file imports it. The exported signatures and the key derivation are unchanged, so every
hint a client following the docs produces still decrypts to the same plaintext.

### Stage 1: the stdlib parser and encryptor, the refusal table, go-jose out of `src/core`

Commit `f9834cb8`.

`DecryptIDTokenHintJWE` keeps its signature and is now a stdlib parser: seven gates in order, each
with its own message so the server log tells a malformed hint from a wrong key. Segment count;
canonical base64url per segment, enforced by re-encoding and comparing because Go's decoder skips
`
` and `
` anywhere in its input; the protected header valid UTF-8, a JSON object, with no
trailing data and no repeated top-level name; `alg` and `enc` compared exactly, the header read into
a map rather than a struct so that `ALG` stays an unknown parameter; `zip` and `crit` refused;
encrypted key empty, iv 12 bytes, tag 16, ciphertext non-empty; then one AES-256-GCM `Open` with the
protected header segment, exactly as received, as the AAD.

The iv gate is load-bearing rather than cosmetic: `crypto/cipher`'s GCM **panics** on a nonce that is
not its nonce size, and the iv arrives on the wire.

`EncryptIDTokenHintJWE` is new and exported, the reference encryptor for the scheme
`integration/endpoints.mdx` documents. Nothing in the binaries calls it; it is what lets tests in
either module build a fixture without a JOSE library, and stage 2's handler and integration fixtures
use it.

The encryption suite is rebuilt around a low-level fixture builder that writes any header and any
segment length, so each refused row differs from the accepted baseline in exactly one thing: 12
accepted rows, 40 refused rows each asserting the message of the gate meant to reject it, four
tamper rows including a substituted-but-valid header that only the AAD binding can refuse, three
allowlist subtests, and the round trip through the new encryptor.

**Tiers.** Unit green on all three modules, 61 `ok`, 1509 pass, 0 fail. Data and integration not
applicable: no query, schema, handler or route moved. Seven mutations run against the new gates with
`where.sh mutate`; all seven caught.

**Worth knowing.** `go mod tidy` dropped go-jose from `src/core` only. `src/authserver` still
requires it for its own logout fixture and `src/adminconsole` still carries the indirect line; both
go in stage 2. Full trace, including four recorded deviations from the plan, in
`log/stage-1.md`.

### Stage 2: the handler fixtures, the integration pair, go-jose out of `src/authserver` and `src/adminconsole`

Commit `d1296bba`.

The handler tests built their JWE fixtures with go-jose, which was the last thing in either module
requiring it. They now encrypt through `encryption.EncryptIDTokenHintJWE`, so the fixture and the
parser it feeds cannot drift; a second encryptor in the handlers package could, and a fixture that
drifts turns a real refusal into a passing test. The classify table builds its fixture before the
table rather than inside the literal, because the encryptor returns an error and a table literal has
nowhere to check it.

No integration test sent an encrypted hint before this, so the whole JWE path was exercised at the
unit tier only, against a mocked database that hands `decryptIDTokenHint` a client row the test
built. That proves the handler consults the parser and nothing about where the secret came from. Two
tests beside the plain-hint one close it: the ID token encrypted under the client's real secret tears
the session down and renders the signed-out page, and the same token under a foreign secret earns the
consent page and no redirect, however good the `client_id` beside it looks. Only the key differs
between them, so nothing but the GCM tag check can be refusing the second.

`go mod tidy` in `src/authserver` and `src/adminconsole` then drops the module.

Tiers: unit on all three modules (5579 pass, 0 fail) and integration on sqlite (1373 pass, 0 fail);
four engines by the check suite on the commit. Both new integration tests were mutation-checked and
both mutations were caught: neutralising the four-dot routing test fails the confirmed one, and
classifying a failed decryption as "no hint supplied" fails the refused one.

## Tiers

**The check suite on `a90c0e83`, the final commit: all thirteen checks pass.** That is the only
place four of these results exist.

|---|---|---|
| Unit, three modules | local, each stage and after each fix | 5584 pass, 0 fail, ~90s |
| Unit, `core` / `authserver` / `adminconsole` / `setup` | check suite | pass |
| Race detector | **check suite only** | `Unit / race` pass |
| Integration, sqlite | local, stage 2 and after round 1's fixes | 1373 pass, 0 fail, ~60s |
| Data + integration, sqlite / mysql / postgres / mssql | **check suite only** | all four pass |
| Lint, vulnerabilities, versions, build validation | **check suite only** | pass |
| Anchors | local, each stage and after each fix | PASS, all 19 rows resolve |
| Mutations | local | 7 in stage 1, 2 in stage 2, 3 across the review rounds; all caught |

Nothing in this change reaches the data layer — no query, schema, migration or route moved — which
is why no stage recorded a data-tier hash and why the four `Tests /` jobs are the whole of that
evidence.

## Deviations from the sealed plan

Seven, all contained, none touching §2's goal. Full text in `log/stage-1.md` and `log/stage-2.md`.

**Stage 1.**

1. **The `go-jose` recount for `jwe_test.go` returns 7, not 0.** All seven are prose in the "keep
   this: reverses the library" markers §5 asked for. The check that means what §4's fanout table
   meant is the import path, `grep -c 'go-jose/go-jose'`, which returns 0 everywhere.
2. **The padding row is refused one gate earlier than predicted.** `=` padding fails
   `base64.RawURLEncoding.DecodeString` outright, so it never reaches the canonicality compare; that
   compare is owned by the three whitespace rows instead. Two gates, two messages, each row asserts
   its own.
3. **The tampered-header row substitutes a valid header rather than flipping a byte.** A flipped
   byte breaks the JSON and would be refused at the header gate, proving nothing about
   authentication. The row swaps in `{"alg":"dir","enc":"A256GCM","cty":"JWS"}`, which passes every
   gate above the cipher, so only the AAD binding can refuse it.
4. **Two rows the plan called reversals of go-jose are parity, and one it did not name is a
   reversal.** go-jose refuses `crit: []` and both duplicate-`alg` orders. The seven genuine
   reversals are: non-empty encrypted key, `zip: DEF` with deflated content, a line break inside a
   segment, a trailing line break in either spelling, and invalid UTF-8 in an unknown value and in
   an unknown name.

Two rows were added beyond the plan (`alg` present but not a string; `alg` and `enc` nested inside
unknown parameters, the only row that fails if the duplicate-name walk stops skipping by depth), and
one property the plan did not state was found: `crypto/cipher`'s GCM **panics** on a wrong-size
nonce, so the 12-byte iv gate stands between an attacker-chosen segment length and a panic in the
logout handler. Under that mutation the suite does not merely fail, it panics.

**Stage 2.**

5. **"The three `test/logout-decrypt` call sites"** are two in `TestDecryptIDTokenHint` and one in
   `TestClassifyIdTokenHint`. The count was right, the attribution was not; all three were rewritten.
6. **`gomod/core` was re-anchored by stage 1 rather than retired**, so this stage re-anchored the
   remaining two the same way. Retiring them would leave §0 with no citation for the files the goal
   is measured on.
7. **Two anchor rows added** for the integration pair. A test the goal names and §0 cannot cite is a
   test the next sweep cannot protect.

### Review

Reviewed by `openai-codex/gpt-5.6-sol` at effort `xhigh`, 3 rounds: one on the plan before any code
existed, two over the whole branch. `LEO_REVIEW_MODE=end` was set, so **no stage was reviewed on its
own** and the two final rounds are the only reading the code got.

|---|---|---|---|
| plan | 1 | 0 | applied to the plan before any code was written |
| final 1 | 2 | 1 | both fixed, `da3481a4` |
| final 2 | 1 | 0 | fixed, `a90c0e83` |

- **significant, conformance (plan round):** the planned header decode would not enforce JOSE's JSON
  rules — `encoding/json` struct matching is case-insensitive, so `ALG` would pass the `alg` gate,
  and the decoder walk accepts invalid UTF-8 by replacement. Applied to the plan before any code:
  the parser `utf8.Valid`s the header and unmarshals into `map[string]json.RawMessage`. Eight probe
  rows against go-jose settled which half is parity and which is a tightening.
- **blocking, security (final 1):** the refusals for an unsupported `alg`, an unsupported `enc` and a
  repeated member name interpolated the whole caller-controlled string with `%q`, and the handler
  logs that error unchanged, so the request logger's 4096-byte cap did not bound the record. An
  unauthenticated caller needs only a known `client_id`; the reviewer measured 200,744 log bytes
  from a 200,000-character `alg`. Fixed in `da3481a4`: `%.*q` at `maxLoggedHeaderValue = 128` runes
  at all three sites, with a regression test asserting the message length is *identical* at 1000 and
  200,000 characters rather than merely under the ceiling.
- **significant, quality (final 1):** the refusal table covered padding and line breaks but not
  non-zero unused trailing bits, the other non-canonical base64url form. Proved by a mutation that
  survived the whole suite. Fixed in `da3481a4`: one row built by an `aliasTrailingBits` helper that
  searches the alphabet for a final character decoding to the same bytes, so the alias is a property
  of the helper rather than an assertion beside it. The reviewer's mutation now fails that row alone.
- **significant, quality (final 2):** that regression test fed ASCII only, and `%q`'s precision
  counts runes of input rather than bytes of output, so the bound could move without the suite
  noticing — confirmed by reproduction, `maxLoggedHeaderValue` 128 → 1000 **survived**. Fixed in
  `a90c0e83`: three cases over two alphabets, `ascii` and `astral` (`\U0010FFFF`, the longest
  rendering `%q` can produce for one rune of the budget). The same mutation is now **caught**, and
  the failures are exactly the three astral subtests.

Full account with all the evidence, including the mutation commands and the tree hashes,
in [this comment](https://github.com/leodip/goiabada/pull/315#issuecomment-5610768960).

## The run's three closing comments

- [Deferred decisions](https://github.com/leodip/goiabada/pull/315#issuecomment-5610761768) — none;
  all eight of §3's decisions were settled at spec time, and two corrections to the sealed record
  are named there.
- [Follow-ups, fold-ins and ceilings](https://github.com/leodip/goiabada/pull/315#issuecomment-5610764667)
  — none, none and none, with the `grep` that shows the last.
- [The code review](https://github.com/leodip/goiabada/pull/315#issuecomment-5610768960) — all three
  rounds and all four findings, with the evidence.

## For the release notes

A client that compressed its encrypted `id_token_hint` will stop working. The parser never inflates,
so a `zip` header parameter is refused rather than processed; go-jose inflated it. The documented
scheme has no compression row and none of the client samples sets one, so no client following the
docs is affected.

Three narrower refusals land with it, all of inputs the docs never described: a non-empty JWE
Encrypted Key segment under `alg` `dir`, a stray line break inside or after the compact form, and
invalid UTF-8 in the protected header.


leodip:
## Deferred decisions

**None.** Nothing was postponed on an assumption.

All eight decisions in §3 of the agreement were raised and settled during the interview, before the
seal: where the encrypt half lives (1), the segment size rules (2), `zip` (3), header parameters
beyond `alg` and `enc` (4), base64url strictness (5), duplicate header names (6), the missing
integration coverage (7), and documentation (8). Decision 6 was decided by the user against the
recommendation; the rest went as recommended.

Neither stage reopened one, and neither raised a new one. Nothing was escalated to this thread, so
there is no comment above this one waiting on an answer, and no part of the change rests on a guess
the run made for you.

Two corrections to the sealed record were made along the way. Both are recorded where a reader of
the diff will meet them, and neither is a decision left open:

- **The fourth tightening against go-jose is invalid UTF-8 in the protected header, not a repeated
  header name.** Stage 1's probe rows established that go-jose refuses a repeated `alg` in both
  orders and refuses `crit: []`, so those are parity rather than reversals; the repeated-name row is
  still load-bearing, but for reversing `encoding/json`'s silent last-wins, which is what decision 6
  chose against. `log/stage-1.md` deviation 4 has the full seven-row list.
- **The `go-jose` recount in §4's fanout table is wrong for `jwe_test.go`.** It returns 7, and all
  seven are prose in the "keep this: reverses the library" markers §5 asked for. The check that
  means what the table meant is the import path, `grep -c 'go-jose/go-jose'`, which returns 0
  everywhere.

leodip:

### Follow-ups to file

**None.** The run found no verified defect or gap outside this change's scope, so nothing went
through the fold-in test and nothing is drafted here. No `gh issue create` commands follow, and no
duplicate search was owed because there was no draft to check for duplicates. If that reads as
suspiciously tidy: the surface this change touches is one file of parser code and its tests, the
whole of it was rewritten rather than extended, and the two review rounds that read it raised four
findings all of which belonged to this change and were fixed in it.

**None during the run.** One thing was folded in *before* the seal and is therefore part of §2's
goal rather than an unrecorded extra, but it is worth naming here because a reader of the diff will
meet it:

- **Decision 7, the missing integration coverage.** No integration test sent an encrypted
  `id_token_hint` before this branch; the whole JWE path was exercised at the unit tier only,
  against a mocked database that hands the handler a client row the test built. That proves the
  handler consults the parser and nothing about where the secret came from. Two tests in
  `d1296bba` close it, `TestLogout_WithEncryptedIdTokenHint_LogsTheUserOut` and
  `TestLogout_EncryptedIdTokenHint_WrongSecret_AsksTheEndUserAndRefusesTheRedirect`, and only the
  key differs between them, so nothing but the GCM tag check can be refusing the second.

**None.**

```
(no matches)
```

The change ships no simplification with a known limit: the parser accepts one shape and refuses
everything else, and every refusal is a spec requirement rather than a bound somebody picked. The
one constant that looks like a ceiling, `maxLoggedHeaderValue = 128`, is not one — it is a security
bound with a test that fails if it moves, added for finding 3 in round 1 and pinned against escape
expansion in round 2.

leodip:
## The code review

`LEO_REVIEW_MODE=end` was set for this run, so **no stage was reviewed on its own and this is the
only reading the code got.** Three rounds: one on the plan before any code existed, two on the whole
branch. Four findings, all four confirmed, all four fixed. Nothing was refuted and nothing was
deferred.

The verdicts live in a gitignored directory that the next run overwrites, so this comment is where
they survive. Pasted whole from `docs/issue-277-hand-rolled-jwe/log/closing.md`, numbers kept.

| Round | Reviewer | Findings | Blocking | Outcome |
|---|---|---|---|---|
| plan | `openai-codex/gpt-5.6-sol` xhigh, fresh | 1 | 0 | applied to the plan before any code |
| final 1 | `pi openai-codex/gpt-5.6-sol` xhigh, fresh | 2 | 1 | both fixed, `da3481a4` |
| final 2 | `pi openai-codex/gpt-5.6-sol` xhigh, resumed | 1 | 0 | fixed, `a90c0e83` |

Round 2 changed no production code, so the round budget's extension condition was not met and the
gate closed there: rounds 1 and 2 are unconditional, and a third is funded only by a round that
moved what ships.

## Review account: plan review, round 1, 2026-09-09

Reviewer `openai-codex/gpt-5.6-sol`, effort xhigh, fresh session, request `277-20260909T220941Z`.
Verdict: findings, one. Unchecked: no implementation to execute yet; unrelated security surfaces
not re-audited; untouched parts of the logout suites not read. All three dismissed for the plan
gate with reasons in `log/plan.md`.

1. **The planned header decode does not enforce JOSE's JSON rules** (conformance, significant).
   `encoding/json` struct matching is case-insensitive, so `ALG` would pass the `alg` gate; the
   decoder walk accepts invalid UTF-8 by replacement. **Confirmed and applied to the plan**
   before any code: verified by eight new probe rows against go-jose (it refuses `ALG` as a missing
   `alg`, ignores `ZIP` and `CRIT`, and accepts invalid UTF-8, so that half is a tightening rather
   than parity). The parser will `utf8.Valid` the header and unmarshal into
   `map[string]json.RawMessage`; seven new test rows and two new mutations pin it. Lands in stage 1.


---

## Final review, round 1: request written, 2026-09-09

Request `277-20260909T232623Z`, `gate: final`, `status: awaiting-review`. Scope is the whole branch
against `origin/main`, three commits, `+788 -124` over ten files. `LEO_REVIEW_MODE=end` was set for
this run, so no stage was reviewed on its own and this is the only reading the code gets. CI is
green at HEAD `d1296bba`: all thirteen checks pass, including `Tests / mysql`, `Tests / postgres`,
`Tests / mssql` and `Unit / race`, which is the four-engine data and integration result no local
run produces.

`tree-hash.sh --verify` against the recorded stage claims:

```
current   production e25f15b0d6b8   unit-tests 888cd3c9e4fa   data-tests 6483c3303341
          integration-tests 986e740d9434   docs 2fbbd8f0cd6f   all c10438b4dcb1

         data-tests, integration-tests not recorded here, so those claims cannot be settled either way
stage 2  unit live           data ?              integration live    docs -
         data-tests not recorded here, so those claims cannot be settled either way

1 of 2 entries record a result against a tree that has since moved.
```

What that costs: stage 1 recorded its unit tier and its seven mutations against a production tree
stage 2 then moved, so those mutation results are the claims that would need re-running to believe.
Stage 2's unit and integration results are live and free. Neither stage recorded a data-tier hash,
which is expected: nothing in this change reaches the data layer, and CI's four `Tests /` jobs are
green at HEAD regardless.


## Final review, round 1: verdict ingested, 2026-09-09

Reviewer `pi openai-codex/gpt-5.6-sol`, effort `xhigh`, fresh session, request
`277-20260909T232623Z`, completed 2026-09-09T23:49:52Z. Verdict `findings`, two findings, all three
axes marked `reviewed`. `unchecked` was empty, and the response's "Not checked" reads "None within
the requested scope", so there is no coverage list to answer. `## Follow-ups` and `## Bookkeeping`
were both "None". Both findings confirmed, both fixed in `da3481a4`.

The reviewer independently re-established the parts of the record that had gone stale: it
re-confirmed go-jose absent from all three module graphs with `go mod tidy -diff`, re-ran the
interop check against go-jose v4.1.4 in both directions, and pushed 20,000 deterministic malformed
inputs through the parser without a panic. That covers stage 1's stale mutation claims from the
`--verify` block in the previous entry, which is the one thing that block said would cost a re-run
to believe.

### Finding 3, security, blocking. Confirmed and fixed.

**Claim:** the refusals for an unsupported `alg`, an unsupported `enc` and a repeated member name
interpolate the whole caller-controlled string with `%q`, and `decryptIDTokenHint` logs that error
unchanged, so the request logger's 4096-byte cap does not bound the resulting record. An
unauthenticated caller needs only a known `client_id`; header validation runs before the
segment-size and GCM gates, so no client secret is required, and the caller receives the ordinary
rejected-hint answer.

**Verified, not taken on trust.** The premise is a live invariant of this repository rather than a
general principle: `core/middleware/middleware_request_logger.go` defines `maxLoggedTarget = 4096`
and its comment states the reason in the same terms, "Without this bound an unauthenticated caller
writes a megabyte to the log per request (#159)". `handler_account_logout.go:151` is
`slog.Error("logout: id_token_hint decrypt failed", "err", err)`, a second record that bound never
reaches, and the three sites at `jwe.go` lines 158, 161 and 300 were unbounded `%q` over strings
taken straight off the wire. The reviewer measured 200,744 log bytes from a 200,000-character `alg`.
Nothing here was refutable.

**Fixed:** all three now quote with a precision, `%.*q` at `maxLoggedHeaderValue = 128` runes. That
is a rune count which `%q` then escapes, so the worst case is 128 astral runes at `\U0010FFFF`, about
1.3 KB, inside the 4096 a whole request target is allowed. A real `alg` or `enc` is under ten
characters, so no legitimate input is ever clipped and every existing short-message assertion in the
refusal table still holds.

**Test:** `TestDecryptIDTokenHintJWE_DiagnosticsAreBounded`, three subtests. Each builds the header
at 1000 and at 200,000 characters and asserts two things: the error stays under the 4096 ceiling,
and the two lengths are *identical*. The second is the stronger half and is deliberate, because a
bound that merely happened to sit under the ceiling at these two sizes would not say the message has
stopped tracking the caller's input.

**Mutation:** reverting the `alg` site to the unbounded `errors.Errorf("... %q ...", alg)` --
`MUTATION CAUGHT`, the new test fails. The constant is documented at its definition with what breaks
if the precision is removed, and the parser comment names #159 rather than this agreement, since the
agreement directory does not survive a clone.

### Finding 2, quality, significant. Confirmed and fixed.

**Claim:** the refusal table covers padding and line breaks but not a base64url spelling with
non-zero unused trailing bits, the other non-canonical form `base64.RawURLEncoding` accepts. The
reviewer proved the gap by mutation: replacing the decode-and-re-encode comparison with a CR/LF-only
check left every existing `EncryptIDTokenHintJWE` and `DecryptIDTokenHintJWE` test passing.

**Verified.** The three canonicality rows in the table were all line-break cases, and the production
comment above the check expressly claims the trailing-bit form: "Go's decoder silently skips \r and
\n anywhere in its input and accepts non-zero trailing bits, so re-encoding and comparing is what
enforces the canonical form." So the code asserted a property no test pinned. Production was correct
throughout; this is a coverage finding and nothing shipped wrong.

**Fixed:** one row, "non-zero trailing bits in the tag segment", built by a new `aliasTrailingBits`
helper. The helper swaps the segment's final base64url character for another that decodes to the
same bytes -- a 16-byte tag spends 22 characters on 128 bits, so its last character carries 2
significant bits and 4 unused ones -- and it searches the alphabet for such a character rather than
computing one, which makes proving the alias decodes identically a property of the helper instead of
an assertion beside it. It calls `t.Fatalf` when no alias exists, so the row can never silently
degrade into a tampering case.

**Mutation:** the reviewer's own, reproduced. Replacing
`if base64.RawURLEncoding.EncodeToString(raw) != part {` with `if strings.ContainsAny(part, "\r\n")
{` -- `MUTATION CAUGHT`, and the failure list is exactly one subtest,
`TestDecryptIDTokenHintJWE_InvalidInput/non-zero_trailing_bits_in_the_tag_segment`. The mutation
that survived the suite before now fails on the new row alone, which is the precise thing the
finding asked for.

### Tiers re-run after the fixes

- `where.sh test --type modules`: green, 5584 pass, 0 fail, about 90s. Covers all three modules,
  `TestGoSourcesAreGofmted` included.
- `where.sh test --type integration --db sqlite`: green, 1373 pass, 0 fail, about 60s.
  `TestLogout_WithEncryptedIdTokenHint_LogsTheUserOut` and
  `TestLogout_EncryptedIdTokenHint_WrongSecret_AsksTheEndUserAndRefusesTheRedirect` both pass, so
  the changed parser is still exercised end to end.
- `check-anchors.sh`: PASS, all 19 rows resolve to exactly one line in the named function. The edits
  moved no cited code.

Tree after the fixes, and after both mutations were restored:

```
production b0f8dd622546   unit-tests 32634072da1b   data-tests 6483c3303341
integration-tests 986e740d9434   docs 2fbbd8f0cd6f   all a7295023b06c
```

`git status` showed only `src/core/encryption/jwe.go` and `src/core/encryption/jwe_test.go`
modified, and `data-tests`, `integration-tests` and `docs` are byte-identical to the values in the
previous entry, which is what says neither mutation leaked outside the file it targeted.

Committed as `da3481a4` and pushed. The four-engine data and integration result for it is the check
suite's to report.


---

## Final review, round 2: verdict ingested, 2026-09-10

Reviewer `pi openai-codex/gpt-5.6-sol`, effort `xhigh`, resumed session, request
`277-20260910T000210Z`, completed 2026-09-10T00:13:00Z. Verdict `findings`, one finding, all three
axes marked `reviewed`. `unchecked` was empty and the response's "Not checked" reads "None within
the requested scope", so there is no coverage list to answer. `## Follow-ups` and `## Bookkeeping`
were both "None". The one finding was confirmed and fixed in `a90c0e83`.

The reviewer re-established the round 1 repairs rather than re-running what was already reported: it
removed the precision from the `alg`, `enc` and duplicate-name errors independently and watched each
mutation fail only its own subtest at roughly 200 KB, mutated `aliasTrailingBits` to select
different decoded bytes and confirmed the new trailing-bits row cannot degrade into a tamper case,
and re-read every error construction in the parser to confirm only those three interpolate a string
the caller chose. It also read CI to a green conclusion at `da3481a4`.

### Finding 1, quality, significant. Confirmed and fixed.

**Claim:** `TestDecryptIDTokenHintJWE_DiagnosticsAreBounded`, the regression test round 1 added
with the bound, feeds `strings.Repeat("A", size)` and nothing else. `%q`'s precision counts runes
of *input*, not bytes of *output*, so an ASCII-only test never exercises the escape expansion that
makes 128 the right number. The bound is therefore free to move without the suite noticing.

**Verified by reproduction, not taken on trust.** The load-bearing fact is asserted in production at
`maxLoggedHeaderValue`'s definition -- "the worst case is 128 astral runes rendering as
\U0010FFFF: about 1.3 KB" -- and no test said it. Running the reviewer's mutation against the
committed tree, `maxLoggedHeaderValue = 128` to `= 1000` on `--type modules --run
TestDecryptIDTokenHintJWE_DiagnosticsAreBounded`: **`MUTATION SURVIVED`**. All three subtests passed
against a limit almost eight times too large, because at 1000 ASCII runes each error renders about
1040 bytes, under the ceiling and identical at both input sizes. So both committed assertions hold
while the bound they exist to pin has been broken. Production was correct throughout and nothing
shipped wrong; this is a coverage finding, and the reviewer's severity is right.

**Fixed:** the three cases now run over two alphabets rather than one, `ascii` (`"A"`) and `astral`
(`"\U0010FFFF"`), at both 1000 and 200,000 runes, keeping the existing pair of assertions per
alphabet: the error stays under the 4096 ceiling, and the two lengths are identical. Six subtests
where there were three. The astral rune is the worst case for a reason the comment beside it now
states: it is the longest rendering `%q` can produce for one rune of the budget, ten bytes, and it
is valid UTF-8, so it reaches the quoting rather than stopping at the parser's UTF-8 gate. Both
demonstrations the finding asked for are recorded here, the survival above and the catch below.

**Mutation, re-run against the fix:** the same `128` to `1000`, now **`MUTATION CAUGHT`**. The
failures are exactly the three astral subtests, at 10042, 10046 and 10047 bytes against a 4096
ceiling, reproducing the reviewer's three numbers precisely; the three ascii subtests still pass,
which is the gap the finding named, now visible in the failure list itself. Round 1's mutation for
this constant, reverting the `alg` site to unbounded `%q`, still fails the test as before.

### Tiers re-run after the fix

- `where.sh test --type modules`: green, 5584 pass, 0 fail, about 90s. All three modules,
  `TestGoSourcesAreGofmted` included.
- `TestDecryptIDTokenHintJWE_DiagnosticsAreBounded` on `--type core`: all six subtests pass, three
  ascii and three astral.
- `check-anchors.sh`: PASS, all 19 rows resolve to exactly one line in the named function.
- Integration was not re-run. The change is one test function in `core/encryption` and the
  `integration-tests` hash is unchanged from round 1's entry, so that tier's recorded result still
  binds.

Tree after the fix, and after the mutation was restored:

```
production b0f8dd622546   unit-tests 81c4d83ce3a0   data-tests 6483c3303341
integration-tests 986e740d9434   docs 2fbbd8f0cd6f   all 70998d40bd63
```

`production` is byte-identical to round 1's recorded value, which is what proves the mutation is
gone and that this round changed no shipped code. `data-tests`, `integration-tests` and `docs` are
unchanged too; `unit-tests` moved, and that single category is the whole of this round.

Committed as `a90c0e83` and pushed. The four-engine data and integration result for it is the check
suite's to report.

### Why the gate closes here

Round 2 changed no production code, so the round budget's extension condition is not met: rounds 1
and 2 are unconditional and a third is funded only by a round that moved what ships. Nothing is left
open -- one finding, confirmed and fixed with both demonstrations, an empty `unchecked`, no
follow-ups and no bookkeeping. The two rounds between them moved production once, on a real security
defect, and hardened the tests twice. `.review/final-review.md` is deleted and the finish runs next.

Issue #281, opened 2026-08-27
Replace mileusna/useragent: prefer Client Hints, keep the raw header, stop baking a parser's guess into the row


leodip:
`mileusna/useragent` is unwanted. This ticket is about what goes in its place, because the answer is not a drop-in swap.

## What it does today

`core/useragent/useragent.go`, three functions, **one caller**: `core/user/usersession_manager.go:89-91`.

```go
DeviceName: useragent.GetDeviceName(r),   // "Chrome 120 (iPhone)"  -> device_name TEXT NOT NULL
DeviceType: useragent.GetDeviceType(r),   // Mobile|Tablet|Desktop|Bot|unknown -> device_type
DeviceOS:   useragent.GetDeviceOS(r),     // "iOS 17.1"             -> device_os
```

Written **once**, at session creation. Never read back by any logic. Displayed concatenated in three templates:

```
{{.DeviceName}} {{.DeviceType}} {{.DeviceOS}}
```

`account_user_sessions.html`, `admin_users_sessions.html`, `admin_clients_usersessions.html`, plus the same string passed into the "end session" confirmation prompt.

**These fields are cosmetic.** Their entire job is helping someone recognise which row is the session they are sitting in, so they end the right one. Precision is not the requirement; recognisability is.

## Why there is no clean drop-in

User-Agent parsing is a **database, not an algorithm**, and browsers deliberately lie in that header. Chrome reports `Mozilla/5.0 ... AppleWebKit/537.36 (KHTML, like Gecko) Chrome/120 Safari/537.36`, naming three engines it is not. Safari on iPadOS reports itself as macOS. Any parser we write is a maintenance treadmill fed by other people's release schedules, and that is exactly the thing worth avoiding.

So the replacement should reduce how much we depend on parsing at all.

## Option A: User-Agent Client Hints, with a parsed fallback (recommended)

Chromium sends three **low-entropy** hints on every request, no opt-in required:

| Header | Example | Gives us |
|---|---|---|
| `Sec-CH-UA` | `"Chromium";v="120", "Google Chrome";v="120", "Not?A_Brand";v="24"` | Brand and major version, structured |
| `Sec-CH-UA-Mobile` | `?1` | `device_type` Mobile vs Desktop, exactly |
| `Sec-CH-UA-Platform` | `"macOS"` | `device_os`, no version |

[...]

PR #316, opened 2026-09-10
refactor(core): replace google/uuid with a stdlib uuidutil, and fail closed on CSPRNG failure

### LINKED ISSUE #211: GenerateSecurityRandomString swallows CSPRNG failure, so no caller can propagate or test it
stringutil.GenerateSecurityRandomString returns "" on CSPRNG failure instead of an error, so each caller needing a usable value repeats an `if id == ""` guard, and none of those guards can be exercised by a test without swapping crypto/rand.Reader globally or adding a test-only seam to production code. Two guards exist today, in SaveLinkMarker (src/authserver/internal/handlers/link_marker.go) and HandleAuthorizeGet (src/authserver/internal/handlers/handler_authorize.go), and both are uncovered. The remaining call sites do not guard at all.

Give the package an error-returning variant, or an unexported reader the tests can substitute within stringutil, so callers can propagate a real error and the failure path can be covered where it is written.

The seam is already present one level down: randomStringFromReader takes an io.Reader and is tested with a failing one. Only the exported wrapper, randomStringFromAlphabet, hides it, and its comment says so: "It preserves the package's historical contract of returning "" when the system CSPRNG is unavailable."

### LINKED ISSUE #278: google/uuid supplies random identifiers only, never UUID semantics

leodip:
Closes #278. Closes #211.

## What this is for

`github.com/google/uuid` was imported by 105 Go files and supplied one thing everywhere: a fresh
random identifier as a 36-character string. Nothing in the tree used UUID semantics beyond that,
except that `models.User.Subject` was typed `uuid.UUID`, which reached the column, the JSON body
and the JWT `sub` claim through `String()` and `MarshalText` as the same canonical string anyway.

This change retires the module for a stdlib-only `core/uuidutil` package with two functions,
`New() string` and `Parse(s string) (string, error)`, and turns `Subject` into a `string`. Every
identifier the server stores, issues or serves stays byte-identical: same 16 bytes from
`crypto/rand`, same version and variant nibbles, same 8-4-4-4-12 lowercase form.

## The rule it establishes

A failure of the system CSPRNG ends the process. It never produces a zero, partial or empty
identifier, and it never turns into a request-level error. `uuidutil.New` reads through
`crypto/rand.Read`, which on Go 1.24 and later crashes the program irrecoverably rather than
returning an error. `stringutil`'s generators, which used to answer the same failure with `""`
(#211), now go through the same call, and the two `== ""` guards written against that contract are
deleted as dead code.

`uuidutil.Parse` accepts only the canonical form. The braced, `urn:uuid:` and 32-hex spellings
that `google/uuid.Parse` accepted are refused, because nothing here produces or receives them.

## Status

**Complete.** All four stages have landed, the whole change has been reviewed over two rounds at the
final gate, and the check suite is green at `1a8c39c0` on all thirteen jobs. Decision 8 was answered
A by the owner on 2026-09-10: the module stays `// indirect` through the SQL Server and sqlite
drivers, so the goal is that no Goiabada file imports it and no module requires it directly. That
goal is met — `grep -rn 'github.com/google/uuid' src --include=*.go` reads 0 and all four `go.mod`
files list it `// indirect` — and so is the folded-in one, with `stringutil` now failing closed.

### Stage 1: `core/uuidutil`, and `fake.UUID` over it
Landed in `819a1737`. `src/core/uuidutil` is a new stdlib-only package: `New() string` draws 16 bytes from `crypto/rand.Read`, sets the version and variant nibbles per RFC 9562 sections 4.2 and 4.1, and formats them 8-4-4-4-12 in lowercase; `Parse(string) (string, error)` accepts only that form, in either case, and returns it lowercased, refusing the braced, `urn:uuid:` and 32-hex spellings with one of three named sentinels. `New` returns no error on purpose: `crypto/rand.Read` has not returned one since Go 1.24 and calls the runtime's fatal handler on a failed read, so a zero or partly random identifier cannot be produced. `fake.UUID` is now a wrapper over `uuidutil.New`, and `fake_test.go` checks through `uuidutil.Parse`, which retires `google/uuid` from `core/testutil/fake`.

Tiers: unit green on all three modules, 5796 pass 0 fail, gofmt guard included. Data and integration not applicable — no query, no migration and no endpoint is in the diff, and nothing outside the package's own tests calls it yet. Three mutations checked and all caught: neutralising either the version or the variant mask fails `TestNew_CanonicalV4`, and rewriting `New` to return `""` on a read error fails `TestNew_CrashesIrrecoverablyOnReaderFailure`, which re-executes the test binary as a child because the real failure is a runtime fatal no `recover` can catch.

Two things worth knowing. `New` uses `_, _ = rand.Read(...)` rather than the bare call in the design sketch, because `errcheck` is enabled for all four modules and would redden the Lint job. And a mutation in this repository has to be gofmt-clean: shortening a line inside an aligned comment block makes `TestGoSourcesAreGofmted` fail in another module first, which reports `MUTATION CAUGHT` without the test under examination ever running.

### Stage 2: `Subject` becomes a string, and the test tree stops importing `google/uuid`
Landed in `c23ef2b4`. `models.User.Subject` and `api.UserResponse.Subject` are `string` now. Every consumer already wanted the string: the column is a varchar on all four engines and was written through `driver.Valuer`, whose `Value()` is `String()`; JSON and the `sub` claim went through `MarshalText`, which emits the same 36 characters; templates rendered `String()`. So the typed value bought parse-on-read validation of the users column and of the admin console's own API responses — neither reachable from outside a deployment — while the rest of the tree spent 84 `.String()` calls unwrapping it. Not one byte that reaches a column, a claim, a JSON body or a template moves.

The guard that mattered is kept. `GetUserBySubject` still compares the stored spelling against the requested one, so a case-insensitive engine collation cannot fold a match OIDC Core section 2 says is case sensitive; its comment now rests on `uuidutil.New` being the column's only writer rather than on `uuid.UUID.Value()`.

The 95 test files stop importing the library in the same commit: fresh values come from `fake.UUID()`, well-formedness assertions from `uuidutil.Parse`, and the 19 `MustParse` literals are literals. `grep -rn 'uuid\.UUID' src --include=*.go` and `grep -rn 'Subject\.String()' src --include=*.go` both read 0.

Tiers: unit green, 5599 pass 0 fail. Data green on mysql, postgres, mssql and sqlite, 2208 pass 0 fail — `TestGetUserBySubject`, `TestGetUserBySubjectIsCaseSensitive` and `TestUnique_UserSubject` on every engine are what say the varchar scans into a `string` and the unique index still refuses a duplicate. Integration green on sqlite, 1373 pass 0 fail, over the `sub`-claim paths, userinfo, the implicit flow and the retyped user-bound-token helpers. One new case, `TestUserResponse_SubjectIsABareCanonicalString`, pins that `subject` marshals as the bare 36-character string and round-trips; mutating its JSON tag fails it.

One deviation worth knowing: `user_creator.go` and the seeder's admin user move to `uuidutil.New` here rather than in stage 3, because a `uuid.UUID` does not assign to a `string` field. That leaves stage 3 six production files, not seven; the auth code, jti, kid, session identifier and DCR prefix are untouched.

### Stage 3: production generators on `uuidutil.New`, and the module graphs tidy
Landed in `c244ec9d`. The eleven remaining production call sites become `uuidutil.New()`: the auth code keeps `strings.ReplaceAll(uuidutil.New(), "-", "")` so its 32-hex-plus-96 shape is unchanged, and the five `jti` claims, the rotator's `kid`, the seeder's two `kid` values, the session identifier and the `dcr_` client identifier are each a one-line substitution producing the same canonical 36-character lowercase v4. No Go file under `src/` imports `github.com/google/uuid` any more, and `grep -rn 'uuid\.UUID' src --include=*.go` reads 0.

`go mod tidy` in all four modules then turns the direct requirement in `core` and `authserver` into `// indirect` and leaves `adminconsole` and `cmd/goiabada-setup` untouched. No `go.sum` moved, because `github.com/microsoft/go-mssqldb@v1.10.0` and `modernc.org/sqlite@v1.57.0` both require the module and keep it in the build — the fact decision 8 was answered about.

Three of the substitution sites had no assertion that could have caught a wrong value: the auth code and the rotator's `kid` were checked with `NotEmpty`, and `generateDCRClientIdentifier` had no test at all. Each is now pinned to what it emits — the auth code's 128 characters, whose first 32 re-hyphenate into something `uuidutil.Parse` accepts and whose remaining 96 are the security alphabet's; the `kid` parsing and round-tripping; and a new `TestGenerateDCRClientIdentifier_IsThePrefixAndAUUID` reading the prefix, parsing the rest, and requiring two calls to differ. Five mutations, one per generation site with a seam over it, were all caught.

Tiers: unit green, 1515 pass 0 fail. Data green on mysql, postgres, mssql and sqlite, 1631 pass 0 fail across the four — the seeder is in the diff and every engine runs it. Integration green on sqlite, 1038 pass 0 fail, over the token issuer, userinfo and DCR paths.

### Stage 4: `stringutil` fails closed by crashing (closes #211)
Landed in `90a9c166`. `GenerateSecurityRandomString`, `GenerateRandomLetterString` and `GenerateRandomNumberString` answered a `crypto/rand` failure with `""`. They now draw through an unexported `io.Reader` whose `Read` calls `crypto/rand.Read`, so a failed draw ends the process rather than handing back a string nobody can tell apart from a real one. That is the same contract `uuidutil.New` has, which is the point of folding #211 in here: the two changes answer one question and answering it twice would leave the tree with two contracts for a while.

The wrapper's `if err != nil { return "" }` branch goes with it, as do the two `== ""` guards at `SaveLinkMarker` and `HandleAuthorizeGet` — the only two call sites that ever checked, out of sixteen. `randomStringFromReader` is untouched: signature, error, doc comment and `TestRandomStringFromReader_ErrorPropagates` all stay, because it is the seam that proves the rejection sampler stops on a short read and a caller can still supply its own source.

Two new cases, one per half. `TestGenerateSecurityRandomString_CrashesIrrecoverablyOnReaderFailure` re-executes the test binary with `crypto/rand.Reader` swapped in the child alone and asserts a non-zero exit carrying the standard library's own fatal message; reverting the wrapper to `io.ReadFull` returning `""` leaves every other case in that file green and fails only this one. `TestMustDraw_PanicsOnShortResult` holds `fake.mustDraw` directly, which survives as the fixture package's own refusal to seed a test with a degraded value. It replaces `TestLetterN_PanicsOnEntropyFailure` and its `DigitN` twin, which swapped `crypto/rand.Reader` and called through `stringutil`: after this change they would take the whole `fake` test binary down with the runtime's fatal handler instead of observing a panic. Both mutations — the wrapper reverted to fail-open, and `mustDraw` reduced to `return s` — were caught.

Tiers: unit green, 1515 pass 0 fail across all three modules. Integration green on sqlite, 1038 pass 0 fail, run because `handler_authorize.go` is in the diff. Data not applicable: no query, migration or schema is touched.

## How this was verified

**Tiers run locally, in the dev container, at every stage gate.** Unit on all three modules,
including the three `gofmt` guards and `go vet`. Data on **all four engines** — mysql, postgres,
mssql and sqlite — at stages 2 and 3, which are the stages that touch the column and the seeder.
Integration on sqlite at stages 2, 3 and 4. Stage 1 ran unit only: nothing outside the new package's
own tests called it yet, and no query, migration or endpoint was in its diff.

**Tiers run only by the check suite.** The four-engine *integration* tier, `Unit / race`, `Lint`,
`Vulnerabilities`, `Build validation` and `Versions` have no local equivalent in this run. All of
them, plus the four `Tests /` and four `Unit /` jobs, are **green at `1a8c39c0`**, which is HEAD.

**Mutation checks.** Every stage mutated its own new code and confirmed a named test failed: the two
UUID nibble masks and the fail-open rewrite of `New` (stage 1), the `subject` JSON tag (stage 2),
five generation sites (stage 3), the `stringutil` wrapper reverted to fail-open and `mustDraw`
reduced to a pass-through (stage 4). The final review re-ran compile-clean mutations at the current
tree, because seven of those were taken on trees later stages moved; all were caught.

## Deviations from the sealed plan

1. **Two production generator lines moved from stage 3 into stage 2** (`user_creator.go`,
   `database_seeder.go`). A `uuid.UUID` does not assign to a `string` field, so the type change does
   not compile without them. Stage 3 owns six production files, not seven.
2. **Stage 3 added coverage the plan did not ask for.** Three substitution sites — the auth code, the
   rotator's `kid`, and `generateDCRClientIdentifier` — had no assertion that could catch a wrong
   value. Each is now pinned. The seeder's two `kid` values are deliberately left unpinned; the
   rotator's pin covers the same contract.
3. **Two subprocess crash tests, in stages 1 and 4, against section 5's rejected list.** Forced by
   plan-review finding 4: section 5 rejected a subprocess test on the stated ground that the OS
   source cannot be made to fail, which reading Go 1.27's `crypto/rand.Read` refuted. Without them
   the branch's central claim had nothing executable behind it.
4. **Goal 1 narrowed, by decision 8**, answered A by the owner. The module cannot leave the module
   graphs while the SQL Server and sqlite drivers require it.
5. **`_, _ = rand.Read(b[:])` rather than the bare call** in the design sketch, because `errcheck`
   is enabled for all four modules and would redden Lint.
6. **`TestParse`'s prose must not name the retired module**, since goal 1's own measure counts
   comments as well as imports.

### Review

Reviewed by `openai-codex/gpt-5.6-sol` at effort `xhigh`, 2 rounds over the whole branch at the final
gate, plus 1 round over the plan before any code was written. No stage was reviewed on its own
(`LEO_REVIEW_MODE=end`).

|---|---|---|---|
| plan, 1 | 5 | 1 | 4 fixed in the plan; 1 escalated as decision 8, answered A |
| final, 1 | 3 | 0 | all confirmed, all fixed, `1a8c39c0` |
| final, 2 | 0 | 0 | clean, all three axes reviewed |

- **blocking, conformance (plan):** `google/uuid` cannot leave the module graphs — the SQL Server and
  sqlite drivers require it in all four modules. Confirmed with `go mod graph`. Escalated as decision
  8; answered **A**, goal 1 narrowed to direct imports and direct requirements.
- **significant, security (plan):** the fail-closed contract had no executable case, and the reason
  section 5 gave for that was false. Fail-open mutants of both `uuidutil.New` and `stringutil` passed
  every planned test. Answered by adding the two subprocess crash tests.
- **significant, quality (plan):** `fake.mustDraw`'s guard would be untested once stage 4 deletes the
  entropy tests. Answered by a direct short-result test in stage 4.
- **significant, quality (final 1):** `TestParse` pinned no *individual* separator position — the one
  malformed row replaced all four hyphens at once, so relaxing the check at any single index left
  another underscore to raise the error and the row passed anyway. Four mutations survived the suite.
  Fixed in `1a8c39c0`: four rows, one per index, each replacing a single hyphen. All four mutations
  are now caught at the position they target.
- **minor, quality (final 1), ×2:** `sessionstore/server_side_store.go` and its test both stated that
  `stringutil` "returns `""` when the CSPRNG fails (#211)" — a contract stage 4 deleted. Both reworded
  in `1a8c39c0` to the reason that now holds.

Nothing was refuted and nothing was deferred at any round.

Full account, with the evidence, in [this comment](https://github.com/leodip/goiabada/pull/316#issuecomment-5617850368).
Deferred decisions: [none, and why](https://github.com/leodip/goiabada/pull/316#issuecomment-5617844594).
Follow-ups, fold-ins and ceilings: [this comment](https://github.com/leodip/goiabada/pull/316#issuecomment-5617848757).

### DIFF
--- a/src/adminconsole/internal/handlers/admingrouphandlers/handler_admin_group_members_add.go
+++ b/src/adminconsole/internal/handlers/admingrouphandlers/handler_admin_group_members_add.go
@@ -126,7 +126,7 @@ func HandleAdminGroupMembersSearchGet(
 		for _, user := range users {
 			usersResult = append(usersResult, UserResult{
 				Id:           user.Id,
-				Subject:      user.Subject.String(),
+				Subject:      user.Subject,
 				Username:     user.Username,
 				Email:        user.Email,
 				GivenName:    user.GivenName,
--- a/src/adminconsole/internal/handlers/adminresourcehandlers/handler_admin_resource_users_with_permission.go
+++ b/src/adminconsole/internal/handlers/adminresourcehandlers/handler_admin_resource_users_with_permission.go
@@ -476,7 +476,7 @@ func HandleAdminResourceUsersWithPermissionSearchGet(
 		for _, u := range annotatedUsers {
 			usersResult = append(usersResult, UserResult{
 				Id:            u.Id,
-				Subject:       u.Subject.String(),
+				Subject:       u.Subject,
 				Username:      u.Username,
 				Email:         u.Email,
 				GivenName:     u.GivenName,
--- a/src/adminconsole/internal/rendertest/render_test.go
+++ b/src/adminconsole/internal/rendertest/render_test.go
@@ -297,9 +297,7 @@ func TestRender_AdminClientWebOrigins(t *testing.T) {
 func TestRender_AdminUsersPaginator(t *testing.T) {
 	out := render(t, "/admin_users.html", map[string]interface{}{
 		"pageResult": adminuserhandlers.PageResult{
-			// Subject is left at its zero UUID: the row only has to render, and
-			// importing google/uuid for it would make adminconsole a direct
-			// consumer of a module it otherwise only inherits.
+			// Subject is left empty: the row only has to render.
 			Users:    []models.User{{Id: 1, Username: "alice", Email: "alice@example.com"}},
 			Total:    73,
 			Query:    "",
--- a/src/authserver/go.mod
+++ b/src/authserver/go.mod
@@ -6,7 +6,6 @@ require (
 	github.com/PuerkitoBio/goquery v1.12.0
 	github.com/go-chi/chi/v5 v5.3.2
 	github.com/golang-jwt/jwt/v5 v5.3.1
-	github.com/google/uuid v1.6.0
 	github.com/leodip/goiabada/core v0.0.0
 	github.com/microsoft/go-mssqldb v1.10.0
 	github.com/pkg/errors v0.9.1
@@ -27,6 +26,7 @@ require (
 	github.com/go-sql-driver/mysql v1.10.0 // indirect
 	github.com/golang-sql/civil v0.0.0-20220223132316-b832511892a9 // indirect
 	github.com/golang-sql/sqlexp v0.1.0 // indirect
+	github.com/google/uuid v1.6.0 // indirect
 	github.com/huandu/go-clone v1.7.3 // indirect
 	github.com/huandu/go-sqlbuilder v1.43.0 // indirect
 	github.com/huandu/xstrings v1.5.0 // indirect
--- a/src/authserver/internal/handlers/apihandlers/handler_api_account_email.go
+++ b/src/authserver/internal/handlers/apihandlers/handler_api_account_email.go
@@ -54,7 +54,7 @@ func HandleAPIAccountEmailPut(
 
 		// Validate email (server-side rules; confirmation is a UI concern)
 		email := strings.ToLower(strings.TrimSpace(req.Email))
-		if err := emailValidator.ValidateEmailChange(email, user.Subject.String()); err != nil {
+		if err := emailValidator.ValidateEmailChange(email, user.Subject); err != nil {
 			writeValidationError(w, r, err)
 			return
 		}
--- a/src/authserver/internal/handlers/apihandlers/handler_api_account_profile.go
+++ b/src/authserver/internal/handlers/apihandlers/handler_api_account_profile.go
@@ -105,7 +105,7 @@ func HandleAPIAccountProfilePut(
 			ZoneInfoCountryName: req.ZoneInfoCountryName,
 			ZoneInfo:            req.ZoneInfo,
 			Locale:              req.Locale,
-			Subject:             user.Subject.String(),
+			Subject:             user.Subject,
 		}
 
 		if err := profileValidator.ValidateProfile(input); err != nil {
--- a/src/authserver/internal/handlers/apihandlers/handler_api_account_profile_picture.go
+++ b/src/authserver/internal/handlers/apihandlers/handler_api_account_profile_picture.go
@@ -119,7 +119,7 @@ func HandleAPIAccountProfilePicturePost(
 		// Return success response
 		response := map[string]interface{}{
 			"success":    true,
-			"pictureUrl": config.GetAuthServer().BaseURL + "/userinfo/picture/" + user.Subject.String(),
+			"pictureUrl": config.GetAuthServer().BaseURL + "/userinfo/picture/" + user.Subject,
 		}
 
 		w.Header().Set("Content-Type", "application/json")
@@ -223,7 +223,7 @@ func HandleAPIAccountProfilePictureGet(
 		}
 
 		if hasPicture {
-			response["pictureUrl"] = config.GetAuthServer().BaseURL + "/userinfo/picture/" + user.Subject.String()
+			response["pictureUrl"] = config.GetAuthServer().BaseURL + "/userinfo/picture/" + user.Subject
 		}
 
 		w.Header().Set("Content-Type", "application/json")
--- a/src/authserver/internal/handlers/apihandlers/handler_api_account_profile_picture_test.go
+++ b/src/authserver/internal/handlers/apihandlers/handler_api_account_profile_picture_test.go
@@ -14,13 +14,13 @@ import (
 	"net/http/httptest"
 	"testing"
 
-	"github.com/google/uuid"
 	mocks_audit "github.com/leodip/goiabada/authserver/internal/audit/mocks"
 	"github.com/leodip/goiabada/core/constants"
 	mocks_data "github.com/leodip/goiabada/core/data/mocks"
 	mocks_handlerhelpers "github.com/leodip/goiabada/core/handlerhelpers/mocks"
 	"github.com/leodip/goiabada/core/models"
 	"github.com/leodip/goiabada/core/oauth"
+	"github.com/leodip/goiabada/core/testutil/fake"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/mock"
 )
@@ -114,7 +114,7 @@ func TestHandleAPIAccountProfilePictureGet_UserNotFound(t *testing.T) {
 
 	handler := HandleAPIAccountProfilePictureGet(database)
 
-	sub := uuid.New().String()
+	sub := fake.UUID()
 	req, _ := http.NewRequest("GET", "/api/v1/account/profile-picture", nil)
 	req = setTokenContext(req, sub)
 	rr := httptest.NewRecorder()
@@ -132,13 +132,13 @@ func TestHandleAPIAccountProfilePictureGet_HasPicture(t *testing.T) {
 
 	handler := HandleAPIAccountProfilePictureGet(database)
 
-	sub := uuid.New()
+	sub := fake.UUID()
 	req, _ := http.NewRequest("GET", "/api/v1/account/profile-picture", nil)
-	req = setTokenContext(req, sub.String())
+	req = setTokenContext(req, sub)
 	rr := httptest.NewRecorder()
 
 	user := &models.User{Id: 1, Subject: sub, Enabled: true}
-	database.On("GetUserBySubject", (*sql.Tx)(nil), sub.String()).Return(user, nil)
+	database.On("GetUserBySubject", (*sql.Tx)(nil), sub).Return(user, nil)
 	database.On("UserHasProfilePicture", (*sql.Tx)(nil), user.Id).Return(true, nil)
 
 	handler.ServeHTTP(rr, req)
@@ -149,7 +149,7 @@ func TestHandleAPIAccountProfilePictureGet_HasPicture(t *testing.T) {
 	err := json.Unmarshal(rr.Body.Bytes(), &response)
 	assert.NoError(t, err)
 	assert.True(t, response["hasPicture"].(bool))
-	assert.Contains(t, response["pictureUrl"].(string), sub.String())
+	assert.Contains(t, response["pictureUrl"].(string), sub)
 
 	database.AssertExpectations(t)
 }
@@ -159,13 +159,13 @@ func TestHandleAPIAccountProfilePictureGet_NoPicture(t *testing.T) {
 
 	handler := HandleAPIAccountProfilePictureGet(database)
 
-	sub := uuid.New()
+	sub := fake.UUID()
 	req, _ := http.NewRequest("GET", "/api/v1/account/profile-picture", nil)
-	req = setTokenContext(req, sub.String())
+	req = setTokenContext(req, sub)
 	rr := httptest.NewRecorder()
 
 	user := &models.User{Id: 1, Subject: sub, Enabled: true}
-	database.On("GetUserBySubject", (*sql.Tx)(nil), sub.String()).Return(user, nil)
+	database.On("GetUserBySubject", (*sql.Tx)(nil), sub).Return(user, nil)
 	database.On("UserHasProfilePicture", (*sql.Tx)(nil), user.Id).Return(false, nil)
 
 	handler.ServeHTTP(rr, req)
@@ -201,7 +201,7 @@ func TestHandleAPIAccountProfilePicturePost_UserNotFound(t *testing.T) {
 
 	handler := HandleAPIAccountProfilePicturePost(database, auditLogger)
 
-	sub := uuid.New().String()
+	sub := fake.UUID()
 	pictureData := createTestPNG(100, 100)
 	req, err := createMultipartRequest("POST", "/api/v1/account/profile-picture", "picture", pictureData)
 	assert.NoError(t, err)
@@ -222,15 +222,15 @@ func TestHandleAPIAccountProfilePicturePost_NoFile(t *testing.T) {
 
 	handler := HandleAPIAccountProfilePicturePost(database, auditLogger)
 
-	sub := uuid.New()
+	sub := fake.UUID()
 	user := &models.User{Id: 1, Subject: sub, Enabled: true}
 
 	req, _ := http.NewRequest("POST", "/api/v1/account/profile-picture", nil)
 	req.Header.Set("Content-Type", "multipart/form-data")
-	req = setTokenContext(req, sub.String())
+	req = setTokenContext(req, sub)
 	rr := httptest.NewRecorder()
 
-	database.On("GetUserBySubject", (*sql.Tx)(nil), sub.String()).Return(user, nil)
+	database.On("GetUserBySubject", (*sql.Tx)(nil), sub).Return(user, nil)
 
 	handler.ServeHTTP(rr, req)
 
@@ -243,17 +243,17 @@ func TestHandleAPIAccountProfilePicturePost_InvalidImage(t *testing.T) {
 
 	handler := HandleAPIAccountProfilePicturePost(database, auditLogger)
 
-	sub := uuid.New()
+	sub := fake.UUID()
 	user := &models.User{Id: 1, Subject: sub, Enabled: true}
 
 	// Create a request with invalid image data
 	invalidImageData := []byte("not a valid image")
 	req, err := createMultipartRequest("POST", "/api/v1/account/profile-picture", "picture", invalidImageData)
 	assert.NoError(t, err)
-	req = setTokenContext(req, sub.String())
+	req = setTokenContext(req, sub)
 	rr := httptest.NewRecorder()
 
-	database.On("GetUserBySubject", (*sql.Tx)(nil), sub.String()).Return(user, nil)
+	database.On("GetUserBySubject", (*sql.Tx)(nil), sub).Return(user, nil)
 
 	handler.ServeHTTP(rr, req)
 
@@ -271,16 +271,16 @@ func TestHandleAPIAccountProfilePicturePost_CreateNew(t *testing.T) {
 
 	handler := HandleAPIAccountProfilePicturePost(database, auditLogger)
 
-	sub := uuid.New()
+	sub := fake.UUID()
 	user := &models.User{Id: 1, Subject: sub, Enabled: true}
 
 	pictureData := createTestPNG(100, 100)
 	req, err := createMultipartRequest("POST", "/api/v1/account/profile-picture", "picture", pictureData)
 	assert.NoError(t, err)
-	req = setTokenContext(req, sub.String())
+	req = setTokenContext(req, sub)
 	rr := httptest.NewRecorder()
 
-	database.On("GetUserBySubject", (*sql.Tx)(nil), sub.String()).Return(user, nil)
+	database.On("GetUserBySubject", (*sql.Tx)(nil), sub).Return(user, nil)
 	database.On("GetUserProfilePictureByUserId", (*sql.Tx)(nil), user.Id).Return(nil, nil)
 	database.On("CreateUserProfilePicture", (*sql.Tx)(nil), mock.MatchedBy(func(pp *models.UserProfilePicture) bool {
 		return pp.UserId == user.Id && pp.ContentType == "image/png"
@@ -298,7 +298,7 @@ func TestHandleAPIAccountProfilePicturePost_CreateNew(t *testing.T) {
 	err = json.Unmarshal(rr.Body.Bytes(), &response)
 	assert.NoError(t, err)
 	assert.True(t, response["success"].(bool))
-	assert.Contains(t, response["pictureUrl"].(string), sub.String())
+	assert.Contains(t, response["pictureUrl"].(string), sub)
 
 	database.AssertExpectations(t)
 	auditLogger.AssertExpectations(t)
@@ -310,7 +310,7 @@ func TestHandleAPIAccountProfilePicturePost_UpdateExisting(t *testing.T) {
 
 	handler := HandleAPIAccountProfilePicturePost(database, auditLogger)
 
-	sub := uuid.New()
+	sub := fake.UUID()
 	user := &models.User{Id: 1, Subject: sub, Enabled: true}
 	existingPicture := &models.UserProfilePicture{
 		Id:          1,
@@ -322,10 +322,10 @@ func TestHandleAPIAccountProfilePicturePost_UpdateExisting(t *testing.T) {
 	pictureData := createTestPNG(100, 100)
 	req, err := createMultipartRequest("POST", "/api/v1/account/profile-picture", "picture", pictureData)
 	assert.NoError(t, err)
-	req = setTokenContext(req, sub.String())
+	req = setTokenContext(req, sub)
 	rr := httptest.NewRecorder()
 
-	database.On("GetUserBySubject", (*sql.Tx)(nil), sub.String()).Return(user, nil)
+	database.On("GetUserBySubject", (*sql.Tx)(nil), sub).Return(user, nil)
 	database.On("GetUserProfilePictureByUserId", (*sql.Tx)(nil), user.Id).Return(existingPicture, nil)
 	database.On("UpdateUserProfilePicture", (*sql.Tx)(nil), mock.MatchedBy(func(pp *models.UserProfilePicture) bool {
 		return pp.Id == existingPicture.Id && pp.ContentType == "image/png"
@@ -370,7 +370,7 @@ func TestHandleAPIAccountProfilePictureDelete_UserNotFound(t *testing.T) {
 
 	handler := HandleAPIAccountProfilePictureDelete(httpHelper, database, auditLogger)
 
-	sub := uuid.New().String()
+	sub := fake.UUID()
 	req, _ := http.NewRequest("DELETE", "/api/v1/account/profile-picture", nil)
 	req = setTokenContext(req, sub)
 	rr := httptest.NewRecorder()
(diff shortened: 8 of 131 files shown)


leodip:
## Decision 8: the `google/uuid` module cannot be removed from the build. Continue with a narrower goal, or stop?

**What the thing is.** Issue #278 asks to retire `github.com/google/uuid`, the third-party library Goiabada uses to mint random identifiers: user subjects, session identifiers, token ids, signing-key ids and DCR client ids. Every use is "give me a fresh random string", so the plan replaces it with a small package in this repository, built on the standard library, that produces the same 36-character identifiers.

**Where it shows.** Nowhere a user can see. The identifiers stored and issued stay byte-identical. What changes is the dependency list in the four `go.mod` files under `src/` and the 105 Go files that import the library today.

**What the plan review found.** The reviewer built the whole change in a scratch tree and ran `go mod tidy`. The library stayed in every module. I checked with `go mod graph` in the dev container and it is right: two drivers Goiabada depends on require it themselves.

```
github.com/microsoft/go-mssqldb@v1.10.0  ->  github.com/google/uuid@v1.6.0
modernc.org/sqlite@v1.57.0 (via modernc.org/libc@v1.75.4)  ->  github.com/google/uuid@v1.6.0
```

So after the change the library is still downloaded, still in `go.sum`, still compiled into the binaries through the SQL Server and sqlite drivers, and still listed in all four `go.mod` files as `// indirect`. The issue's stated outcome, no `google/uuid` anywhere in the module files, cannot be reached without dropping SQL Server or sqlite support. The issue's own count missed this because `go mod why` reports the first Goiabada importer of the package and never the driver edges behind it.

**Who is affected.** Only the goal statement. Nothing about the code change itself is wrong or unsafe. The question is whether the work is still worth doing once its headline outcome is smaller than the issue promised.

**The axis the choice turns on:** what "retired" means for this library.

- **A. Continue, with the goal narrowed.** Retired means no Goiabada source file imports it and no module requires it directly. Goiabada stops depending on the library's API and semantics, so a future breaking change or vulnerability in it matters only as much as it matters to the two drivers, and the direct-dependency count drops by one. The library remains in the build as an indirect requirement of the drivers. Cost: the full change as planned, about 105 files, nearly all mechanical test edits, plus the #211 fix folded in.
- **B. Stop.** Retired means gone from the build, which is not achievable while those drivers need it. Close this PR, leave the tree as is, and put a note on #278 saying it is blocked on the drivers. The #211 fix (`stringutil` returning an empty string when the system random source fails) would then need its own change. Cost: nothing now, and the two open issues stay open.

**Worked example.** Say a CVE lands against `google/uuid` next year. Under A, `go mod graph` still shows the module, a scanner still flags the binary, and the fix is still "bump the driver or wait for it to bump". Under B, the same. The difference A buys is not exposure, it is that Goiabada's own code no longer calls the library, so nothing in `src/` has to change when the drivers move to a different version or a different library.

**What does not differ.** The stored and issued identifiers are unchanged under either option. Neither option touches the database, the wire format or any user-visible behaviour.

**Recommendation: A.** The dependency-reduction series (#268 to #281) has been retiring direct dependencies, and this one is retired by the same measure as the others. I am asking rather than deciding because the issue's premise changed: it promised the module would disappear from the module files, and it will not.

**Also applied from the plan review, no answer needed:** four coverage findings went into the plan. The session-identifier test now rejects the all-zero UUID rather than only an empty string. The `fake` fixture's length guard gets a direct test. And the "a random-source failure ends the process rather than returning an empty identifier" contract gets a real executable test in both the new package and `stringutil`, by re-running the test binary as a child process with a failing reader, which the agreement had assumed could not be done.

leodip:
A.

leodip:
## Deferred decisions

**None.** Nothing was assumed past in order to keep moving. Every question this run met was either
settled in the sealed agreement before any code was written, or escalated and stopped on.

For completeness, the one decision the run itself raised is **decision 8**, and it was not deferred:
it blocked the run. The plan review found with `go mod graph` that `github.com/google/uuid` cannot
leave the module graphs at all, because `github.com/microsoft/go-mssqldb@v1.10.0` and
`modernc.org/sqlite@v1.57.0` (via `modernc.org/libc@v1.75.4`) both require it, in all four modules
including `cmd/goiabada-setup`. That falsified goal 1 as sealed ("appears in no `go.mod`, `go.sum` or
Go file under `src/`"), so the run posted the question and stopped rather than choosing.

You answered **A** on 2026-09-10T08:50:04Z. Goal 1 was reworded the same day to the measure the rest
of the #268–#281 series was retired by: no Go file under `src/` imports the library and no `go.mod`
requires it **directly**; it remains `// indirect` through those two drivers. Section 4's module-graph
paragraph and fanout row, and section 0's `gomod/adminconsole` note, were corrected to say the drivers
hold it. The library's exposure to a scanner is unchanged; what the change buys is that no Goiabada
code calls its API.

Nothing else is open. All seven sealed decisions read **Decided**, and the run raised no others.

leodip:

### Follow-ups to file

**None.** No stage found anything next to the work that it could not carry. Stage 1's log records
"Follow-ups drafted. None"; stage 4's records "None found"; stages 2 and 3 drafted none; and the final
review, across two rounds, left "no follow-up owed by the review itself".

Duplicate search re-run at the finish, in case the tracker moved while the run was working:
`gh issue list --state all --search "uuid"` and `--search "stringutil CSPRNG"`. Nothing new. The
neighbours are unchanged and untouched by this branch: **#279** (pkg/errors) and **#280**
(go-chi/cors) are open and share no lines; **#281** (mileusna/useragent) is open and independent;
**#272** and **#271** are closed.

- **#211, `GenerateSecurityRandomString` swallows a CSPRNG failure.** Folded in by **decision 6** at
  agreement time, landed in `90a9c166`, closed by this PR. Why it fit: it is the same defect this
  issue's own warning describes, and the two changes answer one question — what happens when the
  system CSPRNG fails. Answering it in two pull requests would leave the tree carrying two opposite
  contracts for a while. Decision 6 records that it fails the usual fold-in tests (the generator lives
  in its own package, so `stringutil.go` was not otherwise in the diff, and the shape was a choice) and
  that you chose to carry it anyway; decision 7 settled the shape as *crash*.
- **Three unpinned generation sites, stage 3, `c244ec9d`.** Not a scope change but coverage the plan
  did not ask for, recorded as a deviation before any code moved. The auth code and the rotator's
  `kid` were asserted only with `NotEmpty`, and `generateDCRClientIdentifier` had no test at all, so
  a wrong value at any of the three would have gone unobserved while the stage claimed
  "byte-identical". Added `assertAuthCodeShape`, a `uuidutil.Parse` on the rotator's kid, and
  `TestGenerateDCRClientIdentifier_IsThePrefixAndAUUID`. Each is proven able to fail by its own
  mutation row.
- **Two subprocess crash tests, stages 1 and 4.** A deviation from section 5's rejected list, forced
  by plan-review finding 4: section 5 had rejected a subprocess test "since the OS source cannot be
  made to fail", which reading Go 1.27's `crypto/rand.Read` refuted — swapping `crypto/rand.Reader`
  in a child process induces the fatal without any OS failure, and the reviewer's fail-open mutants
  passed every other planned test. Without these two cases the branch's central claim, that a CSPRNG
  failure ends the process, had nothing executable behind it.

**none.** `git diff --unified=0 origin/main...HEAD | grep '^+.*ceiling:'` matches nothing; this branch
added no `ceiling:` marker.

leodip:
## The code review

`LEO_REVIEW_MODE=end`, so the whole change was reviewed once after the last stage landed, plus one
review of the plan before any code was written. This is the full account, pasted rather than linked:
the verdicts live in a gitignored directory that the next run overwrites, so without this comment the
fix commits would exist with no record of what prompted them, what was refuted, or what was left.

Three rounds in all: one at the plan gate (five findings, all answered, one of them escalated as
decision 8) and two at the final gate (three findings, all confirmed and fixed; then clean).

---

## Review account: plan gate, round 1, 2026-09-10

Reviewer `openai-codex/gpt-5.6-sol`, effort xhigh, fresh session, request `278-20260910T005653Z`.
Verdict `static-review-exhausted`, five findings, all three axes reviewed. It built the planned
stages in a scratch tree and mutation-tested the plan's cases. Unchecked: no diff to hash at a plan
gate; full tiers not run; no browser pass; unrelated security classes checked at call sites only.
All four answered in `log/plan.md`.

1. **`fake.mustDraw` guard left untested once stage 4 deletes the entropy tests.** Quality,
   significant. Confirmed: `mustDraw` reduced to `return s` left the `fake` suite green in the
   reviewer's scratch tree. Fixed in the plan: stage 4 adds a direct short-result test.
2. **Non-empty check accepts the nil UUID.** Conformance, significant. Confirmed: `uuidutil.Parse`
   accepts the nil form by design, so the translated session-identifier assertion was weaker than
   the `uuid.Nil` one it replaced. Fixed in the plan: stage 2 compares against the canonical nil
   string.
3. **`google/uuid` cannot leave the module graphs.** Conformance, blocking. Confirmed with
   `go mod graph` in the container: `go-mssqldb@v1.10.0` and `modernc.org/sqlite@v1.57.0` require
   it, in all four modules including `cmd/goiabada-setup`. Escalated as decision 8 on PR #316:
   continue with goal 1 narrowed to direct imports and direct requirements (A, recommended), or
   stop (B).
4. **No executable case for the process-crash contract, and the stated reason was false.**
   Security, significant. Confirmed against the Go 1.27 `crypto/rand.Read` source: a swapped
   `Reader` in a child process induces `fatal` without an OS failure; the reviewer's fail-open
   mutants of `uuidutil.New` and `stringutil` passed every planned test. Fixed in the plan as a
   deviation from section 5's rejected list: self-exec subprocess tests in stage 1 and stage 4,
   each with a mutation check against the fail-open shape.
5. **`New`'s doc comment would describe behaviour stage 4 removes.** Quality, minor. Fixed in the
   plan: the comment states its own contract only.

Nothing refuted.

## Recorded claims at the final gate, 2026-09-10

Request `278-20260910T103104Z`, code review, gate `final`, round 1, written against
`90a9c166` with the tree clean. `tree-hash.sh --verify` at that instant:

```
current   production 1744ebe5e2f1   unit-tests ef1d873d14a4   data-tests 458d14685f2d
          integration-tests ba7563b9132e   docs 2fbbd8f0cd6f   all ed7ab24370ee

         data-tests, integration-tests not recorded here, so those claims cannot be settled either way
stage 2  unit ?              data ?              integration ?       docs -
         data-tests, integration-tests, unit-tests not recorded here, so those claims cannot be settled either way
         moved since: production
stage 3  unit stale          data stale          integration stale   docs -
stage 4  unit live           data ?              integration live    docs -
         data-tests not recorded here, so those claims cannot be settled either way
```

3 of 4 entries recorded a result against a tree that has since moved, which is what four
stages landing in sequence does: each one moved `production`, and stages 1 and 3 moved
`unit-tests` as well. What that costs is the mutation evidence. Stage 1 recorded four
mutations, stages 2 to 4 one each, and the seven taken before stage 4 landed were taken on
trees that no longer exist, so they are claims about code that has since been edited rather
than about what is on the branch now.

What is not in doubt is the tiers. The check suite is green at `90a9c166` on all thirteen
jobs, which includes the data and integration tiers on sqlite, mysql, postgres and mssql,
plus `Unit / race`, `Lint`, `Vulnerabilities` and `Versions`. That is a result against the
current tree and it supersedes every per-stage tier claim above. It says nothing about
whether the new tests can fail, which is the part the stale hashes bear on and the part the
review is being asked to read.

## Final review, round 1, 2026-09-10

Reviewer `pi openai-codex/gpt-5.6-sol`, effort `xhigh`, fresh session, request
`278-20260910T103104Z`, gate `final`, round 1. Verdict `findings`: three, two quality/minor
and one quality/significant, none security, none conformance. All three axes read `reviewed`.
Every finding carried `judgement: false`, so none was a choice to escalate. All three were
verified against the code, all three stood, and all three are fixed in `1a8c39c0`.

**Finding 3, quality/significant, `uuidutil_test.go:TestParse` pins no individual separator
position.** Confirmed and fixed. The table's only malformed-separator row replaced all four
hyphens at once, so relaxing the check at one index left another underscore to raise
`errHyphen` and the row passed anyway. The reviewer demonstrated it: four mutations relaxing
positions 8, 13, 18 and 23 independently, all four surviving the core suite. Reproduced here
in the same shape, `if c != '-' {` to `if c != '-' && i != <idx> {`, one per index. Fixed by
four rows, each replacing a single hyphen with an underscore and expecting `errHyphen`, plus
a comment saying why the all-four row cannot stand in for them. The same four mutations are
now **MUTATION CAUGHT** at every position, on `--type modules --run TestParse`. This is the
finding that mattered: the parser is the boundary decision 4 drew, and until this commit the
tests admitted a parser that accepts an arbitrary character in the middle of a subject.

**Finding 1, quality/minor, `sessionstore/server_side_store.go:newSessionId`.** Confirmed and
fixed. The comment said `stringutil.GenerateSecurityRandomString` "returns `""` when the
CSPRNG fails rather than an error (#211)", which stage 4 made false: the helper has no error
return at all and `crypto/rand.Read` ends the process first. The comment therefore gave a
false reason for keeping the seam separate, and misdescribed the very defect this branch
closes. Reworded to the reason that now holds: a save has an error return, so an entropy
failure costs the one save rather than the server, and `randReader` keeps that failure
reachable from a test where `crypto/rand` is not.

**Finding 2, quality/minor, `server_side_store_test.go:TestServerSideStore_CSPRNGFailureFailsTheSave`.**
Confirmed and fixed. Same stale contract stated a second time, in the test's explanatory
comment. Reworded to state the property the test pins -- no session is saved with an
identifier or a nonce that did not come from the CSPRNG -- and to name the fatal alternative
`stringutil` now takes.

Nothing was refuted and nothing was deferred.

**`unchecked`, one entry, dismissed.** "No interactive browser pass was performed." Dismissed
with the reviewer's own reason, which is the right one: no template, page or locale string
changed on this branch, and the subject's rendering bytes were checked against the real
retired type in the reviewer's scratch program, with CI's render tests green at HEAD. A
browser pass could only re-observe a string this change proved byte-identical.

**What the round settled about the stale hashes.** The request said the mutation evidence
from stages 1 to 3 was taken on trees that no longer exist. The reviewer answered that
directly rather than trusting it: it re-ran compile-clean mutations at the current tree for
both UUID masks, the fatal-entropy contract, parser case normalisation, broad hyphen
rejection, `stringutil`'s fail-open, and the response subject, auth-code, signing-key,
session-ID and DCR-ID shapes, and each was caught. That supersedes the stale per-stage
mutation claims with evidence against the branch as it stands.

**One bookkeeping correction from the reviewer, recorded rather than acted on.** Stage 3's log
records mutations for the auth-code, signing-key, session-ID and DCR-ID call sites that
replaced the whole `uuidutil.New()` call, which removes those files' only use of the package
and so is caught by the compiler rather than by a test. The reviewer re-ran compile-clean
malformed and truncated variants of each and the named behaviour test failed every time, so
the coverage conclusion those entries drew stands; only their evidence was weaker than
recorded. Noted in stage 3's log entry.

Round 1 was productive: it moved production (a comment that was false), moved tests (four
rows that pin what decision 4 decided), and left nothing contested. Round 2 is unconditional
under the budget and its request follows.

## Final review, round 2, 2026-09-10

Reviewer `pi openai-codex/gpt-5.6-sol`, effort `xhigh`, resumed session, request
`278-20260910T110806Z`, gate `final`, round 2. Verdict **clean**: `findings` empty, and all
three axes -- conformance, quality, security -- read `reviewed`. Nothing to verify, nothing
to refute, nothing deferred. No code, test or documentation moved in this ingest; the tree at
ingest is byte-identical to the tree the round was reviewed at, `production 98ac0877085d`,
`unit-tests 6dbaa69182cd`, `data-tests 458d14685f2d`, `integration-tests ba7563b9132e`,
`docs 2fbbd8f0cd6f`, `all 60c9fc8cc41d`, working tree clean at `1a8c39c0`.

**What it executed, so the absence of findings has a stated scope.** The whole branch through
`review-diff.sh origin/main`, plus a focused `git diff 90a9c166..HEAD` on round 1's repair.
`tree-hash.sh` and `check-anchors.sh` against the agreement. `where.sh ci HEAD`, read
repeatedly as jobs concluded. A `git grep` fanout over the things goal 1 and section 4 promise
are gone or accounted for: `google/uuid`, `uuid.UUID`, `Subject.String`, the module
requirements, the `stringutil` callers, and the `#211` references. A scripted check that each
new malformed-UUID row differs from the canonical form **only** at its named separator
position, which is the property round 1's fix depends on and the one a hand-written table gets
wrong. Then four `where.sh mutate` runs relaxing separator validation independently at indexes
8, 13, 18 and 23. And `git diff --check` with a production-statement classification of the
diff.

That is the reading that earns the verdict: the four mutations it ran are the same four it
demonstrated surviving at round 1, and all four are now caught by the rows that fix targeted.
Its note records the conclusion directly -- round 1's three fixes are correct, every formerly
surviving separator mutation is caught by the intended row, and both reworded comments state
the entropy contract that now holds.

**`unchecked`, one entry, covered.** "The exact-HEAD SQL Server CI job was still in progress
when this review completed. It was green at the reviewed parent, this repair changes no
executable production or data code, and every other exact-HEAD job had passed." Covered here
rather than dismissed: `where.sh ci` at ingest reports **green at
`1a8c39c03bb65289c08a4bdc5bf8da6c4a7c96b6`** on all thirteen jobs, `Tests / mssql` among them,
alongside `Tests / mysql`, `Tests / postgres`, `Tests / sqlite`, the four `Unit` jobs including
`race`, `Lint`, `Vulnerabilities`, `Build validation` and `Versions`. The one job the reviewer
could not wait for has since concluded and passed, so the four-engine data and integration
claim now holds at HEAD and not only at the reviewed parent `90a9c166`.

**Round 2 was unproductive, and that is the gate closing rather than stalling.** The verdict
found nothing, so nothing shipped moved and nothing could have: under the budget an
unproductive round earns no successor, and here there is no successor to want. Rounds 1 and 2
were both unconditional; round 1 was productive and its three findings are all fixed in
`1a8c39c0`; round 2 independently re-established the property round 1 fixed and found no
further defect on any axis. The final gate is closed with two rounds of the four available,
nothing contested, nothing deferred and no follow-up owed by the review itself.

Issue #320, opened 2026-09-11
Streamline logging: one slog convention across the tree, enforced by lint


leodip:
## Summary

Every server package logs through `log/slog` and the #279 rule (error as a value under an attribute, never as text) holds at every site. What sits above that baseline is not one convention but several, grown per file. This issue is to converge on one and make the rules decidable by a lint test, the way `core/errs` and the credential-read rule already are.

Census (production code, `src/`): 207 `slog.Info`, 143 `slog.Error`, 80 `slog.Warn`, 12 `slog.Debug`. No stdlib `log`, no third-party logger. `go vet` reports no key/value mismatches. Neither server installs a handler: both run slog's built-in default (text on stderr via `log`, level Info). Only `core/cmd/schemadump` calls `slog.SetDefault`.

## Inconsistencies

1. **Attribute key for the error.** `"error"` at 66 sites, `"err"` at 5 (`authserver/internal/handlers/handler_account_logout.go` ×4, `core/handlerhelpers/template_funcs.go:140`).
2. **Request id key.** `"request_id"` everywhere except `"request-id"` at `authserver/internal/middleware/middleware_session_identifier.go:42`.
3. **Key casing.** The request logger and the 500 writers use snake_case (`request_id`, `status`, `duration`, `target`, `bytes`, `ip`); handlers use camelCase (`clientId`, `userId`, `sessionIdentifier`, `keyId`, `refreshTokenId`, `codeId`). The same concept has two keys: `clientId` (5 sites) and `clientIdentifier` (12).
4. **Message construction.** Three styles: structured attributes; `fmt.Sprintf` inside the call (77 sites); string concatenation (41). Most are startup lines, but some are per-request:
   - "auth context is missing, redirecting to %v" is copy-pasted 12 times across 8 handler files with the URL formatted into the text.
   - `authserver/internal/handlers/client_display.go:34` formats the error with `%v`, which drops its stack.
   - `adminconsole/internal/handlers/handler_auth_callback.go:79` logs at Info on every login with the base URL concatenated in.
   - `authserver/internal/workers/background_worker.go` formats counts and timeouts into Info messages.
5. **Component tags in the message.** `logout:`, `TokenParser:`, `JSBootstrap:`, `[DEBUG API]`, `config:`, `AuthServer:` are ad hoc prefixes rather than an attribute or a `slog.With` logger.
[...]

PR #321, opened 2026-09-11
feat(logging): one slog convention across the tree, enforced by lint (#320)

leodip:
Issue #320: one `slog` convention across the tree, enforced by sloglint and one small lint.

## What it is for

Goiabada's 441 `slog` calls were written one at a time and it shows: seven spellings of the
same key, messages built with `Sprintf` and concatenation, `err` beside `error`, `request-id`
beside `request_id`, five banner blocks of 52 records, JSON and stack traces pasted into
messages, and no handler installed, so the output shape was Go's `log`-style default and nothing
could raise the level or emit JSON. This change installs one handler per server, injects the
request id from the context so no call site names it, sweeps every call to one vocabulary, fixes
the five correctness items found on the way, and holds the convention with two guards: `sloglint`
over type-checked code, in `.golangci.yml` and run by a new `lint` tier of `run-tests.sh` as well as
CI, and a small `testutil.AssertSlogConvention` for the three things sloglint cannot express.

## The rule it establishes

A `slog` call has a literal, lowercase message with no component prefix; values go in snake_case
attributes, `error` and `request_id` are the fixed names for those two, and one concept has one
key. Inside any function that carries a context or a request, the `*Context` variant is used so
the handler can append `request_id`. Error is "the server could not do what it was asked and
someone must act", Warn is a request refused or a condition met and handled, Info is lifecycle
and summaries, Debug is per-request tracing. `testutil.AssertSlogConvention` holds the decidable
part; per-site level tests through `testutil.CaptureSlog` hold the rest.

Two new settings per server: `GOIABADA_<SERVER>_LOG_LEVEL` (`debug|info|warn|error`, default
`info`) and `GOIABADA_<SERVER>_LOG_FORMAT` (`text|json`, default `text`). An unknown value refuses
startup. Under JSON an error attribute keeps its `core/errs` stack.

**Status:** complete, and ready for review. All seven stages landed (the handler, its config and both mains; the test capture helper; the four structured loggers; the sweep of `core`, of `authserver` and of `adminconsole`; and the lint), the review closed at six rounds and twenty-three findings, and the review of this PR then added one round of two findings, both fixed in the commits after `c629fb61` and described under **Review round of #321** below: the custom lint was replaced by `sloglint` plus a lint about a third its size, and the 26 request-path records written from context-less helpers now carry `request_id`; a third pass, `00310f6e`, added the lint rule that refuses that shape from coming back and named the two request-path writers that still carry none, the audit path (#328) and the transaction records. Production log output has been conformant at HEAD throughout. The convention is a property rather than a moment: sloglint runs over every module locally under `./run-tests.sh --type lint` and in CI, and `testutil.AssertSlogConvention` walks every production file from all three module tiers.

### Stage 1: the handler, its config, and the two mains
Landed in `a0fd0642`.

New package `core/logging`. `Install(level, format)` builds a text or JSON handler at the
configured level and makes it the slog default; an unreadable value returns an error naming the
setting, the value and the accepted list, before anything is installed, and both mains exit 1 on
it. Under JSON an error attribute is rewritten to its `%+v` form, because slog's JSON handler
writes `err.Error()` and would otherwise drop every `core/errs` stack. `WrapRequestID` appends
`request_id` from chi's request id on the record's context, so no call site names it; it keeps the
ungrouped base handler and replays the `WithAttrs`/`WithGroup` calls made on it, because slog
qualifies later record attributes by every open group and a re-wrapping version would write
`operation.request_id` instead.

The client-chosen scalar bound (`MaxLoggedField`, `FieldForLog` and friends, #159) moved from
`core/middleware` into `core/logging` with its comments intact: the handler has to apply it to the
id it injects, since chi's `RequestID` adopts an inbound `X-Request-Id` verbatim.

Config: `GOIABADA_<SERVER>_LOG_LEVEL` (default `info`) and `GOIABADA_<SERVER>_LOG_FORMAT` (default
`text`) on both servers, with matching flags, and the two settings at their defaults in the
`goiabada-setup` generator and the seven compose files. `config.load()` became `loadFrom(fs,
args)` so a test can drive it more than once. Two rows per server table on the
environment-variables page.

Tiers: modules 6044 passed / 0 failed; integration on sqlite 1373 passed / 0 failed, which is what
checks that the reordered mains still reach the listener. Six mutations, one per property the new
tests pin, all caught.

Worth knowing: the text prefix changes from `2026/09/11 21:02:06 INFO` to `time=... level=INFO
msg=...`, which is the cost decision 1 accepted when it chose a real handler over
`slog.SetLogLoggerLevel`. One record still escapes the handler by construction: the warning
`config.load` writes about a removed environment variable, which necessarily runs before the
install.

### Stage 2: the test capture helper

`60ab1b87`. `testutil.CaptureSlog(t)` in `src/core/testutil/slog_capture.go` installs the servers'
own handler, `logging.WrapRequestID` over a recording sink, as slog's default for one test and
restores the previous default on cleanup. `Records()` gives each record's level, message and
flattened attributes; `Text()` renders them the way the text handler does.

It replaces 14 private installs of two shapes that could not assert the same things. A text buffer
cannot see a record's level and cannot tell one record from two, so demoting a `Warn` to `Info`, or
splitting a message away from its attributes into a second record, left every assertion green. Four
hand-written recording handlers modelled only what their own suite needed, and three of the four
dropped `WithAttrs` on the floor. Putting the recorder under the servers' own wrapper is what makes
`request_id` testable at all: an assertion on that attribute now exercises the injection production
runs rather than a second copy of it, which is what every later stage's level tests and level pins
are built on.

19 test files moved rather than the 14 that installed a handler, because five more called those
files' package-level helpers. `apiresponse_test.go` gained an assertion in the move: it had parsed
JSON log lines into maps, which flattened the error attribute to text, and now reads it as an error
value, which the old form could not tell from a site logging `err.Error()`.

Tiers: unit green on all three modules (6058 cases) and again under `-race` with no data race; data
green on mysql, postgres, mssql and sqlite. Integration is not applicable, since no handler, route,
schema or production behaviour changed and the one non-test file added is imported only by tests.
Six mutations against the new helper, every one caught, including dropping `WrapRequestID`, which
leaves a helper that captures records perfectly and silently stops testing the injection.

One thing it settles for later: the lint in stage 7 has to exempt `core/testutil` alongside
`core/logging`, the two mains and `core/cmd/schemadump`. That is forced rather than chosen. The
helper has to be a non-test file, since an `_test.go` file is not importable and 19 test files
across three modules import this one, so it necessarily calls `slog.SetDefault` and `slog.New`
where the lint walks.

### Stage 3: the structured loggers: banners, audit, SQL, API debug
Landed in `65e9900b`. The four sites that were writing something other than a record. 441 slog
calls in production code become 334.

`core/auditlog.LogToConsole` is one writer for the audit console record, called by the authserver's
`AuditLogger` and by the email-collision backfill in `core`, which could not share code any higher
up and so carried a hand-copied envelope struct. The record now carries the event name and the
details map as attributes, where before it put a JSON document in the message field and a collector
consuming the log as JSON had to parse `msg` a second time to reach what it was querying on.

The SQL logger no longer writes bound values. It used to write a second record listing every value
bound to the statement, and nothing at that layer can tell a password hash, a TOTP seed or an
encrypted client secret from a page size: every value the product writes to the database passes
through `ExecSql` or `QuerySql`. The environment-variables page now says so.

The API debug logger writes one record per exchange instead of six, and the method passes through
`logging.FieldForLog`, since it is client-chosen and the middleware is mounted ahead of
authentication. The redaction path is untouched and its tests assert the same content.

Eight banner blocks became one record each, 107 slog calls in all. The plan said five, counting the
five files; those files hold eight blocks and the authserver main alone has four. Seven are now
named functions beside the code they came from so the record can be asserted, and the two that name
environment variables share one list, so a credential added to the bootstrap file cannot reach one
list and not the other. The bootstrap file path is read from the configuration rather than printed
as `./bootstrap/bootstrap.env`, which is the shipped compose files' path rather than the
deployment's.

Nine mutations against the new cases, every one caught, including dropping `FieldForLog` from the
method and forcing the SQL logger's flag true. Tiers: unit 1681 pass; the data tier on all four
engines, because the stage edits `src/core/data/`, 1643 pass in 7m10s; integration on sqlite 1038
pass.

One narrowing is named rather than left to be found: a details value slog's JSON encoder cannot
render is now dropped silently where the old code logged a marshal failure. No call site produces
one, and the database half of `AuditLogger` still marshals and still reports the failure.

### Stage 4: the sweep, core
Commit `d5e7618d`.

All 108 `slog` calls in `core` visited once, in the 28 files that write one. 23 files changed;
`core` now writes 90 records, the 18 that went being 17 collapsed out of the startup path and one
deleted with the defect below. Production calls in the tree fall from 334 to 316. `core` now has no
key that is not snake_case, no message built by `fmt.Sprintf` or concatenation, no capitalised first
rune, no `Component:` prefix, and a `*Context` variant at every site with a context or a request in
scope.

The startup path is the visible change. Each engine's constructor wrote `using database x` and then
four `fmt.Sprintf` lines naming the username, host, port and database name one record each, while
`OpenDatabase` wrote the type twice more above them: seven consecutive records a collector had no
way to join, four on sqlite. One structured record per engine now, plus one naming the type, so a
start writes two. 22 call sites become 5. SQLite's carries the dsn actually in use, where the pair
it replaces printed an empty one on every in-memory start, which is the case where a reader most
needs it.

Three defects closed. `refreshToken` logged "http client is nil" and then dereferenced it on the
next line; it now returns an error, which its caller already handles as "clear the session and
continue". `GetLoggedInSubject` concatenated `debug.Stack()` into a log message, unparseable in a
JSON stream; it passes `errs.New` as the error value, which captures the same stack and prints it
under `%+v`. `ConvertToString` wrote the unconvertible **value** under a key named `type`, so a
record headed "unsupported type" put arbitrary data in the log; it writes the type.

One level moved: an invalid JWT issuer drops from Error to Warn. The middleware exists to meet that
condition and meets it by clearing the session and redirecting, so nobody has to act, and an error
log full of handled conditions is not an error log. Every other core level was reviewed against the
rule and kept, and the five the rule most protects are now pinned by a test rather than left to be
argued about again.

The rate limiter's trip takes a context, so a refusal joins the request that caused it. The request
logger and the three 500 writers stop naming `request_id` themselves: the installed handler injects
it, and two cases assert it arrives exactly once. Five new cases in
`core/middleware/slog_convention_test.go` drive the real middleware under chi's `RequestID` and read
the attribute off the captured record, which is the end-to-end measurement that the correlation
reaches production code rather than only the handler's own unit test.

Seven mutations against the new cases, every one caught, including writing `request_id` twice and
restoring the CSRF middleware's variable message. Tiers: unit 1686 pass; the data tier on all four
engines, because the stage edits `src/core/data/`, 1643 pass in 7m13s; integration on sqlite 1373
pass; the race detector clean over `middleware`, `ratelimit` and `handlerhelpers`.

One rename falsified prose inside the product and was corrected in the same pass: a CSRF remedy
sentence told the operator to "Compare the Origin and requestHost fields below", naming an attribute
that no longer exists.

### Stage 5: the sweep, authserver

`41c619cb`. Every `slog` call in the 26 `authserver` files that write one, visited once: a literal
lowercase message with no component prefix, values out of `fmt.Sprintf` and concatenation into
snake_case attributes under the agreed names, and the `*Context` variant wherever a context or a
request is a parameter of an enclosing function. 168 calls down to 152, with nothing deleted.

- **Sixteen startup records become four.** The version, build date and git commit; the three base
  URLs and the API-debug flag; the time zone, local time and UTC time; and the three-record
  encryption-key failure, whose two prose lines were the remedy an operator had to read as a unit
  and which a JSON deployment received as three unrelated records.
- **A listener failure logs the error as a value** instead of as the message, so `%+v` prints the
  frames `err.Error()` was throwing away.
- **The eight listener-configuration lines become one record per listener.** A reader checking why
  HTTPS is off had to join `https enabled: false` to four separate lines to find which of the four
  settings was the empty one.
- **The logout handler's twenty `id_token_hint` rejections** take the gate as an attribute and one
  literal message, so a collector can group and count them by gate.
- **The eleven copies of `auth context is missing, redirecting to %v`** are one literal with the
  destination as an attribute.
- **The tree's one dashed attribute key is gone**, with the two explicit `request_id` attributes
  the 500 writers were naming: the handler injects it.
- **The client-logo check passes its error as a value** rather than through `%v`, which was
  dropping the stack, and names the client it could not check.

Three levels moved, all Error to Warn and all in the logout path: a `client_id` naming no client,
an `id_token_hint` this server cannot decrypt, and the twenty hint-rejection gates. Each is a
request refused and handled — the caller falls back to a hintless logout and the End-User is still
signed out — and left at Error the gates were the largest single source of expected entries in this
server's error log. Twelve levels are pinned where they already were: the nine bearer rejections
and the two 500 writers.

Level is the one rule the lint cannot hold, so each moved level and each pin has a case through
`testutil.CaptureSlog`, and a new test drives a real handler through chi's `RequestID` and asserts
`request_id` reaches the record with nothing in the test naming it — the end-to-end check decision
2 owes.

Tests: unit 1687 pass; data on sqlite 393 pass; integration on sqlite 1038 pass. Eight mutations
against the new cases, all caught.

One gap this stage found and did not close, drafted as a follow-up rather than widened into here:
fourteen records in `authserver` and about a dozen in `core` are written from handler helpers that
take neither a context nor a request, so they run on a request path and carry no `request_id`.
Decision 2's lint rule permits the plain call there by construction, and closing it means changing
signatures in both swept modules rather than log call sites.

### Stage 6: the sweep, adminconsole

`3494d238`. The same pass over the 6 `adminconsole` files that write a record, and with it the tree
is swept: the census over `src` reports **285 calls, all of them a literal lowercase message with
snake_case attributes** — zero built messages, zero capitalised literals, zero component prefixes,
zero non-snake keys. Before this commit those counts were 27, 5, 0 and 1, and every one of them was
in this module. 59 calls down to 44, with nothing deleted.

- **Twenty-two records become seven.** This main gets the four structured startup records the auth
  server got in stage 1 — `build information`, `using configuration`, `process clock`, and one
  session-key refusal in place of a failure plus two lines of capitalised prose naming the variables
  and the `openssl` command. Its `server.go` gets the two listener-configuration records in place of
  eight one-value-per-line Infos, for the same reason the auth server's did.
- **A failed listener logs the error as a value**, not as the message: it arrives from `errs.Errorf`
  and the frames it captured were going nowhere. This is the last of the correctness items the
  agreement listed.
- **The tree's last non-snake key is gone.** The client-logo lookup's `clientId` is `client_id`, the
  numeric row id's agreed name, and its message loses its capital and its `Failed to`.
- **The two remaining call sites writing `request_id` by hand** stop doing so: both settings-cache
  500s take `ErrorContext`, and the handler reads chi's id off the context. Their fixture now mounts
  chi's `RequestID`, so its two existing `request_id` assertions measure that injection rather than
  an attribute the middleware used to write itself.

One level moves: the code exchange on the admin callback fired at Info on **every** administrator
sign-in, which is per-request tracing in the level reserved for lifecycle and configuration. It is
Debug now, with the auth server address as an attribute rather than concatenated into the message.
Level is the one rule no lint can hold, so a case pins it, and cases pin the logo lookup's Warn
(the page still renders, so nobody has to act) and both settings-cache 500s at Error. A handler
driven through chi's `RequestID` asserts the id reaches the record with nothing naming it, which is
this module's end-to-end check of the same decision.

Tests: unit 1690 pass; data on sqlite 393 pass; integration on sqlite 1038 pass. Six mutations
against the new cases, all caught — including one per `*Context` call, which is the only evidence
that `request_id` arrives through the installed handler rather than through anything the test
arranges.

### Stage 7: the lint, and the agents' pattern

> Superseded in part by the review round of #321, below: rules 1, 3 and 4 and the lowercase half of
> rule 2 moved to `sloglint`, and the rule 6 the final review added was reduced to one refusal. The
> account here is what landed at `397cbac5` and what the final review found against it.

`397cbac5`. `testutil.AssertSlogConvention` in `core/testutil/slog_lint.go`, called from
`core/testutil`, `authserver/internal/server` and `adminconsole/internal/server`, walking all 499
production files under `src/`. Five rules, each decidable from the text: the message is a string
literal; it starts with a lowercase letter, carries no `component: ` prefix and does not open with
`failed to` or `error `; every attribute key is snake_case and none of `err`, `request_id` or
`request-id`; a call inside a function taking a `context.Context` or an `*http.Request`, at any
enclosing depth, is a `*Context` variant; and `slog.SetDefault`, `New`, `Default` and `With` are
refused outside `core/logging`, the two mains, `core/cmd/schemadump` and
`core/testutil/slog_capture.go`.

- **Resolution is by import path, not by the text at the call site.** An aliased `log/slog`,
  `context` or `net/http` is seen, and a dot import of `log/slog` is refused at the import, because
  it leaves no selector for any of the five rules to resolve. The walk, the build-constraint
  exemption and the "walked no files" guard are reused from `errors_lint.go` beside it rather than
  copied, so one answer about what production code means holds for both lints.
- **`core/logging` is exempt from the `request_id` refusal, and from nothing else.** It is the
  handler that injects the key from the context, so a rule refusing it everywhere would refuse the
  mechanism the rule exists to protect. Deliberately narrower than rule 5's allowlist: schemadump
  and the two mains install a handler and still have no business writing the key. This was the
  lint's single finding over the whole swept tree, which is the measurement goal 3 asked for.
- **Three shapes stay outside a parse, marked `ceiling:` with what would close them**, and each is
  pinned as a *passed* fixture so a later reader cannot mistake the gap for coverage: an attribute
  key computed at runtime, an elided composite literal inside a `[]slog.Attr`, and the methods on a
  `*slog.Logger` — which rule 5 is what makes safe, since outside the handler's own five paths there
  is no way to obtain one.
- **Pattern 8 in `CLAUDE.md` and `AGENTS.md`** carries the convention, the level rule and the two
  entry points a reader needs before writing a `slog` call. `AGENTS.md` also regains the
  test-fixtures paragraph it had been missing since #272, so `diff AGENTS.md CLAUDE.md` is now
  empty.

Tests: seam 8 writes 30 fixture files into a temp tree and asserts the exact 56 findings with lines,
one accepted and at least one refused per rule, plus the three evasions the census prototype was
written against — `slog.String("badKey", ...)`, a plain call in a closure nested in a handler, and
`(slog.Info)("x")`. Unit 1694 pass across all three tiers; data on sqlite 393 pass; integration not
applicable, no handler, route or schema touched. Eight mutations, one per rule plus the
`request_id` exemption, all caught — the last of them against the real tree rather than the
fixtures, which is the only evidence that half is live rather than vacuous.

### Final review, round 1: the keys a wrapper carries

Reviewed once, at the end, against the whole branch. Two findings, both real, both fixed in
`b650a086`.

- **The sweep had a blind spot, and it was the biggest single logger on the API surface.** 328
  calls reach one `slog.ErrorContext` by handing their attributes to
  `apiresponse.WriteInternalServerError`, `LogInternalServerError` or
  `apihandlers.writeInternalServerError` as a variadic run. The emission sees only `record...`, so
  the lint had no key to read and **179 key literals in 15 files still spelled `clientId`,
  `groupId`, `userId`, `permissionId` and `resourceId`** while all three tier callers passed. The
  tree was writing `client_identifier` and `clientId` in records of the same request, which is the
  one thing this change exists to end. Swept by an AST rewrite rather than a text substitution,
  because every one of those spellings also names a URL parameter or a JSON field in the same files
  — `"userId"` alone has 118 other occurrences in production code.
- **Rule 6 closes it for good.** A function that forwards a caller's variadic attribute run into a
  record is listed in `slogAttrForwarders` with the index its attributes start at; rule 3 then
  reads the keys at its call sites and inside any `[]any` built and spread into it. Four forwarders
  exist, the fourth being the closure that writes the twenty `id_token_hint` refusals. An
  **unlisted** forwarder is refused rather than ignored, which is what makes this a closed rule
  instead of a snapshot: every other variadic `...any` in production hands its run to `fmt` or to a
  SQL driver, and a new one that reaches a record now fails the lint until it is registered.
- **The nil-HTTP-client panic fix had no test.** Every `refreshToken` case either supplied a client
  or returned before one was used, so the log-then-dereference that guard replaced could have been
  restored with the suite green. It now has `TestRefreshToken_NilHTTPClientFailsClosed`, and a
  mutation removing the guard fails it by panicking.

Tests at `b650a086`: unit 1695 pass across all three tiers; data on sqlite 393 pass; integration on
sqlite 1038 pass. Six mutations, all caught — including two against the real tree, one reverting a
swept call-site key and one reverting the single slice built before spreading, which is the
evidence that the 179 keys are held by the lint rather than by this diff.

### Final review, rounds 2 to 5: the boundaries of the rule that closed it

Rule 6 is the rule round 1 added, and the four rounds after it each found the previous round's
closure of it incomplete. Twelve findings, all confirmed, none of them in what ships: production
log output has been conformant at HEAD since `b650a086`, and every one of these is a false green in
the guard that is supposed to keep it that way. They are listed together because they are one
family.

- **Round 2, one finding, `7b10e151`.** Rule 6 stayed green on six ordinary forms carrying a
  camelCase key, and its `ceiling:` claimed something `reportTrip` had contradicted since stage 4.
  The rule stopped enumerating the shapes it had seen and started refusing every run it cannot
  read.
- **Round 3, three findings, `36fb47c5`.** A run's readability was decided by its assignments
  rather than by every use of it; the forwarder table's own metadata was never checked against a
  declaration; and the tier census walked two kinds and collapsed them by name.
- **Round 4, three findings, `11e99fcd`.** A run identified by name alone, so a spread above its
  declaration or outside its block was a different variable; a composite literal's key subtree
  skipped whole, which a pointer to a run can sit in; a listed forwarder free to rewrite its
  caller's run before spreading it; a method inheriting a registration the table could not have
  been describing; and a tier reachable in a map key or in the tail of an overlapping slice view.
- **Round 5, five findings, `c629fb61`, the last round this gate has.** Rule 6 read a run by its
  spelling where Go reads it by scope: a declaration in an `if` or `switch` header is scoped to
  that statement, an append based on the name being declared reads the outer variable, and a file
  declaring `append` of its own had it credited with the builtin's meaning. It also concatenated
  runs that do not pair off against each other, so a replacement inherited its predecessor's parity
  and a forwarder prepending one element shifted every key its call sites wrote onto the value
  beside it. Rule 3 did not read a key spelled through an alias of `slog.Attr`, through a
  constructor bound to a name, or written onto an attribute after it was built. Rule 4 saw the
  request-bearing types under one spelling only, so a function taking an alias of `context.Context`
  would have lost `request_id` with the lint green. And the rate limiter's tier walk counted routes
  rather than tiers, against a test that asserts an instance count.

Each round resolved the family rather than the forms reported: round 5 closed every statement
header rather than the two demonstrated, and every named-type spelling rather than the two
demonstrated, with a near-miss fixture beside each refusal. What a parse of one file cannot decide
is stated in the two `ceiling:` paragraphs rather than left implied — a computed key, an elided
literal, and a type or builtin declared in a sibling file of the same package — and none of
those shapes exists in this tree.

Tests across the four rounds: twenty-six mutations, all caught, one of which found dead code in its
own round's first fix. `where.sh test --type modules` green at each; the four-engine data and
integration result is CI's, green at `11e99fcd` on all thirteen checks.

## How it was verified

**Locally, in the dev container, at each stage and at each review round:** the three module unit
tiers (`run-tests.sh --type modules`, 1697 tests across 67 packages at the end), the data tier on
sqlite (393), and the integration tier on sqlite (1038). Stage 3 ran the data tier on all four
engines, because it edits `src/core/data/`. Sixty-four mutations across the run, each narrowed to
the test that should catch it; all caught, two of them only after the thing they exposed was fixed.
`check-anchors.sh`: 51 of 51 rows resolve, at every stage and every round.

**Only in the check suite:** the four-engine data and integration matrix (sqlite, MySQL,
PostgreSQL, SQL Server), `Unit / race`, `Lint` (vet, unparam, golangci-lint, gofmt per module),
`Vulnerabilities`, `Versions` and `Build validation`. No local tier covers any of those.

**Green on all thirteen jobs at `c629fb61`**, the commit this pull request goes ready on:
[run 34676131644](https://github.com/leodip/goiabada/actions/runs/34676131644). That result was read
rather than assumed, and it supersedes every stage's local sqlite claim.

## Deviations from the plan

Each was recorded in its stage's log when it happened; collected here so they are in one place.

- **Stage 3 — eight banner blocks, not five, and 107 calls removed, not 52.** The five anchored
  files hold eight blocks (the authserver main alone has four), and decision 6's rule is one record
  per block. The 52 is what the census flags as banner/empty/step-list; the other 55 lines inside
  those blocks are capitalised literals and concatenations that cannot survive the collapse.
- **Stage 3 — seven functions extracted so the banner records could be tested.** Inline in `main()`
  or `Start()` they have no seam: both bind or exit. Two of these records are the only place an
  operator is told which environment variables to set.
- **Stage 3 — `bootstrap_file` is now read from the configuration.** The banner printed the shipped
  compose files' path, so an operator who mounted the file elsewhere was sent to a path that did not
  exist.
- **Stage 3 — one behaviour narrowing, named because it is invisible.** The audit console path used
  to log an Error when `json.Marshal` of the envelope failed; under slog's JSON handler an
  unrenderable details value makes `Handle` return an error the logger discards, so the record is
  dropped silently. No call site produces such a value, and `AuditLogger`'s database half still
  marshals and still reports the failure. Recorded in `LogToConsole`'s doc comment as well as here.
- **Stage 4 — `OpenDatabase`'s five records collapse to one.** Three of the five said the same thing
  as the engine constructor's own record one line later. It removes records an operator may be
  grepping for: `creating mysql database` no longer appears anywhere.
- **Stage 5 — `aud`, `iss`, `typ` and `sub` were kept as attribute keys** where the value is the JWT
  claim of that name. Snake-legal single tokens, one key per concept; renaming them would collide
  with the `client_id` parameter the same record already carries.
- **Stages 5 and 6 — the listener-configuration records and the two startup refusals in `server.go`
  have no test.** They are inline in `Start`, which binds real listeners and calls `os.Exit`. The two
  that were extractable are extracted and asserted.
- **Stage 6 — `initialized settings cache with 30s TTL` keeps its number in the message.** The 30
  seconds is `cacheTTL`, unexported in `internal/cache`; exporting it to make the record carry a
  `ttl` attribute is more surface than one startup line is worth.
- **Stage 7 — rule 3 needed an exemption the sketch did not state.** Run against the tree, the
  lint's single finding over 499 files was `core/logging/logging.go:132`,
  `slog.String("request_id", id)` — the injection decision 2 is built on. `core/logging` is exempt
  from the two injected spellings and from nothing else; `err` stays refused in every file.
- **Stage 7 and final round 1 — pattern 8 is 159 words against a stated "sized like pattern 5"
  (139).** It landed at 150 and round 1 added nine words so the lint's scope sentence says it reads
  the keys a wrapper forwards. Its siblings on the page run 139, 147 and 209, so it is not out of
  scale; the number is on the record either way.
- **Bookkeeping:** the agreement's fanout says 59 files and 27 in `core`; it is 60 and 28, because
  stages 1 and 3 added `core/logging` and `core/auditlog`, which write records the fanout's count
  predates.

## For the release notes

**The log output of both servers changes shape.** Anything that greps, parses or alerts on
Goiabada's log lines needs updating; nothing else about the servers' behaviour changes.

- **The line format.** Both servers now install a handler at startup, so a text line reads
  `time=... level=ERROR msg="..." key=value` instead of Go's `log`-style
  `2026/09/11 19:51:21 ERROR ...`.
- **Two new settings per server**, with matching flags:
  `GOIABADA_AUTHSERVER_LOG_LEVEL` and `GOIABADA_ADMINCONSOLE_LOG_LEVEL` (`debug|info|warn|error`,
  default `info`), and `GOIABADA_AUTHSERVER_LOG_FORMAT` and `GOIABADA_ADMINCONSOLE_LOG_FORMAT`
  (`text|json`, default `text`). **An unknown value refuses startup**, as the session-key validators
  do. The defaults reproduce today's level, so an upgrade that sets neither gets the same records at
  the same level in the new format. Under `json` a `core/errs` stack still reaches the log.
- **Messages and attribute keys moved.** Every message is now a lowercase literal with no
  `component:` prefix, and every key is snake_case with one name per concept: `err` became `error`,
  `clientId` became `client_id` (the row id) or `client_identifier` (the wire string), and the same
  split applies to sessions and to keys. A saved search or a dashboard filter on an old spelling
  will match nothing.
- **`request_id` is on every record written on a request path, with two named exceptions**, injected
  by the handler from chi's request id — the same value already returned to the client in
  `X-Request-Id`. Call sites no longer write it themselves. The helpers that had no context to carry
  it were given one in the review round, so it holds for the authorization error path and logout,
  and a lint rule refuses a new helper written without one. The exceptions are the audit records,
  whose `AuditLogger.Log` takes no context (#328), and the transaction and statement records under
  `RunInTransaction`; both are ceilings recorded in pattern 8, not exemptions.
- **Startup banners are single records.** The five multi-line blocks (missing bootstrap credentials
  on either server, the HTTP-without-TLS warning on either server, and the seeder's generated
  credentials) are now one record each, carrying the content as attributes. `creating mysql
  database` and its siblings no longer appear at all.
- **Logging SQL no longer writes bound values**, only the statement, one record per statement.
  Nothing at that layer can tell a password hash from a page size.
- **The audit console record changed shape.** It was a JSON document pasted into the message; it is
  now `audit event` with `event` and a nested `details`, an object under JSON and Go map syntax
  under text. A collector that parsed the message as JSON must read the fields instead.
- **Verbose API logging writes one record per exchange**, not one for the request and one for the
  response. Redaction (#145) is unchanged.

## Review

Reviewed by `openai-codex/gpt-5.6-sol` at effort `xhigh`. Six rounds: one at the plan gate, then
five over the whole branch. No stage was reviewed on its own (`LEO_REVIEW_MODE=end`), so these
rounds are the only reading this code got.

|---|---|---|---|
| Plan gate | 9 | 1 | all resolved in the plan, before any code existed |
| Final 1 | 2 | 1 | all fixed, `b650a086` |
| Final 2 | 1 | 0 | fixed, `7b10e151` |
| Final 3 | 3 | 0 | all fixed, `36fb47c5` |
| Final 4 | 3 | 0 | all fixed, `11e99fcd` |
| Final 5 | 5 | 0 | all fixed, `c629fb61` |

- **blocking, security (plan gate):** automatic `request_id` injection would have bypassed the #159
  128-byte bound and escaping. Fixed in the plan: the bound moved to `core/logging/field.go` and the
  handler applies it, with oversized and non-printable ids pinned under both formats.
- **blocking, conformance (final round 1):** 328 calls hand their attributes to one of three
  `apiresponse` wrappers as a variadic run, so the lint saw no key literal and **179 camelCase keys
  in 15 files survived the sweep** with all three tiers green. Fixed in `b650a086`: an AST rewrite
  of 136 lines, plus rule 6, which reads the keys at a registered forwarder's call sites and
  **refuses an unregistered one**.
- **significant, quality (final rounds 2 to 5, twelve findings):** one family — rule 6's boundaries,
  and then rule 3's and rule 4's. Each round found the previous round's closure incomplete: a run
  read by spelling rather than by scope, a forwarder free to rewrite the run it forwards, a method
  inheriting a registration, a type reached under a second name. None of them was in what ships —
  production log output has been conformant at HEAD since `b650a086`, and each was a false green in
  the guard that keeps it that way. Fixed in `7b10e151`, `36fb47c5`, `11e99fcd` and `c629fb61`.

Full account, with the evidence and every mutation table, in
[this comment](https://github.com/leodip/goiabada/pull/321#issuecomment-5643998719).

## Review round of #321

The reviewer found no production correctness or security blocker and two things to change. Both
were done rather than deferred.

### The lint was replaced by sloglint plus a lint about a third its size

`slog_lint.go` and its test had reached 3,356 lines, 45% of this change's added lines, and the
history showed why: the pair was 1,070 lines when stage 7 landed and rule 6 with its boundaries
added 2,316 across the five review commits, a single-file parse refusing approximations of what a
type checker answers for free. Decision 12 had rejected `sloglint` on three premises, two of which
were wrong, verified in sloglint v0.12.0 as golangci-lint 2.13.1 bundles it: its `context: scope`
mode matches `context.Context` and `*net/http.Request` parameters at any enclosing depth by their
checked type, which is the `*Context` rule exactly, and `run-tests.sh` can run golangci-lint.

- **`.golangci.yml` enables `sloglint`** with `context: scope`, `static-msg`, `msg-style:
  lowercased`, `key-naming-case: snake`, `forbidden-keys` of `err`, `request_id` and `request-id`,
  and the four forwarders under `custom-funcs` so their callers' keys are read: the API's 500
  writer, its logging half, the `apihandlers` wrapper, and `rejectIdTokenHint`. `_test.go` and
  `mocks/` are excluded as the Go lint excludes them, and `core/logging` is exempt from the
  `request_id` refusal at the one line that injects it. Each rule was planted once in a scratch
  file and reported once before the configuration was trusted; the first probe reported nothing
  because golangci-lint keeps one issue per line and `unused` won every one-liner.
- **`testutil.AssertSlogConvention` keeps what sloglint cannot express**: that the message is a literal at the call rather than a constant, which sloglint's `static-msg` admits and its `msg-style` never reads; the refused `component:`
  prefix and `failed to` / `error ` openers; `slog.SetDefault`/`New`/`Default`/`With` outside the
  handler's owners, which sloglint's `no-global` cannot stand in for because it would also refuse
  the package-level calls the tree uses by design; a `...` spread into a record or a forwarder
  outside the six functions `slogSpreadSites` lists, which is the whole of rule 6 in its honest
  one-line form; and that list held equal to `custom-funcs`, so a forwarder cannot be admitted on
  one side only. Aliases, shadowed builtins, type aliases, closures and the three `ceiling:`
  paragraphs are gone: the type checker resolves them.
- **A `lint` tier in `run-tests.sh`** runs golangci-lint over the four modules with the command CI's
  Lint job uses; `all` includes it, `--race` refuses it, and a missing binary fails the tier rather
  than skipping it. Without this the move would have made the convention CI-only, unlike every
  other guard in the tree.
- **Two production shapes changed to make the keys literal at the call.** The `reject` closure in
  `classifyIdTokenHint` became the package function `rejectIdTokenHint`, since a closure has no
  name for `custom-funcs` to register, and the one conditionally built run in the web origins
  handler became two calls.

### Context threaded through the request-path helpers

The closing pass had counted 26 records written from helpers with neither a `context.Context` nor
an `*http.Request` parameter, uncorrelated by construction, and drafted an issue. The reviewer
called them not exotic, and the user chose to fold the fix in. Every listed helper now takes a
`context.Context` first: `getClientDisplayInfo`, `decryptIDTokenHint`, `clientForPostLogoutRedirect`,
`clientProvenance`, `redirectWillBeEmitted`, `checkRedirectURIEmittable`, `revokeOnAuthCodeReuse`
and `revokeAndAuditAuthCodeReuse`, `RevokeUserAuthState` and `RevokeUserAuthStateTx`,
`validateIdTokenHint`, `ValidateClientAndRedirectURI`, and both `TokenParser` methods, whose
interfaces changed in `core/middleware`, `core/validators` and both servers' `interfaces.go`, with
the two mocks regenerated. `refreshJwks` carries it into `http.NewRequestWithContext`. Three plain
calls stay and are not request-path records: `parseCIDRs` runs at middleware construction, and
`ConvertToString` and `addUrlParam` are template functions, which take no context. sloglint's
`context: scope` now demands the `*Context` variant in every one of the request-path helpers above.

Tests: unit 1699 pass across all three tiers with the new lint's rule table (24 refusals over 18
fixtures, six forwarder-table cases, the table held to the tree), data and integration on sqlite
green, golangci-lint zero issues on all four modules.

### The rule that keeps it closed

`00310f6e`, after the reviewer's third pass. sloglint's `context: scope` demands the `*Context`
variant only where a `context.Context` or `*http.Request` parameter exists, so a helper written
tomorrow with neither is admitted by it, and the shape `65f827c9` fixed could come back with every
guard green. `AssertSlogConvention` rule 5 refuses a plain `slog.Debug`/`Info`/`Warn`/`Error`
anywhere under the twelve request-path directories `slogRequestPathDirs` enumerates (both servers'
`handlers` and `middleware`, `apiresponse`, `apiclient`, and core's `middleware`, `validators`,
`oauth`, `oauthdb`, `handlerhelpers` and `sessionstore`), outside the two functions `slogPlainSites`
names with the reason no context reaches them: `parseCIDRs`, at middleware construction, and
`addUrlParam`, a template function lifted out of the func map to a name so the admission is by file
and function rather than by line. Both tables and the directory list are held to the tree by a
test. Three mutations against the real tree are caught: the closure restored, a new context-less
helper in `handlers`, and `parseCIDRs` dropped from the table.

Two request-path writers are left out of the list and recorded as ceilings in pattern 8 rather than
admitted. The audit path: `auditlog.LogToConsole` and `AuditLogger.Log` take no context, so the
`audit event` record and the three Error records around it carry no `request_id`. Filed as #328
with the count (126 call sites, 202 mock expectations in 28 files, two interfaces) and the
`LogToConsole` design question, since it also serves the startup email backfill. The transaction
and statement records in `core/data/commondb`, under `RunInTransaction` with no context to reach
them short of one on every `Database` method. The release-notes claim below is qualified to name
both. The lint pair is 1,213 lines.

Tests: rule table 29 refusals over 23 fixtures; core's 43 packages green; the admin console's render
tests green, since they drive the template func map; golangci-lint zero issues on core.

## The three closing comments

- [Deferred decisions](https://github.com/leodip/goiabada/pull/321#issuecomment-5643967842) — one,
  forced, plus the related exemption that was derived rather than decided.
- [Follow-ups, fold-ins and ceilings](https://github.com/leodip/goiabada/pull/321#issuecomment-5643981188)
  — one issue filed (#328), two fold-ins, five ceilings with their revisit conditions.
- [The review account](https://github.com/leodip/goiabada/pull/321#issuecomment-5643998719) — all
  six rounds, every finding, every mutation.


leodip:
## Deferred decisions

One decision was taken by the run without asking, and this is it. Nothing else was deferred: every
other question in the agreement was settled at the interview, and no decision was escalated to the
thread during the run.

### 1. Rule 5's allowlist admits `core/testutil/slog_capture.go` — **resolved, forced**

Raised at the plan gate (plan review round 1, finding 7).

**The question.** Decision 11 makes `CaptureSlog` a non-test file in `core/testutil` that calls
`slog.SetDefault` and `slog.New`, because an `_test.go` file is not importable and 19 test files
across the three modules import it. Decision 12 and section 4's rule 5 allow those calls only in
`core/logging`, the two mains and `core/cmd/schemadump`. So the lint as written refuses a file the
same agreement requires.

**The assumption taken.** The allowlist gains that one file, **by path rather than by package**,
with a fixture showing that a second file in `core/testutil` is still refused.

**Why it was not escalated.** Nothing is chosen: neither decision can hold without it, and exactly
one shape satisfies both.

**What reversing it costs.** One string in a list and one fixture. If you would rather the helper
lived elsewhere, moving it means moving the 19 importing suites with it; the lint rule itself does
not change.

One related exemption was derived rather than decided, and is recorded here so it is not a surprise:
stage 7's rule 3 exempts `core/logging` from writing the two injected spellings of `request_id`,
because the injection at `logging.go:132` is the mechanism the rule exists to protect. The exemption
is narrower than rule 5's allowlist — schemadump and the two mains own a handler install and still
may not write that key — and `err` stays refused in every file without exception.

leodip:

Everything this run found beside the work, in three blocks: what was filed, what was annexed into
this change instead, and what was deliberately left unreached with the condition that would make it
worth reaching. Rewritten after the review of #321: the issue this comment first drafted, on the
context-less helpers, was folded in rather than filed and sits in block 2 now, and the two
`ceiling:` markers it listed were on the lint that `ba22b6d1` replaced and left with it.

---

## 1. Filed: one issue

### [#328](https://github.com/leodip/goiabada/issues/328) — Audit records carry no `request_id`, because `AuditLogger.Log` takes no context

Every record written on a request path now carries `request_id`, except the audit path's:
`auditlog.LogToConsole` writes the `audit event` record for every security event, `AuditLogger.Log`
wraps it with three Error records of its own, and neither takes a context. The record that exists to
explain a refusal is therefore the one an operator filtering on a reported `request_id` does not
find. Deferred from this PR because it is an observability improvement rather than a defect, and
because the change is wide for a review-heavy PR: 126 production call sites, 202 mock expectations
across 28 test files, two interfaces and two mocks. It also carries one design question, since
`LogToConsole` serves both the request-driven audit logger and the startup email backfill, which
has no request. The issue names the directories left out of the lint's request-path list for it.
Duplicate search at filing (`gh issue list --state all --search "audit request_id context"`, and
again for `audit log request id`): nothing open or closed but #320 itself. #212, the audit logger's
per-event settings read, touches the same function and is linked as related.

---

## 2. Folded in rather than filed: two

### Request-path log records written from context-less helpers carry no `request_id` — `65f827c9` and `00310f6e` (review round of #321)

This comment first drafted it as the issue to file, with the command that files it. The reviewer of
#321 called the records not exotic, naming the authorization endpoint's error path and RP-initiated
logout, and the user's standing rule is to fold in rather than file. Two commits closed it:

- `65f827c9` threads a `context.Context` through every listed helper and its callers pass
  `r.Context()`: `getClientDisplayInfo`, `decryptIDTokenHint`, `clientForPostLogoutRedirect`,
  `clientProvenance`, `redirectWillBeEmitted`, `checkRedirectURIEmittable`, `revokeOnAuthCodeReuse`,
  `revokeAndAuditAuthCodeReuse`, `RevokeUserAuthState`, `RevokeUserAuthStateTx`,
  `validateIdTokenHint`, `ValidateClientAndRedirectURI`, and both `TokenParser` methods, whose
  interfaces changed in `core/middleware`, `core/validators` and both servers' `interfaces.go`, with
  the two mocks regenerated. `refreshJwks` carries it into `http.NewRequestWithContext`. sloglint's
  `context: scope` then demands the `*Context` variant inside each of them.
- `00310f6e` is the "tighten the lint" half the draft asked for. sloglint admits a helper written
  with neither parameter, so it cannot refuse the shape coming back; `AssertSlogConvention` rule 5
  refuses a plain `slog.Debug`/`Info`/`Warn`/`Error` anywhere under the twelve request-path
  directories `slogRequestPathDirs` enumerates, outside the two functions `slogPlainSites` names
  with the reason no context reaches them (`parseCIDRs` at middleware construction, and the template
  function `addUrlParam`, lifted out of the func map to a name so the admission is by file and
  function rather than by line). Three mutations against the real tree are caught: the closure
  restored, a new context-less helper in `handlers`, and `parseCIDRs` dropped from the table.

**Why it fit here rather than in an issue of its own:** the reviewer's finding was that the PR's
own claim, `request_id` on every request-path record, was false while these stood, so the fix
belongs with the claim. Three plain calls the draft listed were misclassified and stay:
`parseCIDRs` runs at construction, and `ConvertToString` and `addUrlParam` are template functions.

### The API debug logger wrote the request method unbounded — `65e9900b` (stage 3)

Found by the plan reviewer, round 1. `api_debug_middleware.go:debugLog` formatted `r.Method`
verbatim into its first record, and both API groups mount the debug middleware ahead of the
authentication guards, so with verbose API logging on, one unauthenticated request could write a
request-line-sized method to the log. The request logger already bounds the same field through
`fieldForLog` (#159).

**Why it fit here rather than in an issue of its own:** stage 3 was already rewriting `debugLog`
into one record per exchange under decision 9, so the fix is one call to `logging.FieldForLog` in a
function the stage's diff already owned, with an oversized-method case added at seam 7. Nothing was
left to settle and the agreement's §2 stayed true. `gh issue list --state all --search "method log
debug API"` found nothing open or closed.

---

## 3. Ceilings left by this change

Each is recorded where it is decided, in the code or in `CLAUDE.md` pattern 8, with the condition
that would make it worth closing.

- **The audit path carries no `request_id`** — #328 above. Named in the comment on
  `slogRequestPathDirs` in `src/core/testutil/slog_lint.go` as a directory left out of rule 5 for
  it, and in pattern 8. Closes with the issue.
- **The transaction and statement records carry no `request_id`.** `beginning`, `committing` and
  `rolling back transaction` and the `sql` record under `LOG_SQL`, the deadlock-rerun Warn and the
  rollback-failure Warn in `core/data/commondb/db.go` all run under `RunInTransaction` on request
  paths with no context to reach them. The statement log cannot take one short of a context on
  every `Database` method; `RunInTransaction` alone could, at 21 callers, four engines and one mock,
  and only the deadlock Warn would gain. Recorded in pattern 8. **Revisit when a deadlock rerun has
  to be joined to a request in an incident**, and then thread `RunInTransaction` only.
- **Template functions write with no context.** `addUrlParam` in `core/handlerhelpers` and
  `stringutil.ConvertToString` each write one Warn during rendering, and `html/template` calls a
  function with no context to carry `request_id` on. The first is admitted by name in
  `slogPlainSites`; the second sits outside the listed directories and is named in the comment on
  them. **Revisit if either site moves into a handler**, where a context is in scope.
- **Rule 4 compares forwarders by name only.** `msg-pos` and `args-pos` in `.golangci.yml` are not
  read back against the four forwarders' signatures, so a forwarder that grows a parameter with its
  offset left stale would have sloglint read its callers' keys one position out while the rule
  stays green. Documented in the header of `slog_lint.go` with the check that was done instead: a
  camelCase key planted at a call site of each and reported. **Revisit when a fifth forwarder is
  registered or a forwarder's signature changes.**
- **One runtime attribute key.** `reportTrip` in the rate limiter appends `t.keyField`, whose value
  is a literal at each `newTier` call, and sloglint's `key-naming-case` reads literals only. Pinned
  by `TestRateLimiter_EveryTierLogsUnderAConventionalKey` beside it, and `reportTrip` is listed in
  `slogSpreadSites` for the reason. **Revisit when a second site needs a computed key**, since one
  test per site does not scale.

leodip:
## The review account

This is the code review, pasted in full rather than linked, because the artifacts it came from
(`.review/`) are gitignored and overwritten by the next run. `LEO_REVIEW_MODE=end`: no stage was
reviewed on its own, so the six rounds below are the only reading this code got.

Reviewer throughout: `openai-codex/gpt-5.6-sol`, effort `xhigh`. One round at the plan gate, then
five at the final gate against the whole branch (`git diff origin/main...HEAD`: 112 files, 4779
insertions, 1395 deletions at the time of round 1). Every axis — correctness, conformance, security
— was reviewed in every round.

**Totals: 23 findings, all confirmed, none refuted, all resolved.** Nine at the plan gate, resolved
in the plan before any code existed. Fourteen at the final gate: one blocking (round 1, finding 2),
thirteen significant, resolved in `b650a086`, `7b10e151`, `36fb47c5`, `11e99fcd` and `c629fb61`.
Only round 1 changed what ships; every finding after it was a false green in the guard, and
production log output has been conformant at HEAD since `b650a086`. **Sixty-four mutations were run
across the rounds; all were caught, two of them only after the thing they exposed was fixed.**

---

## Plan gate, round 1 — nine findings, all fixed in the plan

Request `320-20260911T201510Z`, fresh session. `unchecked`: no implementation diff exists; the 441
per-site rewrites; `CaptureSlog`'s concurrency, `LogValuer`, duplicate-key and group behaviour; no
engine or browser pass. All four answered in `log/plan.md`.

1. **Automatic `request_id` injection bypasses the #159 128-byte bound and escaping.** Security,
   blocking. Confirmed against `middleware_request_logger.go:fieldForLog` and `security.mdx:72`.
   Fixed in the plan: the bound moves to `core/logging/field.go` and the handler applies it;
   oversized and non-printable ids pinned under both formats and through chi's `RequestID`.
2. **`WithGroup` re-wrap nests `request_id` inside the group.** Conformance, significant. Confirmed
   from the `slog.Handler` contract. Fixed: the wrapper keeps the ungrouped delegate and replays
   recorded `WithAttrs`/`WithGroup` calls after adding the id.
3. **The config test panics on a second `load()`.** Quality, significant. Confirmed at
   `config.go:201-251`. Fixed: a `loadFrom(fs, args)` seam, `load()` unchanged for the servers.
4. **The four flags have no behavioural case.** Quality, significant. Fixed through the same seam,
   one case per flag plus a cross-wire case.
5. **The lint prototype misses aliased imports, `Log`/`LogAttrs`, dot imports and function values.**
   Quality, significant. Confirmed at `probe/census/main.go:32-84`. Fixed in the stage 7 sketch:
   imports resolved by path as `errors_lint.go` does, each form fixtured.
6. **The invalid-input `Install` test cannot observe `SetDefault`.** Quality, significant. Fixed:
   `Install` exercised directly against a sentinel default.
7. **Rule 5 refuses `slog_capture.go`.** Conformance, significant. Confirmed. Resolved as forced —
   see the deferred-decisions comment.
8. **Message checks narrower than decision 4.** Conformance, significant. Fixed: the first rune must
   exist and be lowercase, and the prefix rule is one token then colon-space.
9. **`slog.Attr` literals and nested constructors not read for keys.** Quality, significant.
   Confirmed at `probe/census/main.go:108`. Fixed in the stage 7 sketch, with runtime-computed keys
   marked as the `ceiling:`.

One follow-up raised here (the unbounded method in `debugLog`) and folded into stage 3. No second
round: every finding was resolved in the plan and no production code existed to move.

---

## Final gate, round 1 — two findings, both fixed in `b650a086`

Request `320-20260912T013231Z`, against `397cbac5` (CI green there on all thirteen jobs). Both
findings were verified against the code before anything was changed; neither was refuted. Security
reported nothing.

### Finding 2 — conformance, **blocking**: the shared API error writer still emitted camelCase keys

`apiresponse.LogInternalServerError` appends the caller's `attrs...` onto `record` and spreads that
into one `slog.ErrorContext`, so the keys live at the call sites and the lint, which reads
`record...` and finds no literal there, never saw them.

An AST scan over the three wrappers (`WriteInternalServerError`, `LogInternalServerError`,
`apihandlers.writeInternalServerError`) found **328 calls carrying 179 non-snake key literals in 15
files**: `clientId` 48, `groupId` 37, `userId` 28, `permissionId` 23, `resourceId` 14, then
`attributeId`, `resourceIdentifier`, `groupIdentifier`, `clientIdentifier`, `userCount`,
`userGroupId`, `clientPermissionId`, `auditEvent`, `groupIds`, `groupCount`, `builtInIdentifier`
and `groupPermissionId`. The tree was writing `client_identifier` and `clientId` in records of the
same request — exactly what goal 3 exists to end — with `TestSlogConvention` passing from all three
tiers throughout.

**The family, resolved rather than the instance.** The reviewer reported `mutations_run: []`, so the
boundary was established here rather than taken on trust. Two scans settle it: of the **16**
production functions with a variadic `...any` parameter, exactly **four** reach a record (the three
wrappers plus the `reject` closure in `classifyIdTokenHint`, which writes the twenty `id_token_hint`
refusals); the other twelve hand their run to `fmt` or to a SQL driver. Of the **four** `slog`
emissions spreading a slice, the other three were already snake_case.

**The sweep** was an AST rewrite, not a text substitution, and that is load-bearing: every one of
those spellings also names a URL parameter or a JSON field in the same files (`"userId"` alone has
118 other occurrences in production code), so `sed` would have silently rewritten routes and API
contracts. 136 lines moved, insertions equal to deletions. `clientId` carries `client.Id` and became
`client_id`; `clientIdentifier` carries the wire string and became `client_identifier` — decision
3's split applied rather than a mechanical case change.

**The durable guard is rule 6.** `slogAttrForwarders` lists each forwarder with the index its
attributes start at, resolved by import path for a `pkg.Name` call and by file-or-directory scope
for a bare one. Rule 3 then reads the keys at every call site and in a `[]any{...}` built and spread
in the same body. **An unlisted forwarder is refused**, which is what makes it a closed rule rather
than a snapshot of today's four. Nine fixtures: three call forms, the built slice, the registered
closure, an unlisted forwarder, an anonymous one, and three near misses (a run handed to `fmt`, one
handed to a SQL driver, the same identifier called outside its forwarder scope).

### Finding 3 — quality, significant: the nil HTTP client panic fix had no test

Confirmed: of the four tests reaching `refreshToken`, two supply `mockHTTPClient` and two return
before the client is used, so the guard added at stage 4 was unreached and the log-then-dereference
it replaced could have been restored with the suite green.
`TestRefreshToken_NilHTTPClientFailsClosed` now asserts `refreshed == false` with a non-nil error,
passing `nil` as the interface rather than a typed nil (a `(*mockHTTPClient)(nil)` passes the
`== nil` guard and reaches `Do`, which is a different thing to test).

### Mutations, six, all caught

| File | Replacement | Test | Result |
|---|---|---|---|
| `core/middleware/middleware_jwt.go` | `if m.httpClient == nil {` → `if false {` | `TestRefreshToken_NilHTTPClientFailsClosed` | caught, by panic |
| `core/testutil/slog_lint.go` | `if len(call.Args) > fwd.firstAttr {` → `if false {` | rule table | caught |
| `core/testutil/slog_lint.go` | `if slogForwarderDeclared(name, rel) {` → `if true {` | rule table | caught |
| `core/testutil/slog_lint.go` | the `built.Elts` key read → `_ = built.Elts` | rule table | caught |
| `apihandlers/handler_api_clients.go` | `[]any{"client_id",` → `[]any{"clientId",` | `TestSlogConvention_TheTreeItself` | caught |
| `apihandlers/handler_api_users_search.go` | `"user_count"` → `"userCount"` | `TestSlogConvention_TheTreeItself` | caught |

The last two are the ones that matter: they prove the **real tree**, not only the fixtures, is held
against this exact regression.

### Tiers at `b650a086`

unit (three modules) 1695 pass / 0 fail, 46s · data (sqlite) 393 pass / 0 fail, 33s · integration
(sqlite) 1038 pass / 0 fail, 94s · `check-anchors.sh` 51 of 51. Data ran on sqlite alone because
nothing under `src/core/data/` and no migration moved.

`unchecked`, answered: no level/format binary matrix and no external collector (**dismissed** —
nothing in this round changed the handler, the formats or the startup path); no browser pass
(**dismissed** — the 15 files touched are JSON API handlers and only their key spellings moved); the
four engines not rerun locally (**covered** by the check suite at `b650a086`).

---

## Final gate, round 2 — one finding, fixed in `7b10e151`

Request `320-20260912T022326Z`, resumed session. Conformance and security reported nothing, and the
round 1 repairs were confirmed sound (165 forwarded calls carrying attributes, no camelCase or
hyphenated key left; the 137 repaired API lines changed only a key string).

**Finding 1 — quality, significant: rule 6 was not the closed guard its own comment claimed.**
Stands on all six demonstrated forms plus a seventh found while fixing:

- a dot import of the forwarder's package left the call with no selector to resolve, so the key was
  read by nothing;
- a forwarder assigned to a function value and called through it matched neither the selector nor
  the ident arm;
- `var attrs = []any{...}` was a `*ast.ValueSpec` the assignment pass never visited;
  `append([]any{"clientId", 1}, ...)` had its base at `Args[0]` never read; `other := attrs` and
  `attrs := buildAttrs()` were neither a composite nor an append and were silently skipped;
- the `ceiling:` paragraph read "nothing in this tree writes a computed key today" while `reportTrip`
  had written `attrs = append(attrs, t.keyField, key)` since stage 4. That sentence was false when
  it was written.

**Nothing shipped was wrong** — all four spread sites and all 165 forwarded calls were snake_case at
HEAD, confirmed independently by the reviewer. This is a regression-detection gap in the guard that
*is* goal 5, which is why it was fixed here rather than filed.

**Closed the way the reviewer said it had to be.** Its `why_exhaustive` was explicit: arbitrary Go
data flow cannot be exhausted by adding examples to a parse-only walker, so the family closes only
with a conservative permitted set or a move to type-aware data flow. The conservative set landed, on
both boundaries, as refusals: a dot import of any registered forwarder's package is refused; a
forwarder named anywhere but at a call is refused, qualified and bare; and a run spread into a
record must be the enclosing function's own variadic parameter or a local whose every assignment is
an `[]any` composite, a `make` of one, or an `append` onto itself. A builder's return, an alias, an
append onto a different slice, and a write from inside a closure are each reported as unreadable.
**The tree passed unchanged** — all four production spread sites are readable under the new rule.

**The computed key, pinned where the lint cannot reach.**
`TestRateLimiter_EveryTierLogsUnderAConventionalKey` walks `RateLimiterMiddleware` by reflection for
every `tier` inside it and holds each `keyField` to `^[a-z][a-z0-9_]*$` and to the reserved-key list.
It finds 13 tiers and asserts the count, so a walk that stopped matching cannot pass on an empty map.

**Ten mutations, all caught** (`where.sh mutate --type core`), one per behaviour the fix claims. Two
earlier attempts reported `MUTATION DID NOT COMPILE` (an `if false` inside a `range` left the loop
variable unused) and were re-run as condition flips at the use site. The rule table grew 12 fixtures
and asserts the exact set: 74 findings over 46 files, up from 64 over 35.

Tiers: `--type modules` 6091 passed / 0 failed, `--type data --db sqlite` 534 passed / 0 failed,
`check-anchors.sh` 51 of 51. Integration not applicable: no handler, route, template or schema moved.

---

## Final gate, round 3 — three findings, all fixed in `36fb47c5`

Request `320-20260912T030016Z`. All three quality/significant, all `judgement: false`, so all three
are claims the code settles rather than choices. Conformance and security reported nothing; the
round's own note records that "current production log output remains conformant and no security
defect was found". All three are about the round 2 repair, which was twenty minutes old.

**Finding 1: readability was decided by assignments, not by uses.** Five forms, all confirmed: an
assignment *after* the spread was read as the construction; an inner-block variable of the same name
supplied a readable build for a spread of something else; `attrs[0] = "clientId"` was an
`*ast.IndexExpr` and simply not seen; nothing modelled the run being handed to a helper; and a
listed forwarder shadowing its own variadic parameter had its keys handed to rule 6's declaration
half. **Fixed by making the boundary a rule instead of an enumeration**: a run must be introduced
exactly once in the body it is spread from, and every occurrence of that name — closures included —
must be the introduction, an assignment whose right side the rule reads, an `append` base onto
itself, or the spread. Three positions are skipped because an identifier there cannot be the
variable. `slogRun.built` became `slogRun.decls`, a count, because one is the only answer that
leaves the name unambiguous. One form the reviewer did not name was found while fixing and closed:
a `[]any` **parameter** with an inner block declaring the same name. Seven fixtures, one per form.

**Finding 2: the reflective tier census collapsed and under-reached.** `collectTierKeyFields`
handled exactly `Pointer` and `Struct`, so `Array`, `Slice`, `Map` and `Interface` were walked past,
and a `map[string]string` keyed by tier name meant two tiers of one name were one entry while `len`
stayed 13. Fixed: `[]foundTier` with the path each was reached by, all four kinds traversed, bounded
by a `seen` set (needed, because `ratelimit`'s values point back at themselves). **The guard needed a
guard**: every production tier sits behind a pointer or a struct field, so the test beside it would
pass with `Slice` removed. `TestCollectTierKeyFields_ReachesEveryContainerKind` walks a
purpose-built value holding six tiers across all six kinds, two sharing a name, plus a
self-reference that does not terminate without `seen`.

**Finding 3: the table's metadata was never checked against a declaration.**
`slogForwarderDeclared` matched on name and scope alone, so a second function or closure of a listed
name inherited the registration; and `firstAttr` was never compared to anything, so a registered
signature that grew a parameter would leave every key read one position out. Fixed by
`slogForwarderCensus`: a listed forwarder whose keys were read must have been declared **exactly
once**, **at the argument index the table gives**. Zero declarations beside at least one call is
refused too. Methods are not counted, because a receiver call is a selector.

**Thirteen mutations, all caught.** M13 is recorded as it happened: it **survived** first, which is
how the fixture behind it was found to be testing the occurrence scan rather than the check it was
written for; the fixture was rewritten and the mutation is caught.

|---|---|---|---|
| M1 | `slog_lint.go` | occurrence scan, `tracked && !used[node]` → `tracked && false` | caught |
| M2 | `slog_lint.go` | the `run.decls != 1` / `!run.readable` refusal → `if false` | caught |
| M3 | `slog_lint.go` | variadic shadow, `if run.decls > 0` → `if false` | caught |
| M4 | `slog_lint.go` | `case len(declared) > 1:` → `case false:` | caught |
| M5 | `slog_lint.go` | the `firstAttr` mismatch case → `case false:` | caught |
| M6 | `slog_lint.go` | `case len(declared) == 0 && len(called) > 0:` → `case false:` | caught |
| M7 | `middleware_ratelimiter_test.go` | `case reflect.Slice:` → `case reflect.Chan:` | caught |
| M8 | `middleware_ratelimiter_test.go` | `case reflect.Array:` → `case reflect.Chan:` | caught |
| M9 | `middleware_ratelimiter_test.go` | map range, `iter.Next()` → `false` | caught |
| M10 | `middleware_ratelimiter_test.go` | interface arm, `if !v.IsNil()` → `if false` | caught |
| M11 | `middleware_ratelimiter.go` | the `activate` tier renamed `pwd_ip`, key `clientId` | caught |
| M12 | `middleware_ratelimiter.go` | the `dcr` tier's key → `clientId` | caught |
| M13 | `slog_lint.go` | the `signature[site.name]` refusal → `if false && ...` | survived, then caught |

M11 is the reviewer's own mutation 1 for finding 2, which it recorded as passing against the old
census. It fails now.

Tiers: modules 1697 passing across 67 packages / 0 failures · data (sqlite) 393 / 0 · integration
(sqlite) 1038 / 0 · `check-anchors.sh` 51 of 51.

---

## Final gate, round 4 — three findings, all fixed in `11e99fcd`, plus two more of the same family

Request `320-20260912T035434Z`. Three findings, all quality/significant, all `high` confidence, all
in the guard. The reviewer's own frame: "Current production output remains conformant and no
security defect was found." Its AST census of all registered forwarder call sites found 165
attribute-bearing calls and no bad literal key, and CI was green at `36fb47c5`.

Every one of the six probes behind the three findings was reproduced here as a fixture before
anything was changed, and all three expected lines were missing from the finding set.

**Finding 1, `slogSpreadRunViolations`.** Three escapes under one claim — that round 3's every-use
repair still reads a run by spelling alone. *Use before the declaration*: `decls` counted
declarations anywhere in the body, so a spread above the one declaration was answered for by a
variable not yet declared; fixed by recording `declAt`. *The composite-literal key*: `scan` skipped
a `KeyValueExpr` key subtree whole, on the grounds that a slice is not comparable — true of a slice
and false of a pointer to one, so `map[*[]any]bool{&attrs: true}` stepped over the run; fixed by
scanning any key that is not a bare identifier. *The forwarder's own run*: `site.name == variadic`
returned without consulting `readable`, so a listed forwarder could rewrite its caller's run by
index or append and stay green — in exactly the shape `LogInternalServerError` has; fixed by
tracking the variadic from the start and refusing any write or unmodelled use once bound.

**Finding 2, `slogForwarderDeclarationViolations`.** Every `FuncDecl` name went through a scope-only
registration test while the census excludes `FuncDecl` nodes with a receiver, so a **method**
carrying a listed name inherited the registration, was ignored by the census, and had its call sites
read by nothing — the one arrangement where every rule is green and a caller's keys reach a record
unread. Fixed by reporting a method whatever the table says, with its own message: an entry resolves
a bare name in a scope or a package path through an import, and a receiver is neither.

**Finding 3, `collectTierKeyFields`.** The `Map` arm walked `iter.Value()` only, so a tier held in a
key was reached by nothing, and `visitedValue` identified a slice by its header, so two views of one
backing array collapsed and the tail only the longer one holds was never walked. Fixed: both halves
of a map entry traversed, and a slice bounded per element (base pointer + index × stride against the
element type), which terminates a cycle without suppressing an unvisited tail.

**Two more of the same family, found here rather than reported**, resolving the family rather than
the instance: a **sibling block's declaration** (one declaration, above the spread, and still not the
run that reaches the record, because its block has closed — `slogScopeEnd` now bounds a declaration
by the innermost block, case clause or comm clause); and **an element appended after an
unknown-length tail** (once a run absorbs the caller's, nothing can say whether the next literal is
a key or a value, so a run that grows after absorbing a tail is refused).

**And one piece of dead code, found by its own mutation.** The binding of a variadic spread into a
second listed forwarder was written under `!slogIsRecordCall`, which already answers true for a
listed forwarder: the branch was unreachable, and the mutation that should have caught its removal
survived. It is deleted, the binding happens on the path that was always doing it, and the re-run
mutation is caught.

**Ten mutations, all caught** (`--type core`): the site-above-declaration test deleted; the
`slogScopeEnd` test deleted; the composite-key label test → `if false`; the variadic-integrity
condition → `false`; `if decl.Recv != nil` → `if false`; the unknown-length-tail test deleted; the
`len`/`cap` names changed to ones that never match; the variadic binding at a record → `if false`
(**caught after the dead branch was removed; survived before it**); `iter.Key()` → a discard; the
element stride → a constant zero.

Tiers at `production a2705c397913`: modules 1697 passed / 0 failed, 46s; data (sqlite) 393 / 0, 33s;
`check-anchors.sh` 51 of 51; gofmt clean. **What this round did not change:** no production log
record, message, level, attribute key, handler, configuration or documentation.

---

## Final gate, round 5 — five findings, all fixed in `c629fb61`. The last round this gate has

Request `320-20260912T043816Z`. Five findings, every one quality/significant, every one in the
regression guards: four false-green families in the lint and one false-counting family in the rate
limiter's reflection walk. Each was reproduced before it was believed, in a throwaway fixture tree
run through `findSlogViolations`, and nine of the nine forms the findings name came back green on a
lint that should have refused them. Nothing was contested.

1. **Rule 6 read a run by spelling where Go reads it by scope.** A declaration in an `if` or
   `switch` initializer is scoped to that statement, and no node containing it says so, so reading
   only blocks and clauses handed it the rest of the enclosing body. A short declaration whose
   append is based on its own name reads the *outer* variable, since the name on the left is not in
   scope until the statement ends. And a file declaring `append` of its own had that declaration
   credited with the builtin's meaning. Fixed in `slogScopeEnd` (all five statement headers, `for`,
   `range` and type switches included, rather than only the two demonstrated), in
   `slogReadAppendBase`, and in a new `slogShadowedBuiltins` refusing every run in a file that
   declares `append`, `make`, `len` or `cap`. The tree declares none of the four.
2. **Rule 6 concatenated runs that do not pair off against each other.** Every assignment into a run
   went onto one list, so a replacement inherited its predecessor's parity and a one-element
   predecessor made the replacement's first key a value. And nothing required a variadic tail to
   begin at an attribute boundary, so a forwarder prepending one element shifted every key its call
   sites wrote onto the value beside it, with the call sites innocent and the table's offset stale.
   Fixed by giving `slogRun` **segments** rather than one list, and by refusing the absorption of a
   variadic run onto an odd prefix, reported where the misalignment is created.
3. **Rule 3 did not read three literal attribute keys**: an alias of `slog.Attr`, a constructor
   bound to a name and called through it, and a key written onto an attribute after it was built.
   Fixed by resolving the file's own type declarations, by refusing an attribute constructor used as
   a value (exactly as rule 5 refuses an install), and by `slogAttrKeyWriteViolations`, which
   follows a value this file built as a `slog.Attr` and reads the key written onto it — read rather
   than refused by spelling, because `Key` is an ordinary field name and four production sites write
   it on a model attribute that reaches no record.
4. **Rule 4 saw the request-bearing types under one spelling.** A function taking an alias of
   `context.Context` or of `http.Request` holds the same request and got no `*Context` requirement,
   so its record would lose `request_id` with the lint green. Fixed through the same resolution,
   widened past the two demonstrated spellings: a defined type over either carries the same request,
   a chain of declarations is followed, one leading star is stripped, and whether a star was written
   is no longer read, since a request passed by value carries a request too.
5. **The tier walk counted routes, not tiers.** Confirmed with a probe: two tiers reached both
   directly and through a pointer were reported as four, and the permanent test asserts an instance
   count, so an alias of an existing tier would have failed it without any new tier. Fixed by
   marking the tier by address under an identity separate from the marks on the containers reached
   along the way (the two collide, since a `*tier` and its tier share an address); the walk test
   gained both aliases and still expects nine entries for nine tiers.

**The family, and the boundary this time.** Rounds 1 to 5 are one family — rule 6's, and now rule
3's and 4's, boundaries — and each round's `why_exhaustive` was contradicted by the round after it.
So round 5 resolved the *members* rather than the forms reported: every statement header rather than
the two shown, every named-type spelling rather than the two shown, with a near-miss row beside each
(a conformant replacement whose branches are both read, an even prefix whose callers are still
answered for, a model attribute's `.Key` write that is not read at all). What is left is stated in
the `ceiling:` paragraphs rather than filed: a type alias or a builtin declared in another file of
the same package is outside a parse of one file, and neither shape exists in this tree.

**Tests:** twelve fixtures added to the rule table, one per refused form plus the near misses, and
two more trees for the forwarder parity and its aligned twin (a forwarder cannot be declared two
ways at once). **Nine mutations, all caught**: the statement headers dropped from `slogScopeEnd`;
the declaring case dropped from the append base; `slogShadowedBuiltins` returning nothing; a
replacement appended instead of starting a segment; the parity check made unreachable; the
constructor-as-a-value case made unreachable; the `Key` field name made unmatchable; the alias
lookup made unreachable; and the tier's addressable identity made unreachable. Each applied through
`where.sh mutate`, which restores from a trap and compares by hash; leftovers checked by grep and by
`tree-hash.sh --split`.

**Tiers:** `--type modules` green, all three module tiers including `TestSlogConvention` from each,
~100s. Data and integration not applicable: nothing outside `core/testutil` and one `_test.go` in
`core/middleware` moved, and `data-tests` and `integration-tests` hashes are unmoved since stage 3.

**Round budget: five of five spent.** This gate has no round 6. A finding arriving after this is a
follow-up, not another round.

---

## What the reviewer never checked, across all six rounds

Two entries recurred and were answered every time:

- **No manual browser or external log-collector session.** Dismissed. This change renders nothing,
  and section 5 rejected asserting on stderr from a running server precisely because the handler is
  observable at seam 1 with the same bytes and no process — including the `errs` stack under both
  text and JSON. Every record assertion on the branch goes through `testutil.CaptureSlog`.
- **The four engines were not rerun locally.** Covered, not dismissed: the four-engine data and
  integration result is the check suite's, read rather than assumed at each round's commit and again
  before this pull request went ready. No query, schema or migration semantics moved on this branch
  after stage 3.

Issue #252, opened 2026-08-22
Documentation: the ceremony's state machine, and the two decisions the #239 review closed


leodip:
Found by the #239 authorization ceremony review (deliverables A1 and B1, and findings [F23](https://github.com/leodip/goiabada/issues/239#user-content-f23) and C4). No code change.

Four things that review established which are recorded nowhere durable. The review document itself lives under a gitignored `docs/` directory, so it cannot be cited (that is exactly the problem #187 records), which is why these need a home in the tree.

## 1. `CLAUDE.md` gains the full state transition table

`CLAUDE.md` is the only file that describes the ceremony state machine, and it describes it incompletely: it lists the states and the handlers but not the legal transitions between them, nor which states each handler accepts. The review built the complete table by reading every handler, and it should live next to the description that is already there.

It needs to cover, per transition: the source state, the handler that performs it, the target state, and the condition. Including the branches that are easy to miss, the `AuthStateLevel1ExistingSession` shortcut and the `level2_optional` skip.

## 2. `CLAUDE.md` gains the `AuthContext` field rule

The review's field-by-field audit produced one rule that explains several defects at once and is worth stating as a rule rather than rediscovering:

> A field that accumulates and is not recomputed on every hop must be discarded at a restart.

The set of fields that violate it is empty once PR #238 lands, which is precisely why it should be written down: the invariant is about to be true and nothing will keep it true.

Worth recording alongside it: the request-derived fields are all written once, in a single literal in `HandleAuthorizeGet`, and nothing enforces that today. #248 carries a test to pin it.

## 3. `acr-amr.mdx` records the `mfa` decision

RFC 8176 section 2 defines an `mfa` AMR value:

> Multiple-factor authentication [NIST.800-63-2]. When this is present, specific authentication methods used may also be included.

Goiabada does not emit it, and the review recommends **no change**. The reasoning should be recorded so the question is not reopened: `amr` already carries `["pwd", "otp"]`, from which a relying party can determine that two factors were used, and `acr` carries the level explicitly. Adding `mfa` would introduce a third way to express the same fact, with the attendant risk that the three disagree, and RFC 8176 makes no requirement to emit it.

## 4. `acr-amr.mdx` records the step-up `auth_time` behaviour

[...]

Issue #248, opened 2026-08-22
Ceremony robustness and dead code: two dead states, dead middleware, a dead prefill, a nil dereference, a 500 on the Back button, and a double-code window


leodip:
Found by the #239 authorization ceremony review (findings [F9](https://github.com/leodip/goiabada/issues/239#user-content-f9), [F10](https://github.com/leodip/goiabada/issues/239#user-content-f10), [F12](https://github.com/leodip/goiabada/issues/239#user-content-f12), [F13](https://github.com/leodip/goiabada/issues/239#user-content-f13), [F17](https://github.com/leodip/goiabada/issues/239#user-content-f17) and [F19](https://github.com/leodip/goiabada/issues/239#user-content-f19)).

Six small defects in the ceremony that are cheap together and not worth a pull request each. Nothing here is exploitable; several are dead code that reads as a safety net which is not there.

## 1. A state mismatch answers 500 with a stack trace ([F9](https://github.com/leodip/goiabada/issues/239#user-content-f9))

Eleven sites check `authContext.AuthState != requiredState` and answer `httpHelper.InternalServerError` with a wrapped error. **A user pressing the Back button generates an ERROR log line with a full stack trace and a 500 response.** That is a routine browser action producing an alert-worthy log entry.

**Fix:** one helper, `rejectAuthStateMismatch`, mirroring the existing `rejectCeremonyMismatch`: log at warn with the expected and actual states, render `auth_error.html` with the existing `auth_error.ceremony_mismatch` strings (or a new pair if the wording should differ), and answer 400. Keep `InternalServerError` for a context that genuinely fails to unmarshal.

## 2. `AuthStateInitial` is a state no handler accepts ([F10](https://github.com/leodip/goiabada/issues/239#user-content-f10))

`HandleAuthorizeGet` writes `AuthState: oauth.AuthStateInitial` into the context and saves it before validating anything. No handler accepts that state. Two paths then return without clearing it:

- `ValidateClientAndRedirectURI` fails, `renderErrorUi` renders at 200.
- `IsSupportedResponseMode` fails, `renderErrorUi` renders at 400.

The browser is left holding an auth context in a state every `/auth/*` step answers with 500 (see part 1). Harmless in that nothing can be driven with it, untidy in that the next thing the user does in that browser, short of a fresh `/auth/authorize`, is a 500.

[...]

PR #322, opened 2026-09-11
docs(ceremony): state tables, field rule and agent-docs guard for the auth ceremony (#252)

leodip:
Issue #252: document the ceremony's state machine and the two decisions the #239 review closed.

## What this change is for

`CLAUDE.md` and `AGENTS.md` describe the authorization-code ceremony as a numbered list of states that
"transition in this order", with no gates and no transitions, and open by saying the context is stored
in a session cookie, which #266 made false. This change replaces the list with two tables built from
today's handlers (which handler assigns each state and when; which state each route accepts and what
it answers on a mismatch), states the rule for `AuthContext` fields across a restart and names the
one field that breaks it today (`AuthMethods`, #140), and adds a guard so the two agent files cannot
drift from the constants or from each other. On the site, `acr-amr.mdx` gains two sentences saying
`mfa` is not emitted and what `auth_time` means after a step-up, and `authorization-lifecycle.mdx`
stops saying `auth_time` is the code's creation time.

## The rules it establishes

- `CLAUDE.md` and `AGENTS.md` are byte-identical, and every `AuthState*` constant value appears in the
  Auth States section of `CLAUDE.md`, and every state that section names is a constant. Held by
  `TestAgentDocsAreCurrent` in all three module tiers, through `core/testutil.AssertAgentDocs`.
- `auth_time` is the moment the user last entered a credential. A step-up moves it; only silent
  (`prompt=none`) and SSO reuse keep the earlier value. This documents the code as it has behaved
  since #60; the issue body had it inverted.
- `mfa` is not emitted in `amr`: nothing in RFC 8176 or OIDC Core requires it, and `amr` plus `acr`
  already carry the fact.

**Status:** complete. Both stages landed, the final review is clean at round 2, the one blocking
question it raised was answered by the user and applied, and the check suite is green on the final
commit `25367f84`
([run](https://github.com/leodip/goiabada/actions/runs/34687781866), 13 of 13 jobs). Ready for review.

### Stage 1: the state tables, the field rule and the agent-docs guard
`aa0d6fa9`. The `### Auth States (State Machine)` section of `CLAUDE.md` is now two tables rebuilt
from today's handlers: an assigned-by table of 18 rows (every state, the handler that writes it, and
the condition), and an accepted-by table of 12 rows (route, method, states accepted, status on
mismatch). Both dead constants are marked dead with #248 part 2 as the pointer, and marked
differently because they are dead differently: `initial` is written by the composite literal but
accepted by no gate, `level2_otp_completed` is written by nobody. The SSO shortcut and the
`level2_optional` skip are rows rather than a footnote. The intro sentence now says the context lives
in the server-side store keyed by a cookie (#266).

A new `### AuthContext field rule` section states the rule, names the two restart routes, and names
`AuthMethods` as the one field breaking it today, with #140 as the pointer and an explicit note that
the sentence is owed deletion when #140 lands. `AGENTS.md` is a byte-identical copy, which also
landed the "Test fixtures" paragraph it had drifted without.

The guard is `core/testutil.AssertAgentDocs`: the two files identical, reporting the first differing
line; and the section's roster equal to the `AuthState*` constants parsed out of `auth_context.go`
with `go/parser`. The roster rule reads the *first cell* of each row rather than any occurrence in
the section, because the accepted-by table repeats most values in its `Accepts` column and a looser
rule would still pass after a state's own row was deleted. Transitions are deliberately not checked.
`TestAgentDocsAreCurrent` calls it from all three module tiers; `TestAgentDocs_TheRuleTable` holds
the rules against a temp tree, 10 cases, each asserting the finding's text.

Unit green, 5981 ok 0 fail. Data and integration not applicable: no handler, route, query or
migration moved. Both guard rules mutation-tested and caught (`AGENTS.md` drifted one line; a state
constant's value renamed). One deviation: `check-anchors.sh` went red on `claude/state-intro` because
step 1 rewrote the exact sentence that anchor located, so section 0 was re-swept to the new text.

### Stage 2: the site's `mfa`, step-up `auth_time` and lifecycle sentences
`856ac120`. Three sentences, on two pages.

`concepts/acr-amr.mdx`, after the `amr` examples: Goiabada never emits the `mfa` value, and an
application reads `amr` for the methods and `acr` for the level. RFC 8176 section 2 defines `mfa` and
registers it and asks for nothing further, OIDC Core section 2 only says `amr` values SHOULD come
from that registry, and `["pwd", "otp"]` already says two factors were used.

Same page, after the step-up paragraph: `auth_time` becomes the time the OTP was entered, only
`prompt=none` and plain session reuse keep the earlier time, and no claim carries the first factor's
time after a step-up. `concepts/authorization-lifecycle.mdx` Stage 5 said `auth_time` "reflects when
the authorization code was created", which is false on the SSO reuse arm and imprecise on every other
one, since a consent screen can sit between the credential and the code indefinitely; it now says the
moment the user last entered a credential, and that a reused session keeps the session's time.

Sizes, `check-docs.sh --working`: the AMR section 87 to 112 words, the step-up section 109 to 150,
lifecycle Stage 5 84 to 90. Two sentences each, as the agreement capped them, and no sibling warning
on either page. The step-up sentence was written against a re-run of the two unit tests that pin the
behaviour, not from the review's snapshot.

Unit green, 5981 ok 0 fail, though no Go file moved; the site builds, 43 pages, no new warning. Data
and integration not applicable. Two deviations: the dev container has no `npm`, so the site build ran
on the host against the committed lockfile; and the lifecycle sentence was cut from 39 words back to
35 to hold the "same length" cap the agreement set for it.

### After the final review: `auth_time` is the credential's instant, not the ceremony's
`ca1e9daf`. Round 1 found the two sentences stage 2 wrote disagreeing with the code they described.
They say `auth_time` is the moment the user entered a password or an OTP; the code read the clock
again at `/auth/completed` and used the instant it had already captured as nothing but a yes/no
flag. The browser owns the hop between those two writes, so a tab left sitting after the password
was accepted and resumed later minted a token saying the user had actively authenticated just now —
and a relying party that sent `max_age` and read `auth_time` back to confirm a fresh sign-in was
told it had one, about whoever was holding that browser when it resumed.

The user chose to move the code rather than reword the sentences (**decision 8** on this thread),
which is why a documentation change ends with a handler in it. The reuse arm in
`handler_auth_completed.go` now writes `authContext.AuthenticatedAt`, and `StartNewUserSession`
takes the instant as a parameter and stamps it on the new row. `Started` and `LastAccessed` stay on
now: they measure the session's own life, and pulling them back would shorten it against the idle
timeout and the max lifetime. SSO reuse and `prompt=none` are untouched — nothing was captured, so
the session keeps the timestamp it had. The sentences on the two pages stand exactly as stage 2
wrote them; no documentation changed in this commit.

Both arms are tested with a capture 90 minutes old, which is the only thing that tells the two
instants apart: "now" passes against either behaviour. `TestAuthTime_IsTheInstantTheCredentialWasAccepted`
drives the whole stack on both arms with the browser pausing two seconds after the password is
accepted, and asserts the ID token's `auth_time` falls inside the window the password was submitted
in and strictly before the browser resumed; it reads `sid` to prove the second ceremony really took
the reuse arm. Unit 6184/0, data (sqlite) 537/0, integration (sqlite) 1374/0. Four mutations, all
caught.

This is a deviation from the agreement worth naming: section 4 said "No handler, model or schema
moves", and this commit moves two. It is the user's call, taken with the cost in front of them.

### After round 2: why `OTPKeyURL` is safe across a restart
`25367f84`. Round 2 was clean and raised one follow-up, which was verified against the source and
taken here rather than filed. The field rule listed `OTPKeyURL` among the fields "overwritten before
anything reads it again", on "every arm of `/auth/otp`". `handler_auth_otp.go:121-131` deliberately
does the opposite: the enrolment arm regenerates the key **only when parsing the stored one fails**,
because replacing it on every GET replaced the seed behind a QR code the user had already scanned
(#242 part 3). The conclusion held and only the reason was false; the correct reason is
reachability. `handler_auth_otp.go` is the field's only reader and writer, both its routes gate on
`level2_otp`, and the one transition out of that state clears it, so neither restart route can reach
a stale key. Both agent files edited identically; `cmp CLAUDE.md AGENTS.md` empty. Unit 5986/0, no Go
file moved.

## Tiers

**Locally, in the dev container.** Unit, all three modules, at every stage and at the final commit:
5986 pass / 0 fail at `25367f84`, 6184 / 0 at `ca1e9daf` where the new Go tests landed. Data (sqlite)
537 / 0 and integration (sqlite) 1374 / 0 at `ca1e9daf`, including the new
`TestAuthTime_IsTheInstantTheCredentialWasAccepted` at 4.23s. Both are recorded not applicable at the
two documentation stages and at `25367f84`, which move no Go file, query, handler, route or
migration. Plus the Astro site build at stage 2, 43 pages, no new warning, and `check-anchors.sh`
31 of 31 after every commit.

**In the check suite, and only there.** The data and integration tiers on MySQL, PostgreSQL and SQL
Server; the race detector leg; Lint, Vulnerabilities, Versions and Build validation. All thirteen
jobs are green on the final commit `25367f84`
([run](https://github.com/leodip/goiabada/actions/runs/34687781866)) and were green on `ca1e9daf` too,
which is where the four-engine claim for the new integration test rests.

1. **Section 4 said "No handler, model or schema moves"; two Go files move.** The consequence of
   decision 8, answered by the user on this thread with the cost in front of them.
   `handler_auth_completed.go`, `core/user/usersession_manager.go`, the interface and its generated
   mock, plus the tests.
2. **The field rule is longer than budgeted.** Section 2 asked for "four or five sentences"; it is
   six, and 250 words. Recorded rather than trimmed: two review rounds established that each field
   needs its own reason rather than a shared one, and cutting to the stated length would mean
   dropping something section 2 asked for or restoring a reason the source contradicts.
3. **The site build ran on the host, not in the dev container**, which has no `npm`. Against the
   committed lockfile.
4. **`check-anchors.sh` went red at stage 1** because the step rewrote the exact sentence an anchor
   located. Section 0 was re-swept to the new text rather than the edit being reverted.

### Review

round at the plan gate before any code was written. No stage was reviewed on its own.

|---|---|---|---|---|
| plan | 1 | 5 | 2 | all fixed in the plan, before implementation |
| final | 1 | 4 | 1 | `2f4f2be7`, `ca1e9daf` |
| final | 2 | 0 | 0 | clean; one follow-up it raised was taken in `25367f84` |

- **blocking, security (final 1):** the two new documentation sentences called `auth_time` the moment
  the credential was accepted, while both writes read the clock again at `/auth/completed` — so a
  browser that paused mid-ceremony minted a token claiming a sign-in that had just happened, and a
  relying party using `max_age` was told it was fresh. Escalated as decision 8 rather than fixed
  either way, because the two repairs pointed in opposite directions. Fixed in `ca1e9daf` on the
  user's answer: both arms now write the captured credential instant.
- **significant, conformance (final 1):** the new `mfa` sentence asserted `["pwd", "otp"]` proves two
  factors were used, contradicting this same branch's field rule while #140 is open. Fixed in
  `2f4f2be7`: it now claims only what holds whatever #140 does.
- **blocking, correctness (plan 1), two of them:** the roster guard was specified against `const`
  while the constants sit under `var (`, and against whole-section occurrence, which would still pass
  after a state's own row was deleted; and `SourceRoot` returns `src`, so the checker would have
  looked for the agent files inside it. Both fixed in `plan.md` before implementation.

Full account, with the evidence, the mutation tables and what each round ran, in
[this comment](https://github.com/leodip/goiabada/pull/322#issuecomment-5645300616).

## The rest of the run's account

- [Deferred decisions](https://github.com/leodip/goiabada/pull/322#issuecomment-5645288818): none.
  Decision 8 was escalated and answered rather than assumed; the comment says what reversing it costs.
- [Follow-ups, fold-ins and ceilings](https://github.com/leodip/goiabada/pull/322#issuecomment-5645291380):
  two additions to open issues (#140 and #248), with the `gh` commands that file them; five things
  folded in rather than filed; no ceiling markers.

## For the release notes

`auth_time` in the ID token now names the instant the user's password or OTP was accepted, rather
than the instant the sign-in redirect chain finished. The two are milliseconds apart in an ordinary
sign-in and nothing needs doing on upgrade. They differ when a browser pauses mid-ceremony: such a
sign-in now reports its true age, so a relying party using `max_age` to demand a fresh
authentication will correctly judge it stale and re-prompt, where before it was told the sign-in had
just happened. Single sign-on and `prompt=none` are unchanged: a session reused without
re-authentication keeps the timestamp it had.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -166,17 +166,19 @@ it accepts:
 | `/auth/issue` | `handler_auth_issue.go` | Issues authorization code, redirects to client |
 
 ### AuthContext field rule
-A field on `AuthContext` that accumulates across hops and is not recomputed on every hop must be
-discarded when the ceremony restarts, because a restart sends the browser back to `requires_level_1`
+A field on `AuthContext` that accumulates across hops and is not overwritten before it is next read
+must be discarded when the ceremony restarts, because a restart sends the browser back to `requires_level_1`
 with the abandoned attempt's values still on the context. There are two restart routes:
 `HandleAuthCompletedGet` when no session is reusable and level 1 was never completed, and
 `refuseIssuanceUnusableSession` at `/auth/issue` when the bound session is gone, expired or foreign
 and the request is not `prompt=none`. `AuthMethods` is the one field breaking the rule today:
 `AddAuthMethod` appends to it, nothing recomputes it, and it survives both routes onto the session
 row the second pass creates, so a restarted ceremony can mint `amr` values the second pass never
-earned (#140 fixes this; delete this sentence when it lands). Every other field is either recomputed
-on every hop (`AuthenticatedAt`, `OtpConfigGeneration`, `AcrLevel`, `OTPKeyURL`) or, like
-`ConsentedScope`, coherent by construction rather than by rule. The request-derived fields are
+earned (#140 fixes this; delete this sentence when it lands). No other field is recomputed at every hop
+either; each is instead overwritten before anything reads it again on every path out of a restart —
+`AuthenticatedAt` and `OtpConfigGeneration` at `/auth/pwd`, `AcrLevel` by `SetAcrLevel` at
+`/auth/completed`, `OTPKeyURL` on every arm of `/auth/otp`, the only hops that read it — while
+`ConsentedScope` is coherent by construction rather than by rule. The request-derived fields are
 written once and only at `/auth/authorize` — the composite literal in `HandleAuthorizeGet` plus
 `TargetAcrLevel`, set immediately after validation and nowhere else — which #248 pins with a test.
 
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -97,7 +97,7 @@ Programmatic client registration for MCP servers, native apps, etc.
 
 ## Authentication Flow (Authorization Code)
 
-The auth code flow uses a state machine tracked in `AuthContext` (stored in session cookie).
+The auth code flow uses a state machine tracked in `AuthContext` (stored in the server-side session store, keyed by a cookie (#266)).
 
 ### ACR Levels (Authentication Context Class Reference)
 Defined in `src/core/enums/enums.go`:
@@ -108,20 +108,49 @@ Defined in `src/core/enums/enums.go`:
 Target ACR determined by: `acr_values` param in authorize request → falls back to `Client.DefaultAcrLevel`
 
 ### Auth States (State Machine)
-Defined in `src/core/oauth/auth_context.go`. States transition in this order:
+The values below are the string constants declared in `src/core/oauth/auth_context.go`, and
+`AssertAgentDocs` in `core/testutil/agentdocs.go` holds this section's roster to them: a state
+declared there with no row here, or a row here naming no constant, fails every module's unit tier.
+There is no single order: a ceremony's path depends on the target ACR, the session, and `prompt`.
 
-1. **`AuthStateInitial`** - Entry point at `/auth/authorize`
-2. **`AuthStateRequiresLevel1`** - No valid session, needs level1 auth
-3. **`AuthStateLevel1Password`** - User at password form
-4. **`AuthStateLevel1PasswordCompleted`** - Password verified, deciding next step
-5. **`AuthStateRequiresLevel2`** - Level2 auth needed (based on ACR)
-6. **`AuthStateLevel2OTP`** - User at OTP form (or enrollment)
-7. **`AuthStateLevel2OTPCompleted`** - OTP verified
-8. **`AuthStateAuthenticationCompleted`** - All auth done, checking consent
-9. **`AuthStateRequiresConsent`** - Showing consent screen (if `client.ConsentRequired` or `offline_access` scope)
-10. **`AuthStateReadyToIssueCode`** - Ready to issue code and redirect to client
+| State | Assigned by | When |
+|---|---|---|
+| `initial` | `HandleAuthorizeGet` | the composite literal, before validation. Dead: no gate accepts it, so it is only ever overwritten. #248 part 2 deletes it |
+| `requires_level_1` | `HandleAuthorizeGet` | four sites: a deferred error is parked; `prompt=login`; `id_token_hint` names another user; no valid session |
+| | `HandleAuthCompletedGet` | no reusable session and `!Level1AuthCompleted` (restart route 1) |
+| | `refuseIssuanceUnusableSession` | bound session gone, expired or foreign, and not `prompt=none` (restart route 2) |
+| `level1_password` | `HandleAuthLevel1Get` | unconditional |
+| `level1_password_completed` | `HandleAuthPwdPost` | password verified, user enabled |
+| `level1_existing_session` | `HandleAuthorizeGet` | valid session, hint matches, user enabled. The SSO shortcut: password entry is skipped and `/auth/level1completed` accepts this state directly |
+| `requires_level_2` | `HandleAuthLevel1CompletedGet` | target ACR above the owned session's ACR, or above level 1 with no owned session, or the session's `OtpConfigGeneration` differs from the user's and the target is above level 1 |
+| `level2_otp` | `HandleAuthLevel2Get` | `level2_optional` with OTP enabled, or `level2_mandatory` (enrolment happens at `/auth/otp` if needed) |
+| `level2_otp_completed` | nobody | Dead: declared and assigned nowhere. #248 part 2 deletes it |
+| `authentication_completed` | `HandleAuthLevel1CompletedGet` | no step-up needed |
+| | `HandleAuthLevel2Get` | `level2_optional` and no OTP enrolled, which is the skip that bypasses `/auth/otp` entirely |
+| | `HandleAuthOtpPost` | OTP verified or enrolled |
+| `requires_consent` | `HandleAuthCompletedGet` | `prompt=consent`, or `client.ConsentRequired`, or `offline_access` in scope |
+| `ready_to_issue_code` | `handlePromptNone` | every silent check passed, skipping every hop between |
+| | `HandleAuthCompletedGet` | no consent needed |
+| | `HandleConsentGet` | scope already fully consented, no `offline_access`, no `prompt=consent` |
+| | `HandleConsentPost` | approved with at least one scope |
 
-**Shortcut for existing session**: If user has valid session, flow goes `AuthStateInitial` → `AuthStateLevel1ExistingSession` → `AuthStateLevel1PasswordCompleted` (skipping password entry), then continues from step 4.
+Each route then gates on the state it finds, and answers `rejectAuthStateMismatch` when it is not one
+it accepts:
+
+| Route | Method | Accepts | On mismatch |
+|---|---|---|---|
+| `/auth/authorize` | GET, POST | anything, mints a new context | n/a |
+| `/auth/level1` | GET | `requires_level_1` | 400 |
+| `/auth/pwd` | GET | `level1_password` | 400 |
+| `/auth/pwd` | POST | ceremony id first, then `level1_password` | 400, 400 |
+| `/auth/level1completed` | GET | `level1_password_completed`, `level1_existing_session` | **500**, the one gate c499ec51 missed when it converted the other ten to 400. #248 part 1 |
+| `/auth/level2` | GET | `requires_level_2` | 400 |
+| `/auth/otp` | GET | `level2_otp` | 400 |
+| `/auth/otp` | POST | ceremony id first, then `level2_otp` | 400, 400 |
+| `/auth/completed` | GET | `authentication_completed` | 400 |
+| `/auth/consent` | GET | `requires_consent` | 400 |
+| `/auth/consent` | POST | ceremony id first, then `requires_consent` | 400, 400 |
+| `/auth/issue` | GET | `ready_to_issue_code` | 400 |
 
 ### Flow Handlers (in order)
 | Handler | File | Purpose |
@@ -135,6 +164,21 @@ Defined in `src/core/oauth/auth_context.go`. States transition in this order:
 | `/auth/completed` | `handler_auth_completed.go` | Final auth check, scope filtering, consent check, session bump/create |
 | `/auth/consent` | `handler_consent.go` | User consent screen (if required) |
 | `/auth/issue` | `handler_auth_issue.go` | Issues authorization code, redirects to client |
+
+### AuthContext field rule
+A field on `AuthContext` that accumulates across hops and is not recomputed on every hop must be
+discarded when the ceremony restarts, because a restart sends the browser back to `requires_level_1`
+with the abandoned attempt's values still on the context. There are two restart routes:
+`HandleAuthCompletedGet` when no session is reusable and level 1 was never completed, and
+`refuseIssuanceUnusableSession` at `/auth/issue` when the bound session is gone, expired or foreign
+and the request is not `prompt=none`. `AuthMethods` is the one field breaking the rule today:
+`AddAuthMethod` appends to it, nothing recomputes it, and it survives both routes onto the session
+row the second pass creates, so a restarted ceremony can mint `amr` values the second pass never
+earned (#140 fixes this; delete this sentence when it lands). Every other field is either recomputed
+on every hop (`AuthenticatedAt`, `OtpConfigGeneration`, `AcrLevel`, `OTPKeyURL`) or, like
+`ConsentedScope`, coherent by construction rather than by rule. The request-derived fields are
+written once and only at `/auth/authorize` — the composite literal in `HandleAuthorizeGet` plus
+`TargetAcrLevel`, set immediately after validation and nowhere else — which #248 pins with a test.
 
 ### Deferred error redirects (#213)
 `/auth/authorize` never redirects an unauthenticated browser to a client's `redirect_uri` on a failed
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -177,8 +177,10 @@ row the second pass creates, so a restarted ceremony can mint `amr` values the s
 earned (#140 fixes this; delete this sentence when it lands). No other field is recomputed at every hop
 either; each is instead overwritten before anything reads it again on every path out of a restart —
 `AuthenticatedAt` and `OtpConfigGeneration` at `/auth/pwd`, `AcrLevel` by `SetAcrLevel` at
-`/auth/completed`, `OTPKeyURL` on every arm of `/auth/otp`, the only hops that read it — while
-`ConsentedScope` is coherent by construction rather than by rule. The request-derived fields are
+`/auth/completed` — while `ConsentedScope` is coherent by construction rather than by rule, and
+`OTPKeyURL` by reachability: the enrolment arm of `HandleAuthOtpGet` deliberately keeps a parseable
+key rather than replacing it (#242 part 3), but it is set only while the ceremony sits at
+`level2_otp`, and the one transition out of that state clears it. The request-derived fields are
 written once and only at `/auth/authorize` — the composite literal in `HandleAuthorizeGet` plus
 `TargetAcrLevel`, set immediately after validation and nowhere else — which #248 pins with a test.
 
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -166,17 +166,19 @@ it accepts:
 | `/auth/issue` | `handler_auth_issue.go` | Issues authorization code, redirects to client |
 
 ### AuthContext field rule
-A field on `AuthContext` that accumulates across hops and is not recomputed on every hop must be
-discarded when the ceremony restarts, because a restart sends the browser back to `requires_level_1`
+A field on `AuthContext` that accumulates across hops and is not overwritten before it is next read
+must be discarded when the ceremony restarts, because a restart sends the browser back to `requires_level_1`
 with the abandoned attempt's values still on the context. There are two restart routes:
 `HandleAuthCompletedGet` when no session is reusable and level 1 was never completed, and
 `refuseIssuanceUnusableSession` at `/auth/issue` when the bound session is gone, expired or foreign
 and the request is not `prompt=none`. `AuthMethods` is the one field breaking the rule today:
 `AddAuthMethod` appends to it, nothing recomputes it, and it survives both routes onto the session
 row the second pass creates, so a restarted ceremony can mint `amr` values the second pass never
-earned (#140 fixes this; delete this sentence when it lands). Every other field is either recomputed
-on every hop (`AuthenticatedAt`, `OtpConfigGeneration`, `AcrLevel`, `OTPKeyURL`) or, like
-`ConsentedScope`, coherent by construction rather than by rule. The request-derived fields are
+earned (#140 fixes this; delete this sentence when it lands). No other field is recomputed at every hop
+either; each is instead overwritten before anything reads it again on every path out of a restart —
+`AuthenticatedAt` and `OtpConfigGeneration` at `/auth/pwd`, `AcrLevel` by `SetAcrLevel` at
+`/auth/completed`, `OTPKeyURL` on every arm of `/auth/otp`, the only hops that read it — while
+`ConsentedScope` is coherent by construction rather than by rule. The request-derived fields are
 written once and only at `/auth/authorize` — the composite literal in `HandleAuthorizeGet` plus
 `TargetAcrLevel`, set immediately after validation and nowhere else — which #248 pins with a test.
 
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -97,7 +97,7 @@ Programmatic client registration for MCP servers, native apps, etc.
 
 ## Authentication Flow (Authorization Code)
 
-The auth code flow uses a state machine tracked in `AuthContext` (stored in session cookie).
+The auth code flow uses a state machine tracked in `AuthContext` (stored in the server-side session store, keyed by a cookie (#266)).
 
 ### ACR Levels (Authentication Context Class Reference)
 Defined in `src/core/enums/enums.go`:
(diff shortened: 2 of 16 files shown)


leodip:
## Decision 8: should `auth_time` say when the user typed their password, or when sign-in finished?

`auth_time` is a timestamp Goiabada puts in the ID token saying when the user last proved who they
were. An app that wants a guaranteed-fresh sign-in before something sensitive — a payment, a
password change — adds `max_age=60` to the sign-in link and then reads `auth_time` out of the token
it gets back to check it actually got one.

Users meet it at the end of a sign-in, in the ID token. The two documentation pages that explain it
are **Concepts → ACR and AMR** (the step-up section) and **Concepts → Authorization lifecycle**
(stage 5), and this PR rewrote one sentence on each.

A sign-in has two halves. First the user types the password, or the OTP code, and it is accepted.
Then the browser is redirected onward to a final page that creates or refreshes the session.
`auth_time` is written in the **second** half, never the first:
`src/authserver/internal/handlers/handler_auth_completed.go:143-144` sets
`bumpedSession.AuthTime = time.Now().UTC()` when an existing session is being reused, and
`src/core/user/usersession_manager.go:87` sets `AuthTime: utcNow` when a new session is created.
Both read the clock at that final page. The moment the credential was actually accepted is captured
two hops earlier, by the password handler and the OTP handler, and then used only as a yes/no flag
and thrown away.

Normally the two moments are milliseconds apart and nobody could tell. The gap is the browser's to
choose, though. A tab left sitting after the password is accepted and resumed hours later produces a
token saying the user actively authenticated just now, and an app that asked for a fresh sign-in is
told it got one. Whoever is holding that browser when it resumes is the person the freshness claim
is then about, and that need not be whoever typed the password.

**What the specification says.** OpenID Connect Core 1.0 §2 defines `auth_time` as "Time when the
End-User authentication occurred." §3.1.2.1 defines `max_age` as "the allowable elapsed time in
seconds since the last time the End-User was actively authenticated by the OP." Neither sentence is
a MUST about which instant inside a sign-in counts, so the spec **permits both options below**. Its
wording, "actively authenticated", leans towards A, and A is the behaviour `max_age` exists to
enable. Neither is required.

**The one axis: which instant the number names.**

| | **A** — when the credential was accepted | **B** — when sign-in finished |
|---|---|---|
| What `auth_time` names | the moment the password or OTP was accepted | the moment the final sign-in page ran |
| Can a paused browser move it? | no | yes, by however long it pauses |
| What changes | two lines of handler code, plus tests | two sentences on two docs pages |
| The sentences this PR added | stand as written | get reworded |
| Where the change lands | `handler_auth_completed.go` and `usersession_manager.go`, which this PR's plan put out of bounds | the two docs pages already in this PR |

**Worked example.** An app requires a fresh sign-in for a payment and sends `max_age=60`. The user
types their password at 10:00 and it is accepted; the tab then sits until 12:30 before the redirect
completes. Under **A** the token says `auth_time` = 10:00, the app sees two and a half hours
elapsed, and it sends the user back to sign in again. Under **B** the token says 12:30, the app sees
zero seconds elapsed, and the payment goes through.

**What does not differ.** Single sign-on and `prompt=none` behave identically either way: a session
reused without re-authentication keeps its existing timestamp under both options, which is already
correct, and neither option touches it. Either answer is one commit's work here, and nothing else in
this PR is built on top of it.

**Recommendation: A.** It is what the code's own comment already says it is doing — *"here the user
actually entered credentials again, so refresh AuthTime to reflect that"* — and the value it should
be using is captured two hops earlier and currently discarded. It is also the only option under
which `max_age` means what the spec says it means.

**Why I am asking instead of just doing it.** This PR was scoped as documentation only, and its plan
says in as many words that no handler moves. A is a behaviour change to session and token code in an
authorization server, and it would need its own tests: both session arms driven with a deliberately
old captured time, plus an assertion that the resulting ID token carries that exact value. If you
would rather this PR stayed documentation-only, **B is complete and costs two sentences** — say so
and I will take B here and file A as its own issue.

Round 1's other two findings needed no decision and are already fixed and pushed in `2f4f2be7`: the
`AuthContext` field rule claimed four fields are "recomputed on every hop" when none is, and the ACR
and AMR page claimed `["pwd", "otp"]` proves two factors were used, which this PR's own field rule
explains is not reliable while #140 is open.

leodip:
A. Double check the code reflects this idea.

leodip:
## Decision 8: A, applied in `ca1e9daf`

Both writes now take the instant the ceremony captured instead of reading the clock again.

- **Reuse arm**, `handler_auth_completed.go`: `bumpedSession.AuthTime = authContext.AuthenticatedAt.UTC()`. The existing `userReallyAuthenticated` check is already the non-nil, non-zero guard the dereference needs, so no new condition was added.
- **Create arm**: `StartNewUserSession` takes the instant as a parameter and stamps it on `AuthTime`. `Started` and `LastAccessed` stay on now — they measure the session's own life, and pulling them back would shorten it against the idle timeout and the max lifetime.
- The interface and its generated mock carry the new parameter (regenerated, not hand-edited).

SSO and `prompt=none` are untouched, as you said they would be: nothing is captured, so the session keeps the timestamp it had.

### "Double check the code reflects this idea"

Everything else that touches the value, re-read rather than assumed:

- `/auth/issue` puts `AuthenticatedAt` on the code row and the token issuer signs it as `auth_time` — unbroken chain, and now checked end-to-end by a test rather than by reading.
- `handler_authorize.go`'s `prompt=none` path reads the session's `AuthTime`, which is now the credential's instant. Consistent.
- Both credential handlers still capture, so a step-up names the OTP's instant. That is what decision 1 decided and what the ACR and AMR page says.
- Nothing in production compares `AuthTime` against `Started`. `UserSession.IsValid` uses `Started` and `LastAccessed` only, so the idle timeout, the max lifetime and the server-side `max_age` check are all untouched. The remaining readers are display: the session APIs and the admin console.
- Every documentation sentence about `auth_time` is true as written — the two this PR added, the three on the prompt page, and the model sketch in `CLAUDE.md`. **No documentation changed in this commit**; making those sentences true is what the commit did.

**One interaction worth knowing, and it is not new breakage.** `auth_time` can now be older than the session's `Started` when a browser paused, because the server still measures `max_age` from `Started` while the token reports the credential. An app sending a short `max_age` to a paused browser will judge the token stale and re-prompt — which is the right direction, and the alternative is telling it a stale sign-in was fresh. Moving the server-side check onto `AuthTime` is #243, open and out of scope here.

### Tests

Both arms are driven with a capture 90 minutes old. That is the whole instrument: with "now" the assertions pass against either behaviour, which is how this survived until the review. The new integration test runs a real ceremony on each arm with the browser pausing two seconds after the password is accepted, then asserts the ID token's `auth_time` falls inside the window the password was submitted in and strictly before the browser resumed. It reads `sid` to prove the second ceremony really took the reuse arm rather than quietly exercising the create arm twice.

Unit 6184/0, data (sqlite) 537/0, integration (sqlite) 1374/0. Four mutations, all caught — including replacing the create arm's forwarded value with `nil` and the reuse arm's write with the session's `LastAccessed`, which is the old behaviour expressed so that it compiles.

Also fixed in passing, found while changing the signature: `AssertNotCalled(t, "StartNewUserSession", …)` passed seven `mock.Anything` for an eight-parameter method. That comparison is over the whole argument list, so a count that does not match never matches a call either — it was asserting nothing. Nine now, one per parameter.

Round 2 of the final review is next; the PR stays in draft until it is clean.

leodip:
## Deferred decisions: none

Every decision this run touched is settled, and none was carried by an assumption the user has not seen.

Section 3 holds eight decisions. Seven were settled in the interview before the seal. The eighth was raised by the run, at the final gate, and **escalated rather than assumed**: round 1 of the final review found that the two new documentation sentences called `auth_time` the moment the user entered a credential while the code wrote the moment `/auth/completed` ran, and the two available repairs pointed in opposite directions (reword the docs inside the agreement's no-handler envelope, or move the code). That was a choice, so it was posted on this thread with both options, the OIDC Core 3.1.2.1 text and a worked `max_age=60` example.

You answered on 2026-09-12 with option A. It was applied in `ca1e9daf`, and decision 8 in the agreement records the answer, the reasoning and the rejected option B.

**What reversing it costs, since it is the one decision that moved code.** Reversing means option B: revert `ca1e9daf`, drop `TestAuthTime_IsTheInstantTheCredentialWasAccepted` and the three `TestStartNewUserSession_*` cases, restore the eight-parameter `StartNewUserSession` and its mock, and reword the two sentences on `acr-amr.mdx` and `authorization-lifecycle.mdx` to describe the completion instant instead. One commit to revert, two sentences to rewrite. The cost of doing so is that `max_age` goes back to meaning something OIDC Core's "actively authenticated by the OP" does not: a paused browser resumed hours later mints a token claiming a sign-in that just happened.

One deviation from the agreement follows from that answer and is named in the body: section 4 said "No handler, model or schema moves", and `ca1e9daf` moves two Go files plus an interface and its generated mock. You chose that with the cost in front of you.

leodip:

Duplicate search re-run at the close (`gh issue list --state all --search "amr restart AuthMethods"`): #140 and #248 are both still open and still the right homes. Nothing new appeared beside them, so **no new issue is filed** — both follow-ups are additions to issues that already exist.

### Follow-up 1 — add to #140 (open)

Found at the final gate, round 1, finding 4. Not folded in: it fails fold-in test 2, there is something to settle. The example is correct the moment #140 lands, so the options are to leave it alone and fix #140, or to publish a caveat about a live unfixed defect on a public page. That is a choice, and choices are not folded in quietly.

This also carries decision 4's seal item: the sentence PR #322 added to `CLAUDE.md` / `AGENTS.md` naming `AuthMethods` as the field that breaks the restart rule is owed deletion when #140 lands.

```bash
gh issue comment 140 --body 'Two notes from #252 / PR #322, which documented this defect rather than fixing it.

**1. Closing this issue also makes a documentation example true again.** `site/src/content/docs/concepts/acr-amr.mdx`, under "Checking ACR and AMR in your application", ships a JavaScript example whose comment reads `// User used OTP during authentication` and which logs `MFA was used` on `idToken.amr.includes("otp")`. While this issue is open that inference is unsound in exactly the way the issue describes: a ceremony that restarts carries `AuthMethods` forward, so a second pass verifying only a password can still mint `otp` into `amr`. Closing this issue makes the example true again and needs no documentation change. If it is going to stay open for a while, the example needs a caveat instead. Either way the check belongs in this issue scope rather than being rediscovered later. (PR #322 corrected a separate sentence on the same page that had asserted `["pwd", "otp"]` proves two factors were used; the example itself was left alone deliberately, for the reason above.)

**2. A sentence in `CLAUDE.md` and `AGENTS.md` is owed deletion when this lands.** The `### AuthContext field rule` section added by PR #322 names `AuthMethods` as the one field breaking the restart rule today, with this issue as the pointer, and says so in its own text. Delete that sentence as part of the fix; the agent-docs guard does not check it, so nothing will fail if it is forgotten.'
```

### Follow-up 2 — add to #248 (open)

Decision 7's seal item. Not folded in: it is a code change on a surface this documentation change does not touch (fold-in test 1), and it is #248 part 1's own residue.

```bash
gh issue comment 248 --body 'From #252 / PR #322, which built the ceremony state tables from today handlers and counted the gates.

Part 1 of this issue is smaller than it looks: c499ec51 already converted ten of the eleven auth-state gates from 500 to 400 (`grep -c "rejectAuthStateMismatch(httpHelper" src/authserver/internal/handlers/handler_*.go`, sum 10). The eleventh, the two-state gate in `HandleAuthLevel1CompletedGet` at `/auth/level1completed` accepting `level1_password_completed` and `level1_existing_session`, still answers 500 (`grep -c "does not match any required state" src/authserver/internal/handlers/handler_auth_level1.go` = 1). The c499ec51 commit message says "ten sites in seven files" and never mentions this one, so it was missed rather than kept deliberately. PR #322 records the 500 in the accepted-by table in `CLAUDE.md` with this issue as the pointer; that table row changes to 400 when this lands.'
```

- **`25367f84` — the `OTPKeyURL` clause in the field rule.** Round 2 raised it as a follow-up on a round-budget argument, not a substantive one. Verified against the source first, and the reviewer was right: the paragraph said `OTPKeyURL` is overwritten on every arm of `/auth/otp`, and `handler_auth_otp.go:121-131` deliberately does the opposite, regenerating only when the stored key fails to parse, because replacing it on every GET broke a QR code the user had already scanned (#242 part 3). It is prose this change wrote, it is the same family as round 1's finding 3 which repaired other false reasons in the same paragraph, and there was nothing to settle. The conclusion held; only the reason was false, and the correct reason is reachability: `handler_auth_otp.go` is the field's only reader and writer, both its routes gate on `level2_otp`, and the one transition out of that state clears it, so neither restart route can read a stale key.
- **`2f4f2be7` — round 1's findings 3 and 4.** The field rule's "recomputed on every hop" was false of all four fields it named; it now states the invariant that actually holds, with the handler named per field. And the new `mfa` sentence had asserted `["pwd", "otp"]` proves two factors were used, which this same branch's field rule explains is unreliable while #140 is open; it now claims only what holds whatever #140 does.
- **`ca1e9daf` — a test that could not fail.** `handler_auth_completed_test.go` passed seven `mock.Anything` to `AssertNotCalled(t, "StartNewUserSession", ...)` for what was then an eight-parameter method. `AssertNotCalled` compares the whole argument list, so a count that does not match the method's never matches a call either: the assertion passed however often the method ran. Nine now, one per parameter. Found while changing that method's signature, fixed in the same commit.
- **`aa0d6fa9` — the drifted `AGENTS.md` paragraph.** The "Test fixtures" paragraph was missing from `AGENTS.md` (decision 3). Copied across in the same stage, because the new byte-identity guard would otherwise fail on it.
- **`856ac120` — the false `auth_time` sentence on `authorization-lifecycle.mdx`.** Decision 5: same docs surface, same subject, one sentence, nothing to settle. A follow-up would have been a run for one line.

`none`. `git diff --unified=0 origin/main...HEAD | grep '^+.*ceiling:'` returns nothing: this change adds no marker, because it introduces no bound that a later change would need a trigger to revisit.

leodip:
## The review account

This is the code review for this branch. It ran at the end, over the whole change, so this comment is the only place its content exists: the verdicts live in a gitignored directory that the next round overwrites. Reviewer `openai-codex/gpt-5.6-sol` at effort `xhigh` throughout.

### Summary

|---|---|---|---|---|
| plan | 1 | 5 | 2 | all fixed in `plan.md` before any code was written |
| final | 1 | 4 | 1 | 2 fixed in `2f4f2be7`; the blocking one and its twin escalated as decision 8, then fixed in `ca1e9daf` |
| final | 2 | 0 | 0 | clean; one follow-up it raised was verified and taken here, `25367f84` |

No stage was reviewed on its own. Round 2 found nothing and changed no production code, so no round 3 was funded.

---

### Plan gate, round 1 (2026-09-11)

Request `252-20260911T212154Z`, `type: plan-review`. Five findings, all with a forced answer, nothing escalated. `unchecked` held three entries: no implementation to run (dismissed, plan gate), no runtime probes (dismissed, no handler or schema moves at that point), and the `auth_time` probe not re-run because it targeted another checkout (covered: rewritten and re-run here, green).

1. **Blocking.** The roster guard was specified against `const` while the declarations sit under `var (`, and against whole-section occurrence, which misses a deleted assigned-by row. Confirmed and fixed in the plan before implementation: parse `var` or `const` `GenDecl`s, require set equality with the first-cell state tokens, add the `var` fixture and the deleted-row negative.
2. **Blocking.** `SourceRoot` returns `src`, so the checker would have looked for the agent files inside `src`. Confirmed and fixed: `filepath.Dir(SourceRoot(t))`.
3. **Significant.** `probe/stepup-auth-time.sh` hardcoded the primary checkout's container. Confirmed and fixed: container and mount read from `where.sh env`, and it refuses a container that does not mount this checkout. Re-run green.
4. **Minor.** Only `level2_otp_completed` was marked dead; `initial` is the second. Confirmed and fixed.
5. **Significant.** The probe's grep pipeline masked a failing `go test`. Confirmed and fixed: no filter between the test and the exit status.

No refutations.

---

### Tree-hash verification before the final review (against `856ac120`)

```
current   production 33d5a98f8447   unit-tests e7492bbfc493   data-tests 9eeb7116e6e2
          integration-tests e3ced8adbedb   docs 79aba98d1411   all e7c5b1a0fb59

stage 1  unit live   data ?   integration ?   docs stale   (moved since: docs)
stage 2  unit live   data ?   integration ?   docs live
```

One category stale, and it is the expected one: stage 1's `docs` hash moved when stage 2 edited the two `.mdx` pages, which is stage 2's work landing on top of stage 1's. Nothing stage 1 asserted depends on it — stage 1 touched no `.mdx` file and ran no site build. The mutation evidence is live at both stages: `unit-tests` is live for both and `production` was unchanged since stage 2 recorded it, so the six mutations run against `agentdocs.go` still held against the tree under review. `data-tests` and `integration-tests` are `?` because no stage recorded one: no stage touched `src/core/data/`, a migration, a handler, a route or a schema, so both were recorded not applicable rather than skipped. The four-engine suites ran in CI regardless and were green.

---

### Final gate, round 1 (2026-09-11)

Request `252-20260911T232113Z`, fresh session. Verdict `findings`, four. All three axes `reviewed`. Worth naming from its `ran` list: it re-ran `TestAgentDocs` and then **mutated `agentdocs.go` three ways itself** — disabling the byte-identity comparison, the declared-state-missing detection and the unknown-roster-state detection — and reported all three caught. That is the guard this change added, independently established rather than taken on the run's word.

**Finding 1, conformance, minor. Confirmed, deferred into decision 8.** Both new `auth_time` sentences call it the moment the credential was entered, and the value actually written was the moment `/auth/completed` ran. Verified: `handler_auth_completed.go:143-144` set `bumpedSession.AuthTime = time.Now().UTC()` on the reuse arm and `usersession_manager.go:87` set `AuthTime: utcNow` on the create arm; the captured `authContext.AuthenticatedAt` was read at line 95 only as a boolean and never written through.

**Finding 2, security, blocking. Confirmed, escalated as decision 8.** The same code read as a defect: a browser can pause between the credential POST and `/auth/completed`, and the timestamp the token signs is the resume. The window is bounded only by the server-side session row's expiry — `AuthContext` carries no ceremony start time and no ceremony expiry, in any of its 30 fields. Not auto-resolved, on three grounds: session and token behaviour is never auto-resolved; the delay is chosen timing by whoever holds the browser rather than something no party can steer, so the rarity qualifier does not apply; and the forced-repair exception does not fit, because the reviewer supplied two incompatible forced answers — finding 1 says reword the docs, finding 2 says change the handler — which is the definition of a choice. Decision 1 did not cover it either: that settled that a step-up moves `auth_time` forward at all, not which instant inside the step-up it names.

**Finding 3, conformance, minor. Confirmed and fixed in `2f4f2be7`.** The field rule's "recomputed on every hop" is false of all four fields it named. Verified by inventory: every production write to the four is at `handler_auth_pwd.go:291,296`, `handler_auth_level2.go:84`, `handler_auth_otp.go:79,129,412,434,443`, `handler_auth_completed.go:157,159,298,300,319`, `handler_authorize.go:582,762,771,773,777` and `auth_context.go:249,259` — not one of them runs on every hop. The paragraph now states the invariant that is actually true and is what makes the restart argument work: each is overwritten before anything reads it again on every path out of either restart route, with the handler named per field.

**Finding 4, conformance, significant. Confirmed; fixed in `2f4f2be7` in part, the rest is follow-up 1.** The new sentence asserted `["pwd", "otp"]` proves two factors were used, which this same change's field rule explains is unreliable while #140 is open — two statements on one branch contradicting each other. The reviewer demonstrated the mechanism by initialising the existing issuance-restart test with `AuthMethods: "pwd otp"` and watching a restart carry it through. The sentence now claims only what holds whatever #140 does. The forced answer's second half, qualifying the page's pre-existing JavaScript example, was **not** applied: that example predates this branch, and whether a public page should carry a caveat about a live unfixed defect is a choice rather than a correction.

**`unchecked`, one entry, answered.** "Did not drive a delayed `/auth/completed` request through a real browser and decode the resulting ID token." It belonged to findings 1 and 2; the escalation named that end-to-end assertion as part of what option A owed, and the integration test below delivers it.

---

### Decision 8 answered, and findings 1 and 2 resolved (`ca1e9daf`)

**Answer.** `leodip` (`authorAssociation: OWNER`), 2026-09-12T09:28:40Z, in full: "A. Double check the code reflects this idea." Option A was one of the two offered, so it is an answer rather than an instruction, and the sentence after it asks for verification of the option chosen.

- Reuse arm: `bumpedSession.AuthTime = authContext.AuthenticatedAt.UTC()`. `userReallyAuthenticated` is already the non-nil, non-zero guard, so no new check was needed.
- Create arm: `StartNewUserSession` gained an `authenticatedAt *time.Time` parameter and stamps it on `AuthTime`. `Started` and `LastAccessed` stay on now. Nil or zero falls back to now; the one production caller cannot produce either, because `/auth/completed` refuses to mint a session without `Level1AuthCompleted` and the password handler sets `AuthenticatedAt` beside it.
- `interfaces.go` and the regenerated mock (mockery v3.7.4; only that one mock changed).

**The verification the answer asked for.** Every other place that touches the value was re-read rather than assumed: `handler_authorize.go:769-773` (`prompt=none`) reads the session's `AuthTime`, consistent and untouched; `handler_auth_issue.go:717` and `oauth/code_issuer.go:97` forward it onto the code row and `oauth/token_issuer.go:703,845` sign it as `auth_time`, an unbroken chain now verified end to end by test rather than by reading; both credential handlers still capture, so a step-up names the OTP's instant, which is what decision 1 decided; no production code compares `AuthTime` against `Started`, and `UserSession.IsValid` uses `Started` and `LastAccessed` only, so the idle timeout, the max lifetime and the server-side `max_age` check are untouched; every documentation sentence about `auth_time` on the site and in the agent files was re-read and all are true as written, so this commit owed no documentation.

**Known interaction, not a defect introduced (#243).** `auth_time` can now be older than the session's `Started` when a ceremony paused, because `UserSession.IsValid` measures `max_age` from `Started` while the token reports the credential. A relying party sending a short `max_age` to a paused browser will see a token it judges stale and re-prompt. That is the correct direction — the alternative is telling it a stale sign-in was fresh — and moving the server-side check onto `AuthTime` is #243's, which is open and out of scope.

**Mutations, four, all CAUGHT**, with `production` re-hashed to `d3857fc9923d` afterwards, unchanged, so nothing was left behind:

| File | Replaced | With | Result |
|---|---|---|---|
| `core/user/usersession_manager.go` | `if authenticatedAt != nil && !authenticatedAt.IsZero() {` | `if false {` | CAUGHT |
| `core/user/usersession_manager.go` | `!authenticatedAt.IsZero()` | `true` | CAUGHT |
| `handlers/handler_auth_completed.go` | `authContext.AuthenticatedAt)` | `nil)` | CAUGHT |
| `handlers/handler_auth_completed.go` | `bumpedSession.AuthTime = authContext.AuthenticatedAt.UTC()` | `bumpedSession.AuthTime = bumpedSession.LastAccessed` | CAUGHT |

`AuthTime: authTime,` replaced with `AuthTime: utcNow,` was tried first and did not compile, since it orphans the variable; the condition mutation is the same defect expressed so that it builds.

---

### Final gate, round 2 (2026-09-12): clean

Request `252-20260912T095053Z`, session resumed, started 09:52:54Z, answered 10:05:00Z. **Verdict: clean, zero findings.** All three axes `reviewed`.

**What it ran**, so the absence of findings has a stated scope. It did not merely re-run the tiers this run reported: it executed the new tests, wrote its own mutations, and drove the ceremony over real HTTP.

- The three new `TestStartNewUserSession_*` cases in `core/user`; the handler's existing-session re-auth, new-session success, SSO-preservation, zero-`AuthenticatedAt` and OTP-only invalid-session-restart cases; and the new integration test on sqlite.
- **Two mutations of its own**, independent of this run's four: `authenticatedAt.UTC()` to a direct assignment, caught by `TestStartNewUserSession_AuthTimeIsNormalisedToUTC`; and the `authTime` fallback to zero, caught by `TestStartNewUserSession_NoCapturedInstantFallsBackToNow`. Both CAUGHT. It recorded the checkout returning to `8c9cc2cfbd3a` after every probe.
- Static fanout sweeps across the credential writers, completion, session persistence, issuance, both token-claim builders, docs, dependencies, schemas and logs; `cmp AGENTS.md CLAUDE.md`; `git diff --check`; `git status`.

It traced the credential instant end to end and confirmed the properties the change claims: the password and OTP handlers are the only credential-time writers, OTP replaces the password time when it is the last credential, `Started` and `LastAccessed` remain session timestamps, plain SSO reuse and `prompt=none` preserve the prior time, and the nil/zero fallback is unreachable from its one production caller.

**`unchecked`, two entries, both answered.**

1. *"No interactive browser pass was performed."* **Dismissed.** The branch touches no browser-facing file: the diff is two agent files, two `.mdx` pages and nine Go files. No template, JavaScript, CSS or cookie attribute moves. What a browser would have exercised — redirects, cookies, session reuse, a deliberate pause, code redemption and the token claims — the new integration test drives over real HTTP.
2. *"Spot-checked the new integration test only on SQLite."* **Covered by CI rather than by re-running it locally.** The check suite at `ca1e9daf` was green across all thirteen jobs, including `Tests / sqlite`, `mysql`, `postgres` and `mssql`, each of which runs the data and integration tiers for its engine. A check run names its own commit and no session can write it, which is why the four-engine claim rests there.

**The one follow-up it raised was verified and taken here** rather than filed, as `25367f84`; the reasoning is in the follow-ups comment. Its tree-hash evidence: `production` stayed at `d3857fc9923d`, unchanged from the tree round 2 reviewed, which is what shows the fix moved prose only.

PR #329, opened 2026-09-12
feat(audit): carry the request id on audit records and rows

### LINKED ISSUE #212: AuditLogger.Log does a settings read and an insert per event, with no bound of its own

### LINKED ISSUE #328: Audit records carry no request_id, because AuditLogger.Log takes no context
#320 made the installed slog handler read chi's request id off a record's context and append it as
`request_id`, so every record written on a request path is joined to the request logger's record
without any call site naming it. The audit path is the one request-path writer that still cannot be
joined, because nothing on it carries a context.

**What is uncorrelated.** `auditlog.LogToConsole` (`src/core/auditlog/auditlog.go`) writes the
`audit event` record for every security event the server raises, and `AuditLogger.Log`
(`src/authserver/internal/audit/audit.go`) wraps it and writes three Error records on its own
failure paths (settings unreadable, details unmarshallable, row not persisted). The rate limiter's
`AuditLogger` in `core/middleware` is the same shape. `Log(auditEvent string, details map[string]any)`
takes no context, so an operator who filters the log on the `request_id` a user reported finds the
request logger's record and not the `auth_failed_pwd` event beside it, whose details are the email
alone. The record that exists to explain a refusal is the one that cannot be joined to the request.

**Why it was not folded into #321.** It is an observability improvement, not a functional or
security defect: the event carries its own identifying details, and the request logger's record is
still found. The change is mechanical but wide, and #321 was already review-heavy:

| | count |
|---|---|
| production call sites of `auditLogger.Log` | 126 |
| test files with `Log` mock expectations | 28 |
| expectations needing a leading `mock.Anything` | 202 |
| interfaces changed and mocks regenerated (`handlers.AuditLogger`, `middleware.AuditLogger`) | 2 |

There is also one design question a sweep would paper over. `LogToConsole` serves two callers that
cannot share code any higher up: the authserver's `AuditLogger`, on a request, and the email
backfill in `core/data/commondb/email_backfill.go`, at startup, with no request. The comment on
`LogToConsole` records the choice not to take a context for exactly this reason. Closing this
means deciding whether the backfill passes `context.Background()` to one entry point or the
package grows a second one, and saying which in that comment.

**What closing it takes.** `Log(ctx context.Context, auditEvent string, details map[string]any)`
on both interfaces, `r.Context()` at every call site, `LogToConsole(ctx, ...)` writing through
`slog.InfoContext`, the three Error records switched to `ErrorContext`, both mocks regenerated
from `src/authserver/.mockery.yaml`, and the 202 expectations given a leading `mock.Anything`.
sloglint's `context: scope` then holds the `*Context` variant inside every one of these functions
for free. Last, add `authserver/internal/audit` and `core/auditlog` to `slogRequestPathDirs` in
`src/core/testutil/slog_lint.go`, whose comment names them as the directories left out for this
issue, so `TestSlogConvention` refuses a plain call there from then on.

**Anchor:** `src/core/auditlog/auditlog.go`, `LogToConsole`, locate by `It takes no context
because AuditLogger.Log has none`.

**Related:** #212 (open) is the audit logger's per-event settings read, in the same function; the
two changes touch the same lines and could land together.

Found in the review of #321.

leodip:
Closes #328. Folds in #212 item 2 (the per-event settings read); #212 stays open for item 1.

## What this change is for

An audit record written while serving a request could not be tied to that request: `AuditLogger.Log`
took no context, so the console record carried no `request_id` even though the installed slog handler
injects one from the request context for every other record, and the `audit_logs` row had no column
for it. An operator holding a `request_id` from a user's report found the request logger's record and
nothing else.

## The rule it establishes

`AuditLogger.Log` and `auditlog.LogToConsole` take a `context.Context` first, and every one of the
126 call sites passes the request's. The slog handler appends `request_id` to the console record and
to the three failure records with no site naming the id. `authserver/internal/audit` and
`core/auditlog` join the request-path directories the slog lint refuses a plain call in, and a new
lint refuses a `Log` call whose first argument is `context.Background()` or `context.TODO()` in those
directories. `Log` reads the settings off the request context when the settings middleware put them
there, and reads the row only when it did not.

The persisted row carries the same id in a `request_id` column, stored exactly as the log renders it
(escaped, clipped to the log's bound), so the value on the admin page is the value an operator greps
for. The admin API returns it as `requestId` and filters on it by exact match; the audit log viewer
shows it in a column and filters on it beside the event select. The id is whatever the client or
proxy sent in `X-Request-Id`, or one Goiabada made when none was sent, so it ties an event to a log
line and is not proof of who sent the request.

## Status

**Complete, and ready for review.** All four stages landed: the context is threaded, the two lints
are in place, the `audit_logs` row carries the request id on all four engines, the admin API returns
it and filters on it, and the admin console's viewer shows it in a column of its own and filters on
it beside the event select.

The whole change was then reviewed once, as one branch rather than stage by stage. Round 1 came back
clean with zero findings across all three axes; its one follow-up was folded in rather than drafted
(below), and round 2 read that fold-in alone and found one test-coverage gap, fixed in `4ca584a1`.
The check suite is green on that commit, all thirteen checks including the data and integration
tiers on all four engines. Nothing is outstanding.

### Stage 1: `AuditLogger.Log` and `LogToConsole` take the request's context
Landed in `6d2cd400`.

`Log(ctx, event, details)` on both interfaces and both hand-written implementers, with the signature
replaced rather than joined by a second method, so the compiler finds every call site and no new site
can opt out of correlation. All 126 production sites pass the request's context: 124 handler sites
pass `r.Context()`, `revokeAndAuditAuthCodeReuse` and `reportTrip` pass the `ctx` they already hold.
The three revocation helpers in `revocation.go`, which took neither a request nor a context, gained
`ctx` first and their 8 callers pass `r.Context()`. `LogToConsole` writes through `slog.InfoContext`
and the three failure records through `slog.ErrorContext`; no site names `request_id`, because
`core/logging` owns the injection. The email-collision backfill passes `context.Background()`, so its
record carries no request id, which is what a startup event is.

`authserver/internal/audit` and `core/auditlog` join `slogRequestPathDirs`, and both comments
recording the omission as a ceiling are rewritten. The compiler forces the parameter but not that the
context is the request's, so the new `testutil.AssertAuditLogContext` refuses `context.Background()`
or `context.TODO()` as the first argument of a `.Log` call in those directories, resolving through the
file's import map so an aliased `context` is caught too. It runs from all three module tiers.

Two fold-ins, both decided at the seal. Decision 4: the orphan `AuditLogger` mock at
`src/core/audit/mocks/` is deleted — no `.mockery.yaml` produced it and nothing imported it, and a
mock asserts nothing about the interface it came from, so this change would have left it a stale copy
of a signature that no longer exists. Decision 5: `Log` takes the settings row off the request
context when `MiddlewareSettings` put it there and reads the row only when it did not, which closes
#212 item 2. The nil check beside the type assertion is load-bearing, since the assertion succeeds on
a typed nil pointer and dereferencing one would panic in the audit path of every event.

Tiers: unit green on three modules (6131 pass, 0 fail), lint green on all four modules, data green on
sqlite, integration green on sqlite. The lint tier is the one that proves the conversion carried:
sloglint's `context: scope` refuses a plain record inside a function holding a context, so a missed
`slog.Info` in either body would have failed it. Ten mutations, one per new seam and one per case
pinning a decision, all `MUTATION CAUGHT`. The check suite was red on this commit, on one test in all three unit jobs: `CLAUDE.md`
and `AGENTS.md` are co-maintained identical copies and only the first was edited. Fixed in
`ca1c0a3f`, which applies the same line to the other file. The guard that caught it,
`TestAgentDocsAreCurrent`, landed on main after this branch forked, so no local tier runs it; CI
builds the pull request's merge commit, which does. Four-engine data and integration are the check
suite's on `ca1c0a3f`.

Worth knowing: the 202 mock expectations took a leading `mock.Anything`, and 18 `args.Get(1)` reads
inside `.Run` callbacks plus two `recordEvent` closures shifted by one index with them — without that
they would have type-asserted the context where they expected the event name. Five bare
`AssertNotCalled(t, "Log")` were deliberately left alone: adding an argument there would weaken
"never called" into "never called with three arguments".

### Stage 2: the `audit_logs` row carries the request id
Landed in `e486a7c9`.

`models.AuditLog` gains `RequestId string` tagged `db:"request_id"`. Every statement over the table
is built from `sqlbuilder.NewStruct`, so both inserts, the PostgreSQL and SQL Server `CreateAuditLog`
overrides and the paginated select carry the new field with no statement text edited. `Log` sets it
to `logging.FieldForLog(chimiddleware.GetReqID(ctx))`, which is byte for byte what `core/logging`
puts on the record: the id is client-chosen, since chi adopts an inbound `X-Request-Id` verbatim, so
it is escaped to printable ASCII and clipped at 128 bytes with a counted marker before it reaches
either target. Storing the raw string would make the page and the log disagree on exactly the values
an attacker picked. The startup backfill sets nothing, so its row reads `''`, as does every row
written before the column.

Migration 000045 adds the column `NOT NULL DEFAULT ''` at width 256 (`TEXT` on SQLite, as
`audit_event` already is there) and `idx_audit_logs_request_id`, on all four engines, with MySQL's
and SQL Server's collations spelled and SQL Server's default constraint named so the down can drop
it; the down drops the index, then the constraint, then the column. A nullable column was rejected:
the JSON layer renders NULL as `""` anyway, and this table carries no nullable string today. Nothing
reads the index until stage 3's filter. The four `schema.golden` files were regenerated after
`where.sh reset-dbs`, so a discarded number recorded in a long-lived database could not be skipped in
silence.

Tiers: unit green (1712 ok, 0 fail), data green on all four engines (mysql 414, postgres 414, mssql
425, sqlite 394, 0 fail), integration green on sqlite (1038 ok). Four engines locally because a
migration is exactly the change engine differences are the risk of. `where.sh migrations` green. Four
mutations, one per behaviour the stage made observable — storing chi's raw id, dropping the column's
`NOT NULL DEFAULT ''`, making the index unique, and inventing an id on the backfill's row — all
`MUTATION CAUGHT`.

Two fold-ins, both forced by the new migration number. The migrate subcommand's tests stated the
schema head as the literal `000044` in eighteen assertions, so any migration breaks them; they are
bumped to `000045`, while `rollbackFloor` stays at 44 by its own rule that it moves only when a
release makes a one-way data change. That makes the floor one below the head for the first time, so
the two comments asserting they were equal now say otherwise. And the schema-parity allowlist, which
records how many columns each known cross-engine idiom covers, took `request_id` under its existing
"sqlite: a string column carries no declared length" rule: count 91 to 92, with its digest.

### Stage 3: filter by request id on the data layer and the admin API
Landed in `7fc63f11`.

`GetAuditLogsPaginated` takes the request id as a fifth argument, through the `Database` interface,
`commondb`, `mssqldb`'s own body and the three delegating wrappers. The match is exact, because the
id is client-chosen and came off one log record: a fold would list one request's audit entries under
another request's id.

Two of the four engines decide `=` differently from Go, and the two halves of the answer differ.
`commondb` adds the equality to the page and the count and drops a folded row after the scan, which
is the guard every other exact string lookup in this tree applies. That cannot work on SQL Server,
which pads for `=` under every collation it has, including the binary ones: this body pages with
`OFFSET/FETCH`, so a row dropped after the fetch leaves a short page, and the count query has no rows
to drop at all. Measured against the live engine during the plan review, a guard applied after the
fetch gave an empty page under a total of 1. So `mssqldb` carries a byte-exact predicate,
`CAST(request_id AS VARBINARY(512))`, beside the equality in both statements, with the equality kept
so the index added in stage 2 can still seek.

What that does not reach is marked rather than left unmarked: MySQL folds a decomposed accent against
its precomposed spelling at equal length, which the page's guard drops and the count still counts, so
a page total could read one higher than the rows listed. Nothing reaches it today, because the only
writer stores the id as `logging.FieldForLog` renders it and that percent-escapes every byte outside
printable ASCII, so no combining mark survives into the column. It is a `ceiling:` with its trigger,
and a MySQL-only binary predicate is the shape that would close it.

The API reads `requestId` off the query and returns the row's id as `requestId`; `openapi.yaml` gains
the property, as required, and the operation gains the parameter. The admin console's `ApiClient`
takes the parameter here so the console compiles against the new API, appended query-escaped because
the value is whatever the client put in `X-Request-Id`; its viewer passes `""` until stage 4.

Tiers: unit green (6138 pass, 0 fail), data green on mysql, postgres, mssql and sqlite (2288 pass, 0
fail), integration green on sqlite (1378 pass, 0 fail), lint green on all four modules. Four engines
locally because SQL Server's fold is the whole risk of the stage. Five mutations, all `MUTATION
CAUGHT`; the one that matters is neutralising the byte-exact predicate, which turns the plan review's
finding into a property the mssql data tier holds rather than a claim in a document.

Nine data cases cover the filter alone, combined with the event, matching nothing, and empty meaning
no filter, plus the guard: a stored id and the same id with one trailing space, filtered by each at
page size 1, where the page must list only the byte-equal row and the total must equal the page
length, on every engine and with no `dbType()` branch. Four integration cases prove the chain through
the API alone, since reading `audit_logs` directly would pass with the endpoint broken: a `PUT` that
audits itself carrying `X-Request-Id` is found by that id and narrows further with `auditEvent`, a
300-byte header is found by the log's clipped rendering of it and not by the raw value, and a request
with no header still carries the id chi generated and is searchable by it.

Folded in: the REST API page's sample request said `pageSize=20`, a name neither the handler nor the
OpenAPI operation reads. It says `size=20` now.

### Stage 4: the audit log viewer shows and filters by request id
Landed in `3949969e`.

The viewer handler reads `requestId` off the query and passes it to both `GetAuditLogsPaginated`
calls, the clamp's re-query included, so paging a filtered list cannot silently drop the filter. The
template gained a text input beside the event select, a request id column between Event and Details
showing a dash for an entry not written on a request, a clear link that appears for either filter,
and a filter script that builds its target through `URLSearchParams`. Two catalog strings in each
catalog.

The id is client-chosen, so nothing concatenates it. The paginator's base link is now built in the
handler with `url.Values` rather than with `printf` in the template: `addUrlParam` parses the string
it is handed, so an id holding `&page=9` would have arrived as a second parameter and every page link
would have carried two, with the handler reading the one the id smuggled in. `apiclient` escapes
`auditEvent` beside `requestId`, the ride-along the plan put here; event values come from a fixed
list, so a test case carrying a separator as the event was added so that escaping rests on a test
rather than on the list staying fixed. The echoed id is rendered by `html/template`, pinned by a case
asserting `"><script>` comes back escaped.

Tiers: unit green (1717 pass, 0 fail), lint green on all four modules, data green on sqlite (396
pass, 0 fail). Integration not applicable: no auth server handler, route or schema moved, and the
console is invisible to that tier. Six mutations, all `MUTATION CAUGHT`.

Two departures from the plan's sketch, both forced by the code. The paginator link moved into Go for
the reason above. Seam 8's HTML assertions live in `internal/rendertest` rather than in the page
test, because that test drives a mocked `HttpHelper` and renders nothing, so "renders escaped" and
"renders a dash" are not expressible there; `rendertest` renders the real template over the real
funcmap and fails on a leaked catalog key, which holds the two new strings for free. The template's
filter script is executed by no tier -- `rendertest` parses HTML and is not a browser -- so it was
checked by reading, as the plan review recorded it would be; the server-side half of the same
property is pinned by the link builder's and the API client's cases.

Also on the same docs page: the REST API bullet said the endpoint has an "optional `auditEvent`
filter", which stage 3 falsified. It names both filters now.

**CI was red on this stage, fixed in `1f20a132`.** The `Lint` job's `Generated CSS is committed`
step failed on `3949969e`: the new filter row uses `w-5/12`, which no admin console template had
used before, so the committed `web/static/main.css` -- generated by Tailwind from the class names
the templates use -- held no rule for it and the two filter cells would have shipped with no width.
Regenerated with the pinned tailwindcss v4.3.3; only that rule moves. That step was CI-only, which
is why every local tier passed, so `./run-tests.sh --type lint` now runs the same regeneration and
comparison, proven able to fail against the stale file.

### Folded in at the final review gate: the 500 records' filter values were unbounded

`79024ea7`. The final review raised one follow-up, and it turned out to sit on a line stage 3 had
just edited, so it was folded in rather than drafted.

`HandleAPIAuditLogsGet` put the caller's `auditEvent` query value straight onto its 500 log record,
and `HandleAPIUsersSearchGet` did the same with `query` at three sites. Both come off the query
string, so both are as long and as arbitrary as the client makes them: a 4096-byte value landed
whole, with whatever control characters or line separators it carried. #159 put every other
client-chosen scalar attribute through `logging.FieldForLog`, which escapes then clips at 128 bytes
with a counted marker, so these four were the last path where a request's own query string reached
the log verbatim -- and it left the audit-logs record reading half bounded and half not, since stage
3 had put the new `requestId` filter on that very call through `FieldForLog` while `auditEvent`,
three arguments to its left, went on raw.

The family was enumerated before folding, not assumed: a sweep over both servers' `handlers`, both
`middleware` packages, `handlerhelpers`, `apiclient` and `apiresponse` for a request-derived string
used as a log attribute returned exactly these four sites. Nothing an ordinary request logs changes;
a value with nothing to escape and nothing to clip comes back unaltered, which the third new test
pins.

`apihandlers/api_error_log_bounds_test.go` is new: three cases at the two handlers' error paths,
asserting the escaped-and-clipped attribute written out literally rather than through `FieldForLog`
itself, so removing the bound fails them instead of following them. Unit 1720/0, lint green, data
(sqlite) 396/0, integration (sqlite) 1039/0. Two mutations, one per handler, unwrapping
`logging.FieldForLog`: both caught.

Corrected by `4ca584a1`, which round 2 of the review bought: three of the four bounded sites are on
`HandleAPIUsersSearchGet`, and the one test case reached only the first of them, because the other
two sit behind a successful search. Two more cases, one per branch. The production change did not
move; the coverage did. Unit 1722/0.

## Tiers

Green on `4ca584a1`, the final commit.

| Tier | Where it ran |
|---|---|
| Unit, all four modules, plus `Unit / race` | locally at every stage, and in the check suite |
| Lint (`golangci-lint`, `sloglint`, the tree lints, the Tailwind regeneration) | locally at every stage, and in the check suite |
| Data, sqlite | locally at every stage |
| Integration, sqlite | locally at every stage |
| **Data and integration on mysql, postgres and mssql** | **check suite only** — there is no local four-engine tier, so this is where migration 000045, the byte-exact SQL Server predicate and the collation pins are established |
| Build validation, Vulnerabilities, Versions | check suite only |

Check suite on `4ca584a1`: thirteen of thirteen pass.
[Run 34716273694](https://github.com/leodip/goiabada/actions/runs/34716273694).

`check-anchors.sh`: 46 of 46 settled. Mutations: each stage ran its own, and the final review re-ran
five of its own design against the tree as it now stands, because four stages landing in sequence had
left every stage's mutation evidence recorded against a tree that had since moved. All caught.

Three, all recorded in full in the log and none of them a narrowing:

1. **The SQL Server count ceiling that section 4.3 described is gone.** The plan-gate review showed
   the planned shape gives an empty page with a total of 1 for a filter differing by a trailing
   space on live SQL Server. Stage 3 carries a byte-exact predicate on both the page and the count
   instead, so SQL Server has no ceiling. A narrower MySQL one remains and is reported in the
   fold-ins comment.
2. **Six fold-ins**, two decided at the seal (the orphan mock, #212 item 2) and four found while
   building (the migrate tests' schema head, the schema-parity digest, the REST sample's `pageSize`,
   the unbounded 500-record filter values). Each is listed with its commit and its justification in
   the fold-ins comment.
3. **`rollbackFloor` stays at `000044` while the head moves to `000045`**, so the floor is one below
   the head for the first time. Deliberate: the floor moves only when a release makes a one-way data
   change, and this migration is not one. Two comments that had asserted the two were equal were
   rewritten.

## Review

Reviewed by `pi openai-codex/gpt-5.6-sol` at effort `xhigh`. One round at the plan gate before any
code, and two at the final gate over the whole branch — `LEO_REVIEW_MODE=end`, so no stage was
reviewed on its own.

|---|---|---|---|---|
| 1 | plan | 3 | 0 | all fixed in the plan, before implementation |
| 1 | final | 0 | 0 | clean; all three axes `reviewed` |
| 2 | final | 1 | 0 | fixed, `4ca584a1` |

- **significant, conformance (plan):** the `requestId` exact-match guard ran after pagination and
  `COUNT(*)` was never guarded, so SQL Server would show a total one higher than the rows listed.
  Answered before any code: stage 3 carries a byte-exact predicate on both queries, with a
  four-engine data case at page size 1.
- **significant, quality (plan):** the audit-context lint matched `context.Background` by spelling,
  so an aliased import would have passed. Answered before any code: the qualifier is resolved by
  import path, with aliased fixtures refused.
- **minor, conformance (plan):** the docs promised a request id on every console record, which the
  startup backfill's does not carry. Qualified to an event raised while serving a request.
- **significant, quality (final round 2):** two of the four newly bounded log sites were bounded but
  unpinned — removing either bound left every test green. Confirmed by reproducing the mutation here
  (**SURVIVED**), fixed in `4ca584a1`, and the same mutations now come back **CAUGHT**.

Nothing was refuted at any round. Full account, with the evidence and the executed scope of the
clean verdict, in [this comment](https://github.com/leodip/goiabada/pull/329#issuecomment-5648453248).

## The three closing comments

- [Deferred decisions](https://github.com/leodip/goiabada/pull/329#issuecomment-5648426937) — none;
  all nine decisions were settled at the seal, and the two that went against the run's recommendation
  are restated there.
- [Follow-ups and fold-ins](https://github.com/leodip/goiabada/pull/329#issuecomment-5648431585) — no
  follow-ups to file, six fold-ins with their commits, and the one ceiling this branch adds.
- [The review account](https://github.com/leodip/goiabada/pull/329#issuecomment-5648453248) — all
  three rounds, every finding and its disposition, with the mutation evidence.

## For the release notes

- **Migration `000045` adds `audit_logs.request_id`** on SQLite, MySQL, PostgreSQL and SQL Server.
  It is applied automatically by Goiabada's own migrator at startup; there is no operator step, and
  nothing breaks. Rows written before the upgrade, and the startup backfill's row, carry the empty
  string, which the viewer shows as a dash meaning "not written on a request". Rolling the schema
  back below 000045 drops the column and loses the ids recorded in it; `rollbackFloor` is unchanged
  at 000044, so rollback remains supported.
- **The request id shown in the audit log viewer is client-supplied when the client supplies it.**
  Goiabada adopts an inbound `X-Request-Id` header and generates one only when it is absent — the
  behaviour it already had for the application log, now persisted and searchable. It therefore ties
  an event to a log line; it is not proof of who sent the request. A deployment that wants the value
  to be trustworthy should have its proxy overwrite the header rather than pass it through. The
  stored value is escaped and clipped exactly as the log renders it, so it is always printable ASCII
  and at most 161 bytes however long or strange the inbound header.


leodip:
## Deferred decisions

**None.** All nine decisions in section 3 of the agreement were settled at the seal, before any code
was written, and no stage raised a new one. Nothing in this branch was built on an assumption the
run took to keep moving, so there is nothing here for you to confirm or reverse after the fact.

For the record, the two that went against the run's own recommendation, since those are the ones
worth re-reading if you want to revisit anything:

- **Decision 8 (the surface).** The run recommended display only; you chose to add the `requestId`
  filter to the API and the viewer. That brought the index on four engines, the
  `GetAuditLogsPaginated` signature change across five implementations and the data mock, the page
  control and its two catalog strings. All landed in stages 3 and 4.
- **Decision 9 (whose request id).** The run recommended adopting `X-Request-Id` only from trusted
  proxies; you chose to keep chi's behaviour unchanged and asked for proper handling before the id
  is stored. What that handling is, in full: the id reaches the row and the page only as
  `logging.FieldForLog` renders it (printable ASCII, at most 161 bytes), through a parameterized
  insert, rendered by `html/template` on the page and by the JSON encoder on the API; the filter
  value is a parameterized query argument and is never stored. What stays open by choice, and is
  said in one clause in the docs, is that the id is a claim the client made and not proof of which
  request wrote the row.

leodip:
## Follow-ups and fold-ins

### Follow-ups to file: none

Everything this run found beside the work was folded into it rather than deferred, per the fold-in
tests in the agreement's section 9. The tracker was re-searched at the closing pass
(`audit request_id`, `audit_logs request id`, `pageSize rest-api sample`, all states) and turned up
only #328 itself and the two closed issues this change builds on, #320 and #283. Nothing is left to
file, so there is no `gh issue create` command in this comment.

### Folded in, six items

Each is production or test code in this branch that the agreement did not originally scope. They are
listed here because a fold-in you were never shown is a fold-in you cannot object to.

1. **The orphan `AuditLogger` mock at `src/core/audit/mocks/`.** Decision 4, landed in `6d2cd400`.
   A second generated mock that no `.mockery.yaml` produced and no file imported. It compiled only
   because a mock asserts nothing about the interface it was generated from, so this change would
   have left it a silently stale copy of a signature that no longer exists. Deleted whole, directory
   included, in the stage that regenerated the live mock.

2. **#212 item 2, the per-event settings row read.** Decision 5, landed in `6d2cd400`. Every audit
   event cost a full settings row fetch to read two booleans, while on every `/auth`, `/account` and
   API route that row was already on the request context. Once `Log` had the context, reading it was
   one type assertion plus a nil check; the fallback is today's row read, taken on the five root
   registrations and three limiter tiers that never pass through `MiddlewareSettings`. **#212 stays
   open** for its item 1, bounding or queueing the audit write, which is out of scope here.

3. **The migrate subcommand's tests pinned the schema head at `000044`.** `e486a7c9`. Eighteen
   assertions stated the head as a literal, so the first migration to land after it failed six
   tests. Bumped to `000045`. `rollbackFloor` was deliberately left at 44 — its own comment says it
   moves only when a release makes a one-way data change, which 000045 is not — so the floor is one
   below the head for the first time and two comments asserting they were equal were rewritten.

4. **The schema-parity allowlist's count and digest.** `e486a7c9`. `audit_logs.request_id` is `TEXT`
   on SQLite against `varchar(256)` elsewhere, which is the existing "sqlite: a string column
   carries no declared length" idiom. Count 91 to 92, digest to `4dbc0da396ffd7d1`. The guard exists
   so a new column is admitted under a named idiom deliberately rather than passing silently.

5. **The REST docs sample named `pageSize`, the endpoint reads `size`.** `7fc63f11`. Found by the
   reviewer at the plan gate. `rest-api.mdx` showed
   `?page=1&pageSize=20&auditEvent=user_login` while `HandleAPIAuditLogsGet` reads `size` and
   `openapi.yaml` declares `size`, so a caller copying the sample silently got the default page
   size. Pre-existing; folded in because stage 3 edits that same sample line to add `requestId`.

6. **The 500 log records' client-chosen filter values were unbounded.** `79024ea7`, corrected by
   `4ca584a1`. Raised by the reviewer at the final gate, verified, and taken here rather than
   drafted. `HandleAPIAuditLogsGet` wrote the caller's `auditEvent` query value straight onto its
   500 record, and `HandleAPIUsersSearchGet` wrote `query` the same way at three sites — all of them
   as long and arbitrary as the client makes them (4096 bytes captured verbatim in the reviewer's
   scratch assertion). #159 put every other client-chosen scalar through `logging.FieldForLog`, so
   this was the last path where a request's own query string reached the log raw. It also made the
   audit-logs record read half bounded and half not: stage 3 had put the new `requestId` filter on
   that very call through `FieldForLog` while `auditEvent`, three arguments to its left, went on
   raw. Four call sites wrapped. The family was enumerated by an AST-shaped sweep rather than
   assumed, and the sweep was falsified against the pre-fix tree — four hits at `79024ea7^`, zero
   now.

`src/core/data/commondb/audit_log.go:169` — the count query cannot apply the page's Go-side
exact-match guard, which needs the rows, so on MySQL a stored id differing from the filter only by a
collation fold at equal length (a decomposed accent against its precomposed spelling) would be
counted and not listed, showing as a page total one higher than the rows shown. SQLite and
PostgreSQL have no such fold; SQL Server's are closed in the statement by `mssqldb`'s own
byte-exact body. Revisit when a second writer can store a non-ASCII id, or when the total is read
for anything but the pager; a MySQL-only binary predicate beside this equality is the next shape.

Nothing reaches it today, and that is measured rather than argued: the only writer stores the id as
`logging.FieldForLog` renders it, which percent-escapes every byte outside printable ASCII, and the
reviewer ran a MySQL assertion over `utf8mb4_0900_as_cs` covering the empty string and every one-
and two-byte printable ASCII fragment — the whole alphabet the column can hold — finding distinct
collation weights throughout.

leodip:
## The review account

Under `LEO_REVIEW_MODE=end` the whole change was reviewed once, after the last stage landed, rather
than stage by stage. This is that reading in full, plus the plan gate that preceded any code. The
verdicts live in a gitignored directory that the next run overwrites, so this comment is the only
surviving record of what an independent reader found here.

Reviewer throughout: `pi openai-codex/gpt-5.6-sol` at effort `xhigh`.

| Round | Gate | Scope | Findings | Blocking | Outcome |
|---|---|---|---|---|---|
| 1 | plan | the plan, before any code | 3 | 0 (2 significant, 1 minor) | all three fixed in the plan before implementation |
| 1 | final | whole branch vs `origin/main`, 7 commits, 144 files | 0 | 0 | clean |
| 2 | final | one commit, `79024ea7` | 1 | 0 (1 significant) | confirmed and fixed, `4ca584a1` |

---

### Plan gate, round 1 — fresh session, request `328-20260912T144413Z`

Verdict: findings, 3. All three confirmed and fixed in the plan, so none of them ever reached the
code.

1. **The `requestId` exact-match guard runs after pagination, and `COUNT(*)` is never guarded.**
   Conformance, significant. The reviewer's own probe showed a stored `corr` and a filter `corr `
   give an empty page with total 1 on the live SQL Server under the planned shape
   (`probe/sqlserver_filter_total.sh`). The run's answering probe showed equality plus a `VARBINARY`
   cast, on both the page and the count, keeps exactly the byte-equal row for the exact value, a
   trailing space, two accent spellings and a zero-width space
   (`probe/sqlserver_binary_predicate.sh`). Stage 3 carries the byte-exact predicate on SQL Server
   and a four-engine data case with both guard rows at page size 1. The count ceiling the design's
   section 4.3 had described for SQL Server is gone as a result; what remains is the narrower MySQL
   one reported in the fold-ins comment.
2. **The Console docs row promised a request id on every console record, and the startup backfill's
   has none.** Conformance, minor. Fixed in the plan: the stage 1 docs line qualifies it to an event
   raised while serving a request.
3. **The audit-context lint matched `context.Background` and `context.TODO` by spelling, so an
   aliased import would pass.** Quality, significant. Fixed in the plan: the lint resolves the
   qualifier by import path through the package's existing `qualifiedCall`, with aliased fixtures
   planted and refused.

Nothing refuted, nothing deferred, nothing escalated. `unchecked` held five entries, all of the form
"there is no implementation yet"; each is answered in `log/plan.md`.

---

### Final gate, round 1 — fresh session, request `328-20260912T185124Z`, 19:14:38Z to 19:29:14Z

**Verdict: clean. Zero findings.** All three axes `reviewed`. What it executed, so the absence of
findings has a stated scope: the complete 144-file diff against `origin/main`, classified by file
class; `check-anchors.sh`, 46 anchors; `where.sh ci`, green at `1f20a132`; `where.sh migrations`;
focused container tests across core `auditlog` and `testutil`, the authserver's `audit` and
`handlers`, and the admin console's client, handler and real renderer; the sqlite data tier for
`Migration000045_AuditLogsRequestId`, `GetAuditLogsPaginated_RequestIdFilter` and `CreateAuditLog`;
the sqlite integration tier for `TestAPIAuditLogsGet_RequestIdFilter`; and a MySQL collation
assertion over `utf8mb4_0900_as_cs` covering the empty string and every one- and two-byte printable
ASCII fragment, which is the whole alphabet the new column can hold, finding distinct collation
weights throughout. That last one is executed evidence that the documented MySQL count ceiling
cannot fire for any value this change can persist, which the run had argued but not measured.

**It re-ran the mutation evidence.** The run's four stage log entries each recorded mutations
against the tree as it stood at that stage, and `tree-hash.sh --verify` at the moment the request
went out showed all four entries stale — which is what four stages landing in sequence produces, but
it made the mutation evidence the weakest claim in the record. The tier claims were re-established
for free by CI at HEAD; the mutations were not. The reviewer wrote five of its own against the tree
as it now stands:

| Mutation | Result |
|---|---|
| remove the persisted context id from `audit.go` | four audit row cases fail |
| reverse the lint predicate in `audit_context_lint.go` | its five fixture refusals fail |
| `context.Background()` at a real handler call in `handler_token.go` | the tree lint fails |
| remove SQL Server's byte-exact predicate | page and total exactness fail on the mssql data test |
| omit the row request id from the viewer template | the real localized render test fails |

So the weakest claim in the record is now the strongest, and it is current.

**`unchecked`, both entries answered.**

1. *No interactive browser or visual responsive-layout pass.* **Dismissed.** This repository has no
   browser tier and no JavaScript execution tier — the two tests named with `JS` in
   `static_lint_test.go` are source-text assertions, not a JS engine — so there is nothing this
   change could have run and nothing it withheld. The gap is the repository's standing shape, not
   this change's. What the risk actually rests on was covered by tiers that did run: the escaping by
   the real localized render test (and by the reviewer's own mutation of the viewer cell, which it
   caught), the URL building by the API client's `httptest` case. The script added is eighteen lines
   of `URLSearchParams` plus one `window.location.href`, the same shape as the event-select filter
   beside it.
2. *The repetitive mock expectation and handler call-site edits were checked by classification,
   greps, compilation, CI and representative seams rather than file by file.* **Covered — this is
   the design's own answer rather than a concession.** Seam 3 says in as many words that four cases
   are written and not 126, because the other 122 are held by the compiler, which requires a context
   argument at every site, and by `AssertAuditLogContext`, which refuses `context.Background()` and
   `context.TODO()` anywhere under the request-path directories. The reviewer then established that
   boundary itself rather than taking the argument: its `handler_token.go` mutation put a real
   `context.Background()` at a real call site and the tree lint caught it.

**Round 2 was requested for one reason, and it was not the findings.** The verdict was clean and
nothing in it was outstanding. But the round-1 follow-up had been folded in as production code
(`79024ea7`, the 500-record bounds), and under `LEO_REVIEW_MODE=end` the change gets exactly one
reading — shipping four newly edited production lines past the end of it would defeat that.

---

### Final gate, round 2 — session resumed, request `328-20260912T194741Z`, scope the single commit `79024ea7`

Verdict **findings**, 1. All three axes `reviewed`, `unchecked` **empty**, follow-ups **none**,
bookkeeping **none**.

**Finding 1 — quality, significant, confidence high. Confirmed and fixed in `4ca584a1`.**

*Claim.* `TestHandleAPIUsersSearchGet_ErrorRecordBoundsTheQuery` reaches only the initial
`SearchUsersPaginated` failure. `79024ea7` separately wrapped `query` on the `GetGroupById` and
`UsersLoadGroups` error records too, and removing either of those bounds leaves every test green. The
production code is correct; the coverage is not.

*Verified, not taken on trust.* Reading the handler settles the mechanism — the two annotation
records sit behind a successful search, and the existing case fails at the first database call, so it
can never reach them. Then reproduced by mutation on this tree rather than relying on the reviewer's
run: replacing `logging.FieldForLog(query)` with `query` on the "failed to get group by ID" record
came back **MUTATION SURVIVED**. The finding is exactly right.

*Fixed.* Two cases in `api_error_log_bounds_test.go`, one per branch. Each lets the search return one
user, asks for group-membership annotation, then fails that branch's own call, and asserts the
`query` attribute equals the escaped-then-clipped string **written out literally** rather than
computed through `FieldForLog` — which is what makes the case fail when the bound is removed instead
of following it. Each also pins the neighbour attribute its record carries, `group_id` and
`user_count`, at the int64 slog widens every integer to.

*Evidence the fix is worth its green.* The same two mutations, re-run after the cases landed:

| Mutation | Before the fix | After |
|---|---|---|
| "failed to get group by ID", drop the bound | **SURVIVED** | **CAUGHT** |
| "failed to load user groups", drop the bound | not run | **CAUGHT** |

All four sites of the family are now both bounded and pinned, where two were only bounded.

*The family, not the instance.* The finding's `why_exhaustive` claimed an AST scan found five
request-derived values at four log calls and that these two were the complete remainder. Rather than
accept that, the sweep was re-run independently and, more usefully, **falsified against the pre-fix
tree**: a scan for a variable assigned from `r.URL.Query().Get`, `chi.URLParam`, `r.PostFormValue`,
`r.FormValue` or `r.Header.Get` and then used as a log attribute value returns **four** such values
at `79024ea7^` and **zero** on the current tree. A sweep that finds nothing is worthless unless it
can be shown to find something, and this one finds precisely the sites the fold-in fixed.

*Why there is no round 3.* This round moved tests only: `tree-hash.sh --split` reads
`production 9ca396f61669` both before and after, with only `unit-tests` changed
(`af31bdce2e09` to `23dba34c2646`). Under the reviewer reference's section 5 that does not fund
another round, and with `unchecked` empty and no follow-ups raised there is nothing left to put in
one.

**Nothing was refuted at any round.** Every finding across all three rounds was confirmed, and each
was either fixed in the plan before code existed or fixed in a named commit.

PR #330, opened 2026-09-12
feat(sessions): replace mileusna/useragent, keep the raw User-Agent on the session

### LINKED ISSUE #281: Replace mileusna/useragent: prefer Client Hints, keep the raw header, stop baking a parser's guess into the row

leodip:
Closes #281.

## What this change is for

Goiabada stored three labels per user session, `device_name`, `device_type` and `device_os`, produced by `github.com/mileusna/useragent` from the request's User-Agent. The issue asked for the dependency to go. Grounding found that the three labels were not cosmetic: `StartNewUserSession` deletes every other session of the same user whose three labels and IP address match the new one, so the parser's exact output decided which sessions replaced each other.

This change moves that "same device" key onto the raw header and the IP address, stores the raw `User-Agent` on the session row, derives the display labels from User-Agent Client Hints where a browser sends them and from a short fixed User-Agent parser otherwise, exposes the raw header through the session API and as a tooltip on the three session pages, and removes the dependency from all three modules.

## The rules it establishes

- **Same device** means the same raw User-Agent (bounded to 512 bytes) and the same IP address. The three labels are display only and may change shape freely.
- **Every string written from a request header is bounded to its column width on a rune boundary**, after UTF-8 repair, by one helper, `useragent.Bound`. `codes.user_agent` gets the same treatment, which closes a live 500 on `/auth/issue` for a header of 513 bytes or more.
- **The label parser is allowed to be wrong and must not grow into a database.** Client Hints are read per RFC 8942 and the UA-CH draft (ignore what is not understood, accept any brand in any order); the fallback is an ordered list of seven browser tokens and seven platform substrings.
- Sessions created before the upgrade carry an empty `user_agent` and keep their old labels. They are not swept by a later login and expire on their own.

## Status

**Complete.** All three stages landed and were traced, the whole branch was reviewed at a single
end-of-run gate over four rounds, and the check suite is green on `cd0d08de`, which is HEAD:
[run 34719319348](https://github.com/leodip/goiabada/actions/runs/34719319348), all thirteen jobs —
`Tests /` on sqlite, mysql, postgres and mssql, `Build validation`, `Lint`, `Vulnerabilities`,
`Versions`, `Unit / race` and the four `Unit /` module jobs.

Three closing comments carry what the body cannot:

- [Deferred decisions](https://github.com/leodip/goiabada/pull/330#issuecomment-5648855997) — **none**, plus the two places where the sealed agreement was overridden inside the run.
- [Follow-ups, fold-ins and ceilings](https://github.com/leodip/goiabada/pull/330#issuecomment-5648858943) — **nothing left to file**, three defects folded in, no ceilings. The issue that comment drafted was folded in afterwards, at the user's direction, and the fold was then widened to the whole tier; see stage 4.
- [The review account](https://github.com/leodip/goiabada/pull/330#issuecomment-5648876442) — every round, every finding, and the evidence that answered it.

### Tiers

|---|---|---|
| unit, three modules | local, dev container | 1747 pass, 0 fail |
| lint, four modules | local | clean |
| data | local, **sqlite only** | 397 pass, 0 fail |
| integration | local, **sqlite only** | 1040 pass, 0 fail |
| `where.sh migrations` | local | all rules pass on all four migration sources |
| `check-anchors.sh` | local | 20 of 20 rows resolve |
| **data and integration, four engines** | **the check suite**, on `cd0d08de` | green on sqlite, mysql, postgres and mssql |
| race detector, govulncheck, version check, build validation | **the check suite** | green |

The four-engine data and integration result for this branch is CI's and nowhere else, which is why
it names its commit. The `Tests / mysql` job first failed on `cd0d08de` inside
`setup-test-container` with exit 35, skipping its build and both test steps; the rerun succeeded and
all four of its steps are recorded green in the review comment.

### Review

Reviewed by `openai-codex/gpt-5.6-sol` at effort `xhigh`, **4 rounds over the whole branch**. No
stage was reviewed on its own (`LEO_REVIEW_MODE=end`); a fifth round was available and was not
needed. A plan-gate round ran before any code existed and its six findings were answered by editing
the plan.

|---|---|---|---|
| 1 | 3 | 1 | all fixed, `09c3788f` |
| 2 | 2 | 0 | both fixed, plus a crash the fix introduced, `55993a7d` |
| 3 | 2 | 0 | one fixed `cd0d08de`, one an evidence repair |
| 4 | 1 | 0 | resolved by a CI rerun; no code changed |

- **blocking, security:** a device label is attacker-chosen text (UA-CH 3: a server "MUST accept arbitrary values") that reached an `innerHTML` sink through the End Session modal on two session pages, so markup stored in your own session's label ran in the origin of any administrator who listed it. Fixed in `09c3788f`: both call sites pass `escapeHtml(device)`, the helper already in `static/utils.js` and already pinned by `static_lint_test.go`.
- **significant, conformance:** the Client Hints reader accepted several malformed structured fields and refused two valid ones — a parameter key read on the HTTP token grammar, three hints read with `Header.Get` (first field line only), a base64 alphabet check with no decode, and no Date or Display String branch, which UA-CH now requires via RFC 9651. Fixed in `09c3788f` and `55993a7d`. Refusing a valid field is not neutral here: it falls back to Chromium's frozen User-Agent and reports the wrong platform.
- **significant, quality:** fixing the Display String branch introduced a panic in `Labels` for a `Sec-CH-UA` ending at `v=@` — a header anyone can send. Fixed in `55993a7d`, bounds-checked at the indexing site, and covered by a test that cuts a header carrying every bare-item form at every byte in turn.
- **significant, conformance:** a `v` parameter of any structured-field type but a String was rendered into `DeviceName` as its serialization (`Chrome %"12%2e3"`). Fixed in `cd0d08de`: only a String carries a version, and no other form answers a value at all.
- **significant, quality:** the verification record described the pre-rebase tree, and separately the required MySQL check had not executed on HEAD. Both were evidence repairs rather than code defects; the whole branch was re-run at the gate and the MySQL job was rerun to completion.

None of the nine findings was refuted or contested, and none was escalated.

### Deviations from the sealed agreement

- **The migration is `000046`, not `000045`.** `origin/main` took `000045_audit_logs_request_id` from #328 while this branch was in flight, and the branch was rebased onto it. Sections 0 to 5 are frozen at the seal and still read `000045`; `log/stage-1.md` carries the correction.
- **`userAgent` is `required` in `openapi.yaml`**, where decision 6 said it would not be. The plan review found the schema lists every always-emitted field as required, and the field has no `omitempty`. The wire is unchanged.
- **Three defects were folded in** rather than filed: the `codes.user_agent` truncation (named in goal 5 and decided at decision 5, before any code) and a dead `endSessionClick` parameter on the admin client sessions page that made its "you are ending your own session" warning unreachable. Both are set out in the follow-ups comment. The third, the integration tier's fixture addresses, is stage 4 below: it was drafted as an issue and then folded in on the user's instruction, against that comment's own reasoning, which stays there as written.
- Seam 7 recorded that no console session-handler test infrastructure existed, so the tooltip was to be covered by handler tests and reviewed by eye. Review round 1 built more than that: `rendertest` renders the three real session templates through the real renderer and funcmap, and asserts the escaping.

### Stage 1: the raw header column, and the sweep keyed on it

Landed in `d8cf29b3`.

`user_sessions.user_agent` exists on all four engines (migration `000046`), holding the request's
User-Agent as the browser sent it, repaired to valid UTF-8 and cut to 512 bytes on a rune boundary
by the new `useragent.Bound`. `StartNewUserSession` writes it and the "same device" sweep now
compares it and the IP address and nothing else; `device_name`, `device_type` and `device_os` keep
today's values and become display only. `mileusna` is untouched in this stage: the labels change
shape in stage 2.

The bound is in **bytes** rather than runes because one byte bound satisfies every engine at once:
PostgreSQL and MySQL count characters and a value of at most N bytes has at most N characters, and
SQL Server's `nvarchar` counts UTF-16 units where a 4-byte rune is two of them. The UTF-8 repair
runs before the cut, never after, because PostgreSQL and MySQL both refuse a value carrying a stray
latin1 byte (which RFC 9110 10.1.5 permits in a User-Agent) and repairing afterwards would push the
result back over the bound.

Pre-upgrade rows read `''` and are not backfilled; a later login from such a device does not sweep
its legacy row, which expires on its own, and a client sending no header matches every legacy row on
its address, as it does today.

**Folded in:** an over-long User-Agent made code issuance fail. `codes.user_agent` is
`varchar(512)` and `CreateAuthCode` wrote the header untruncated, so a browser sending 513 bytes or
more got a 500 from `/auth/issue` after a completed ceremony. It now writes
`useragent.Bound(input.UserAgent, 512)` at that one writer of the column.

Three test files outside the change's own surface needed edits, none behavioural: `migrate_test.go`
spelled the head schema version as a literal in nine places and now derives it from
`Migrator.Head()`; the SQLite unbounded-string allowlist rule in the schema-parity test is re-counted
by one; and `migration_000039`'s Go-layer assertion migrates to the head first, because
`models.UserSession` always names every column the head declares.

Tiers: unit green; **data green on mysql, postgres, mssql and sqlite**, including a new four-engine
test for the migration that round-trips 512 bytes of 4-byte runes; integration green on sqlite,
where the two "second device" fixtures that fake a second device by varying only the User-Agent pass
unchanged and are the sweep's regression guard at the endpoint; lint green; `where.sh migrations`
green. Seven mutations run against the new tests, all caught.

### Stage 2: `core/useragent` rewritten on the stdlib, and `mileusna` gone

Landed in `14f8cd99`.

`useragent.Labels(r)` replaces `GetDeviceName`, `GetDeviceType` and `GetDeviceOS`, and parses the
request once rather than three times. **Client Hints first:** Chromium freezes its User-Agent,
reporting `Windows NT 10.0` on Windows 11 and `Android 10; K` on every phone with the minor version
digits zeroed, so `Sec-CH-UA`, `Sec-CH-UA-Mobile` and `Sec-CH-UA-Platform` are the only truthful
platform source for the majority browser. Nothing but Chromium sends them, and only over a secure
context, so the ordered User-Agent list behind them is the whole story for Firefox, Safari,
plain-HTTP deployments and every non-browser client.

The labels change shape, which is now safe because stage 1 took the sweep off them:
`Chrome 120.0.0.0` becomes `Chrome 120`, `Windows 10.0` becomes `Windows`, `Safari 17.1 (iPhone)`
becomes `Safari 17`, and an unknown OS becomes the empty string rather than a single space. Brand
plus major version from **either** path, so the same browser reads the same over HTTP and HTTPS; no
OS version for the same reason; `Bot` is dropped, because no crawler completes a login ceremony.

**`Sec-CH-UA` is read as a structured field and refused when it is not one**, which RFC 8942 2.2
makes equivalent to an absent header: the caller falls through to the User-Agent rather than storing
a half-read value. The review rounds moved that reader onto RFC 9651, which is what the current UA-CH
draft cites, and added the two bare-item forms it dispatches on that RFC 8941 does not. Each label is
then cut to its own column: 256, 32 and 64 bytes.

The ordered token list is deliberately short and deliberately allowed to be wrong: a browser it does
not know falls through to the first RFC 9110 product token of the header, which is a worse label and
not a broken one. That is also what keeps the two integration fixtures that fake a second device
with an invented token telling themselves apart.

`github.com/mileusna/useragent` is now absent from all three `go.mod` and `go.sum`.

Tiers at the stage: unit green; integration green on sqlite; data green on sqlite; lint green.
Thirteen mutations, all caught — one of them only after two earlier attempts at the same gate
survived, because both had picked a malformed header that the parser refuses one step later anyway.
The row that pins it, and why it looks contrived, is spelled out in the test.

### Stage 3: the raw header exposed through the API and the three session pages

Landed in `27bc794c`.

The column was write-only until here. `UserSessionResponse` and `EnhancedUserSessionResponse` now
carry `userAgent`, filled by `ToUserSessionResponse` and by the three auth server session handlers
that build the enhanced shape by hand. The field has no `omitempty` and `openapi.yaml` lists it as
required, which is one claim in two places: a session created before the column existed carries an
empty string, and an empty value is an answer rather than an absent property. The document's
`EnhancedUserSessionResponse` is an `allOf` over `UserSessionResponse`, so that is one property
edit, not two.

The admin console carries it to all three session pages, where the Device cell gains a `title`
attribute holding the header. The three labels in that cell are a parser's guess, so two sessions
from different machines can read alike; the raw string is what tells them apart, which is the
recognisability the labels exist for. A tooltip rather than a fourth column keeps the table its
current width and adds no catalog string.

The header is a request header copied into the row verbatim, so it is attacker-chosen text rendered
on an admin page. A render test over the three real embedded templates asserts the escaping rather
than assuming it: a header carrying a quote and an angle bracket must appear escaped inside
`title="..."` and nowhere else on the page. Review round 1 found the sibling case this missed — the
*label*, equally attacker-chosen, reaching the End Session modal's `innerHTML` — and closed it.

Tiers at the stage: unit green; integration green on sqlite; data green on sqlite; lint green over
four modules. Six mutations, one per seam, all caught. The one worth naming is the json tag: the
integration tests decode with the same structs the server encodes with, so a tag renamed or given an
`omitempty` would leave every endpoint assertion green while every real client read nothing, and only
the map-decoding case at the unit tier can see it.

One unrelated flake was seen and not reproduced: two tests in `api_users_search_test.go` collided on
the fixed email addresses their fixture hardcodes. Two further runs of the tier on the same tree
were clean. It was drafted as a follow-up, then folded in as stage 4 below.

### Stage 4: every integration user fixture draws a unique address

Landed in `8e6500a0`. This is the drafted follow-up, folded in on the user's instruction and then
widened, also at the user's direction, from the one file to the whole tier.

`users.email` is UNIQUE and this tier spells its fixture addresses as literals, which is only safe
while every row carrying one is deleted again. Two things make that fail. **Within a run**, two
fixtures sharing an address collide the moment either one's cleanup is missed — the observed defect:
`createTestUsers` inserted three users at fixed addresses for five tests in that file, the inserts
were `assert.NoError` so a collision returned three users that were never created, and a failed
insert leaves `Id == 0` while the cleanup discarded the delete's error, so one missed cleanup
cascaded into every later caller and the run failed far from wherever it started. **Across runs**,
`run-tests.sh` drops the sqlite file on its EXIT trap but never drops `goiabada_integration` on
mysql, postgres or mssql, so a row leaked there fails the same test on every run afterwards. A
cleanup can always be missed — a `t.FailNow` between the insert and the defer, a panic, a killed run
— so the fix is on the address, not on the cleanup.

The five callers now use `createTestUsersWithSuffix`, which already sat beside the helper it
replaces, with its inserts promoted to `require.NoError`. 102 addresses across the rest of the tier
go through a new `uniqueEmail`, which keeps the descriptive literal so a leftover row still names the
test that wrote it. Eight sites that use an address twice on purpose draw it once into a variable;
the normalization case derives its uppercase padded input from the drawn address, so what it tests is
unchanged. Five literals stay, all in requests asserted to be rejected, where nothing reaches the
database and in one the length *is* the assertion. `users.username` is not unique and every subject
is `fake.UUID()`, so the address is the whole of the exposure.

The rule is enforced rather than documented: `TestIntegration_UserFixturesDrawUniqueEmails` refuses a
hardcoded address in any `models.User` literal in the tier, on the AST rather than the text, with a
floor so a walk covering nothing cannot pass. `TestIntegration_UniqueEmailDrawsADistinctAddress` pins
the helper itself, which nothing else can see — a `uniqueEmail` returning its argument leaves every
other test green. `TestAPIUsersSearch_FixtureIsIndependentAcrossCalls` builds the search fixture
twice with no cleanup between, which is the shape a missed cleanup leaves behind.

Tiers at the stage: integration green on sqlite, 1382 pass 0 fail, over four runs; lint green over
four modules. Four mutations, all caught — a reintroduced literal, a walk pointed at an empty tree, a
`uniqueEmail` that returns its argument, and a restored fixed address in the search fixture, the last
reproducing the original `UNIQUE constraint failed: users.email (2067)`.


leodip:
## Deferred decisions: none

Nothing in this run was left for you to settle. All seven decisions in section 3 of the agreement
were sealed `Decided` at the interview, no stage raised a new one, and no review finding escalated:
every finding across four review rounds came back either `judgement: false` with a populated
`forced_answer` (settled by the code) or was answered by evidence. The run never posted a blocking
question, so there is nothing here awaiting an answer before the pull request can merge.

Two places where the sealed text was overridden **inside** the run are listed here anyway, because
they change what section 3 says and you should not have to find them in a log entry.

| What | Sealed text | What landed | Why | Cost of reversing |
|---|---|---|---|---|
| `userAgent` in `openapi.yaml` | decision 6: "not in the required list, since legacy rows carry `''`" | declared **required** | The plan review (finding 2) confirmed the schema lists every always-emitted field as required, and the parenthetical confused an empty *value* with an absent *property*. The field has no `omitempty`, so the server always emits it. | One line in `openapi.yaml`. The wire does not change either way. |
| Migration number | decision 4 and section 0: `000045_user_sessions_user_agent` | `000046_user_sessions_user_agent` | `origin/main` took `000045_audit_logs_request_id` from #328 while this branch was in flight, and the branch was rebased onto it between review rounds 2 and 3. | A renumber of four SQL file pairs, four golden rows, the data test name and the migrate-version cases. Not worth reversing; `where.sh migrations` passes and reports `000047` free. |

Sections 0 to 5 are frozen at the seal and still read `000045`; `log/stage-1.md` carries the
correction, and review round 3's finding 7 is where it was caught and recorded.

leodip:

Everything this run found next to the work, in three blocks: what it left for you to file, what it
fixed here instead of filing, and what it capped deliberately.

---

### 1. Drafted, not filed — the user-search integration fixtures collide on fixed emails

Duplicate search re-run at the finish: `gh issue list --state all` over `createTestUsers`,
`integration fixture flaky email` and `endSessionClick` returns nothing. #136 is the closest, and it
is the data-tier sibling of this defect, closed, with a fix that does not reach this helper.

**Title:** `Flaky: the user-search integration fixtures share three fixed email addresses, and one missed cleanup cascades`

**Labels:** `bug`, `go`

**Body:**

> `TestAPIUsersSearch_NoResults` and `TestAPIUsersSearch_SpecialCharacters` failed once during a
> `./run-tests.sh --type integration --db sqlite` run on 2026-09-12, both with
> `unable to insert user: ... UNIQUE constraint failed: users.email (2067)` raised from
> `createTestUsers`. Two further full runs of the tier on the same tree were clean, and so was the
> run one stage earlier, so it is intermittent rather than a break.
>
> The mechanism is in the helper. `createTestUsers` in
> `src/authserver/tests/integration/api_users_search_test.go` inserts three users at hardcoded
> addresses: `john.doe@test.com`, `jane.smith@test.com` and `disabled@test.com`. Five tests in that
> file call it, each deleting its users in a `defer`. Three facts combine, none a bug alone:
>
> 1. **The addresses are fixed**, so any two callers whose rows overlap in the database collide.
>    Every other fixture in that package draws a unique address, usually from `fake.UUID()`.
> 2. **The inserts are asserted with `assert.NoError`, not `require.NoError`**, so a collision marks
>    the test failed and the helper keeps going, returning users that were never created.
> 3. **A failed insert leaves `user.Id == 0`**, and the cleanup is `_ = database.DeleteUser(nil, user.Id)`
>    with the error discarded, so it deletes nothing and does not say so.
>
> Together they make the failure cascade and outlive its cause: one caller that fails to clean up
> leaves rows behind, the next caller collides on them, cannot delete them either, and passes the
> same leftovers on. A run can therefore fail in a test far from whichever one first went wrong, and
> re-running clears it only because sqlite gets a fresh file per run. On mysql, postgres and mssql
> `run-tests.sh` never drops `goiabada_integration`, so leftovers there persist between runs.
>
> This is the integration-tier sibling of #136, which was the same class of defect at the data tier
> (`TestSearchUsersPaginated` colliding with accumulated rows in the never-reset shared databases)
> and is closed. The fix there does not reach this helper.
>
> **What closing it would take:** give `createTestUsers` unique addresses, in the shape of the
> `createTestUsersWithSuffix` helper that already sits beside it in the same file, and promote the
> three inserts to `require.NoError` so a fixture that did not get built stops its test instead of
> handing it three phantom users. The five call sites do not change: none of them searches for one of
> the three addresses by name.

**Why it was not folded in:** it rides on no surface this change touches. `api_users_search_test.go`
covers a different endpoint, and the only fixture #281 edits is `createTestUserSession`, which
inserts sessions. Fixing it here would put an unexplained hunk in a user-agent change.

**Files it:**

```bash
gh issue create \
  --title 'Flaky: the user-search integration fixtures share three fixed email addresses, and one missed cleanup cascades' \
  --label bug --label go \
  --body 'Body as quoted in https://github.com/leodip/goiabada/pull/330#issuecomment-5648858943'
```

---

### 2. Folded in rather than filed — two, both with their commit

**a. An over-long User-Agent made code issuance fail.** Commit `d8cf29b3` (stage 1).

`codes.user_agent` is `varchar(512)` and `CreateAuthCode` wrote the request header untruncated, so a
browser sending 513 bytes or more got a 500 from `/auth/issue` **after a completed ceremony** —
verified against PostgreSQL ("value too long for type character varying(512)") and MySQL ("Data too
long for column 'user_agent'") during grounding. It now writes `useragent.Bound(input.UserAgent, 512)`
at the one writer of the column. **Why it fit:** it is the same rune-safe bound this change creates
for `user_sessions.user_agent`, eight lines on a helper already being written, with nothing to
settle; a follow-up would have cost a whole run for it while the 500 stayed live. Named in goal 5 of
the agreement and decided as a fold-in at decision 5, before any code was written.

**b. The admin client sessions page could never warn you were ending your own session.** Commit `55993a7d`.

Reported by review round 2 as belonging to a different change. `admin_clients_usersessions.html`
declared `endSessionClick(elem, evt, userSessionId, email, device, isCurrent)` while its End Session
button called it with five arguments, `(this, event, UserSessionId, UserEmail, IsCurrent)`. The
boolean bound to `device` and `isCurrent` was always `undefined`, so
`adminconsole.admin_clients.usersessions.modal_current_warning` could never be appended although
`HandleAdminClientUserSessionsGet` sets the flag correctly. JavaScript pads a short call with
`undefined` rather than failing, so nothing reported it. `device` is dead in that file — it appears
in the signature and nowhere in the body — so deleting it from the signature makes the existing call
site correct. The same mismatch is on `main`, so it predates this change. **Why it fit:** it rides on
a file stage 3 already edits and a test file round 1 already added to, there is nothing to settle,
one deletion is the whole fix, and the reviewer itself offered it as foldable. Covered by
`TestRender_SessionPagesPassEveryArgumentEndSessionClickDeclares`, which *counts* each page's
rendered argument list against its rendered declaration on all three pages, so the same drift in
either direction fails there too. Restoring the six-parameter signature was run as a mutation and was
caught.

---

### 3. Ceilings left by this change

**none.** `git diff --unified=0 df19ad97~1..HEAD | grep '^+.*ceiling:'` over this run's seven commits
returns nothing.

leodip:
## The review account

`LEO_REVIEW_MODE=end` was in force, so **no stage was reviewed on its own**: the plan was reviewed
before any code existed, and the whole branch was then read four times at the final gate. This is the
only reading the code got, and the verdicts live in a gitignored directory that the next run
overwrites, so the account is pasted here in full rather than pointed at.

**Reviewer, every round:** `openai-codex/gpt-5.6-sol`, effort `xhigh`.

|---|---|---|---|---|
| plan | 1 | 6 | 0 (2 forced) | all six fixed in the plan before any code was written |
| final | 1 | 3 | **1** (security) | all fixed, `09c3788f` |
| final | 2 | 2 | 0 (both forced) | both fixed, plus a crash the fix introduced, `55993a7d` |
| final | 3 | 2 | 0 (both forced) | one fixed `cd0d08de`, one an evidence repair |
| final | 4 | 1 | 0 | resolved by a CI rerun; no code changed |

Nine findings at the final gate. **None was refuted and none was contested**; none was escalated to
you. The review converged rather than being capped — a fifth round was available and was not needed.

---

### Plan gate, round 1 — six findings, all answered by editing the plan

Request `281-20260912T145855Z`, fresh session. Verdict `static-review-exhausted`; conformance,
quality and security all `reviewed`. `unchecked` held four entries, all dismissed with reasons
recorded in `log/plan.md`. No code existed yet, so every disposition is a plan edit.

1. **conformance, minor.** The probe's Client Hints reader accepted invalid `sf-string` escapes and
   trailing bytes after a member. Confirmed in `probe/labels.go`. → stage 2 gates the grammar per RFC
   8941 3.3.3, one table row per gate.
2. **conformance, minor.** `userAgent` was planned as *not* required in `openapi.yaml`. Confirmed:
   the schema lists every always-emitted field as required. → stage 3 declares it required. This
   overrides decision 6's parenthetical, which confused an empty value with an absent property; the
   wire is unchanged.
3. **quality, significant.** The tooltip and the console client mapping had no output-level test.
   Confirmed: `rendertest.render` exists, `GetUserSession` was untested. → stage 3 gains three render
   rows and one client stub test.
4. **quality, significant, forced.** Two sweep fixtures kept the label key. Confirmed at
   `usersession_manager_start_test.go:339` and `:417`. → stage 1 step 6.
5. **quality, significant, forced.** The `Bound` tests did not force repair-before-cut. Accepted on
   the reviewer's scratch demonstration. → stage 1 step 1 gains `Bound("\xe9a\xe9a", 4)`.
6. **quality, significant.** The `userAgent` wire key was not pinned independently of its struct tag.
   Confirmed. → stage 3 gains a map-decoding row per response struct.

---

### Final gate, round 1 — three findings, all real, all fixed in `09c3788f`

Request `281-20260912T182448Z`, fresh session, started 18:26:50Z. The whole branch against
`origin/main`.

**The verdict was a salvage, and that governs how it reads.** The reviewer was cut off before it
reported and the file was rebuilt from `.review/findings.jsonl`, which it had appended to as it
confirmed each finding. `axis_coverage` therefore reads `skipped` on all three axes and `ran` is
empty — not because nothing was examined, but because the run never reached the point where it
records what it examined. **The three findings below are a floor, not the review's last word.**

**Finding 3 — security, blocking. A device label is attacker-chosen text reaching an `innerHTML` sink.**
Confirmed and fixed. `Labels` derives the name and OS from `Sec-CH-UA` and `Sec-CH-UA-Platform`, and
UA-CH 3 says a server "MUST accept arbitrary values for each", so any user can store markup in their
own session's label by completing a login with hand-written headers. On `account_user_sessions.html`
and `admin_users_sessions.html` the End Session button hands the label to `endSessionClick`, which
concatenates it into the message `showModalDialog` assigns to `innerHTML`. `html/template` escapes
the label for the JavaScript string literal in the `onclick` attribute and stops there, so the markup
ran in the origin of whoever opened the list — **including an administrator reviewing another user's
sessions**.

Verified rather than taken on trust: `escapeHtml` already exists in `static/utils.js` on
`origin/main`, is loaded by both layouts, is already the idiom at another `showModalDialog` call site
(`escapeHtml(err.error_description)`), and its body is pinned by `static_lint_test.go`. Both call
sites now pass `escapeHtml(device)`. The reviewer's `forced_answer` ruled out the two alternatives and
both rulings hold: source-side filtering would contradict the arbitrary-values requirement that is
decision 3's premise, and escaping inside `showModalDialog` would break its other callers, which pass
markup deliberately. The third session page is the negative and the reviewer had it right:
`admin_clients_usersessions.html` passes the user's email to its modal, not the device label.

**Findings 1 and 2 — conformance, minor.** Both made the reader accept a header that is not a
structured field and then label a session from it, where RFC 8942 2.2 says to ignore a hint the server
cannot understand and fall through to the User-Agent. That fallback is not a nicety: it is the only
path for Firefox, Safari and every plain-HTTP deployment.

1. `parseParam` read its key on the wider HTTP token grammar and its value by scanning to the next
   delimiter, so `"Google Chrome";V="120"` took `V` as a key and `v=@junk` became the version
   `@junk`. It now reads an RFC 8941 3.1.2 key — `( lcalpha / "*" ) *( lcalpha / DIGIT / "_" / "-" / "." / "*" )`
   — and a 3.3 bare item in all six forms, with the sf-integer and sf-decimal digit counts actually
   set, and `*SP` rather than OWS after the semicolon.
2. All three hints were read with `Header.Get`, which answers the first field line alone. A
   `Sec-CH-UA` legitimately split after its GREASE brand read as a one-brand list, so the session was
   labelled from the arbitrary value UA-CH 8.2 requires be present; two `Sec-CH-UA-Platform` lines
   read as the first rather than as the invalid Item they combine into. `fieldValue` now joins with
   `", "` before parsing, per RFC 9110 5.3 and RFC 8941 4.2, which makes a repeated Item hint fail its
   own gate and be ignored — the outcome 8942 2.2 asks for.

**Tiers on the fix:** unit 1723 pass / 0 fail; integration sqlite 1039 / 0; lint 0 issues; anchors
20/20. Four-engine data and integration are the check suite's.

**Mutations, and the one that survived:**

| Mutation | Result |
|---|---|
| `!isKeyStart(h[i])` dropped from the parameter gate | caught |
| `isKeyStart` widened to `isTokenChar` | **survived**, then caught (below) |
| `parseBareItem` given a case making `@junk` parse | caught |
| `fieldValue` reverted to `h.Get(name)` | caught |
| `escapeHtml(device)` reverted to `device` in the account template | caught |

The survival is the part of the round worth keeping. The row written to pin the key grammar used an
uppercase key, `"Google Chrome";V="120"`, and that row is refused whether the reader is strict or
not: `V` starts no key in RFC 8941 and continues one in neither grammar, so a widened reader takes an
empty key, the cursor never moves, and the member is refused by the list's comma gate instead — the
row passed against deliberately wrong code. The case where the two readers actually disagree *and the
parse runs on* is a key starting with a **digit**, which starts no key but does continue one: a
widened reader takes `1x` as a key, reads its value, finishes the list cleanly and labels the session
`Chrome 120`. One row, `"Google Chrome";v="120";1x=2`, closes it, and the same predicate covers `-`,
`.` and `_`. Re-run against that row, the mutation is caught. The row's comment records why it is the
only row that can catch it, so a later reader does not delete it as a duplicate of its neighbours.

**`unchecked`, one entry:** "the review was cut off before it reported, so what it covered is not
recorded." **Filed as the round's open question and answered by running round 2 rather than by
argument.** No session can reconstruct what a dead one read, and dismissing it would have dismissed
the possibility that the review had more to say — on a change whose one blocking finding was a stored
XSS this run had shipped through three unreviewed stages.

---

### Final gate, round 2 — two conformance defects, plus a crash the fix introduced; all in `55993a7d`

Request `281-20260912T192938Z`, started 19:31:45Z, completed 19:49:47Z. Verdict `findings`, and
unlike round 1 it reports its own coverage: `conformance`, `quality` and `security` all `reviewed`.
**That was the whole reason round 2 was asked for, and it is answered: the branch has had one
complete reading.** Both findings came back `judgement: false` with a populated `forced_answer`, so
both were the code's to settle.

**Finding 4 — conformance, minor. The reader still accepted malformed structured fields.** Confirmed
in two halves, and only after the normative text was fetched rather than recalled.

The wire-reachable half is `parseByteSequence`. RFC 9651 4.2.7 step 6 checks the base64 alphabet and
step 7 then says: *"Let binary_content be the result of base64-decoding b64_content, synthesizing
padding if necessary. If base64 decoding fails, parsing fails."* Only step 6 was implemented. A probe
over the tree as it stood answered `("Chrome 120", "Desktop", "Windows")` for
`Sec-CH-UA: "Google Chrome";v="120";other=:A:` against a Firefox User-Agent — one base64 character is
six bits, which is no whole byte — and the same for `:=A==:`. So a malformed hint was used where RFC
8942 2.2 says to ignore it.

The second half is top-level whitespace: RFC 9651 4.2 steps 2 and 6 discard *SP* around the whole
field, where OWS is 4.2.1's, discarded around a list's commas. **Not wire-reachable**, and the
reviewer said so itself, having run a raw HTTP/1 probe establishing that Go strips field-line edge
whitespace before constructing a server request. Fixed anyway, with its two test rows labelled as
pinning the grammar rather than a wire case, because the alternative is a comment claiming
equivalence to an algorithm the code does not implement.

**Finding 5 — conformance, minor. The reader targeted the RFC UA-CH no longer uses.** Confirmed, and
the premise was checked at the source: the current UA Client Hints draft's 3.1, 3.7 and 3.9 each now
cite **RFC 9651**, which obsoletes RFC 8941 and whose 4.2.3.1 dispatches on eight leading characters
rather than six — step 6, `"@"` for a Date, and step 7, `"%"` for a Display String. `parseBareItem`
had neither branch, and the probe confirmed the cost: `;seen=@1659578233` and `;note=%"hello"` each
made an otherwise ordinary Chrome-on-Windows hint fall through to the User-Agent and report
`("Firefox 121", "Desktop", "Linux")`. Falling back is not neutral here — Chromium freezes its
User-Agent, which is decision 2's entire premise — so this is a valid field answered with a wrong
platform. Both forms are now parsed by the 4.2.9 and 4.2.10 algorithms as written, including the two
rules easiest to get wrong: a Date whose integer is a Decimal fails (4.2.9 step 4), and a Display
String's pct-encoding takes lowercase hex only, with a backslash standing for itself rather than
escaping.

**One deliberate narrowing, recorded because it is not obvious.** `parseDisplayString` answers the
source span, not the decoded value. 4.2.10 rejects anything outside VCHAR and SP in the *encoded*
text but places no limit on what the octets decode to, so a display string is a valid field whose
value can be a NUL byte — and PostgreSQL refuses a text value carrying U+0000 outright, which would
have turned a crafted header into a failed session insert.

**The third defect, which this round's own fix introduced.** Fixing finding 5 added a crash:
`parseDate` discards the `@` per 4.2.9 and handed `parseNumber` an index one past the end, and
`parseNumber` indexed `s[i]` unguarded, so a `Sec-CH-UA` ending at `v=@` panicked inside `Labels` — a
header anyone can send, on the session-creation path. It was found by acting on the round's second
`unchecked` entry rather than dismissing it. Bounds-checked in `parseNumber`, at the site that does
the indexing rather than at its two callers. Covered by `TestLabels_NoPrefixOfAHintCanPanic`, which
cuts a header carrying **every** bare-item form at every byte in turn and asserts `Labels` answers,
twice, with the same triple — that is the family ("a look-ahead reading past a field the sender ended
early") rather than the one row, so a look-ahead added to any form's parser is covered on the day it
is written.

**`unchecked`, both answered:**

1. *"No interactive browser pass executed the rendered modal DOM."* **Dismissed.** The reviewer's own
   reason stands and the evidence is in the tree: `rendertest` renders the three real session
   templates through the real renderer and funcmap, `escapeHtml` is pinned independently, and round
   1's mutation removing `escapeHtml(device)` was caught.
2. *"No differential fuzz campaign against an independent full RFC 9651 implementation was run."*
   **Covered, and it paid.** A differential campaign needs a second implementation, which means the
   dependency this issue exists to remove, so that literal form was never available. What it pointed
   at — that nothing proved the hand-written parser safe on arbitrary input — was reachable without
   one, and doing it found the `v=@` panic above.

**Tiers, all re-run after the last edit so one tree carries the whole claim** (`production 80901b7cafab`):
unit 6348 pass / 0 fail (~48s); lint 0 issues across four modules (~33s); data sqlite 537 / 0 (~36s);
integration sqlite 1374 / 0 (~1m40s); anchors 20/20. The data tier ran on sqlite alone: the round
touched no migration and nothing under `src/core/data/`.

**Eleven mutations, every one CAUGHT**, each narrowed to the tests that should notice: the base64
decode gate; top-level SP on the list, at the Item start and at the Item end (each reverted to
`skipOWS`); the Date branch and the Display String branch (each dispatched on a character no field
carries); Date-rejects-a-Decimal; Display String UTF-8 validity; lc-hexdig lower-case-only; the
`parseNumber` bound; and the folded-in template fix. A twelfth attempt, deleting the
Date-is-not-a-Decimal check outright, reported `MUTATION DID NOT COMPILE` (it left a variable declared
and unused) and was rerun in the form above — recorded because a non-compiling mutation proves nothing
and must not be read later as one that passed.

---

### Final gate, round 3 — one parser defect fixed in `cd0d08de`, one evidence repair

Request `281-20260912T202502Z`, resumed session, started 20:41:52Z, completed 20:56:39Z. All three
axes `reviewed`; `unchecked` empty; no follow-ups, no bookkeeping. Both findings `judgement: false`
with a populated `forced_answer`.

**Finding 8 — conformance, minor. Non-String `v` parameters were displayed as serialized structured
field syntax.** Confirmed, fixed. `parseBareItem` answered every bare-item form's source text and
`parseBrands` assigned it whenever the parameter key was `v`, so a structurally valid `Sec-CH-UA`
whose `v` was any type but a String put that type's serialization in `DeviceName`:
`Chrome @1659578233`, `Chrome %"12%2e3"`, `Chrome ?1`, `Chrome :MTI=:`. The display string is the
sharp case, because its pct-encoding hides the `.` the major version is cut at, so the whole span
survived the cut.

The normative claim was fetched rather than recalled, since the agreement quotes UA-CH 3.1 but not the
type of `v`. UA-CH 4.1.4: "Let parameter be a dictionary… Set parameter['param_key'] to 'v'. Set
parameter['param_value'] to version," where the same algorithm's earlier step makes `version` "a
string", and the `NavigatorUABrandVersion` dictionary it builds declares `version` a `DOMString`.
Every example spells it `v="73"`. So a `v` of another type is a parameter this package cannot use, not
a version spelled unusually.

The fix has two halves and either alone closes the defect: `parseBareItem` now answers the item's
*type*, and `parseBrands` assigns a version only for a String; and no form but the String carries a
value out of `parseBareItem` at all, which let `parseDate`, `parseDisplayString` and
`parseByteSequence` drop their span return entirely. Keeping both is deliberate — the type gate states
the protocol rule where the rule belongs, and the absent value means a bare-item form added later
cannot leak text into a label by forgetting that gate. It also retires the old comment's argument
(that answering the display string's span was what kept a decoded NUL out of a column PostgreSQL
refuses) and replaces it with a structural answer. The assignment resets `major` before the gate,
because parameters are a dictionary (RFC 9651 3.1.2): a repeated `v` is the later value, including
when the later value is the unusable one.

Nine rows added to `TestLabels_ClientHints`, each pairing the hint header with a Firefox User-Agent so
the expected label proves the hints path still ran. **Seven of the nine fail against the code as it
stood before the fix**, run and recorded before the fix was written: the Date, Display String,
Boolean, Byte Sequence, Token and Integer forms, and the repeated-`v` row. The two that already passed
are the valueless boolean `;v`, which read as no version by accident, and the control row where the
later repeated `v` is the String.

| Mutation | Result |
|---|---|
| `major = ""` → `_ = major` | CAUGHT, the repeated-`v` row |
| `if isString {` → `if !isString {` | CAUGHT, every row carrying a version |
| the boolean branch's `return "", false, i+2, true` → returning the span | CAUGHT, the boolean row |

A fourth, `if isString {` → `if true {`, reports MUTATION DID NOT COMPILE and proves nothing: the flag
becomes unread and Go refuses the destructuring that produces it.

**Finding 7 — quality, significant. The final-gate verification record described the pre-rebase tree.**
Confirmed, and the larger of the two. The branch was rebased onto #328 after the round's request was
written, so the commit that request names was no longer on the branch and all six split tree hashes
had moved; nobody inside the run was placed to notice. Two things followed, both recorded rather than
argued: **the migration is `000046`, not `000045`** (`origin/main` took `000045_audit_logs_request_id`
from #328, so four SQL pairs, the golden rows, the data test — now
`TestMigration000046_UserSessionsUserAgent` — and the migrate-version cases were renumbered); and
**the stage tier results are superseded rather than repaired**, since re-running them per stage would
answer for trees that no longer exist. The whole branch was run once instead, against the tree
actually at the gate:

| Tier | Engines | Result |
|---|---|---|
| unit, three modules | n/a | 1747 pass, 0 fail |
| lint | n/a | clean, four modules |
| data | sqlite | 397 pass, 0 fail |
| integration | sqlite | 1040 pass, 0 fail |
| `where.sh migrations` | all four sources | all rules pass |
| `check-anchors.sh` | n/a | 20 of 20 rows resolve |

**The four-engine claim is CI's**, and a check run names its own commit, which is the one piece of
evidence the rebase did not invalidate.

**A red check suite, read immediately after that entry, and it was not the code.** `Tests / mysql`
failed inside the `setup-test-container` composite action with exit 35, and its `Build`, `Data tests`
and `Integration tests` steps all reported **skipped**, so no test of this change ran on that engine.
Everything that did run passed. `gh run rerun --failed` refuses while a workflow is still running, so
the rerun was recorded as owed rather than argued away.

---

### Final gate, round 4 — one finding, resolved; the review converged and closed

Request `281-20260912T211515Z`, resumed session, completed 21:23:12Z. Verdict `findings`, 1 finding.
All three axes `reviewed`.

**The verdict was salvaged from the archive, not the mailbox.** The reviewer wrote
`verdict-issue281-stagefinal-round4.json` and its prose, but `.review/verdict.json` was never written,
so the ingest reported "no verdict yet" and the driver wrote `.review/rounds-stalled.md`. The archived
verdict carries `request_id: 281-20260912T211515Z`, this round's request exactly, so it answers it
unambiguously and was ingested normally. This is the second time a completed round at this gate failed
to reach the mailbox — round 1 was recovered from `findings.jsonl` the same way. **The stalled marker
was bookkeeping about a missing file, not a contested finding**; taken at face value, a clean
converged review would have been recorded as a budget-exhausted stall and its only finding never
answered.

**Finding 9 — quality, significant. The required MySQL check did not execute on HEAD.** Confirmed, and
now **resolved**. The parent commit was green, but HEAD changes production parser code, so the
parent's check does not name HEAD. The reviewer's forced answer: rerun the job and require it to pass
on HEAD before approval, and if setup fails again, repair the environment rather than treating skipped
steps as green. Not a code defect — the run had already recorded the red as a setup failure with a
rerun owed. The rerun of workflow `34719319348` concluded, checked against the reviewer's explicit
warning rather than the job's summary conclusion alone:

| Step | Result |
|---|---|
| 5. `./.github/actions/setup-test-container` | success (was exit 35) |
| 6. Build | success (was skipped) |
| 7. Data tests | success (was skipped) |
| 8. Integration tests | success (was skipped) |

**What the round checked and found sound**, recorded so the absence of further findings has a stated
scope. *Conformance:* the full `55993a7d..cd0d08de` diff reconciled against every earlier finding;
typed handling compared across String, Token, Integer, Decimal, Boolean, Byte Sequence, Date, Display
String, valueless Boolean and repeated `v`; the round 3 reproduction (`v=@1659578233`, `v=%"12%2e3"`,
`v=?1`, `v=:MTI=:`) now yields `Chrome` while String versions still yield `Chrome 120`; RFC 9651
parameter-dictionary semantics and the UA-CH brand-version construction rechecked; migration 000046,
the raw-header persistence and sweep, the API and OpenAPI fanout, the dependency removal and the
console display unmoved, with migration rules and all 20 anchors passing. *Quality:* the nine new
non-String and duplicate-`v` rows pass; a 20-second native Go fuzz run executed **678,934** arbitrary
brand/mobile/platform/User-Agent combinations with no panic, nondeterminism, invalid-UTF-8 label or
width violation; a Display String `%00` assertion confirms a decoded NUL cannot reach any label; the
three recorded mutations each caught. *Security:* the attacker-controlled brand and platform sinks
still pass through the independently tested `escapeHtml`, all three raw User-Agent tooltips remain in
autoescaped attribute contexts, and the fix changes no security decision, endpoint, query, token path
or logger.

**`unchecked`: one entry, covered** — "MySQL data and integration behavior on HEAD was not checked by
CI because setup failed before those steps." That is finding 9, and the rerun above closes it.

Reviewer's follow-ups: **none**. Bookkeeping: **none**. Nothing deferred.

leodip:
## Correction to the follow-ups comment: item 1 was folded in, not filed

[The follow-ups comment](https://github.com/leodip/goiabada/pull/330#issuecomment-5648858943) reasoned that the user-search fixture collision should be filed rather than fixed here, because it rides on no surface this change touches. That reasoning was overridden: the user asked for it to be fixed in this branch, and then asked for the fix to be widened from the one file to the whole tier. It landed in `8e6500a0` as stage 4. **Nothing from this run is left to file.**

Two corrections to what that comment said, found while doing the work:

- **"Every other fixture in that package draws a unique address" is wrong.** 128 addresses across the tier were hardcoded. What made `createTestUsers` different is not that it was the odd one out — it is that *five tests in one run shared its three addresses*, so its collision was intra-run and self-propagating. Every other hardcoded address was used by exactly one test, where the exposure is only across runs, on the three engines whose database is never dropped.
- **It missed a sibling in the same file.** `TestAPIUsersSearch_MultiplePages` inlines eight users at `testuser1..8@pagination.test`, fully fixed, with the same discarded-delete cleanup. Same class, weaker exposure. Fixed in the same commit.

What landed, beyond what the draft prescribed:

| | Draft | Landed |
|---|---|---|
| Scope | `createTestUsers`, 3 addresses | the tier, 102 addresses through `uniqueEmail` + the search and pagination fixtures |
| Inserts | `require.NoError` | same |
| Guard | none proposed | an AST lint refusing a hardcoded address in any `models.User` literal, with a floor; a test pinning `uniqueEmail` itself; a test building the search fixture twice with no cleanup between |

Five literals stay, all in requests asserted to be rejected (400/401/404) — nothing reaches the database, and in one the length *is* the assertion. The eight sites that repeat an address deliberately (a duplicate-rejection target, an authenticator account name, an asserted value) draw it once into a variable.

Four mutations, all caught. The last one restores a fixed address in the search fixture and reproduces the original failure exactly: `UNIQUE constraint failed: users.email (2067)`.

Tiers: integration green on sqlite, 1382 pass 0 fail, over four runs; lint green over four modules. The four-engine result for stage 4 is CI's.

Issue #347, opened 2026-09-12
[Core refactor 16/29] Move SMTP and email delivery to authserver


leodip:
## Goal

Move core/communication, its authentication helpers, mocks, and tests to authserver.

## Rationale

Only authserver sends email; adminconsole asks it to do so through API endpoints. SMTP credentials, encryption, templates, transport policy, and settings models belong to the service that performs delivery. The only adminconsole edge is an unused interface removed in #333.

## Implementation guidance

1. Move EmailSender, SendEmailInput, SMTP PLAIN/LOGIN auth helpers, mocks, and tests into a focused authserver package.
2. Preserve context cancellation, dialing, timeouts, TLS/start-TLS behavior, SMTP authentication selection, sender/recipient/header construction, body encoding, and connection cleanup.
3. Preserve decryption of stored settings and the distinction between configuration, authentication, transport, and message failures.
4. Preserve audit and logging level/message/attribute behavior, especially secret redaction.
5. Keep interfaces at authserver call sites and update account registration, verification, reset, settings-test, and API handlers.
6. Do not change templates, localized content, subject lines, or delivery retry behavior.

## Compatibility and non-goals

- This is a structural refactor only. Do not change business rules, OAuth/OIDC behavior, HTTP routes, status codes, response bodies, persistence behavior, configuration names, logging semantics, or security policy.
- Preserve public serialized representations, including JSON fields, session payloads, cookie names, token claims, database values, and error classification.
- Move or regenerate tests and mocks with the code they cover. Do not weaken assertions to make the move pass.
- Avoid temporary forwarding aliases unless a concrete compatibility consumer requires one; any temporary exception must name its removal issue.
- Keep every commit buildable where practical and finish with gofmt and tidy only in modules whose imports changed.

## Acceptance criteria

- Core/communication is removed and adminconsole has no SMTP dependency.
- Existing SMTP unit tests and Mailpit-backed integration coverage pass.
- No credentials or message content are newly logged.

- The unit tier passes for core, authserver, and adminconsole.
- The repository-wide gofmt and error/logging convention guards pass.
- Relevant race, data, and integration tests continue to pass.
- The PR description lists old and new package paths and confirms runtime behavior is unchanged.

## Dependencies

Depends on #333 and #345.

## Architecture guard (added by #332)

`ARCHITECTURE.md` at the repository root records the module boundary. It is not a description: its
tables are parsed by `AssertArchitecture` and checked against the real import graph by every module's
unit tier.

[...]

Issue #346, opened 2026-09-12
[Core refactor 10/16] Move the auth server's application services and leaf packages


leodip:
## Goal

Move the auth server's database-backed application services and its authserver-only leaf packages out of core: `core/user`, `core/useragent`, `core/communication`, `core/otp`, `core/imaging` and `core/uithemes`.

## Rationale

These are application services and adapters, not cross-process contracts. Session reuse, permissions, revocation, ACR/AMR and user creation are server policy; SMTP credentials, transport and delivery belong to the process that sends the mail; OTP verification and enrollment are authoritative authentication logic; image validation protects the auth server's upload endpoints; and the UI-theme catalog is served through an authserver settings API. The admin console asks for all of it over HTTP.

## What this issue absorbed

This issue replaces the former #347 (SMTP and email delivery) and #348 (OTP, imaging and UI themes). Both are closed and point here.

## Why this is one issue and not several

All six packages have the same consumer set and no adminconsole production consumer:

| package | prod + test LOC | imported by |
|---|---|---|
| `core/user` | 3528 | `authserver` handlers, accounthandlers, apihandlers, server |
| `core/useragent` | 1818 | `core/user` only, once #339 lands |
| `core/communication` | 2195 | `authserver` handlers, accounthandlers, apihandlers, server |
| `core/otp` | 1065 | `authserver` handlers, apihandlers, middleware, server, tests/integration |
| `core/imaging` | 433 | `authserver/internal/handlers/apihandlers` |
| `core/uithemes` | 40 | `authserver/internal/handlers/apihandlers`, tests/integration |

Each is a `git mv` plus an import rewrite over the *same* four or five authserver handler packages. Doing that three times over instead of once buys no review value. The former #348 already grouped three leaf packages on exactly this reasoning; this issue extends the same argument to the other three.

## Implementation guidance

1. Move each service into a cohesive authserver internal package. Avoid one generic catch-all: `UserCreator`, `PermissionChecker` and `UserSessionManager` have distinct responsibilities, and email, OTP, imaging and themes are separate concerns again.
2. Move consumer interfaces to the packages that consume them and generate the mocks there, with `src/authserver/generate-mocks.sh` at the pinned version.
[...]

Issue #332, opened 2026-09-12
[Core refactor 1/29] Define and enforce the shared-kernel boundary


leodip:
## Goal

Document the target ownership rules for core, authserver, and adminconsole, and add an automated dependency guard before moving implementation packages.

## Rationale

The module direction is correct, but package placement inside core is inconsistent: reusable contracts sit beside auth-server adapters, persistence, and provider services. Go compiles an imported package as a whole, so one misplaced file can leak database drivers or provider dependencies into the admin console. An explicit, tested boundary prevents later PRs from relying on taste and turns temporary exceptions into a measurable burn-down list.

## Implementation guidance

1. Add `ARCHITECTURE.md` at the repository root describing allowed module edges and the definitions of shared contract, application service, adapter, persistence model, and wire DTO. It goes at the root, beside `README.md`, `CLAUDE.md` and `AGENTS.md`, rather than under `docs/`: `/docs/` is gitignored on purpose, since it holds the per-issue working documents that are deliberately not pushed.
2. Record the intended final owner of every top-level core package.
3. Add a guard, preferably through core/testutil and called by all three module unit tiers, that inspects Go imports or package metadata rather than raw substring matches.
4. Allow temporary dependencies only as exact package edges. Put the follow-up issue number beside each exception.
5. Treat core as a shared kernel: code may remain there only when both processes consume it or when it is an intentionally stable cross-process contract.
6. Include the complete 1/29 through 29/29 issue checklist in `ARCHITECTURE.md` or this issue after all issue numbers exist.

## Compatibility and non-goals

- This is a structural refactor only. Do not change business rules, OAuth/OIDC behavior, HTTP routes, status codes, response bodies, persistence behavior, configuration names, logging semantics, or security policy.
- Preserve public serialized representations, including JSON fields, session payloads, cookie names, token claims, database values, and error classification.
- Move or regenerate tests and mocks with the code they cover. Do not weaken assertions to make the move pass.
- Avoid temporary forwarding aliases unless a concrete compatibility consumer requires one; any temporary exception must name its removal issue.
- Keep every commit buildable where practical and finish with gofmt and tidy only in the modules whose imports changed.

## Acceptance criteria

- The target dependency graph is documented and unambiguous.
- The guard fails on a deliberately introduced forbidden edge and passes on the current tree through explicit, issue-linked exceptions.
- No production package is moved in this PR.

- The unit tier passes for core, authserver, and adminconsole.
[...]

PR #362, opened 2026-09-13
refactor(core): remove dead interfaces and the import edges they held open

### LINKED ISSUE #333: [Core refactor 2/29] Remove dead interfaces and accidental shared dependencies
## Goal

Remove unused cross-layer interfaces, mocks, constructor dependencies, and test-only production files that make auth-server code appear shared.

## Rationale

Several admin-console interfaces mention core user, communication, validator-input, and persistence-model types but have no production consumer. Address and phone validators also receive database dependencies they do not read. These false edges obscure the real dependency graph and would force later moves to carry compatibility code for APIs that are not actually used.

## Implementation guidance

1. Prove each candidate is unused with production-reference searches before deleting it.
2. Remove the unused admin-console interfaces for profile, email, address, phone, email sending, and user creation when no handler or constructor consumes them.
3. Remove the unused database fields and constructor arguments from AddressValidator and PhoneValidator while preserving every validation rule and error.
4. ~~Review AdminUserDatabase, AdminGroupDatabase, and AdminClientDatabase in core/data.~~ **No work: these three symbols do not exist.** `grep -rn 'AdminUserDatabase\|AdminGroupDatabase\|AdminClientDatabase' src/` returns nothing, in production or test. Verified while implementing #332. If a future census finds a similar interface, narrow it at the consuming package rather than deleting it outright.
5. ~~Delete core/audit/mocks.~~ **No work: `core/audit` is not in the repository.** `git ls-files src/core/audit` is empty. The directory appears only in a working tree that has run the mock generator, so there is nothing tracked to delete; if it is unwanted, the change belongs in whatever generates it. The *category* this point describes does have one live instance: `core/mocks/test_fs_mock.go`, a production file providing a test double, imported only by `_test.go` files in `adminconsole/internal/handlers`, `authserver/internal/handlers` and `core/handlerhelpers`. Its consumers are live, so **do not delete it** — move it to the tests that own it, or leave it. `ARCHITECTURE.md` records `core/mocks` as `kernel`; changing that requires changing the row.
6. Regenerate or relocate mocks only for interfaces that remain reachable.
7. Update architecture exceptions from issue 1 to reflect the smaller real graph.

## Acceptance criteria

- No unused production interface keeps an import on core/user, core/communication, or validator request types in adminconsole.
- Address and phone validator behavior is byte-for-byte equivalent at the error/message boundary.
- A production and test import census is attached to the PR description.

## Dependencies

Depends on #332.

## Architecture guard (added by #332)

`ARCHITECTURE.md` at the repository root records the module boundary, and its tables are checked
against the real import graph by every module's unit tier.

**No exception row belongs to this issue.** Point 7 above ("update architecture exceptions from
issue 1") therefore resolves to no work unless points 2 or 3 remove a package-to-package edge that
the table lists. Check the live table before assuming otherwise. A row left standing for an edge
that no longer exists fails the unit tier of all three modules.

Deleting a package that still has a row — or adding one that has none — also fails: every top-level
`core` package needs exactly one ownership row.

## Epic context

<details>

- [#333 — Remove dead interfaces and accidental shared dependencies](https://github.com/leodip/goiabada/issues/333) (this issue)
- [#336 — Move rate limiting and its middleware to authserver](https://github.com/leodip/goiabada/issues/336)
- [#338 — Separate shared OAuth client primitives from provider code](https://github.com/leodip/goiabada/issues/338)

</details>

leodip:
Closes #333.

## What this is for

`adminconsole/internal/handlers/interfaces.go` declares 14 interfaces, and nine of them have no
consumer anywhere — no implementation is asserted against them, no handler takes one, no test names
one. They are not inert. Four of them are the only reason the admin console's handler package
imports `core/user`, `core/communication` and `core/validators` at all, so two of the sixteen
temporary exception rows in `ARCHITECTURE.md` exist to grant edges that nothing but a dead
declaration carries. A tenth dead thing sits behind the ninth: `adminconsole/internal/tcputils` has
zero importers in the entire repository, and commit `f8960857` already deleted its mock as unused
and left the package standing.

The same defect turns up in fields. `golangci-lint`'s `unused` runs with
`field-writes-are-uses: true` by default, which means a field the constructor writes and nothing ever
reads counts as used. Four production fields survived review on that: the unread `database` on
`AddressValidator` and `PhoneValidator`, `UserSessionManager.codeIssuer`, and
`SettingsCache.lastFetchErr`. `codeIssuer` is the whole of the production `core/user -> core/oauth`
edge.

So this removes declarations that are kept alive by nothing but their own definition, and the import
edges and exception rows they were holding open.

## The rule this establishes

**A declaration with no consumer is a defect, and three guards now say so rather than a reviewer.**

- `testutil.AssertNoDeadInterfaces`, called from both modules' `internal/handlers`, fails the unit
  tier for an interface no file outside its own declaration names. It resolves references **by
  package, not by name** — a bare-name census credits `handlers.TCPConnectionTester` with four uses
  that belong to `tcputils.TCPConnectionTester`, a different type in a different package, and calls
  a dead interface live. That error is exactly how the first census of this issue went wrong.
- `unused.field-writes-are-uses: false` in `.golangci.yml` closes the field half, so a
  write-only field fails `--type lint`. Every finding it then reports is either deleted or carries a
  `//nolint:unused` naming why the field is consumed whole.
- `AssertArchitecture`, landed by #332, already fails an exception row whose edge no longer exists.
  That is what turns the two stale rows from a documentation chore into a test failure, and it is why
  deleting the interfaces and deleting the rows have to be the same commit.

A fourth change is the reason the first is affordable: every guard that runs in more than one module
is now pattern A, a helper in `core/testutil` plus a thin per-module caller. `credential_read_lint_test.go`
and `template_lint_test.go` were the only two guards in the tree written out twice, and lifting them
means the new guard makes three uses of one shape rather than a third copy of a walk.

A fifth change, folded in after the review, is what makes any of them trustworthy. Until now no test
anywhere observed a guard **failing**: the rule tests exercise each guard's finder, and the half that
reports is reached only by callers that walk a clean tree, so blinding it disabled the guard across
every module with nothing going red. Thirteen guards — the nine in `core/testutil` and the four
written in place under `core/data` and `apihandlers` — now report through a `testutil.Reporter` and
are driven from a rule test against a tree that must fail, a tree that must pass, and the walk that
reached nothing.

## What is deliberately not here

- **`unused.exported-is-used: false`.** It reports zero findings on a tree that still contains all
  nine dead interfaces, so it is inert in this golangci-lint build. Setting it would be a line nobody
  could attribute a behaviour to, which is why the interface half needs a guard of its own.
- **Moving `core/mocks/test_fs_mock.go`.** Nine `_test.go` files across all three modules import it,
  and a `_test.go` file cannot be imported from another package, so "moving it to the tests that own
  it" means three copies. It reaches no production binary.
- **Converting any single-site guard to pattern A.** `openapi_contract_lint_test.go` (1601 lines),
  `migration_rules_test.go` (2507) and two others have no second copy to reconcile, and lifting them
  would put one application's policy in the shared kernel — the failure mode `ARCHITECTURE.md` rule 2
  exists to refuse.
- **A standing dead-package guard.** `tcputils` was found by a one-off import-graph sweep and is
  deleted by hand; `AssertArchitecture` already covers the `core` half.

## Behaviour

Unchanged, everywhere. No route, status code, response body, header, string, template, claim, column
or configuration name is touched. No package moves module, so no import path a consumer could depend
on changes. The one `.golangci.yml` edit alters only what the linter reports.

## Status

**Complete and ready for review.** All six stages landed, the whole change was reviewed twice and the
second round came back clean, the check suite was green on `8883642d` across all thirteen jobs, and the
one follow-up closing round 2 raised has since been folded in rather than filed, landing on `82d3ac21`.
The
nine dead interfaces, the `tcputils` package and the two stale exception rows are gone, so are the four
constructor dependencies nothing read; the linter setting that hid all four is on with every finding it
raises resolved; the dead-interface guard that stops a tenth arriving is in both handler packages; and
the credential-read and template guards are each written once in `core/testutil`, with every module
keeping its own policy at its call site; and every tree-wide guard in the repository, the ten this
branch did not write included, is now driven against a tree that must fail it. Each section below
carries its stage's commit and test result.

## Import census

The census the issue asks for, measured on the branch rather than predicted.

| edge | before | after |
|---|---|---|
| `adminconsole/internal/handlers` to `core/user` | only `UserCreator` | **gone**; the package now imports `core/config`, `core/constants`, `core/customerrors`, `core/errs`, `core/models`, `core/oauth`, `core/sessionstore` and nothing else from `core` |
| `adminconsole/internal/handlers` to `core/communication` | only `EmailSender` | **gone** |
| `adminconsole/internal/handlers` to `core/validators` | only the four validator interfaces | **gone** |
| `adminconsole/internal/handlers` to `core/models` | the four dead interfaces, plus `handler_admin_get_permissions.go` and `types.go` | **kept**, and its exception row with it |
| `core/user` to `core/oauth`, production | only `UserSessionManager.codeIssuer` | **gone**; no non-test file under `src/core/user/` names `core/oauth` |
| `adminconsole/internal/tcputils` | zero importers, production or test | **package deleted** |
| `core/mocks` | nine importers, every one a `_test.go` file, across all three modules | **unchanged**; it reaches no production binary, which is why `test_fs_mock.go` stays where it is rather than becoming three copies |

**No package moves module**, so no import path a consumer could depend on changes, no ownership row
moves, and no `reachable today` value in `ARCHITECTURE.md` changes. Runtime behaviour is unchanged
everywhere: no route, status code, response body, header, string, template, claim, column or
configuration name is touched.

### Stage 1: delete the nine dead interfaces, the dead package and their two exception rows
Landed in `37a330f8`.

`adminconsole/internal/handlers/interfaces.go` declared fourteen interfaces and nine had no consumer:
`OtpSecretGenerator`, `ProfileValidator`, `EmailValidator`, `EmailSender`, `AddressValidator`,
`PhoneValidator`, `PasswordValidator`, `UserCreator` and `TCPConnectionTester`. All nine deleted, the
file down to the five survivors `HttpHelper`, `AuthHelper`, `IdentifierValidator`, `TokenParser` and
`TokenExchanger`. With them went the four imports no other file in the package carries:
`core/models`, `core/communication`, `core/user` and `core/validators`. `core/models` stays imported
*by the package* through `handler_admin_get_permissions.go` and `types.go`, so its exception row
stands and only two rows went stale.

`adminconsole/internal/tcputils` deleted entirely — zero importers anywhere by the real import graph,
production or test, and no call site for `NewTCPConnectionTester`. Commit `f8960857` had already
removed its mock as unused and left the package standing.

`ARCHITECTURE.md`: the `adminconsole/internal/handlers -> core/communication` (#347) and
`-> core/user` (#346) exception rows deleted, plus the two prose sentences the change falsifies —
sixteen rows becomes fourteen, and the five core packages the admin console imports that reach
`core/data` becomes four, dropping `core/user`. Neither issue is closed by this; each loses an edge
it no longer has to carry. `AssertArchitecture` reads the table in both directions, so the rows and
the interfaces had to move together: either half alone reddens every module's unit tier.

Tests: unit on all three modules, 6502 pass and 0 fail, with `TestArchitectureBoundaryHolds` and
`TestGoSourcesAreGofmted` green in each; lint tier, 0 issues across the four modules. Data and
integration are not applicable — this stage touches no data code, no migration, no schema, no route
and no handler body.

### Stage 2: drop the four unread constructor dependencies
Landed in `87b24258`.

Four production fields were written by a constructor and read nowhere. `golangci-lint`'s `unused`
misses all four on its defaults, where `field-writes-are-uses: true` counts the constructor's write
as a use; they were found by mechanising the issue's own point 1 over the whole tree rather than over
the two candidates it names.

`AddressValidator` and `PhoneValidator` each held a `data.Database` with zero reads, while the other
four database-taking validators in the package read theirs between 2 and 25 times. Both are now empty
structs with zero-argument constructors, which is the package's existing shape rather than a new one:
`NewIdentifierValidator()` and `NewPasswordValidator()` already take none. Six call-site lines lose an
argument, both files lose their `core/data` import, and the two test files lose a mock database they
never programmed. Every validation rule, error code and error argument is unchanged, and the two
table tests pass with all thirty cases untouched — the diff on those files is the removed import, the
removed `mockDB` line and the constructor line, and nothing else. `core/validators` keeps its
`core/data` edge through four other files, so nothing moved at package granularity.

`UserSessionManager` held a `*oauth.CodeIssuer` that nothing read, and `usersession_manager.go` was
the only production file in `core/user` importing `core/oauth`. Removing the field therefore deletes
the **entire production `core/user -> core/oauth` edge**, which is what the issue's title asks for. No
exception row covered that edge and none was owed: both packages are non-kernel rows, so no rule
fires on it either way. The `codeIssuer` local in `initRoutes` survives — it is still handed to
`HandleIssueGet`.

`SettingsCache` held `lastFetchErr`, assigned in three places and read in none. The error the fetch
path returns is untouched.

Tests: unit on all three modules, 6502 pass and 0 fail; lint, 0 issues across the four modules; data
and integration on sqlite, 397 and 1043 pass and 0 fail. Integration is not strictly applicable —
no endpoint, route, status or body changed — but `routes.go` carries two of the six call sites, so it
was run to prove `initRoutes` still wires a working server rather than merely compiling. The seam
test `TestValidateAddress` was mutation-checked with a length bound neutralised and caught it.

### Stage 3: set `unused.field-writes-are-uses: false` and resolve its test findings
Landed in `2161f288`.

`unused` ran on golangci-lint's defaults, and the relevant default is `field-writes-are-uses: true`:
a field the constructor writes and nothing reads counts as used. That is the whole reason the four
production fields stage 2 deleted survived review. The setting is now `false`, so the shape is a
finding rather than a silence, and the comment beside it in `.golangci.yml` names those four so the
next reader knows what the line is paying for.

Turning it on reports 25 test fields. **Seven are dead and are deleted with their writes:** `bodyRan`
in the four `RunInTransaction` stubs (written in all four, read in none), `stubHttpHelper.notFound`,
`failingResponseWriter.code`, and `crossUserBrowser.passwordA` — that last one being why the shape is
easy to miss, since the same-named local three lines down is live and stays. **The other eighteen are
consumed whole rather than by selector,** so `unused` cannot see the use and deleting them would gut
the assertion: a reflection fixture whose every field is a container kind the walk under test must
reach, a map key read by the map's own equality, and two structs compared against positional
literals. Each keeps a `//nolint:unused` with the reasoning in prose above it. No assertion was
weakened in either direction.

Decision 6 left one question to the run, and it was measured rather than assumed: **one directive on
the `type X struct {` line covers the whole declaration.** Four directives silence eighteen
field-level findings, and `nolintlint` reports none of the four as unused — which it would have done
had each covered only its own line, and the eighteen findings would have stood as well.

`exported-is-used` is still deliberately not set. The agreement measured it as inert on a tree that
still held all nine dead interfaces; re-measured here on a tree without them, it is still zero
findings in all three modules, so it would be a line nobody could attribute a behaviour to. The
dead-interface half is stage 4's guard instead.

Tests: lint, 0 issues across the four modules — the tier that owns this stage; unit on all three
modules, 6502 pass and 0 fail; data and integration on sqlite, 547 and 1381 pass and 0 fail.
Integration ran because one of the edited files is an integration test. The new check was
mutation-tested: a constructor-written, never-read field added back to `AddressValidator` is reported
by `unused` with this setting, and not reported without it, which is what shows the setting and not
something else is doing the catching.

### Stage 4: `AssertNoDeadInterfaces`, called from both handler packages
Landed in `454c8656`.

Stage 1 deleted nine dead interfaces; nothing stopped a tenth arriving tomorrow. `unused` says
nothing about an exported declaration, and the two settings that might have were measured over the
unmodified tree, where all nine were still present, and between them reported none of them. So this
stage adds the guard: `testutil.AssertNoDeadInterfaces(t, dirs ...string)` holds every interface
declared under the directories it is given to having at least one reference somewhere in its own
module, production or test. It is called from `adminconsole/internal/handlers` and
`authserver/internal/handlers`, and **both are green on arrival**, which is an independent
confirmation of the census by a different method.

**Liveness is `go/types` object identity, never a matching spelling.** That is not a refinement: the
census that grounded this issue first read `TCPConnectionTester` as live, because a bare-name search
credited it with four references that belonged to `tcputils.TCPConnectionTester`, a different type
in a different package. Two more shapes fail the same way in the other direction — an unqualified
name a local shadows, and a selector whose base is a local wearing the import's name — and all three
are answered by asking `go/types` which object an identifier actually bound to. What the search
excludes is the declaration itself and nothing else, not the file it sits in, so a second
declaration in `interfaces.go` consuming the first still counts as a consumer. Structural
satisfaction deliberately does not: an interface no code names is what dead means here.

**No import is ever read.** Each package is checked with a stub importer that returns an empty
package for every path, which works because both questions this guard asks are lexical rather than
semantic — identifier resolution in Go needs no import to be correct. `go/importer` is not used at
all, so `core/testutil` stays on `go/ast`, `go/parser`, `go/token`, `go/types`, the standard library
and `core/errs`, and takes no new edge. A package outside the guarded directories is parsed only
when one of its files literally spells a declared name, which is sound because Go binds an
identifier by its text; that is why a module-wide walk costs 0.45s and 0.91s at the two call sites.

**Two shapes are reported rather than skipped**, because a false failure gets investigated while a
false pass is a guard that has quietly stopped guarding: a dot import in a file carrying a declared
name (there is none in the tree today, which is what makes refusing it free), and a declaration
whose type does not resolve. One ceiling is recorded on the function itself: the search space is one
module per declaring directory, which suffices only because both call sites guard an `internal/`
package, and the revisit trigger is a caller naming a directory outside one.

The rule table covers thirteen fixture interfaces across five packages of a fixture module, nine
live and four dead, asserted as an exact set — including the three rows the plan review made
non-optional — plus a walk that reaches no files, a dot import, and a declaration outside the
guarded dirs that is searched but not held.

Tests: unit on all three modules, 6508 pass and 0 fail across 67 packages, +6 on the previous stage
— the tier that owns this stage; lint, 0 issues across the four modules; data on sqlite, 554 pass
and 0 fail. Integration is not applicable: three of the four new files are `_test.go` and the fourth
is `core/testutil`, which reaches no production binary. Three mutations, all caught — a dead
interface added back to the real `interfaces.go`, and two that break the guard's cross-package and
same-package resolution in turn, which is what shows the rule table depends on identity rather than
on spelling.

### Stage 5: lift the credential-read guard into `core/testutil`
Landed in `bed2336e`.

`credential_read_lint_test.go` was one rule written twice, 141 lines in the admin console and 124 in
the auth server, 115 of them differing and nearly all of that the two rationales rather than the
walk. The walk, the `_test.go` exclusion, the accessor matching, the finding text and the coverage
floor now live once as `core/testutil.AssertNoCredentialQueryFallback(t, names []string, dirs
...string)` (249 lines), with a rule table beside it in `core/testutil/credential_read_rules_test.go`
(277 lines, 8 tests). The two files become thin callers, 71 and 89 lines, each keeping the half that
must differ: its name list, and the reasoning for it at the call site.

Why the lists must differ, and therefore why the names are a parameter rather than a shared
constant: the admin console's carries `state`, `error` and `error_description` because those reach it
only at `/auth/callback`, which is registered POST-only and arranged with `response_mode=form_post`
precisely so they arrive in a body. The auth server's must not, because `handler_authorize.go` reads
`.FormValue("state")` and OIDC Core 3.1.2.1 requires the authorization endpoint to accept GET.

Two gaps the plan review demonstrated in the old guard are closed by the lift. Both are green on the
tree as it stands, so neither changes production code.

- **The auth server's list gains `code`, `code_verifier`, `refresh_token` and `client_secret`.**
  `handler_token.go` reads all four from `r.PostForm.Get` and no list protected any of them, so a
  regression to `.FormValue` would have left every tier green. None of the four has a lawful query
  source: RFC 6749 4.1.3 and section 6 put the token and refresh requests "in the HTTP request
  entity-body", RFC 7636 4.5 puts `code_verifier` in that same request, and RFC 6749 2.3.1 says
  client credentials "can only be transmitted in the request-body and MUST NOT be included in the
  request URI". `code` is an authorization *response* value (4.1.2), so the GET allowance that
  protects `state` at `/auth/authorize` does not reach the token endpoint.
- **The walk gains the merged-form accessors.** It matched `.FormValue` alone while its own rationale
  named merged `r.Form` as the unsafe source, so a protected read moved from `r.PostForm.Get` to
  `r.Form.Get` passed the whole tier. Four shapes are now refused per name — `.FormValue`,
  `.Form.Get`, `.Form.Has` and `.Form[...]` — against their four body-only counterparts, which the
  leading dot keeps admitted. `.Form.Has` is one beyond what the plan named and is called out in the
  stage log as such: it answers whether a credential was supplied from the merged map, `.PostForm.Has`
  is a live idiom here, and a copy-paste between the two is exactly this regression.

Being lexical, the walk cannot follow the merged map through a name — `form := r.Form` then
`form.Get("password")` matches nothing — which is recorded as a `ceiling:` with its revisit trigger
rather than left to be discovered. No site in the tree binds either map to a local today.

**Tests.** Unit green, 6516 ok 0 fail, with `TestHandlers_NoCredentialQueryFallback` running in both
modules and the 8 new rule tests in `core/testutil`. Lint green, run although the stage is unit-only
because it adds a non-test file to `core/testutil`, where `AssertArchitecture` and the tightened
`unused` both have an opinion; the helper adds no import outside the standard library. Data and
integration not applicable: no query, handler, route or schema is touched.

**Mutations, eight, all caught.** The three the plan review used to demonstrate the gaps — the `code`
and `client_secret` reads moved to `.FormValue`, and the `password` read moved to `r.Form.Get` — plus
five on the helper, one per rule it claims. The fifth survived on the first run and the test was at
fault rather than the code: a quoted name carries its own closing quote, so the accessor's closing
delimiter is load-bearing only for a name spelled as a constant. That added a case and corrected the
helper's own comment.

### Stage 6: lift the shared template guard into `core/testutil`
Landed in `2a9f731c`.

Both servers' `web/template_lint_test.go` carried the same three rules and the same walk under them:
markup in a `{{define "title"}}` block, which renders literally in the browser tab; a template naming
`csrfField` after #155 deleted the bind; and a page layout hardcoding `<html lang="en">` instead of
reading the active locale. All three bodies were byte-identical, as were the `walkHTMLTemplates`
helper and the `titleDefineRe` regexp. The only difference between the two files was in the doc
comments, and it was each module's own evidence rather than a difference of rule — so unlike the
credential guard there is no per-module policy here, and what stays at each call site is that
module's evidence, its pointers, and the FS it holds the rules against.

`AssertTemplatesNoHTMLInTitle`, `AssertTemplatesNoCsrfField` and
`AssertTemplatesHtmlLangNotHardcoded` now live in `core/testutil/template_lint.go` with a rule table
of six tests beside them. Each takes the caller's `//go:embed` FS rather than a path under the source
root, which is what keeps the lift behaviour-preserving: the embedded set is what the binary renders
from, so a template outside the embed pattern is not a page anyone can reach and a disk walk would
guard a different set in both directions. `authserver/web/template_lint_test.go` goes 96 lines to 33
and `adminconsole/web/template_lint_test.go` 152 to 84, with all seven `TestTemplates_*` names
unmoved.

The walk is exported as `WalkHTMLTemplates` because a fourth caller reaches it: the admin console's
`TestTemplates_RedirectURIAndWebOriginCellsAreText`, a rule about two of its own client pages (#105).
That rule is one application's policy and stays in `adminconsole/web` untouched — its guarded set,
its banned strings, its `seen` floor and its whole comment — while only the walk beneath it is
shared. Sharing the walk and not the rule is the line ARCHITECTURE.md rule 2 draws, and it is the
same line that kept `openapi_contract_lint_test.go` and the other single-site guards out of the
kernel.

Neither copy had a coverage floor, so a walk reaching no `.html` file is now fatal. All four rules
pass vacuously over an empty set and both call sites have been green for releases, so nothing in any
tier could distinguish a guard covering every page from one covering none, and the embed pattern is
the single point of failure for all of them.

**Goal item 9 is complete with this stage.** Every guard that runs in more than one module is now
pattern A — a helper in `core/testutil` with a thin caller per module — and nothing is written twice.

**Tests.** Unit green on three modules, 1850 ok 0 fail, with all seven `TestTemplates_*` still
running in their own modules and six new rule tests in `core/testutil`. Lint green on four modules.
Unit under `-race` green, 1850 ok 0 fail. Data and integration run on sqlite, 398 and 1044 pass,
although the stage touches no handler, route, schema or query.

**Mutations, four, all caught.** One per rule lifted — the title block's markup check, the
`csrfField` spelling, the lang rule's email exemption — plus one on the walk they share, its `.html`
filter. The rule table is built on `fstest.MapFS` fixtures walked through the production walk, so
every row runs through the same code the four real callers reach.

### After the review: guard the kernel's bearer-token form read
Landed in `8883642d`.

Not a stage. The final review found that both callers of the newly shared
`AssertNoCredentialQueryFallback` walk their own module's `internal/` tree, so no walk reached `core` at
all — and `core/middleware/middleware_jwt.go` reads `access_token` out of a form body. The product line
was already correct, `r.PostFormValue`, body only; what was missing is anything that would notice a
regression to the merged `r.FormValue`, and the reviewer demonstrated that gap by making exactly that
mutation and watching every tier stay green.

Two independent answers, because the lint and the behaviour are different claims.
`TestMiddleware_NoCredentialQueryFallback` in `core/middleware` is the kernel's caller of the shared
guard, with `access_token` as its whole list, walking `core` with `core/middleware` named beneath it as a
coverage floor; the file records why each of the six sibling form reads under `src/` is deliberately
absent from that list. `TestJwtAuthorizationHeaderToContext_PostQueryTokenIgnored` in
`middleware_jwt_test.go` pins the accessor rather than the spelling: a POST carrying a form body *and*
`?access_token=querytoken` is the one request where the two accessors disagree, and it must reach the
parser with no bearer token. The reviewer's mutation now reddens both, separately.

The same commit fixes the documentation this branch had falsified — pattern 5 of `CLAUDE.md` and
`AGENTS.md`, which still said `handler_authorize.go` is exempt for `code` and still located the guard in
each `internal/handlers` package — and one pre-existing miscount in `ARCHITECTURE.md` that `origin/main`
carries too.

### After the review: a harness that can watch a guard fail
Landed in `77224d9f`, `2d12aa1e`, `269dd9d5` and `82d3ac21`.

Not a stage, and folded in rather than filed. Closing round 2 raised one follow-up: every tree-wide
guard splits into a finder, which walks the tree and returns findings, and a reporting half, which
turns those into `t.Errorf` and turns an empty walk into `t.Fatalf`. The rule tests exercise the
finder; the reporting half is reached only by the real callers, and every one of those walks a tree
that is clean and passes. So no test anywhere observed a guard **failing**, and a defect in the
handful of lines that report disabled it across every module with nothing going red. Demonstrated on
`8883642d`: replacing `if len(dead) == 0 { return }` with `if true { return }` in
`dead_interface_lint.go` left the entire core tier green.

It was written up as ready-to-file on the reasoning that driving a `t.Errorf`/`t.Fatalf` wrapper from
a test meant a `testing.TB` parameter on nine signatures and every caller, a subprocess re-exec
harness, or a `t.Run` capture that does not survive `Fatalf`. That reasoning is superseded by the
issue body's own refinement, which found the fourth option, and this implements it: a small
`Reporter` interface — `Helper`, `Errorf`, `Fatalf`, which is the measured surface, since the
reporting halves call those three and nothing else. It is an interface rather than a `testing.TB`
parameter because `testing.TB` carries an unexported method precisely to prevent an outside
implementation, so no recording fake satisfies it. `*testing.T` satisfies `Reporter` for free, so
**no exported signature changes and no call site moves.**

`testutil.RunGuard` runs a guard on its own goroutine, because that is the only way to reproduce what
`Fatalf` means. The real one is `Logf` followed by `FailNow`, and `FailNow` is `runtime.Goexit`: the
goroutine stops there, its defers run, and nothing after the call executes. A recorder that merely
recorded the message and returned would run code the guard's author had already decided was
unreachable — a harness that lies in the direction that matters, since it would pass a guard whose
fatal path falls through. A panic inside a guard is caught and re-raised on the caller's goroutine
carrying the original stack, so a broken fixture does not take the binary down unattributed.

**Thirteen guards, not the three this branch added.** Nine in `core/testutil`, plus `core/data`'s
begin-transaction, benign-sentinel and page-offset lints and the auth server's API error-code
contract lint, which are the same defect in a different shape: a `Test` function that is itself the
unexercised wrapper. Each owes three cases — a tree that must fail, a tree that must pass, and the
walk that reached nothing. The clean-tree case is not decoration: without it, a harness that called
everything a failure would satisfy every "this fixture fails" assertion.

Five things the work turned up that the follow-up had not:

- **`AssertGofmted` had no fixture test at all.** Its whole test was ten lines against the real tree,
  so it was untested on both halves rather than one. It now has `findUnformatted`, and with it the
  leniency test the split makes possible: a tree of nothing but unparseable files reads as an *empty
  walk* rather than as a clean one.
- **`SourceRoot`'s ascent** is split the same way, with both answers pinned. A guard rooted at the
  wrong directory is exactly the quiet pass its own doc comment warns about.
- **Two guards answer the empty walk differently, and should.** `AssertNoCredentialQueryFallback`
  reports each named directory as its own floor through `Errorf`, so a tree renamed out from under
  one of three call sites names that directory rather than shrinking the guard to the two that still
  resolve; only an unnamed walk is fatal. `AssertAgentDocs` has no walk at all — it reads two named
  files — so an unreadable document is a finding rather than a fatal.
- **`AssertSlogConvention`'s missing `.golangci.yml` is a violation, not a walk error.** The test was
  written expecting a fatal and the tree was right: the file being absent is a fact about the
  repository a reader can act on, and phrasing it as "walking failed" would send them looking for a
  broken tree.
- **`AssertArchitecture` has two fatal seams, reached in order.** A missing `go.mod` fails the graph
  builder before the empty-graph refusal is reachable, so they are separate cases rather than one.

Two changes of substance beyond the split. Findings that were `t.Fatalf` in the four in-place lints
are now `Errorf`: with nothing after them the effect was identical, but `Fatalf` is the category for
a walk that covered nothing and `Errorf` for a thing the walk found, and having all thirteen agree is
the point of the harness. And the API error-code lint's file floor moves from a literal `60` into a
named constant passed in, so a rule test can drive the refusal itself; a second test holds the
constant to the surface it guards, since a floor drifting far below what it covers stops being a
floor. That lint was also the only one of the thirteen with no finder at all — its walk, parse and
comparison were one function — so `findAPIErrorCodes` is new.

`CLAUDE.md` and `AGENTS.md` gain a **Guard shape** section, because the next guard is written from
that document and the seam is invisible from the outside: a guard missing it is green in every tier.

**Mutation evidence.** Each of the thirteen reporting halves blinded in turn, through `where.sh
mutate`. Every one of these survived before this work and is caught after it.

| guard | mutation | result |
|---|---|---|
| dead interfaces | `if len(dead) == 0` → `if true` | **caught** |
| dead interfaces, empty walk | `if files == 0` → `if files < 0` | **caught** |
| credential read | `range reads` → `range reads[:0]` | **caught** |
| gofmt | `if len(unformatted) > 0` → `if false` | **caught** |
| templates | `range findCsrfField(...)` → `range findCsrfField(...)[:0]` | **caught** |
| architecture, empty graph | `if len(graph.prod) == 0` → `if false` | **caught** |
| errors, slog, agent docs, audit context | report branch blinded | **caught** |
| begin-transaction, benign-sentinel, page-offset, API error codes | report branch blinded | **caught** |

The first row is the mutation the follow-up cites as having left the whole core tier green.

Tiers: `--type modules`, `--type modules --race` — the harness is concurrent, so this one matters —
and `--type lint`, 0 issues across all four modules. 1,774 insertions, 87 deletions, no production
code touched.

## Tiers

| tier | where | result |
|---|---|---|
| unit, `core` / `authserver` / `adminconsole` / setup | check suite on `8883642d`, and locally per stage | green |
| unit under `-race` | check suite (`Unit / race`) | green |
| lint, four modules | check suite (`Lint`), and locally per stage | green |
| data, sqlite + MySQL + PostgreSQL + SQL Server | check suite (`Tests / *`) | green |
| integration, all four engines | check suite (`Tests / *`) | green |
| build validation, vulnerabilities, versions | check suite | green |
| data and integration, sqlite | locally, after the final-review fix | 397 and 1043 pass, 0 fail |
| unit, unit under `-race`, lint | locally, after the folded-in guard harness | green; 0 lint issues |
| thirteen guard mutations | locally, `where.sh mutate`, one per guard | all caught |

Every local run covered sqlite only; the four-engine result is the check suite's, and it is green on
`8883642d`. `where.sh ci` at `8883642d`: **13 of 13 jobs pass.** The four commits after it touch test
files, `CLAUDE.md` and `AGENTS.md` only — no production code — so the tiers re-run locally are the unit,
race and lint ones the change can affect.

## Review

Reviewed by `openai-codex/gpt-5.6-sol` at effort `xhigh`, two rounds over the whole branch, plus one
round at the plan gate before any code landed. `LEO_REVIEW_MODE=end`, so no stage was reviewed on its own.

|---|---|---|---|
| plan gate, 1 | 4 | 0 | all fixed in the plan before any code landed; one amended agreement decision 9 |
| final, 1 | 1 | 0 | fixed, `8883642d` |
| final, 2 | 0 | 0 | clean |

- **significant, security (plan gate):** the auth server's credential-read list protected none of `code`,
  `code_verifier`, `refresh_token` or `client_secret`, all four read by `handler_token.go`, so a
  regression to `.FormValue` on any of them would have left every tier green. Fixed as part of stage 5's
  lift: all four join the list, `state` stays out, and decision 9 was amended rather than built on.
- **significant, security (plan gate):** the same guard matched `.FormValue` only, so a protected read
  moved to the merged `r.Form.Get` passed the whole tier. Fixed in `bed2336e`: four merged accessors are
  refused per name against their four body-only counterparts.
- **significant, quality (plan gate):** the sketched dead-interface guard counted a matching spelling as
  a reference. Fixed: liveness is `go/types` object identity, with three fixture cases made mandatory.
- **minor, conformance (plan gate):** the planned agent-docs sentence claimed a reach the guard does not
  have. Fixed: it names the two call sites.
- **significant, security (final, round 1):** no caller of the shared credential guard reached `core`, so
  the kernel's `access_token` form read was unguarded. Fixed in `8883642d`: a kernel caller with a
  coverage floor, plus a behavioural test on the accessor. The reviewer's mutation is caught by each,
  independently.

Full account, with every mutation, table and refutation, in
[this comment](https://github.com/leodip/goiabada/pull/362#issuecomment-5651343873).

## Deviations

- **Decision 9 was amended mid-run**, at the plan gate, on the reviewer's evidence: the auth server's
  credential list gains `code` where the sealed text said it must not carry it. The premise that
  exclusion rested on is falsified by the tree, and the amendment is recorded in the decision itself.
- **Stage 5 refused a fourth accessor the plan did not name**, `.Form.Has`, for the reason the other
  three exist.
- **`probe/unused-settings.sh` can no longer run against this branch.** It patches a `settings.unused`
  block into `.golangci.yml`, and stage 3 put that block in the file, so the patch now produces a
  duplicate YAML key. It was left measuring the pre-change tree it was written for, with a header note,
  rather than rewritten to silently reproduce a different `.out`. The standing check is
  `golangci-lint --enable-only unused` with no patch, which is what `--type lint` runs.
- Nothing else departed from the sealed agreement.

## Where the rest of the record is

- [Deferred decisions](https://github.com/leodip/goiabada/pull/362#issuecomment-5651329750) — there are
  none; the comment says which ten were settled and which one was amended.
- [Follow-ups, fold-ins and ceilings](https://github.com/leodip/goiabada/pull/362#issuecomment-5651332572)
  — five things folded in rather than filed, and the two `ceiling:` markers this branch added with their
  revisit triggers. **Superseded on one point:** the one ready-to-file issue it describes, the `Assert*`
  wrappers never being exercised against a failing tree, was folded in as well rather than filed, and its
  "why it is not folded in" note no longer holds. See *After the review: a harness that can watch a guard
  fail* above, and the amendment comment on this pull request.
- [The review account](https://github.com/leodip/goiabada/pull/362#issuecomment-5651343873) — every
  round, every finding, and the evidence for each disposition.
- `docs/issue-333-dead-interfaces/` — the agreement, the per-stage logs and the probes. `/docs/` is
  gitignored in this repository, so that directory is not on the branch and the three comments above
  are the whole of the record that travels with the pull request.


leodip:
## Deferred decisions: none

The run deferred nothing. All ten decisions in `docs/issue-333-dead-interfaces/agreement.md` section 3
were settled before the agreement was sealed, and `log/closing.md` carries no section 8, because the
run raised no new question it had to proceed past on an assumption.

For the record, the ten and who raised each: delete all nine dead interfaces rather than the six the
issue names (run), zero-argument validator constructors (run), leave `core/mocks/test_fs_mock.go`
where it is (user), fold in `UserSessionManager.codeIssuer` (run), fold in `SettingsCache.lastFetchErr`
(run), set `unused.field-writes-are-uses: false` with no test exclusion (user), a dead-interface guard
in both modules (run), the guard lives in `core/testutil` (run), standardize only the two genuinely
duplicated guards (user), and delete `adminconsole/internal/tcputils` (run).

One was amended mid-run, at the plan gate, and the amendment is recorded in the decision itself rather
than left implicit: decision 9 originally said the auth server's credential list must carry neither
`state` nor `code`. The `code` half rested on a premise the tree falsifies, so the auth server's list
gained `code`, `code_verifier`, `refresh_token` and `client_secret`, and `state` stayed out. The
reasoning and the RFC citations are in decision 9 and in the review comment below.

leodip:

### Follow-up, one, ready to file

Found while closing the reviewer's bookkeeping item in final review round 2, not raised by the
reviewer. Duplicate search re-run against the tracker at the finish: nothing open or closed covers it
(the only issue whose title mentions a lint guard is #320, closed).

**Why it is not folded in.** It needs a design choice this change cannot make in passing — driving a
`t.Errorf`/`t.Fatalf` wrapper from a test means either a `testing.TB` parameter on all nine guard
signatures and every caller, a subprocess re-exec harness, or a `t.Run` capture that does not cleanly
survive `Fatalf`. And it is the first step rather than the whole fix: three of the nine guards are in
this diff and six are not, so fixing only the three would reintroduce exactly the two-shapes-for-one-job
split stages 5 and 6 spent themselves removing.

**Title:** the lint guards' `Assert*` wrappers are never exercised against a failing tree

**Labels:** enhancement, go

**Body.**

`core/testutil` carries nine repo-wide guards in the pattern A shape, `Assert<X>(t, ...)` walking
`SourceRoot(t)`: `AssertGofmted`, `AssertArchitecture`, `AssertNoLegacyErrors`, `AssertSlogConvention`,
`AssertAgentDocs`, `AssertAuditLogContext`, and the three #333 added or lifted,
`AssertNoDeadInterfaces`, `AssertNoCredentialQueryFallback` and the template assertions.

Each splits into a **finder**, which walks the tree and returns findings, and a **wrapper**, which
turns those findings into `t.Errorf` and turns an empty walk into `t.Fatalf`. The rule tests in
`core/testutil` exercise the finder against fixture trees and assert on its return value:
`findDeadInterfaces` at `dead_interface_rules_test.go:161`, `findLegacyErrorUses` at
`errors_lint_test.go:334`, `check(...)` at `architecture_rules_test.go:153`. The wrapper is reached
only by the real callers, and every one of those walks the real tree, which is clean and passes.

So no test anywhere observes a guard **failing**. A defect in the roughly five lines that report, or a
future edit to them, disables the guard across every module with nothing going red. Demonstrated on
`8883642d`: replacing `if len(dead) == 0 { return }` with `if true { return }` in
`dead_interface_lint.go`, so the guard can never report a dead interface, leaves the entire core module
tier green — `ok github.com/leodip/goiabada/core/testutil 15.877s`. The equivalent blinding of
`credential_read_lint.go` (`range reads` to `range reads[:0]`) and of `template_lint.go` survives its
own rule test likewise. The credential guard is the one that matters most: it is the sole enforcement
of `CLAUDE.md` pattern 5, which exists to stop a bearer token or a password reaching a proxy log.

This is pre-existing and repo-wide rather than introduced by #333, whose three guards match the
precedent its own agreement named. `AssertNoLegacyErrors` is that precedent and has the same shape:
`errors_lint_test.go` tests `findLegacyErrorUses` against fixtures and calls `AssertNoLegacyErrors(t)`
only at line 390, against the real clean tree.

One refinement on the above, from checking all nine rather than the four cited:
`TestAgentDocs_AMissingAgentsFileFails` (`agentdocs_test.go:202`) reads like the exception and is not
one — it calls the finder, `checkAgentDocs(root)`, and asserts on the returned slice, exactly as the
others do. And **`AssertGofmted` has no fixture-level rule test at all**: `gofmt_lint_test.go` is ten
lines whose whole body is `AssertGofmted(t)` against the real tree. So it is untested on both halves,
not just the wrapper, and it is the weakest of the nine rather than one with a finder test to build on.

**Closing it** means one harness all nine can use, letting a rule test assert that a given fixture tree
makes the wrapper fail and that a clean one does not. The narrowest shape that avoids touching nine
signatures is to have each wrapper report through a small internal `reporter` interface satisfied by
`*testing.T`, with the rule tests passing a recording fake; the alternatives are a `testing.TB`
parameter or a subprocess harness. Then one case per guard for the failing tree and one for the empty
walk — the case `AssertNoDeadInterfaces`'s own seam already singles out as "must fail rather than pass"
and which is currently asserted only indirectly, through `assert.Zero(t, files)` on the finder at
`dead_interface_rules_test.go:185`.

**File it with:**

```bash
gh issue create \
  --title "the lint guards' Assert* wrappers are never exercised against a failing tree" \
  --body-file <path to the body above>
```

Five things were found beside the work and dealt with here instead of becoming issues.

- **`UserSessionManager.codeIssuer`** — `87b24258`. Same defect as the two validator fields, and
  removing it deletes the entire production `core/user -> core/oauth` edge, which is the issue's own
  title. Agreement decision 4.
- **`SettingsCache.lastFetchErr`** — `87b24258`. Same defect, three assignments and no read; leaving it
  would make the issue about where the symptom sits rather than what it is. Agreement decision 5.
- **The four token-endpoint credentials and the merged-form accessors** — `bed2336e`. Raised by the plan
  review as two findings against the guard being lifted; fixing them *is* the lift, since the old
  duplicated guard is the thing being replaced. Agreement decision 9 was amended in place rather than
  built on.
- **`ARCHITECTURE.md`'s "#350 owns nine" against a ten-row table** — `8883642d`. Pre-existing
  (`origin/main` carries the same error), one word, in a sentence stage 1 had already edited.
- **`CLAUDE.md` and `AGENTS.md` pattern 5** — `8883642d`. This change falsified it twice over: `code` is
  now on the auth server's forbidden list, and the guard is no longer "per module in each
  `internal/handlers` package". Documentation this branch broke is this branch's to fix.

Two `ceiling:` markers were added, each with its revisit trigger in the tree.

- `src/core/testutil/credential_read_lint.go:68` — being lexical, the walk cannot follow the merged map
  through a name: a handler writing `form := r.Form` and then `form.Get("password")` reads the query and
  matches nothing, because the receiver is no longer spelled `Form`. No site in the tree binds `r.Form`
  or `r.PostForm` to a local today, and resolving one would mean type-checking every package under
  `dirs`. *Revisit when* a handler binds either map to a variable, or a finding shows a read reaching
  the query through one.
- `src/core/testutil/dead_interface_lint.go:57` — the search space is one module per declaring
  directory, so an interface declared outside an `internal/` tree could be referenced from a sibling
  module and this walk would call it dead. Nothing is in that position today, both call sites guarding
  an `internal/handlers` package, and widening the search would cost a fourth-module scan on every run.
  *Revisit when* a caller names a directory outside `internal/`.

leodip:
## The review account

`LEO_REVIEW_MODE=end`: no stage was reviewed on its own, so the two final rounds below are the only
independent reading this code got, plus one plan-gate round that ran before any code landed. Pasted in
full here because `.review/` is gitignored and overwritten by the next gate.

---

### Plan gate, round 1 — 2026-09-12

`openai-codex/gpt-5.6-sol`, effort `xhigh`, fresh session, request `333-20260913T010526Z`, round 1 of a
2-round plan budget. **Four findings, no follow-ups.** All four confirmed against the tree and fixed in
the plan before any code landed; none refuted, none deferred. Every one arrived with a populated
`forced_answer`, so nothing was escalated.

**1. `AssertNoDeadInterfaces` confused a matching spelling with a reference (significant, quality).**
The sketched criterion excluded the whole declaring file, so a live interface consumed by a second
declaration in `interfaces.go` reads as dead; and it counted any unqualified identifier of the same
spelling, so an unrelated local `EmailSender := 1` elsewhere makes a dead interface read as live.
*Fixed:* liveness is object identity — a reference that binds to the declared type, resolved through
`go/types` `Info.Uses`, excluding only the declaration's own name identifier. Both `go/types` and
`go/importer` are standard library, so the guard still adds no dependency edge to the kernel. Three
fixture cases became mandatory: a same-file consumer, an unrelated same-package identifier of the same
spelling, and an import alias shadowed by a local value. An unresolvable shape is reported rather than
skipped.

**2. The planned `CLAUDE.md` / `AGENTS.md` sentence overstated the guard's reach (minor, conformance).**
It would have said a consumerless handler interface "fails every module's unit tier", but the guard is
called from the two applications' `internal/handlers` packages only, so the core and setup tiers never
run it. *Fixed:* the sentence names the real fanout. Adding a core caller to rescue the wider wording
was rejected — it would move application policy into the kernel, which `ARCHITECTURE.md` rule 2 refuses.

**3. The authserver credential-read guard omits four token-endpoint credentials (significant,
security).** `handler_token.go` reads `code`, `code_verifier`, `refresh_token` and `client_secret` from
`r.PostForm.Get`, and the auth server's forbidden-name list protected none of them, so the guard stays
green if any one regresses to `r.FormValue` and starts accepting that credential from the request URI —
into browser history, `Referer`, and every proxy and CDN access log in front of the deployment. The
reviewer demonstrated it rather than asserting it: mutating the `code` read and the `client_secret` read
to `.FormValue` left the entire authserver internal tier green, this guard included, while the same
mutation on the already-listed `password` failed it. `code` had been excluded on the belief that OIDC's
GET allowance covered it; it does not — RFC 6749 4.1.2 makes `code` an authorization *response* value
and 4.1.3 requires the client to return it in the token request body, and `handler_authorize.go` never
reads `code` at all. *Fixed:* all four names join the auth server's list, `state` stays out for the
reason it always had, and agreement decision 9 was amended so no later stage builds from the falsified
sentence. Green on the current tree: no production file reads any of the four with `.FormValue`.
Production behaviour unchanged; this is the regression machinery, not the handler.

**4. The same guard misses direct reads from the merged `r.Form` (significant, security).** It matched
`.FormValue(name)` only, although its own rationale names merged `r.Form` as the unsafe source and
`handler_token.go` calls `ParseForm`. Moving the protected `password` read from `r.PostForm.Get` to
`r.Form.Get` left the whole authserver internal tier green, that guard included; moving it to
`.FormValue` failed the guard — so the blind spot is the accessor, not the policy name, and even a
listed credential could be read from the query unnoticed. *Fixed:* the lifted helper rejects
`.Form.Get(name)` and direct `.Form[name]` alongside `.FormValue(name)`, still admitting `.PostFormValue`
and `.PostForm.Get`. (Stage 5 added a fourth, `.Form.Has`, beyond what the review named.) Green on
arrival — no production site uses a merged accessor today. Alias propagation remains a stated lexical
ceiling.

`unchecked` held three entries, all dismissed with reasons: the six stages had no diff yet, which is
what a plan gate is; no suites were re-run because no code had landed and CI was green on both
`b175d0a7` and the opening commit `73ec84c3`; and no browser pass, because the change moves no route,
status, body, header, string or template.

---

### Final review, round 1 — request `333-20260913T040714Z`, against `2a9f731c`

`openai-codex/gpt-5.6-sol`, effort `xhigh`, fresh session, gate `final`, round 1 of 5. Verdict:
**findings**, one — `security` / `significant` / confidence `high`, `judgement: true` with a populated
`forced_answer`, so a finding the code settles rather than a question for the user. All three axes
reported `reviewed`.

#### Finding 1: the shared credential guard omitted the kernel's bearer-token form read

**Claim.** `core/middleware/middleware_jwt.go:121` reads `access_token` out of a form body, and both
callers of the newly shared `AssertNoCredentialQueryFallback` walk their own module's `internal/` tree,
so no walk reached `core` at all. The product line was correct — `r.PostFormValue`, body only — but a
regression to the merged `r.FormValue` left every tier green. Demonstrated rather than asserted:
replacing that read with `r.FormValue` and running `TestHandlers_NoCredentialQueryFallback` reported
**MUTATION SURVIVED**, all tests passing.

**Confirmed, not taken on trust.** Verified at the anchor: the read is `r.PostFormValue("access_token")`
inside the `r.Method == http.MethodPost` arm, and neither caller's `dirs` reaches it. The family was
re-enumerated independently over every non-test `.FormValue`, `.Form.Get`, `.Form.Has`, `.Form[`,
`.PostFormValue`, `.PostForm.Get`, `.PostForm.Has` and `.PostForm[` site under `src/`, and the
reviewer's boundary holds: outside the two covered trees there are exactly seven sites, and
`access_token` is the only credential among them. The other six are accounted for rather than waved past
— `middleware_ratelimiter.go` reads `"email"` and `"username"` as account keys, which #219 requires the
limiter and the handler it protects to read identically, and `"grant_type"` to select a branch;
`handlerhelpers/http_helper.go` reads a caller-supplied key for #109's query-or-body lookup and owns no
credential policy of its own.

**Fixed on the branch in `8883642d`**, per the forced answer, in two independent places:

1. `src/core/middleware/credential_read_lint_test.go`, new: `TestMiddleware_NoCredentialQueryFallback`,
   the kernel's caller of the shared guard, with `"access_token"` as its whole list. It walks `core`
   with `core/middleware` named beneath it as a coverage floor, the same shape as the auth server's
   caller, so the read moving to another kernel package stays covered while a rename that empties
   `core/middleware` fails here rather than shrinking the walk in silence. The file carries why each of
   the six sibling reads is deliberately absent, and the RFC 6750 2.3 / RFC 9700 4.3.2 sentences that
   make a bearer token in a URL worth a guard.
2. `src/core/middleware/middleware_jwt_test.go`:
   `TestJwtAuthorizationHeaderToContext_PostQueryTokenIgnored`, the behavioural half the lint cannot
   give. The guard asserts a spelling is absent; this asserts the accessor is right. A POST with
   `Content-Type: application/x-www-form-urlencoded`, body `other_param=value` and
   `?access_token=querytoken` is the one request where the two accessors disagree, and it expects no
   parser call and no bearer token in the context. The pre-existing `PostBodyIgnoredForGetRequest` case
   does not pin this: it is refused by the method check two branches earlier and never reaches the read.

**Evidence the fix is worth its green.** The reviewer's exact mutation re-run against each new test
separately — both now redden where the tier was green before:

| file | replacement | run | result |
|---|---|---|---|
| `src/core/middleware/middleware_jwt.go` | `r.PostFormValue("access_token")` to `r.FormValue("access_token")` | `TestMiddleware_NoCredentialQueryFallback` | **MUTATION CAUGHT** |
| `src/core/middleware/middleware_jwt.go` | same | `TestJwtAuthorizationHeaderToContext_PostQueryTokenIgnored` | **MUTATION CAUGHT** |

`tree-hash.sh --split` afterwards reports `production ed53de37c932`, identical to the value the round was
reviewed at, so both mutations are proven restored and no production byte moved for this fix.

#### Documentation this change had falsified, found while fixing the above

Not a reviewer finding, and owed regardless. Pattern 5 of `CLAUDE.md` and `AGENTS.md` read
"`handler_authorize.go` is exempt for `state` and `code`". Decision 9's plan-gate amendment made that
false during this run — `"code"` is on the auth server's forbidden list now, and `handler_authorize.go`
reads `.FormValue("state")` at two sites and `code` at none. The same sentence also still located the
guard "per module ... in each `internal/handlers` package", which stages 5 and 6 made wrong. Both files
now name the shared helper, its three callers including the kernel's, the four token-endpoint
credentials, and why `state` is exempt where `code` is not, citing RFC 6749 4.1.2 and 4.1.3. The two
files remain byte-identical, held by `AssertAgentDocs`.

#### Bookkeeping, fixed in passing

The reviewer noted `ARCHITECTURE.md` says #350 owns nine of the fourteen exception rows while the table
carries ten, and correctly classed it as bookkeeping rather than a finding. Pre-existing: `origin/main`
says "Sixteen rows, and #350 owns nine" over the same ten rows. Corrected to ten; the `arch/row-count`
anchor was re-swept, and `check-anchors.sh` passes all 18 rows.

#### `unchecked`, answered

- *No interactive browser pass.* **Dismissed.** No template byte, route, response or user-facing string
  changed; this round's additions are two test files and three markdown files.
- *No local four-engine data or integration tier.* **Covered twice over.** CI green at `2a9f731c` on all
  four engines, and this round's `production` hash is byte-identical to that commit's. Both tiers were
  re-run locally on sqlite anyway.
- *No dynamic endpoint probes for fail-open, replay, revocation, redirect, PKCE, authorization, rate
  limits, oracles or audit.* **Dismissed**, on the reviewer's own stated ground: static tracing found no
  changed executable branch in those classes, and `production ed53de37c932` is unchanged by this round.
- *Arbitrary syntactic evasions of the lexical guards.* **Covered by an existing ceiling**, stated at
  `credential_read_lint.go:68` with its revisit trigger.

#### Tiers re-run after the fix

| tier | command | result |
|---|---|---|
| modules (all three) | `where.sh test --type modules` | 1785 pass, 0 fail |
| lint | `where.sh test --type lint` | green over all four modules |
| data | `where.sh test --type data --db sqlite` | 397 pass, 0 fail |
| integration | `where.sh test --type integration --db sqlite` | 1043 pass, 0 fail |

No follow-ups: the reviewer's `## Follow-ups` section reads "None".

---

### Final review, round 2 — request `333-20260913T044446Z`, against `8883642d`

`openai-codex/gpt-5.6-sol`, effort `xhigh`, resumed session, gate `final`, round 2 of 5. Verdict:
**clean**, zero findings, all three axes `reviewed`, `## Follow-ups` reads "None".

Clean in a stated scope. What it ran: the whole branch re-read hunk by hunk against round 1's reading;
`check-anchors.sh`, 18 of 18 resolving; the import-graph probe re-run, confirming the admin console
binary no longer reaches `core/user`, `core/communication` or `tcputils`;
`go test ./testutil -run Test(Dead|NoCredential|Template)`; and — the part that earns the verdict rather
than restating it — a copy of the core module into scratch with only `PostFormValue("access_token")`
changed to `FormValue`, then each of round 1's two new tests run **alone** with its own fresh build
cache. Both reddened independently, which is what round 1 claimed and could not then prove.

Verified independently: `where.sh ci` at `8883642d`, all 13 jobs green. `tree-hash.sh --split` before and
after reading the verdict reports `production ed53de37c932` and `all c13e0cba211a`, matching the
request's recorded block exactly, so the reviewer's mutations stayed in its scratch copy and no branch
byte moved for this round.

#### `unchecked`, four entries, each answered

| entry | answer |
|---|---|
| no interactive browser pass | **Dismissed.** `git diff --stat origin/main...HEAD -- '*.html'` is empty: the branch changes no template byte. Stage 6 moved the template *guard*, not any template |
| no local live-database tier repeated | **Dismissed.** The branch touches no query, migration or `schema.golden`. The four-engine result is CI's by construction, and it is green at HEAD |
| no dynamic endpoint probes for replay, revocation, redirect, PKCE, authorization, rate-limit, oracle or audit paths | **Dismissed.** The ten changed non-test Go files are deletions of unreferenced declarations, fields, imports and constructor arguments, plus three guard helpers, which are test infrastructure no production binary imports |
| lexical guard evasions not exhaustively generated | **Dismissed as an already-recorded ceiling**, not a gap — `credential_read_lint.go:68` carries it in the tree |

#### The reviewer's one bookkeeping item, confirmed and closed

It reported that the mutation runs recorded across the stage logs used `--run` against a **module** tier,
while `run-tests.sh` documents that `--run` is ignored there, so those runs did not establish what they
claimed. Checked at the source rather than accepted: `src/authserver/run-tests.sh:62-63` reads "`--run`
only affects `data` and `integration` runs ... It is ignored for module-level test runs." **The reviewer
is right.** Those runs proved the *tier* reddens under the mutation, which is the property CI actually
enforces; they did not prove the *named test* is what reddened it.

Closed rather than annotated for the three guards this change delivers. Each guard's report path was
blinded and its own rule test run **alone**, with a fresh count, in the container:

| guard | mutation | isolated test | result |
|---|---|---|---|
| `dead_interface_lint.go` | `if len(dead) == 0 {` to `if true {` | `TestNoDeadInterfaces_TheRuleTable` | **SURVIVED** |
| `credential_read_lint.go` | `range reads` to `range reads[:0]` | `TestNoCredentialQueryFallback_TheAccessorTable` | **SURVIVED** |
| `template_lint.go` | `range findHTMLInTitle(...)` to the same sliced to zero | `TestTemplates_TheTitleTable` | **SURVIVED** |

Three survivals is a finding about the tests, so it was run down rather than filed as a number. The
proximate cause is that the mutations aimed at the wrong layer: every rule test asserts on the **finder**
and never drives the `Assert*` wrapper the mutations blinded —
`dead_interface_rules_test.go:161` calls `findDeadInterfaces(root, dirs)`, takes back
`dead, blocked, files, err`, and checks the table with `assert.ElementsMatch`. The rule tables are sound
and exhaustive at the layer they claim, and the survivals say nothing against them. What the survivals do
expose is real and is the follow-up in the comment above: blinding `AssertNoDeadInterfaces` so it can
never report and running the whole core tier gives
`ok github.com/leodip/goiabada/core/testutil 15.877s`. Both mutation scripts restored from a backup under
a trap, and `tree-hash.sh --split` after each reports `production ed53de37c932` unchanged, so every
mutation written this round is proven gone.

**Round 2 is unproductive by the reviewer contract's measurement, and correctly so:** the verdict found
nothing, so no production byte moved in response to a reviewer finding. That is the gate passing, not the
gate stalling. Nothing was refuted, and nothing was deferred.

---

### Stale per-stage claims, and what re-established them

`tree-hash.sh --verify` at the final gate — the one check no session could run from inside a stage,
because each stage's tiers were recorded against a tree the stages after it then moved:

```
current   production ed53de37c932   unit-tests bc13097705fb   data-tests 1e42e555ec0e
          integration-tests 167c3aebaf1c   docs f985655280b0   all a2241bf080e3

5 of 6 entries record a result against a tree that has since moved.
```

Every per-stage unit claim is stale, which is expected: production moved under each one as the next stage
landed. Stage 4's docs claim was the only row still live. The claims that matter are re-established at
HEAD rather than re-argued — the check suite is green on `8883642d` across all thirteen jobs, covering
four engines where the local runs covered sqlite, which is a stronger result than any of the stale rows
claimed.

The mutations are the exception and are not re-established by CI. What holds them is that the later
stages are additive: stages 3 to 6 add a linter setting, three guards in `core/testutil` and their
callers, and touch none of the production code stages 1 and 2 deleted from. The per-stage entries in
`docs/issue-333-dead-interfaces/log/` carry each file, replacement and verdict.

leodip:
## Amendment: the one follow-up is folded in, not filed

This amends [Follow-ups, fold-ins and ceilings](https://github.com/leodip/goiabada/pull/362#issuecomment-5651332572).
That comment lists one ready-to-file issue with its `gh issue create` — *the lint guards' `Assert*`
wrappers are never exercised against a failing tree* — and a "why it is not folded in" note. **No issue
was filed. It is folded in**, on `77224d9f`, `2d12aa1e`, `269dd9d5` and `82d3ac21`. The full account is
in the pull request body under *After the review: a harness that can watch a guard fail*; this comment
records only what changed about the disposition.

### Why the deferral no longer held

The note gave three ways to drive a `t.Errorf`/`t.Fatalf` wrapper from a test and rejected all three: a
`testing.TB` parameter on nine guard signatures and every caller, a subprocess re-exec harness, or a
`t.Run` capture that does not cleanly survive `Fatalf`.

The fourth option is in the drafted issue body itself, a paragraph below that note, and it costs none of
those things: *"have each wrapper report through a small internal `reporter` interface satisfied by
`*testing.T`."* Every exported `Assert*` keeps its `*testing.T` and delegates to an inner function
taking the interface, so **no exported signature changes and no call site moves.** The note was written
before that refinement and was not revised against it.

The interface is three methods — `Helper`, `Errorf`, `Fatalf` — which is measured rather than minimized:
the reporting halves in `core/testutil` call those three and nothing else, the two `t.Error(f)` sites
normalizing to `Errorf("%s", f)`. It is exported rather than internal, which the draft did not
anticipate, because four of the thirteen guards live outside that package.

### The second half of the deferral, which did hold, and is answered by doing all thirteen

The note's other argument was that fixing only the three guards in this diff would reintroduce the
two-shapes-for-one-job split stages 5 and 6 spent themselves removing. That is right, and it is why the
scope is **thirteen guards rather than three or nine**: the nine in `core/testutil`, plus `core/data`'s
begin-transaction, benign-sentinel and page-offset lints and the auth server's API error-code contract
lint. Those four were outside the drafted issue's scope; they are the same defect in a different shape,
a `Test` function that is itself the unexercised wrapper, and leaving them would have left the split the
argument warned about.

### Corrections to the drafted issue body

- It says nine guards are in the pattern A shape. Counting the three template assertions as one, that is
  right for `core/testutil`; the repository total is thirteen.
- It calls `AssertGofmted` "the weakest of the nine … untested on both halves". Confirmed, and it needed
  a finder before it had a reporting half worth driving: `findUnformatted` is new.
- It proposes "one case per guard for the failing tree and one for the empty walk". Each guard got a
  third, the tree that must **pass**. Without it, a harness that called everything a failure would
  satisfy every failing-tree assertion, so the two cases the draft names cannot carry each other.
- `SourceRoot` is not in the drafted scope and is included anyway. Its ascent is the root every guard
  walks from, its two `t.Fatalf` paths were unexercised, and a guard rooted at the wrong directory is
  the quiet pass its own doc comment warns about.

### Evidence

Each of the thirteen reporting halves blinded in turn through `where.sh mutate`, including the exact
mutation the draft cites as having left the whole core tier green on `8883642d` — `if len(dead) == 0 {
return }` to `if true { return }` in `dead_interface_lint.go`. All thirteen are now **caught**. The
per-guard table is in the pull request body.

Tiers: `--type modules`, `--type modules --race`, `--type lint` (0 issues, four modules). The four
commits touch test files, `CLAUDE.md` and `AGENTS.md` only.

### Follow-ups from this fold-in

None. The scope closes the defect for every tree-wide guard in the repository, and `CLAUDE.md` now
carries a **Guard shape** section so the next one is written with the seam rather than without it.

Issue #337, opened 2026-09-12
[Core refactor 6/29] Decouple shared HTTP and localization helpers from authserver state


leodip:
## Goal

Narrow core handler-helper, middleware, and i18n interfaces so shared packages no longer depend on AuthContext or persistence models.

## Rationale

AuthContext and persistence models remain in core partly because broadly shared helper packages accept their concrete types. Those are dependency-direction problems rather than evidence that the types are shared. Narrowing the seams first lets later moves relocate auth-server state and models without duplicating shared rendering, CSRF, JWT, and localization behavior.

## Implementation guidance

1. Inventory each exported method in handlerhelpers.AuthHelper and split it by consumer.
2. Move auth-context load/save/clear responsibilities behind an authserver-local interface; keep only authentication/authorization and redirect capabilities used by both applications in shared code.
3. Replace i18n dependencies on concrete AuthContext and models.Settings with minimal locale/translation readers or explicit values.
4. Replace shared middleware dependencies on persistence models with focused claims/session/settings values.
5. Keep API shape and helper behavior stable at all call sites; this is interface narrowing, not a handler rewrite.
6. Move or regenerate mocks beside the new interface owner.
7. Preserve request-context logging, error renderers, localization fallback, template behavior, CSRF behavior, JWT claims, and redirects.

## Compatibility and non-goals

- This is a structural refactor only. Do not change business rules, OAuth/OIDC behavior, HTTP routes, status codes, response bodies, persistence behavior, configuration names, logging semantics, or security policy.
- Preserve public serialized representations, including JSON fields, session payloads, cookie names, token claims, database values, and error classification.
- Move or regenerate tests and mocks with the code they cover. Do not weaken assertions to make the move pass.
- Avoid temporary forwarding aliases unless a concrete compatibility consumer requires one; any temporary exception must name its removal issue.
- Keep every commit buildable where practical and finish with gofmt and tidy only in the modules whose imports changed.

## Acceptance criteria

- Core/i18n and the shared portions of handlerhelpers/middleware do not import auth ceremony or persistence model types.
- Both servers render the same localized messages and make the same auth decisions.
- The next issues can move AuthContext and models without introducing reverse module imports.

- The unit tier passes for core, authserver, and adminconsole.
- The repository-wide gofmt and error/logging convention guards pass.
- Relevant race, data, and integration tests continue to pass.
- The PR description identifies the old and new package paths and confirms that runtime behavior is unchanged.

## Dependencies

[...]

Issue #335, opened 2026-09-12
[Core refactor 4/29] Move authserver settings and CORS middleware out of core


leodip:
## Goal

Relocate the settings-loading and database-backed CORS middleware to authserver while leaving middleware genuinely used by both applications in core.

## Rationale

Settings lookup and allowed-origin lookup are auth-server policies backed by the auth database. The admin console neither owns nor executes that behavior. Keeping these files in core/middleware makes a generally shared HTTP package depend on persistence and prevents core from having a coherent boundary.

## Implementation guidance

1. Move middleware_settings.go and middleware_cors.go, with their tests, into authserver/internal/middleware.
2. Preserve constructor signatures at call sites unless narrowing is needed only to break an import cycle.
3. Keep cookie reset, CSRF, JWT/session auth, no-store, real-IP, request logging, and security headers in core when both processes still use them.
4. Reproduce middleware registration in server.go and routes.go without changing order.
5. Preserve database query timing, origin normalization, preflight behavior, headers, status codes, caching, and logging.
6. Add or retain an order-sensitive server test where reordering changes CORS, CSRF, authentication, proxy IP, or response headers.
7. Remove the exact architecture exceptions introduced by #332 for these edges.

## Compatibility and non-goals

- This is a structural refactor only. Do not change business rules, OAuth/OIDC behavior, HTTP routes, status codes, response bodies, persistence behavior, configuration names, logging semantics, or security policy.
- Preserve public serialized representations, including JSON fields, session payloads, cookie names, token claims, database values, and error classification.
- Move or regenerate tests and mocks with the code they cover. Do not weaken assertions to make the move pass.
- Avoid temporary forwarding aliases unless a concrete compatibility consumer requires one; any temporary exception must name its removal issue.
- Keep every commit buildable where practical and finish with gofmt and tidy only in the modules whose imports changed.

## Acceptance criteria

- Core/middleware no longer imports core/data because of settings or CORS.
- Authserver route behavior and middleware order are unchanged.
- Adminconsole no longer compiles the moved middleware or its persistence dependencies.

- The unit tier passes for core, authserver, and adminconsole.
- The repository-wide gofmt and error/logging convention guards pass.
- Relevant race, data, and integration tests continue to pass.
- The PR description identifies the old and new package paths and confirms that runtime behavior is unchanged.

## Dependencies

Depends on #332 and #333.

## Epic context

[...]

Issue #360, opened 2026-09-12
[Core refactor 16/16] Move remaining authserver-only utilities and close the core refactor


leodip:
## Goal

Move any remaining authserver-only utility packages, eliminate every temporary architecture exception, tidy the module graphs, and verify the final shared-kernel boundary.

## Rationale

Earlier issues move code with strong application identities first. Some generic-looking utilities remain in core only because persistence and provider code consumed them during the transition — encryption, hashing, OIDC provider helpers, RSA utilities, URL validation and UUID helpers. This issue uses the actual post-refactor consumer graph to decide their owner, and proves that core is shared by design rather than by history.

## What this issue absorbed

This issue absorbs **#187** (tracked files that mislead or churn) and **#226** (`utils.js` duplicated between the two modules, with the authserver copy unused). Both land on ground this issue already owns in its points 6 and 7 — see *Work folded in from other issues* below. Both are closed and point here.

## Why this is one issue and not several

**This issue was kept separate deliberately.** It is the closer: the census, the tidy, and the burn-down of the exception table to zero. Its content is only knowable once everything else has landed.

## Implementation guidance

1. Repeat a symbol-level and package-level consumer census across production and test code.
2. Move authserver-only packages — `core/encryption`, `core/hashutil`, `core/oidc`, `core/rsautil`, `core/urlutil`, `core/uuidutil` — when the final graph confirms no adminconsole consumer. Move test-only helpers to the tests that own them.
3. Keep genuinely shared packages — API wire contracts, shared OAuth client values, errors, logging, localization and catalog data, shared middleware and session mechanics, common string and reference-data helpers — only when `ARCHITECTURE.md` justifies them.
4. Remove every temporary exception from the architecture guard. Any remaining exception requires a new explicit issue and rationale; an untracked waiver does not complete this issue.
5. Run `go mod tidy` in core, authserver, adminconsole and `cmd/goiabada-setup`. Verify that database drivers, sqlbuilder, OTP and image libraries, and provider-only crypto dependencies are no longer pulled into adminconsole through core.
6. Update `CLAUDE.md` / `AGENTS.md`, docs, diagrams, scripts, Makefiles, CI paths, Dependabot and module configuration, and package comments.
7. Review the complete series diff for accidental logic changes, and remove obsolete aliases, mocks and directories.
[...]

Issue #153, opened 2026-08-08
The CSS build harvests class names from the whole Go module, not from the template globs its config declares


leodip:
Both apps build their stylesheet with `tailwindcss -c ./web/tailwindcss/tailwind.config.js`, whose `content` key lists only `./web/template/*.{html,js}` and the partials and layouts beside it. Because `input.css` uses `@import "tailwindcss"` (Tailwind v4), automatic source detection also runs and scans the whole module, so the generated `web/static/main.css` reflects every file in the tree rather than the templates.

### How it was demonstrated

Executed rather than reasoned. Writing the string for Tailwind's screen-reader-only utility into a `_test.go` file under `src/authserver/tests/integration/`, with every template byte-identical, added an eleven-line rule for it to `src/authserver/web/static/main.css`. Restoring the generated file and rebuilding brought the rule straight back, and it disappeared only when the string was removed from the Go source. No template in the module contains that class (`grep -rn "sr-only" src/authserver/web/template/` returns nothing).

### Why it matters

Two consequences, and the second is the one that matters:

1. Rules ship that nothing renders, which is only bytes.
2. A class deleted from a template keeps shipping for as long as any Go file, comment included, happens to contain the word, so the generated file stops tracking the templates and the config's stated intent silently does not hold.

It also means an unrelated Go change can show up as a diff in a generated CSS artifact, which is how this was found.

### Both apps

`src/adminconsole` has the identical `build.sh` line, the identical config and the same `@import "tailwindcss"`, so it has the same behaviour.

### Suggested fix

Pin the sources so the stylesheet is a function of the templates, for example `@import "tailwindcss" source(none);` plus explicit `@source` lines in each app's `input.css`, then diff the regenerated `main.css` in both apps to confirm nothing rendering today loses a rule.

Found during stage 2 of #109, which worked around it by not spelling the token out. Filed from the closing pass of #109 (PR #150).

leodip:
Absorbed into #429 (6/20), in the package-review refactor plan #423.

This issue's body and every comment are carried there verbatim, under *Carried from #153*, and its acceptance criteria are part of that issue's. Nothing is dropped.

**Why it moved:** the package review of `70f95150` reached the same code independently. Its finding group R7 lands in the same files, so one PR closes both. The plan numbers its parts in dependency order, and this work is part 6.

Closing this one so the work has a single home.

PR #365, opened 2026-09-14
refactor(authserver): move rate limiting and its middleware out of core

### LINKED ISSUE #336: [Core refactor 5/29] Move rate limiting and its middleware to authserver
## Goal

Relocate core/ratelimit and middleware_ratelimiter.go to authserver as one behavior-preserving unit.

## Rationale

Rate limiting is applied only to auth-server credential and unauthenticated endpoints. Its correctness is coupled to route policy, audit throttling, client-IP extraction, cleanup, and concurrency behavior. It is an auth-server security control rather than a shared application primitive, and leaving it in core makes the admin console compile code it never uses.

## Implementation guidance

1. Move the limiter, store, gate, IP-key helpers, middleware, mocks, and tests into focused authserver internal packages.
2. Move GetClientIPFromRequest with the implementation if no shared production caller remains; do not duplicate it.
3. Preserve all limiter names, keys, thresholds, windows, failure-only accounting, success behavior, cleanup lifecycle, proxy handling, and Retry-After/429 responses.
4. Preserve the one-audit-event-per-key-per-window gate and the exact audit payload fields.
5. Rewire route construction without changing middleware order or endpoint coverage.
6. Keep concurrency tests with the implementation and run them under the race detector.
7. Update slog guard path lists/custom functions only for the package relocation, without changing record levels, messages, or attributes.

## Acceptance criteria

- Core has no ratelimit package and no rate-limiter middleware.
- The same endpoints are limited with the same keys and limits.
- ./run-tests.sh --type modules --race passes, including limiter/gate concurrency tests.

## Dependencies

Depends on #335 so auth-only middleware moves happen in small, reviewable PRs.

## Epic context

<details>

- [#333 — Remove dead interfaces and accidental shared dependencies](https://github.com/leodip/goiabada/issues/333)
- [#336 — Move rate limiting and its middleware to authserver](https://github.com/leodip/goiabada/issues/336) (this issue)
- [#338 — Separate shared OAuth client primitives from provider code](https://github.com/leodip/goiabada/issues/338)

</details>

leodip:
Closes #336.

## What this is

Rate limiting is auth-server policy that lives in `core`. `src/core/ratelimit/` holds the counting
primitive — a sliding-window limiter, its in-process store, a fixed-window audit gate and
`CanonicalizeIP` — and `src/core/middleware/middleware_ratelimiter.go` holds the policy built on
top of it: thirteen tiers, ten `Limit*` wrappers, the failure-only reservation machinery, three
rejection classes and the key derivations. Nothing outside the auth server uses either. The admin
console imports `core/middleware` from three files and names no rate-limiter symbol.

This moves both into the auth server: `src/authserver/internal/ratelimit/` for the counting
package, and `src/authserver/internal/middleware/` — the package #335 created for the auth server's
own middleware — for the policy. `GetClientIPFromRequest` travels with the file it lives in, still
exported, and its two handler callers change an import line each.

## The rule it establishes

`core` holds what both processes consume or what is an intentionally stable cross-process contract.
Rate limiting is neither, so it moves, and `ARCHITECTURE.md` stops recording the intent and starts
recording the fact: the `core/ratelimit` ownership row goes, and `core/middleware` becomes `kernel`
rather than `split`, which is what makes the architecture guard start holding the package to kernel
purity. The one edge that survives that flip, `core/middleware -> core/models`, is booked as a
temporary exception against #337 rather than left invisible.

## How it works

The safety argument is the relocation itself: a budget, a key, a window, a threshold or an
accounting rule cannot change if the statement producing it is the same statement in a different
directory. Every limiter name, key derivation, budget, window, failure-only accounting rule,
cleanup lifecycle, proxy handling, `Retry-After` value and 429 body is what it was, and the auth
server's twelve route registrations are unchanged in wrapper, order and coverage. No caller can
observe this change, so no user-facing documentation moves.

Two edits are content rather than relocation, and both are called out in the diff:

- The API rejection arm built the `api.ErrorResponse` envelope by hand because the helper that
  writes it is unexported in the auth-server module and core could not see it. In its new home it
  can, so it calls `apiresponse.WriteError`. Both versions were written into a recorder and
  compared: same status, same headers, same body byte for byte.
- `TestRejection_APIClass` asserted `error_description` was non-empty. It now asserts the exact
  string, so the swap above is pinned rather than remembered.

### Stage 1: move the limiter and its middleware into the auth server

Landed in `cfea993e`. The eight counting files and their tests now live in
`authserver/internal/ratelimit`; the policy middleware and its tests now live in
`authserver/internal/middleware`, with all six consumers rewired and all twelve route wrappers
unchanged. The slog and credential/API/architecture guards moved with the ownership boundary, and
the destination middleware package now loads i18n for its localized rejection tests.

The API 429 arm also folded in the now-available `apiresponse.WriteError` helper. Its output was
probed byte-for-byte against the prior inline envelope, and the unit case now pins the exact message.

Green locally: modules, modules under the race detector, lint, data on SQLite, and integration on
SQLite. No implementation or scope deviations; `go mod tidy` was a no-op. The multi-engine data and
integration matrix is the pushed commit's check suite.

## Verification

Green locally on stage 1, in the dev container: `where.sh test --type modules` (68 packages, 1,862
PASS lines), the same tier under `--race`, `--type lint` (0 issues across all four modules),
`--type data --db sqlite` (397 PASS / 32 SKIP) and `--type integration --db sqlite` (1,045 PASS).
Plus `check-anchors.sh` (24/24), `git diff --check`, `diff CLAUDE.md AGENTS.md` empty, and a
mutation — the API writer's exact description changed — caught by `TestRejection_APIClass`.

Green in the check suite on `cfea993e`, which is what the review read and what is pushed: all
thirteen jobs, including `Tests / mysql`, `Tests / postgres`, `Tests / mssql`, `Tests / sqlite`,
`Unit / race`, `Lint`, `Vulnerabilities` and `Versions`
([run 34857004897](https://github.com/leodip/goiabada/actions/runs/34857004897)). MySQL, PostgreSQL
and SQL Server on the data and integration tiers are the suite's alone; nothing local claimed them.

`Tests / mysql` needed a re-run of the unchanged commit. Attempt 1 failed in `Install Tailwind CLI`
inside `.github/actions/setup-test-container` with curl exit 22 after 26.3 s — the retry budget spent
against the GitHub release CDN — with `Build`, `Data tests` and `Integration tests` all skipped, so
nothing of this change ran on that job while the other three engines passed through the same step.
No fix commit: there was nothing in the tree to fix, and the tree hashes CI ran are identical to the
reviewed ones. A follow-up for the uncached download is drafted in the adjacent-work comment below.

### Review

Reviewed by `openai-codex/gpt-5.6-sol` at effort `xhigh`, 1 round over the whole branch against
`origin/main`. `GBD_REVIEW_MODE=end`, so no stage was reviewed on its own.

|---|---|---|---|
| 1 | 0 | 0 | clean on all three axes: conformance, quality, security |

No blocking or significant finding was raised. The one `unchecked` entry — that no interactive
browser or live reverse-proxy pass was performed — was dismissed with evidence: no template or `web/`
file is in the diff at all, and neither file carrying proxy-derived key handling moved, so a browser
or proxy pass could only observe behaviour no line of this change touches. The reviewer ran its own
mutation against the one file this change authored rather than moved, `test_main_test.go`, and
confirmed it is load-bearing.

Full account, with what the reviewer ran and every claim it verified, in
[this comment](https://github.com/leodip/goiabada/pull/365#issuecomment-5666299927).

## Deviations

None in implementation or scope. Two operational notes: the dev container was created without the 43
`containerEnv` values `devcontainer.json` declares, and `where.sh` supplied them per tier — an
environment workaround, not a narrower configuration; and the `TestMain` probe had to be run by
temporarily substituting it for the landed `TestMain`, since compiling both correctly reports
duplicate definitions, with the landed file restored before verification.

Ceilings left by this change: none.

## Reading this PR

- [Deferred decisions](https://github.com/leodip/goiabada/pull/365#issuecomment-5666286693) — none;
  the six settled ones and what reversing the two substantive ones would cost.
- [Adjacent work](https://github.com/leodip/goiabada/pull/365#issuecomment-5666293056) — what was
  folded in, the one follow-up ready to file with its `gh issue create`, and what was left alone.
- [Review account](https://github.com/leodip/goiabada/pull/365#issuecomment-5666299927) — round 1 in
  full.

## Status

Complete. Stage 1 of 1 landed in `cfea993e`, reviewed clean at the final gate, check suite green on
all thirteen jobs. Nothing is owed for the release notes: no configuration name, endpoint, status
code, header or body changes, so an operator upgrading has nothing to do.


leodip:
## Deferred decisions

**None.** All six decisions in section 3 of the agreement were settled during the interview, before
any code was written, and the run raised no new one and deferred nothing. Nothing here was built on
an assumption you have not already seen and agreed to.

For the record, since this is the list you would otherwise be reading, the six and what they
settled:

| # | Question | Settled as |
|---|---|---|
| 1 | Where does the limiter land, and as how many packages? | Two packages, mirroring today: `authserver/internal/ratelimit` for the counting primitive, `authserver/internal/middleware` for the policy |
| 2 | Where does `GetClientIPFromRequest` go? | With its file, still exported; its two handler callers change an import line each |
| 3 | What does `ARCHITECTURE.md`'s `core/middleware` row say afterwards? | `kernel`, with one temporary-exception row for `core/middleware -> core/models` against #337 |
| 4 | The API rejection builds the error envelope by hand — fold in? | Folded in: it now calls `apiresponse.WriteError`, measured byte-identical |
| 5 | `MiddlewareNoStore` is core-resident with one consumer — fold in? | Discarded: one consumer today is not the membership test `ARCHITECTURE.md` uses; #360's census owns it |
| 6 | The moved tests stub the templates — swap for the real ones? | Discarded the swap, corrected the comment: real templates would couple 23 tests to the auth server's markup |

Decisions 3 and 4 are the two worth a second look, since they are the only ones that change
anything beyond a file path. Decision 3 makes the architecture guard start enforcing kernel purity
on `core/middleware`: if you would rather leave the row as `split` and bump its issue to #337 — which
is literally what #335 did to this row — that is a one-line revert of the ownership row plus deleting
the exception row, and it costs the guard coverage described in the agreement. Decision 4 is about
six lines in a file the change was moving anyway; reversing it means restoring the inline
`api.ErrorResponse` encode and relaxing `TestRejection_APIClass` back to a non-empty check.

leodip:
## Adjacent work: folded in, filed, and left

### Folded in

**Decision 4 — the API 429 arm now uses the shared envelope writer.** Landed in `cfea993e`, in the
same commit that moved the file.

`reject`'s API class hand-encoded `api.ErrorResponse` with `encoding/json`, and the comment above it
said why: the helper that writes that envelope is unexported in the auth-server module, and core
could not see it. The move makes that sentence false — the file now sits in
`authserver/internal/middleware`, which already imports `authserver/internal/apiresponse` — so the
arm calls `apiresponse.WriteError(w, "Too many requests. Please wait and try again later.",
"TOO_MANY_REQUESTS", http.StatusTooManyRequests)` and the comment is deleted rather than rewritten,
having no subject left.

Identical output was measured rather than argued: both versions written into an
`httptest.ResponseRecorder` agree on the status, on every header name and value, and on the body
byte for byte, trailing newline included.

```
status=429
header Cache-Control=["no-store"]
header Content-Type=["application/json"]
body="{\"error_code\":\"TOO_MANY_REQUESTS\",\"error_description\":\"Too many requests. Please wait and try again later.\"}\n"
```

Why it was folded rather than filed: the file was already open, the comment had to be touched either
way, a reviewer reading the moved code would ask the question, and deferring six lines would cost an
issue, an agreement, a run and a merge. `TestRejection_APIClass` previously asserted only that
`error_description` was non-empty, which would have passed even if the message had changed; it now
pins the exact string, so the swap is held by a test rather than by this paragraph.

### Ready to file

**1. CI refetches the Tailwind CLI from a third-party CDN on every job and never caches it.**
Labels: `bug`. Found while resolving the check suite on stage 1.

Duplicate search re-run at the finish: `gh issue list --state all --search "tailwind"` returns only
#153, which is about the CSS build harvesting class names from the whole Go module rather than from
its declared globs — a different defect in a neighbouring area, not this one. `--search "cache CI
flaky"` returns nothing.

Not folded in: no workflow file is in this change's diff, so it fails the first fold-in test —
nothing about it rides on a surface this change touches, and no tier this change runs would catch it
going wrong. It needs three call sites edited, a cache step keyed off `versions.yaml`, and proof
that the cache-miss path still installs a working binary, which is a stage of its own.

```bash
gh issue create \
  --title "CI refetches the Tailwind CLI on every job and never caches it" \
  --label bug \
  --body 'CI installs the Tailwind CLI by downloading it from the tailwindlabs GitHub release on every job that needs it, and never caches it. The command appears three times (`.github/actions/setup-test-container/action.yml`, and `.github/workflows/check.yml` lines 257 and 625), and the composite action is used by eight jobs, so a single run fetches the same pinned binary about ten times from a third party. `curl -sfL --retry 3 --retry-delay 5` gives roughly a 26-second budget; when the CDN exceeds it the job fails at setup, before anything is built, and the whole database leg is lost to an outage that has nothing to do with the commit. This happened on run 34857004897: `Tests / mysql` failed with exit 22 while sqlite, postgres and mssql passed through the same step, and re-running the unchanged commit passed.

The repository already solves this shape twice with `actions/cache`, keyed off the versions pinned in `versions.yaml` — linter binaries at `check.yml:158`, Go modules and build output at `setup-test-container/action.yml:43`. The Tailwind CLI is pinned the same way and is byte identical across jobs, so caching it on `tailwind-${{ runner.os }}-${{ versions.tailwind }}` fits the existing pattern and removes nine of the ten fetches. Keep the download as the cache-miss path, and consider `--retry-all-errors` there so a 4xx is retried as well as the transient set curl already covers.

Found while resolving a red check suite on #336 (PR #365).'
```

### Left alone, deliberately

- **`MiddlewareNoStore`** stays in `core/middleware` even though the auth server is its only
  consumer (decision 5). One consumer today is not the membership test `ARCHITECTURE.md` uses, and
  #360 point 1 is the tree-wide consumer census that should decide it — not a sample of one taken on
  an issue about rate limiting. Its three `r.Use` mounts are untouched, so every response they cover
  keeps its `Cache-Control: no-store` and `Pragma: no-cache`.
- **The moved tests keep their `fstest.MapFS` template stub** (decision 6). Swapping in the auth
  server's real templates is now possible, but it would change what 23 tests assert against and
  couple them to markup, so they would break when a designer edits `auth_error.html`. Only the
  comment explaining the stub was corrected.
- **`middleware_jwt.go`'s dependency on `models.Settings`** is booked as a temporary-exception row
  naming #337 rather than narrowed here. That is #337 point 4's work. Note for whoever lands #337:
  its body carries an "exception rows this issue must delete" table with two rows, and this adds a
  third — the guard will fail #337 until that row goes.

**None.** `git diff --unified=0 origin/main...HEAD | grep '^+.*ceiling:'` returns nothing.

leodip:
## Review account

`GBD_REVIEW_MODE=end`, so no stage was reviewed on its own: this one round over the whole branch
against `origin/main` is the only reading the code got.

**Round 1 — clean, zero findings.** Request `336-20260914T145613Z`, `type: code-review`,
`gate: final`, written 2026-09-14T14:56:13Z against HEAD `cfea993e`. Reviewer
`openai-codex/gpt-5.6-sol` via pi, effort `xhigh`, fresh session, started 2026-09-14T14:57:33Z and
completed 15:08:55Z. All three axes read `reviewed`: conformance, quality, security. No finding to
confirm, defer or refute, and no follow-up raised.

**Which recorded claims still held when the request went out.** `tree-hash.sh --verify`:

```
current   production ae2324b2b20d   unit-tests 05bc9b70c69a   data-tests 1e42e555ec0e
          integration-tests a07c73173c89   docs dbc699ad7ad6   all ddf2d3cc9333

stage 1   unit live   data live   integration live   docs live

all 1 entries still match the tree. Every recorded result stands as written.
```

Stage 1 is the only stage, so no later stage could have moved the tree under it, and every tier
result and every mutation it recorded was measured against exactly the trees on disk.

**What the absence of findings is worth, which is its coverage.** The reviewer did not re-run the
stage's own report. From its `ran` list: `review-diff.sh` against `origin/main` plus `name-status`,
`numstat`, `diff --check` and baseline-to-HEAD source comparisons; `tree-hash.sh` and
`check-anchors.sh` against the agreement; `where.sh ci HEAD`; fanout, stale-path,
route-registration, documentation, catalog, module-boundary, logging-source, reflection, migration
and schema greps; `go list` in the container for both destination packages; `spec.sh 6585 4`;
`where.sh test --type modules` filtered to `TestRejection_APIClass`,
`TestSlogConvention_RateLimitTripIsWarnAndJoinsTheRequest` and
`TestInitRoutes_LimitersAreRegisteredOnTheProductionRoutes`, whose wrapper ran and passed the full
modules tier; and **its own mutation** — `where.sh mutate` on
`src/authserver/internal/middleware/test_main_test.go`, bypassing `i18n.LoadBundle`, which made
`TestRejection_BrowserClass` fail on the missing English message, after which the tree was restored
by hash. That last item is the load-bearing one: the new `TestMain` is the single file this change
authored rather than moved, and the reviewer established independently that it is doing work.

**Its substantive checks**, each verified against the diff before being accepted:

| Claim | Evidence |
|---|---|
| All eight `core/ratelimit` files moved byte-identical | `git diff -M` reports `R100` on all eight |
| `middleware_ratelimiter.go` changed only as agreed | `R097`; no statement change beyond the import rewrite, three comment corrections and decision 4's `WriteError` swap |
| Route wiring unchanged | `routes.go` identical but for the constructor's package qualifier; `grep -c "rateLimiter\.Limit"` still 12 |
| The agreement's anchors still resolve | all 24, exactly once each |
| The agent documents stay byte-identical | `diff CLAUDE.md AGENTS.md` empty |
| RFC 6585 §4 still satisfied | its three requirements met by the unchanged `Retry-After` response |

**`unchecked`, one entry, answered rather than carried.**

> No interactive browser or live reverse-proxy deployment pass was performed.

Dismissed, with the reason, because neither surface has a changed line to observe and that is
checkable rather than asserted. No template or `web/` file is in the branch diff at all
(`git diff --name-only origin/main...HEAD -- '*.html' '*.tmpl' 'src/*/web/'` is empty), so the
browser-class rejection renders through byte-identical markup. Proxy-derived key handling is two
files and neither moved: `src/core/middleware/middleware_realip.go`, which reads the forwarding
headers, is absent from the diff entirely, and `ip.go` carrying `CanonicalizeIP` is `R100`;
`GetClientIPFromRequest` sits in the relocated file and appears in none of its hunks. Production
route wiring is the one-qualifier change above, covered by `routes_ratelimiter_test.go` driving a
real chi router over all 12 registrations, and by CI's four database jobs green at `cfea993e`. A
browser or proxy pass could only observe behaviour no line of this change touches. Not filed: there
is no defect here, and manufacturing a follow-up for a pass nobody is owed would put a permanent
issue on the tracker for the absence of a check with nothing to check.

**Why round 1 was the last round.** Clean is conditional on `findings` being empty, all three axes
reading `reviewed`, and `unchecked` holding nothing the gate depended on. The first two hold
outright. The third holds because the agreement's section 5 seams are the gate's dependency and none
of them is a browser or proxy pass: seam 1 is the counting package's 13 in-package tests, seam 2 the
23 middleware tests driven as an `http.Handler`, seam 3 `initRoutes` on a real chi router, seam 4
the handler credential-failure seams, seam 5 the tree-wide guards. Every one was executed by the
run, and in part again by the reviewer.

**Bookkeeping, fixed in passing.** The reviewer noted that `agreement.sh digest` reported no
as-built note for stage 1 although `log/stage-1.md` carries a populated one. That was a format
mismatch, not a missing note — `digest` matches a bold run at the start of a line and the entry used
a heading — and the entry now opens `**As-built.**`. No code, tier or hash is affected; the
agreement directory is not in the checkout.

Issue #344, opened 2026-09-12
[Core refactor 9/16] Move provider and account validators to authserver


leodip:
## Goal

Move every database-backed and auth-only validator out of `core/validators`, leaving behind only the two symbols the admin console actually uses.

## Rationale

`core/validators` mixes provider endpoint validation (authorization request, token request) with account and profile validation (email uniqueness, profile persistence, password policy, canonical address, phone) and two small shared helpers. Because Go compiles an imported package whole, every consumer of the package inherits its dependency on `core/data`, `core/encryption`, `core/models` and the OAuth provider surface. That single edge is one of the five paths by which all four database drivers reach the admin console's binary.

## What this issue absorbed

This issue replaces the former #345, which split the same package along a line that does not exist in the consumer graph. It also absorbs **#230** and **#124**, two non-epic issues that live in this package and are better done before the move than after — see *Work folded in from other issues* below. All three are closed and point here.

## Why this is one issue and not several

The admin console's production code references exactly two symbols from this package: `validators.ContainsAngleBrackets` and `validators.NewIdentifierValidator`. The whole job is therefore one sentence — keep `identifier_validator.go` and `angle_brackets_validator.go`, move the rest — and splitting it in two gains nothing.

Worse, the split put `authorize_validator.go` and `token_validator.go` in one issue and `email_validator.go` and `profile_validator.go` in the other, but **all four** import `core/data`. So the former #344 would have landed while `core/validators` still imported `core/data`, and its own acceptance criterion ("core/validators no longer imports core/data") would only have been half-true. The package also has one `test_main_test.go` and one `mocks/` directory, which a split reorganizes twice.

## Implementation guidance

1. Re-run the production consumer census before moving anything and use it as the source of truth; the census above is accurate as of the regrouping.
2. Keep `IdentifierValidator` and `ContainsAngleBrackets` in `core/validators`. Move everything else — `authorize_validator.go`, `token_validator.go`, `email_validator.go`, `profile_validator.go`, `password_validator.go`, `address_validator.go`, `phone_validator.go`, their input types, mocks and complete tests — to authserver.
3. Do not land them all in one package. The endpoint validators belong with the provider code moved in #339; the account and profile validators belong with the account services. Split the `mocks/` directory and `test_main_test.go` to match.
4. Move `core/phonecountries` with the phone validator. After this issue its only other consumer, `authserver/internal/handlers/apihandlers`, is in the same module.
[...]

Issue #339, opened 2026-09-12
[Core refactor 8/16] Move the OAuth provider to authserver


leodip:
## Goal

Move the whole provider half of the OAuth code — `core/oauthprovider` (ceremony context, authorization-code issuance, signing-key rotation, token issuance) and `core/oauthdb` (the database-backed token parser) — into the auth server.

## Rationale

#338 already drew the line between the shared OAuth client surface and the provider surface, and put the provider half in its own package. What remains is to carry that package across the module boundary. None of it is a cross-process contract: `AuthContext` is the auth server's authorization state machine, code and token issuance write provider-owned grant state, key rotation owns private-key lifecycle, and the `oauthdb` parser validates tokens by reading the auth server's signing keys. The admin console consumes public JWKS over HTTP and issues nothing.

## What this issue absorbed

This issue replaces the former #340, #341, #342 and #343, which split the same package move into four. They are closed and point here.

## Why this is one issue and not several

`core/oauthprovider` holds four production files — `auth_context.go` (349 lines), `code_issuer.go` (145), `signing_key_rotator.go` (189) and `token_issuer.go` (1380) — plus `mocks/`. Its only importers are seven `authserver` packages (`internal/handlerhelpers`, `internal/handlers`, `internal/handlers/apihandlers`, `internal/handlers/mocks`, `internal/middleware`, `internal/server`, `tests/data`) and its own `mocks/`. There is no adminconsole edge and no core edge into it.

`core/oauthdb` is 484 lines across three files and is imported by exactly one file, `authserver/internal/server`.

Moving these one file at a time therefore means four passes over the same seven consumer packages, with three intermediate states in which `core/oauthprovider` is half-empty and the auth server imports both the old and the new path. The review surface is the diff, and the diff is a path rename; the reviewer's question is "did anything change?", which `git diff -M` answers identically for four files or one. Keep the per-file commit granularity inside this issue instead.

## Implementation guidance

1. Move `auth_context.go`, `code_issuer.go`, `signing_key_rotator.go`, `token_issuer.go`, their tests and their mocks to authserver-owned provider packages. Prefer one commit per file so each rename stays reviewable.
2. Move `core/oauthdb/token_parser.go` and its full test suite with them.
3. Do not land a generic catch-all package. Ceremony state, issuance and key rotation have distinct responsibilities; give them packages that say so.
[...]

PR #367, opened 2026-09-15
refactor(oauth): split shared client primitives from provider code

### LINKED ISSUE #338: [Core refactor 7/29] Separate shared OAuth client primitives from provider code
## Goal

Define the stable OAuth/OIDC surface actually shared with adminconsole and separate it from auth-server provider implementation.

## Rationale

Adminconsole is an OAuth client/resource server: it needs token response values, PKCE, token exchange, and JWT/JWKS validation. It does not issue codes or tokens or rotate provider signing keys. Those roles currently share core/oauth, so provider files leak database dependencies into a package the admin console legitimately imports. TokenResponse also participates in gob-encoded admin sessions, making serialized type identity a compatibility constraint.

## Implementation guidance

1. Inventory every exported core/oauth symbol and record its production consumers.
2. Keep stable shared value types such as TokenResponse, JwtToken, and JwtInfo at a compatible package path unless explicit legacy gob registration is proven by test.
3. Keep or extract shared client-side token exchange, JWT/JWKS validation, and PKCE helpers into cohesive shared packages.
4. Leave provider-only AuthContext, issuers, rotators, and validators for issues 8 through 13.
5. Preserve JSON names, gob registration, accepted JWT algorithms, issuer/audience/expiry checks, clock behavior, JWKS refresh/cache behavior, HTTP client injection, form encoding, client authentication, response limits, and error mapping.
6. Replace broad mocks with consumer-owned narrow interfaces without changing runtime wiring.
7. Add compatibility fixtures for token JSON and gob-encoded session values.

## Acceptance criteria

- Core OAuth client packages have no dependency on core/data or provider persistence models.
- Adminconsole compiles against only the documented shared client/protocol surface.
- Existing admin sessions containing OAuth token responses remain readable.

## Dependencies

Depends on #337. Complete before provider code starts moving.

## Epic context

<details>

- [#333 — Remove dead interfaces and accidental shared dependencies](https://github.com/leodip/goiabada/issues/333)
- [#336 — Move rate limiting and its middleware to authserver](https://github.com/leodip/goiabada/issues/336)
- [#338 — Separate shared OAuth client primitives from provider code](https://github.com/leodip/goiabada/issues/338) (this issue)

</details>

leodip:
Closes #338.

## What this is for

`core/oauth` holds two unrelated things. One half is what an OAuth *client* needs — token response
values, PKCE, JWT and JWKS validation — and the admin console is exactly that. The other half issues
authorization codes, issues tokens and rotates signing keys, and three of the package's twelve
production files import `core/data` to do it.

Because Go's dependency unit is the package, the admin console links all four database drivers and
`go-sqlbuilder` to get at a struct with seven string fields. `core/oauth` is one of exactly two paths
it takes to `core/data`; `core/validators` is the other, and #344 owns that one.

This change draws the line. The shared half keeps the `core/oauth` import path and becomes kernel; the
provider half moves to a new `core/oauthprovider`, which #339, #341, #342 and #343 then empty one file
at a time.

## The rule it establishes

**`core/oauth` is the shared OAuth/OIDC surface and nothing else.** Its ownership row in
`ARCHITECTURE.md` reads `kernel`, which means the architecture guard starts checking its production
imports on every module's unit tier — an exemption it had only because a `split` package has nothing
yet to check at package granularity. After this it imports `core/errs` and `core/hashutil`, and the
second is recorded as a temporary exception owned by #360, beside the `core/handlerhelpers` row that
is the other half of the same pair.

The shared half keeping its import path is deliberate and is the whole reason the direction was chosen.
`TokenResponse` is the only type either binary gob-registers, the admin console stores one in its
session, and gob derives the registered name from the **import path**. Moving it would break every
existing admin session; leaving it costs nothing. A test in this change pins that by asserting the
registered name appears verbatim in the encoded bytes, so whoever moves these types next breaks
loudly rather than silently.

## It is not purely structural

One thing here is behaviour, and it is named rather than smuggled. Three calls the admin console makes
to the auth server read a peer response with no bound, and the code-for-token exchange builds an
`http.Client` with no timeout on every call, so a peer that accepts the connection and never answers
hangs the callback handler indefinitely. This change bounds all three at ten seconds and 1 MiB — the
values the admin console already applies to the same endpoint on every page load — and runs the two
single-use grants on a context detached from the browser's, because abandoning a read after the auth
server has already burned the grant loses the only copy of what it issued.

The issue's own non-goals say a structural refactor would not do this. It is here because it rides on
the same surfaces and the reviewer reads no new layer for it.

## Status

**Ready for review.** All five stages landed, the whole change was reviewed twice at the final gate,
both rounds are fully disposed with nothing refuted and nothing deferred, and the check suite was green
on `9cf5e2bd`. One commit landed after that review, `3462ff0c`, which builds the follow-up this run had
drafted rather than filed; it is described in its own section below and is not covered by either review
round, and the suite is green on it.

### Stage 1: split `core/oauth`, leaving the shared half in place
`74eda30b`

Thirteen provider files moved to `src/core/oauthprovider/` as `package oauthprovider` — `AuthContext`
and the auth states, `response_type.go`, `CodeIssuer`, `SigningKeyRotator`, `TokenIssuer`, their tests,
and the `TestMain` whose data cipher two of them need. The eleven shared files stayed put, so every one
of the 141 files naming only shared symbols is untouched and `TokenResponse`'s gob identity is unchanged
by construction. `git` records each move as a rename.

The dependency runs one way: `token_issuer.go` and its test gained a `core/oauth` import for
`TokenResponse` and `JwtToken`, and nothing left in `core/oauth` names a provider symbol. `core/oauth`
now imports `core/errs` and `core/hashutil` and nothing else, and gained a package doc comment saying
what it is. Thirty-seven consumers changed an import line, a qualifier or both, including three that are
easy to miss: two name moved types in comments only, and one reaches the issuer solely through the
relocated mock package.

`ARCHITECTURE.md` took six edits — `core/oauth` to `kernel`, a `core/oauthprovider` row at #339, the
`core/oauth -> core/hashutil` exception row, the row count, and the two prose sites this change
falsified. `CLAUDE.md` and `AGENTS.md` took five each and remain byte-identical, which the guard checks.
`slog_lint.go` gained the new directory and `agentdocs.go` follows `auth_context.go` to it.
`CreateCodeInput` now carries the constraint between #339 and #341 in its own words.

Tiers: unit green (6877 passing — the tier that runs the architecture guard), lint green (0 issues,
four modules), data green on sqlite (554), integration green on sqlite (1394). The four-engine data and
integration verdict is the check suite's.

Goal 1 was checked against the closure probe rather than argued: the admin console now has exactly one
path to `core/data`, through `core/validators`, and all five driver rows still read `yes`.

One deviation worth knowing: `generate-mocks.sh` could not be run as the plan asked. The dev
container's mockery is ahead of whatever generated the committed mocks and rewrites fourteen of them
that this change does not touch, all with the same boilerplate hunk. The three intended mock changes
were applied in the committed files' own style instead, verified by regenerating and diffing against
the result. The version drift is a separate problem and is drafted as a follow-up.

### Stage 2: pin the wire, and delete the auth server's dead gob registration
`a190102f`

New `src/core/oauth/wire_test.go`, seven cases. The gob one encodes a `map[interface{}]interface{}`
keyed by `constants.SessionKeyJwt` and holding a populated `TokenResponse` — the shape
`handler_auth_callback.go` stores — and asserts the encoded bytes contain the literal
`github.com/leodip/goiabada/core/oauth.TokenResponse`, rather than recomputing gob's naming rule,
then decodes it back and checks all seven fields. That string is in every admin session in flight,
because gob derives a registered name from the import path; the constant holding it carries the
reasoning, so whoever moves these types next meets the cost at the moment they try rather than
after a release.

The JSON table is the shape `core/api/session_wire_test.go` established: `TokenResponse` populated
and zero (`{}`, which is all seven `omitempty` tags at once), `Jwk` populated and zero (all six
keys present, since none carries `omitempty`), `Jwks` populated and zero (`{"keys":null}`, the nil
slice rather than `[]`). Thirteen wire field names had been covered only by integration tests
reading a live response.

The auth server's `gob.Register(oauth.TokenResponse{})` is deleted with the two imports only it
needed: that process writes three session values and all three are strings, and
`git log -S"SessionKeyJwt" -- src/authserver` returns no commit ever, so no released version wrote
one. The admin console keeps its own, which is live, and now says so — deleting one of two
identical-looking lines and leaving the other unexplained reproduces the same confusion facing the
other way.

Tiers: unit green, 6885 passing 0 failing. Data and integration not applicable — no query,
handler, route, migration or schema is touched. Three mutations, all caught; the one against the
gob name is a mutation of the test rather than of production, and the log says why no production
mutation can model a relocated package.

### Stage 3: six cases for `ExchangeCodeForTokens`, no production change
Commit `f2a48bf6`. One new file, `src/core/oauth/token_exchanger_test.go`. No production code
changed.

`ExchangeCodeForTokens` is the admin console's entire authorization-code exchange, and nothing
exercised it: the only reference to it outside the implementation was a hand-written stub in an
admin console lint test. Six cases now do, at the `tokenEndpoint` seam as an `httptest.Server`,
which is the shape `token_parser_jwks_test.go` already uses in the same package: the posted form
and content type, a full token body decoded field by field, `{}` accepted as a zero value, an
empty body refused as a parse error, a non-200 carrying its body into the error, and an
unreachable endpoint.

They are deliberately characterization tests, written against what ships today, because stage 4
gives this function a `ctx` parameter and an injected HTTP client and something has to carry its
behaviour across that. Each negative case names the mechanism meant to refuse it and varies one
thing from a passing case: the non-200 answers JSON that would unmarshal cleanly, so only the
status check can refuse it, and an accepted `{}` sits beside the refused `""` so neither reads as
"short bodies fail".

Tiers: unit green, 6893 passing 0 failing; data green on sqlite, 554 passing. Integration not
applicable — one `_test.go` file, no handler, route or schema. Three mutations, against the grant
type, the status gate and the JSON decode; all three caught.

### Stage 4: bound the three outbound reads, and inject the exchanger's client
`3b942346`

The part of this change that is not structural, and the one the summary above names rather than
smuggles. All three calls the admin console makes to the auth server now carry a ten second
deadline and read at most 1 MiB: the code-for-token exchange, the JWKS fetch and the refresh grant.
The two values are exported constants in `core/oauth` taking what
`adminconsole/internal/apiclient/session_client.go` already applies to the same endpoint — three
calls doing the same thing against the same `/auth/token` must not carry three different numbers —
and neither becomes configuration.

The two grants are detached from the browser's context with
`context.WithTimeout(context.WithoutCancel(r.Context()), TokenExchangeTimeout)`. `authorization_code`
and `refresh_token` are single use: the server has burned the grant by the time it answers, so
abandoning the read when the browser disconnects loses the only copy of what it issued and leaves
the administrator holding a revoked token. `WithoutCancel` rather than `context.Background()` so
`request_id` still reaches every record below. The JWKS fetch is an idempotent read and keeps the
request's own context, gaining only the bound.

`ExchangeCodeForTokens` takes a `ctx` first parameter and `NewTokenExchanger` an injected
`*http.Client`; `routes.go` builds one configured client for the parser, the exchanger and the JWT
middleware. That construction is a named function rather than an inline literal because for the
JWKS fetch it is the only bound there is — both grants build their own deadline, but a browser
context carries none — and a mutation zeroing its timeout survived until a test could reach it.

Tiers: unit green, 6903 passing 0 failing; lint clean across all four modules; data green on
sqlite, 554 passing; integration green on sqlite, 1394 passing — applicable here, since the stage
changes a handler and the route wiring. Seven mutations, all caught. Every oversized-body case
serves a finite, *valid* body of cap + 1024 bytes, so removing a cap flips both the byte count and
the outcome rather than hanging the tier, and each has a benign sibling under the cap so it cannot
be read as "large answers are refused".

### Stage 5: `core/middleware` generates its own `tokenParser` mock
`3edf63cb`. Goal 6. `core/middleware`'s tests were driven by mocks generated from the auth server's
declarations: `core/oauth/mocks.TokenParser` comes from `authserver/.mockery.yaml`, and a kernel
package's tests reused it. The concrete harm is that the auth server narrowing its own
`TokenParser` breaks a kernel package's tests.

`src/core/.mockery.yaml` now declares `core/middleware`'s unexported `tokenParser` beside the
`AuthHelper` entry already there, emitting into `core/middleware/mocks`. The entry carries
`structname: "TokenParser"` because the file's global `structname` template would otherwise emit a
lowercase type name, unusable from outside the generated package. The 39 uses in
`middleware_jwt_test.go` and `middleware_jwt_error_page_test.go` are repointed off the `mock_oauth`
alias and both import lines dropped; both files already imported `mock_middleware` for `AuthHelper`.

`core/validators` is deliberately left alone: ~100 more renames in a file #344 relocates to the
auth server, where its mock regenerates anyway. So `core/oauth/mocks.TokenParser` stays live and
stays generated from the auth server's config, and nothing is orphaned.

Unit green (6903), data green on sqlite (554). Integration not applicable — no handler, route or
schema is touched. One mutation against `middleware_jwt.go`'s `DecodeAndValidateTokenResponse` call,
caught, which is what distinguishes "the repointed tests pass" from "the repointed tests still run".

**Also in this commit, and not part of that change:** regenerating produced a diff on 16 committed
mocks unrelated to this stage. They predated the mockery `3.8.0` that `versions.yaml` pins, so
`generate-mocks.sh` was not idempotent against its own tree and anyone running it collected 16 files
of churn. The whole diff across all 16 is the v3.8.0 template's `t.Helper()` propagation, in each
constructor and inside the `t.Cleanup` closure — nothing semantic. Folded in rather than drafted
because it rides on the surface this stage already operates, there is nothing to settle, and
regenerating is the whole of the fix.

### After the stages: what the final review changed

The whole branch was reviewed at the end rather than stage by stage, and three commits came out of it.

- **`ed4046e3`** — two test gaps from round 1, no production change.
  `TestNewJWKSTokenParser_DefaultsHttpClient` asserted only `NotNil`, so stage 4's default timeout could
  be removed with the package still green; and neither grant test actually pinned the ten second
  deadline, the callback's lower bound being `TokenExchangeTimeout - time.Minute`, which is negative.
- **`e02eb735`** — decisions 14 and 15, both escalated by round 1 and both answered in this thread.
  `response_type.go` and its test go back to `core/oauth`, which takes `core/oauthprovider` out of the
  admin console's production closure entirely and leaves exactly one package there importing `core/data`;
  and the refresh path now hands its detached, deadlined context to `sessionStore.Get` and
  `sessionStore.Save`, so the ten seconds bounds the token call and the write that records it as one
  operation. **#339 now takes `response_type.go` from `core/oauth`, not from `core/oauthprovider`.**
- **`9cf5e2bd`** — round 2's two test gaps, no production change. The refresh test permitted a *fresh*
  persistence deadline, and neither detached-context case pinned the request-value preservation that is
  the entire reason both sites use `context.WithoutCancel` rather than `context.Background`.

### After the review: the drafted follow-up, folded in instead
`3462ff0c`

The one issue this run had drafted rather than filed is built here, so nothing is left to file. Stage 5
regenerated sixteen stale mocks on the pinned generator and left the tree current; what it could not do
was keep it that way, or make the tree say which generator had written it.

`src/mockery-header.txt` now opens every generated mock, inlined by `template-data.boilerplate-file` in
the two `.mockery.yaml` files — a hook mockery's own testify template already has, so no template is
forked and none is inherited. It names the pin, `./version-manager.sh update` writes it, and
`generate-mocks.sh` refuses any mockery but the pinned one, so the stamped version is the generator that
actually ran rather than a literal somebody remembered to change.

Two checks, because they catch different things and the second cannot run everywhere:

- every module's unit tier runs `TestGeneratedMocksArePinned`, holding the pin, the header and every
  generated file to each other through `core/testutil.AssertGeneratedMocksArePinned`. Three file reads
  and a walk: no generator, no git, so the drift that actually happened is caught before a push;
- the lint tier and CI's Lint job run the generator itself and fail on a tree it changed, which is the
  only half that catches an interface newly named in a config. Both sit beside the `versions.yaml` and
  Tailwind gates they are shaped after. Locally it compares digests rather than using git, because a
  worktree's `.git` names the main checkout by host path and git cannot run in the container — which
  also means a passing run writes nothing and needs no restoring trap, there being one `EXIT` trap in
  that script already.

The rule keys on mockery's own `Code generated ... DO NOT EDIT` marker rather than on the `*_mock.go`
name, which is what leaves `core/mocks/test_fs_mock.go` — hand-written, despite the name — and the
countries and timezones tables, which have generators of their own, out of it.

Nineteen mocks gained the two header lines and nothing else; the whole diff across all of them reduces
to those two lines plus the blank one after. `CLAUDE.md` and `AGENTS.md` take a paragraph each and
remain byte-identical, and the guard count in both plus three Go comments goes from thirteen to
fourteen.

## Verification

| Tier | Where | Result |
|---|---|---|
| unit, all four modules | local, `e02eb735`, `9cf5e2bd`, `3462ff0c` | 6927 passing, 0 failing |
| data, sqlite | local, `e02eb735` and again at `3462ff0c` | 554 passing |
| integration, sqlite | local, `e02eb735` | 1394 passing |
| lint, four modules | local, `e02eb735` and again at `3462ff0c` | clean |
| gofmt, vet, unparam, four modules | local, `3462ff0c` | clean |
| anchors | local | 35 of 35 |
| data + integration on mysql, postgres, mssql | **check suite** | green |
| race, vulnerabilities, versions, lint, four unit jobs | **check suite** | green |

The check suite is green on the head commit, `3462ff0c`, and was green on `9cf5e2bd`: twelve jobs pass
on each, and `Build validation` is skipped on both, being push-to-`main` only by design rather than
something this branch turned off. An earlier revision of this section counted it among the green
thirteen, which was wrong in the same way on both commits. The four-engine data and integration tiers
run only in the suite; no local run covers them and none is claimed.
`gofmt`, `vet` and `unparam` are in the table for `3462ff0c` because the local lint tier does not run
them and that commit touches the `Lint` job itself.

Eight mutations, all caught. Five against the new guard — each finding blinded in turn, the empty-walk
fatal blinded, the newline dropped from the header prefix comparison (without it `mockery v3.8.0.1`
satisfies `mockery v3.8.0.`), and the `tools:` section rule defeated in the `versions.yaml` scan. Three
end to end: bumping the pin and applying it left 19 of 19 mocks stale and the generator refusing to run,
and an interface added to a `.mockery.yaml` with nothing regenerated was caught by the lint gate and,
correctly, **not** by the unit guard — which is what shows the two halves are not redundant.

- **Decision 15 reverses decision 3.** `response_type.go` lives in `core/oauth`, not
  `core/oauthprovider`. This supersedes two frozen paragraphs: §2's out-of-scope list and §4's fanout
  both still say thirteen files move and name these two among them. **Eleven moved**, and `core/oauth`
  keeps thirteen files rather than eleven. §0's `oauth/response-type` anchor is repointed.
- **Decision 14-b narrows decision 12's reach.** The detached deadline covers the refresh path's
  persistence and the callback's exchange alone.
- **Stage 1 could not run `generate-mocks.sh`** as its plan asked: the pinned mockery rewrote sixteen
  committed mocks the change did not touch. Its three intended mock changes were hand-applied in the
  committed files' style and verified by regenerating and diffing. Stage 5 then regenerated all sixteen
  on the pinned version, and `3462ff0c` added the stamp and the two checks that keep them that way —
  a fold-in no §2 goal describes, called out in its own commits and in the follow-ups comment below.
  It is the one thing here that came from neither the agreement nor the review: the run had drafted it
  as an issue, and the owner asked for it to be built rather than filed.
- **Goal 9 is a behaviour change in a structural refactor**, which the issue's non-goals forbid. It is in
  §2 by name, decided at the interview rather than by the run, and the section above says what it does.

### Review

Reviewed by `pi openai-codex/gpt-5.6-sol` at effort `xhigh`, 2 rounds over the whole branch. No stage was
reviewed on its own.

|---|---|---|---|
| 1 | 4 | 1 | 2 fixed, `ed4046e3`; 2 escalated as decisions 14 and 15, answered by the owner, built in `e02eb735` |
| 2 | 2 | 0 | both fixed, `9cf5e2bd` |

- **blocking, conformance:** `core/oauthprovider` was still in the admin console's production closure
  through `core/validators`, so the binary compiled the very issuer files this issue exists to remove.
  Escalated as decision 15, answered `15-a`, fixed in `e02eb735`: that package is now out of the closure
  and one package there reaches `core/data`.
- **significant, conformance:** the detached deadline stopped at the HTTP call, so a refresh could fetch
  a replacement token it could not write down. Escalated as decision 14, answered `14-b`, fixed in
  `e02eb735`.
- **significant, quality (×4):** four test gaps, each demonstrated by a mutation the reviewer had run
  green — the JWKS client's default timeout, both grants' deadline assertions, a permitted *fresh*
  persistence deadline, and request-value preservation at both detached sites. All fixed in `ed4046e3`
  and `9cf5e2bd`, each with the mutation re-run here and now **CAUGHT**.

Nothing was refuted and nothing was deferred.

## The three closing comments

- [Deferred decisions](https://github.com/leodip/goiabada/pull/367#issuecomment-5682060641) — none open;
  decisions 14 and 15 restated with what reversing each costs.
- [Follow-ups, fold-ins and ceilings](https://github.com/leodip/goiabada/pull/367#issuecomment-5682073297)
  — nothing left to file: the one drafted issue is built in `3462ff0c`. Two fold-ins; no ceilings.
- [The review account](https://github.com/leodip/goiabada/pull/367#issuecomment-5682103859) — every
  finding of both rounds with its evidence, and what the record's own tier claims are worth.


leodip:
## Decision 14: when an administrator's browser goes away mid-request, should the token the auth server just issued still reach the console's session?

**What this is about.** The admin console signs an administrator in by trading a one-time sign-in code with the auth server for a set of tokens, and keeps them signed in by trading the old refresh token for a fresh one whenever the access token has expired. Both trades are one-shot on the auth server's side: it destroys what it was given at the moment it answers, burning the code in the first case and revoking the old refresh token in the second.

**Where you would meet it.** Signing in to the admin console, where the browser comes back to `/auth/callback` once the auth server is done; and any later console page load after the access token has expired, where the refresh runs in the background before the page renders.

**What happens today.** This change gave both trades a ten second deadline and detached them from the browser's own request, so a browser that goes away — a closed tab, a Stop, a dropped connection — does not take the trade with it. That covers the outbound HTTP call and nothing after it. Both flows then return to the browser's request for the work that makes the answer usable:

- the background refresh writes the new tokens with `m.sessionStore.Save(r, w, sess)` (`src/core/middleware/middleware_jwt.go:316`), and the server-side session store hands that request's context to the database (`src/core/sessionstore/server_side_store.go:517`). Once the browser has gone that context is already cancelled, so the write is refused.
- the sign-in callback validates the tokens and rotates the session cookie on the same cancelled request (`src/adminconsole/internal/handlers/handler_auth_callback.go:109` and `:155`).

The reviewer reproduced the first against the real server-side session store: the detached refresh returned new tokens, and the save then failed with `unable to save the session: unable to create the browser session: context canceled`. The test covering that path passes because its session-store double saves happily whatever the context says.

**Who this costs.** An administrator whose browser abandons a page load while the console is refreshing in the background. The auth server has revoked their old refresh token and issued a new one; the console cannot write the new one down, so it keeps holding the dead one. Their next click signs them out and they log in again — the exact outcome the ten second deadline was added to prevent.

**The choice** is how far past the HTTP call the detached deadline is carried.

| | A: both flows | B: refresh only | C: neither |
|---|---|---|---|
| a background refresh keeps its new token when the browser has gone | yes | yes | no |
| a sign-in finishes server-side when the browser has gone | yes | no | no |
| can leave a signed-in session row whose cookie nobody received | yes, on an abandoned sign-in | no | no |
| what changes | the detached context threaded through both flows' validation and save | threaded through the refresh; the callback's comment and test name narrowed to what they actually claim | both comments and test names narrowed |

**Worked example.** An administrator on a flaky connection opens a console page. The console has to refresh first, and the browser gives up half a second in. Under A and B the fresh token lands in their session and their next click works. Under C the auth server has revoked the old token, the console never records the new one, and their next click signs them out.

The other half of the same example: that administrator closes the tab a moment after approving the sign-in. Under A the console finishes building the signed-in session server-side, but the cookie naming it can never be delivered, so a row holding their tokens sits there until it expires. Under B and C the sign-in simply fails and they sign in again; the code they burned buys nothing either way.

**Recommendation: B.** The refresh is where the harm is real, and carrying the deadline into the save is the whole of the fix. In the callback there is no earlier console session to protect, so carrying it further there buys an unreachable session row rather than a signed-in administrator.

**What does not differ.** A browser that is still there sees no difference under any of the three. Each is a handful of lines and reversible in one commit. A failure anywhere on these paths still clears the console session and sends the administrator to sign in rather than showing an error.

**Why I am asking rather than just doing it.** A and B both mean the console writes to an administrator's session for a request the browser has abandoned, which is session-lifetime behaviour rather than a bug fix. If you would rather the console never wrote anything for a request nobody is waiting for, C is that answer, and it costs one comment and two test names.

leodip:
## Decision 15: the admin console's program still contains the auth server's token-issuing code. Close that here by leaving one file where it was, or leave it to the issue that moves the validators?

**What this is about.** Goiabada builds two programs out of one shared library directory: the auth server, which issues authorization codes and tokens and rotates signing keys, and the admin console, which is only a client of it. The long-running effort this issue belongs to is to stop the admin console's program containing auth-server code at all, database layer included.

**Where you would meet it.** Nowhere at runtime. It is about what ends up inside the admin console binary, and about what the architecture document tells the next reader is inside it.

**What happens today.** This change split the shared OAuth code in two: the client half keeps the `core/oauth` name, and the provider half — issuing codes, issuing tokens, rotating signing keys, all of it touching the database — became `core/oauthprovider`. The admin console no longer names any of that itself. But one file went with the provider half: `response_type.go`, sixty lines with no dependencies of its own, which reads `code`, `token` and `id_token` out of an authorize request. The shared validator package calls it (`src/core/validators/authorize_validator.go:152` and `:266`), and the admin console compiles that package for two unrelated helpers (`src/adminconsole/internal/server/routes.go:38` and `src/adminconsole/internal/handlers/adminresourcehandlers/handler_admin_resource_permissions.go:224`).

So the provider package is still compiled into the admin console's binary, and it still reaches the database:

```
$ go list -tags production -deps ./cmd/goiabada-adminconsole   # inside src/adminconsole
... includes github.com/leodip/goiabada/core/oauthprovider

direct importers of core/data in that closure:  core/oauthprovider, core/validators
the only importer of core/oauthprovider there:  core/validators
```

The stated, checkable goal for this change was that the admin console's build reaches the database through exactly one package. Every route does enter through the validators, which is how the run read it when it landed, but there are two packages in the binary that touch the database rather than one, and `ARCHITECTURE.md` now tells a reader there is one.

**Who this costs.** Nobody today: the console already links all four database drivers through the validators, so the program behaves identically and is much the same size either way. What it costs is the claim. The point of this issue was that the admin console stops compiling code that issues tokens, and today it still compiles all of it under a new name.

**The choice** is where the response-type helper lives.

| | A: move it back to `core/oauth` | B: leave it in `core/oauthprovider` |
|---|---|---|
| provider package in the admin console binary | gone | stays, until the issue that moves the validators lands |
| packages in that binary touching the database | one, the validators | two |
| `ARCHITECTURE.md` | as written now | needs correcting to say two |
| what moves | two files back to where they were before this change, five files change a qualifier | nothing |
| the later issue that takes this file | takes it from `core/oauth` | takes it from `core/oauthprovider`, as its text says today |

**Worked example.** Thirteen provider files moved in this change — the code issuer, the token issuer, the signing key rotator and their companions. Under A the admin console's binary contains none of them the day this merges. Under B it contains all thirteen until a later issue moves the validator package out of the shared directory, and until then anyone reading the architecture document is told otherwise.

**Recommendation: A.** `core/oauth` is where that file lived before this change, it has no imports of its own so it costs the shared half nothing, and it is what makes this change's own criterion true now instead of after another issue. The argument for putting it on the provider side was that only the auth server names it, which is true of the source and not of the binary.

**What does not differ.** No runtime behaviour either way, and the admin console still links all four database drivers either way, through the validators. Both are reversible in one commit, and no test changes meaning under either.

**Why I am asking rather than just doing it.** A reverses a decision taken when this work was agreed, and it hands a slightly different job to the later issue that was going to take this file. If you would rather this change stayed exactly inside the line it drew and let the validator move clean it up, B is that answer and it costs three sentences of documentation.

leodip:
14-b
15-a

leodip:
## Deferred decisions

**None.** No stage proceeded past a question on a stated assumption; the run's closing log carries no
`Status: **Open**` entry, and section 3 of the agreement has no open item.

Two decisions *were* escalated rather than deferred, both raised by the final review's first round and
both answered by you in this thread at 2026-09-15T13:43:43Z (`14-b` / `15-a`). They are recorded as
decisions 14 and 15 and are built, in `e02eb735`. Restated here so this comment is self-contained:

**Decision 14 — how far past the HTTP call does a one-shot grant's detached deadline reach?**
Answered **B, the refresh path only**. `refreshToken` now hands `sessionStore.Get` and
`sessionStore.Save` a request carrying the detached, deadlined context, so the ten seconds bounds the
token call and the write that records its result as one operation. The authorization callback keeps
`r.Context()` for everything after the exchange — there is no earlier console session to protect there,
and finishing a sign-in server-side would write a row of administrator tokens under a cookie that can
never be delivered. Both the production comment and the test's comment were narrowed to claim the
exchange alone.

Reversing to A (carry it through the callback too) is one context substitution in
`handler_auth_callback.go` plus its test's assertions. Reversing to C (carry it no further than the
call in either flow) puts back the gap the review found: a refresh that fetches a replacement token it
cannot write down.

**Decision 15 — `response_type.go`'s placement kept the whole provider package in the admin console's
build.** Answered **A**, which reverses decision 3: `response_type.go` and its test went back to
`core/oauth`. `core/validators` calls `ParseResponseType` and the admin console links `core/validators`
for two unrelated helpers, so the file's placement was deciding the contents of a binary rather than the
ownership of a concept. Verified, not argued: `go list -tags production -deps ./cmd/goiabada-adminconsole`
no longer contains `core/oauthprovider`, and exactly one package in that closure imports `core/data`,
`core/validators` — which is goal 1's checkable, met as written.

This supersedes two frozen paragraphs rather than rewriting them: §2's out-of-scope list and §4's
fanout still say thirteen files move and name these two among them. Eleven moved. **#339 now takes
`response_type.go` from `core/oauth`, not from `core/oauthprovider`.**

Reversing costs the two files back and five qualifier changes, and re-opens the two-importer reading of
goal 1 until #344 lands.

leodip:

**Superseded by `3462ff0c`.** This comment used to carry one issue ready to paste — the residual half of
a draft this run had half-answered itself. The owner asked for it to be built on this branch instead, so
there is now **nothing to file**, and the draft is recorded below as the fold-in it became. The original
text is kept underneath it, struck through in substance rather than deleted, because the reasoning it
gives for *not* doing this in #338 is the thing the decision reverses and a reader should be able to see
both halves.

### Follow-up to file — none

Duplicate search, re-run at the finish: `gh issue list --state all --search` on `mockery`,
`generate-mocks`, `mock generator` and `generated code drift`. The only recurring hit was **#187** (open,
repository hygiene: citations to a gitignored doc, two committed binaries, an orphan mocks directory),
about different artifacts. Nothing to track, because nothing is left over.

- **The mockery template drift itself**, commit `3edf63cb`, stage 5. Sixteen committed mocks
  regenerated on the pinned `3.8.0`; the entire diff across all sixteen is the v3.8.0 testify template's
  `t.Helper()` propagation — the `if helper, ok := t.(interface{ Helper() }); ok` guard in each
  constructor and the same inside the `t.Cleanup` closure — verified by reducing the unified diff to its
  distinct changed lines. Nothing semantic, no signature change. Folded in because it rides on the
  surface the stage already operates (running the generator *is* step 3 of that stage), there is nothing
  to settle, and regenerating is the whole of the fix rather than its first step. The alternative was to
  `git checkout` sixteen files, deliberately holding the tree stale against its own pinned generator, and
  file an issue asking someone to run one command.

- **What that left undone**, commit `3462ff0c`, after the review. Both gaps the draft named, closed:

  1. **The generated header now carries the version.** `src/mockery-header.txt` is inlined above
     mockery's own marker in all nineteen generated files, through `template-data.boilerplate-file` in the
     two `.mockery.yaml` files. That hook is mockery's own — the v3 testify template reads it before
     emitting anything — so nothing forks or vendors a template. `./version-manager.sh update` writes the
     file from `tools.mockery`, which makes it the pin rather than a literal, and `generate-mocks.sh`
     refuses any mockery but the pinned one, which makes the stamp the generator that actually ran.
  2. **Two jobs regenerate or check.** CI's `Lint` job runs `generate-mocks.sh` and fails on a dirty
     tree, beside the `versions.yaml` and Tailwind gates it is shaped after; `./run-tests.sh --type lint`
     does the same locally, comparing digests because git cannot run in the dev container. And a third
     thing the draft did not ask for, because it is the half that needs neither the generator nor git:
     every module's unit tier runs `TestGeneratedMocksArePinned`, holding the pin, the header and every
     generated file to each other. That is the one that catches what actually happened — the pin moving
     without the mocks following — and it catches it before a push rather than in CI.

  The draft's "why not in #338" said §2 asks nothing about mocks and that adding a CI job is a choice
  about the check suite rather than part of the split. Both remain true; the owner made that choice
  explicitly rather than by default, which is what changed. Eight mutations back it, all caught, listed
  in the pull request's Verification section.

Two other fold-ins were decided at the interview rather than found by the run, and are named for
completeness: the outbound HTTP bounds (decision 9, §2 goal 9, stage 4), which is the behaviour change
this pull request's summary calls out, and the auth server's dead `gob.Register` (decision 7, stage 2).

**None.** `git diff --unified=0 main...HEAD | grep '^+.*ceiling:'` returns nothing; this branch adds no
`ceiling:` marker.

---

<details>
<summary>The original draft, for the record</summary>

**Title:** Nothing stops the committed mocks drifting from the pinned mockery again, and a clone cannot tell which generator wrote them

**Labels:** `enhancement`, `go`

**Body:**

`src/authserver/versions.yaml` pins `mockery: "3.8.0"` and the dev container installs it, but nothing
checks that the committed `*_mock.go` files were produced by that version. #338 hit the consequence:
sixteen committed mocks predated the pin, so `generate-mocks.sh` was not idempotent against its own tree
and any change that regenerated mocks for its own reasons collected sixteen files of unrelated churn.
Stage 5 of #338 (`3edf63cb`) regenerated all of them, so the tree is current *today*. Nothing keeps it
that way.

Two gaps remain:

1. **No job regenerates and diffs.** The `Versions` job in `.github/workflows/check.yml` reads tool
   versions out of `versions.yaml` and stops there; no job runs `generate-mocks.sh` and checks the tree
   is clean afterwards. The next mockery bump reintroduces exactly the same drift, and it will again be
   discovered by whoever happens to regenerate.
2. **The generated header carries no version.** It reads:

   ```
   // Code generated by mockery; DO NOT EDIT.
   // github.com/vektra/mockery
   // template: testify
   ```

   So from a clone there is no way to tell which generator the tree is on. The drift is invisible until
   somebody runs the generator.

Suggested fix: a CI step that runs `src/authserver/generate-mocks.sh` and fails on a dirty tree, plus
emitting the mockery version into the generated header so the next drift is readable from a clone rather
than only from a run.

Why not in #338: #338's §2 asks for nothing about mocks, and adding a CI job is a choice about the check
suite rather than part of the split. What #338 *could* do without settling anything — regenerate on the
pinned version — it did.

</details>

leodip:
## The review account

This run used `GBD_REVIEW_MODE=end`: no stage was reviewed on its own, and the whole branch was
reviewed twice at the final gate. That makes this comment the entire code review, so it is written to
stand alone — the verdicts live in an untracked directory on one machine and are pushed nowhere.

Reviewer: `pi openai-codex/gpt-5.6-sol`, effort `xhigh`, both rounds. Six findings total, one blocking.

| Round | Session | Findings | Blocking | Outcome |
|---|---|---|---|---|
| 1 | fresh, request `338-20260915T125000Z` | 4 | 1 | 2 fixed in `ed4046e3`; 2 escalated as decisions 14 and 15, answered by the owner, built in `e02eb735` |
| 2 | resumed, request `338-20260915T140147Z` | 2 | 0 | both fixed in `9cf5e2bd` |

Nothing was refuted, and nothing was deferred. The gate closed after round 2 because round 2 changed no
production code: `production 7961523db1f9` is byte-identical before and after it.

---

### Round 1

Every finding arrived with an executed demonstration rather than a reading — three Go overlays the
reviewer wrote and ran, and one import-graph enumeration.

**Finding 1 — conformance, significant, no forced answer.** Decision 12 detaches the two one-shot grants
from the browser's context, but both flows returned to `r.Context()` for the work that makes the result
usable. The refresh saves through `m.sessionStore.Save(r, w, sess)` (`middleware_jwt.go:316`), and
`ServerSideStore.Save` hands `requestContext(r)` to the backend (`server_side_store.go:517`), so a
cancelled inbound request means the replacement token cannot be written.

*Verified here, not taken on trust:* those three lines read as stated, and the reviewer's overlay against
the real `ServerSideStore` failed with `unable to save the session: unable to create the browser session:
context canceled`. `TestJwtSessionHandler_RefreshesOnACancelledRequestContext` passed anyway because
`mockSessionStore.Save` returns nil whatever the context says.

*Disposition: escalated as decision 14* — sessions and one-shot tokens are not auto-resolved territory,
the reviewer left `forced_answer` empty, and the two flows genuinely differ. Answered `14-b` by the owner
and built in `e02eb735`. See the deferred-decisions comment on this PR.

**Finding 2 — quality, significant, forced. Confirmed and fixed, `ed4046e3`.**
`NewJWKSTokenParser`'s nil-client arm gained `Timeout: TokenExchangeTimeout` in stage 4, but
`TestNewJWKSTokenParser_DefaultsHttpClient` asserted only `NotNil`, so restoring a bare `&http.Client{}`
left the whole `core/oauth` package green. The test now asserts
`assert.Equal(t, TokenExchangeTimeout, tp.httpClient.Timeout)`, the shape its sibling
`TestNewTokenExchanger_DefaultsANilClientToTheConfiguredTimeout` already used.
*Demonstration, re-run here rather than trusting the reviewer's:* `where.sh mutate
src/core/oauth/token_parser_jwks.go`, nil arm replaced with `&http.Client{}` → **MUTATION CAUGHT**.

**Finding 3 — conformance, blocking, no forced answer.** `core/oauthprovider` was in the admin console's
production binary closure, reached only through `core/validators` (which calls `ParseResponseType` at
`authorize_validator.go:152` and `:266`), and it imports `core/data` directly — so that closure had two
direct `core/data` importers where §2 goal 1's checkable says one.

*Verified independently:* `go list -tags production -deps ./cmd/goiabada-adminconsole` included
`core/oauthprovider`; enumerating direct imports across the closure gave exactly
`core/oauthprovider -> core/data` and `core/validators -> core/data`. Stage 1 had recorded the same three
facts and read goal 1 as satisfied, on the grounds that every route enters through `core/validators`.
Both readings are defensible and the disagreement was not about the graph. What is not a matter of
reading: the admin console's binary still compiled `code_issuer.go`, `token_issuer.go` and
`signing_key_rotator.go` — the thing this issue exists to stop.

*Disposition: escalated as decision 15*, because it bears on a §2 goal, which makes the agreement wrong
rather than the code. Answered `15-a` by the owner and built in `e02eb735`; `core/oauthprovider` is now
out of that closure entirely and one package reaches `core/data`.

**Finding 4 — quality, significant, forced. Confirmed and fixed, `ed4046e3`.** Neither grant test pinned
the ten second deadline: the callback test's lower bound was `oauth.TokenExchangeTimeout - time.Minute`,
which is negative and therefore held for any deadline at all, and the refresh test recorded only
`hadDeadline`. The callback's tolerance is now one second; `recordingTransport` records the deadline
itself and the refresh test asserts its remaining duration against `TokenExchangeTimeout`.
*Demonstration:* both mutations re-run here — `oauth.TokenExchangeTimeout` divided by ten at the detached
`context.WithTimeout` call in `middleware_jwt.go` and in `handler_auth_callback.go`, each against its own
test → **MUTATION CAUGHT** both times.

**`unchecked`, four entries, each answered.** (1) Did not read every import-only hunk or generated mock
body — *dismissed*, the qualifier fanout is the compiler's to check and CI checks it on four engines; the
mocks carry `//go:build !production` and the reviewer verified the tags. (2) Did not rerun broad tiers
locally — *covered*, four engines are CI's and the unit tier was re-run after the fixes. (3) No
interactive browser disconnect; callback cookie delivery is a design question — *covered, and it is
decision 14*. (4) No security defect class omitted — *covered*, nothing owed.

---

### Round 2

Both findings are about the tests rather than the code. The reviewer states plainly that "the two
production repairs are sound" and "no security defect was found". Fixed in `9cf5e2bd`.

**Finding 1 — the refresh test permitted a *fresh* persistence deadline. Confirmed, fixed.**
Decision 14-b bounds the token call and the write that records its result as one operation, but
`TestJwtSessionHandler_RefreshesOnACancelledRequestContext` asserted only
`time.Until(saveDeadline) <= oauth.TokenExchangeTimeout` — which any deadline under ten seconds
satisfies. A fresh ten second context minted after the token response stayed green and could stretch the
operation to nearly twenty seconds, while the case's own comment claimed "and it is the same deadline,
not a fresh one". The case now records the absolute deadline at all three seams — the transport, and
`Store.Get` and `Store.Save` through the mock `Run` hooks — and asserts
`transport.deadline.Equal(getDeadline)` and `transport.deadline.Equal(saveDeadline)`.
*Demonstration:* the reviewer's own mutation, which it had run **green**, re-run through `where.sh
mutate` — `detachedReq := r.WithContext(ctx)` replaced by a context minted after the token response →
**MUTATION CAUGHT**.

**Finding 2 — neither detached-context case pinned request-value preservation. Confirmed, fixed.**
Both sites use `context.WithoutCancel` rather than `context.Background` specifically so `request_id`
survives, and both production comments say exactly that — but neither test inspected anything except
cancellation and deadlines, so substituting `context.Background()` left `core/middleware` and
`adminconsole/internal/handlers` entirely green. `grep -rn WithoutCancel --include=*.go src/` returns
exactly the two construction sites, one per state-changing grant, so the finding's exhaustiveness holds.
Fixed using the real mechanism rather than the sentinel the forced answer suggested: each case now puts a
request id on the inbound context under `chimiddleware.RequestIDKey` and asserts `chimiddleware.GetReqID`
returns it at the seam — the exchanger in the callback case, and the transport plus `Store.Get` plus
`Store.Save` in the refresh case. That pins the property decision 12 actually chose.
*Demonstration:* both of the reviewer's mutations, each of which it had run **green** —
`context.WithoutCancel(r.Context())` → `context.Background()` in `middleware_jwt.go` and in
`handler_auth_callback.go`, each against its own test → **MUTATION CAUGHT** both times.

**`unchecked`, four entries, each answered.** (1) No interactive browser disconnect; a real
`ServerSideStore` overlay was used for refresh — *dismissed*, the overlay is the stronger observation and
is what caught round 1's first finding; the callback half is excluded by the owner's answer at 14-b.
(2) Did not rerun broad local tiers — *covered*, the full unit tier was re-run (6903 passing) and this
round touched two `_test.go` files only; `data-tests` and `integration-tests` hashes are unmoved.
(3) Did not reread every import-only hunk or mock body from round 1 — *dismissed*, round 1 read the whole
branch and these are the cheapest hunks in the diff: a wrong qualifier does not compile. (4) No security
defect class omitted — *covered*.

---

### What the record's own tier claims are worth

Run at each gate with `tree-hash.sh --verify`, and stated rather than papered over: **all five stage log
entries record their tiers against trees that have since moved**, because each stage landed and then the
ones after it did. Stage 2's entry recorded only an `all` hash, so its result cannot even be attributed
to a tier.

What settles that is not an argument: the branch has a full local tier run against its own tree at
`e02eb735` — unit 6903, data on sqlite 554, integration on sqlite 1394, lint across four modules, anchors
35 of 35 — and the check suite is green at the head commit on all thirteen jobs.

One limit is named rather than covered: **the per-stage mutations are evidence about a test's power at
the moment they ran**, and `--verify` marks every one of those `unit-tests` rows stale. The five run at
rounds 1 and 2 are current. The earlier ones were not re-run, because re-proving each costs a tier run
for tests whose subject has not been edited since.

Issue #353, opened 2026-09-12
[Core refactor 13/16] Move database selection and factory composition to authserver


leodip:
## Goal

Move `NewDatabase` / `OpenDatabase` and database-configuration mapping to authserver, leaving the `Database` contract and the engine implementations in core for now.

## Rationale

Only the auth server opens the application database, but `src/core/data/database.go` currently combines the large `Database` interface with factory imports for all four engines and `core/config`. Extracting the composition removes the shared module's reason to select drivers, and makes it possible to move the adapters afterwards.

## Why this is one issue and not several

**This issue was kept separate deliberately.** It is the only remaining issue in the epic that is not a move, and it is the highest-value one: it is the single cut point that removes five foreign modules from the admin console's binary, six issues before the persistence packages themselves move.

`src/core/data/database.go` is the only production file in `core` that imports an engine package — every other importer is an auth-server test:

```
src/core/data/database.go:12-18
	"github.com/leodip/goiabada/core/config"
	"github.com/leodip/goiabada/core/data/commondb"
	"github.com/leodip/goiabada/core/data/migrator"
	"github.com/leodip/goiabada/core/data/mssqldb"
	"github.com/leodip/goiabada/core/data/mysqldb"
	"github.com/leodip/goiabada/core/data/postgresdb"
	"github.com/leodip/goiabada/core/data/sqlitedb"
```

Importing `core/data` at all therefore compiles all four drivers, and five of the core packages the admin console imports in production reach it — `core/oauth`, `core/user`, `core/middleware`, `core/validators`, `core/sessionstore` — so closing any one of those paths achieves nothing. Point 7 below closes all five at once.

## Implementation guidance

1. Split the factory and composition functions from the `Database` interface without changing the interface yet.
2. Move engine selection, DSN and config mapping, create/open behavior, `LogSQL` wiring and supported-engine error handling into authserver.
3. Update authserver main, the migrate command, the data-test setup and any setup tooling to call the new factory.
4. During this transition the factory may instantiate engine packages still located under `core/data`. Record those temporary edges against #354.
5. Preserve exact defaults, DSN construction, connection-pool behavior, SQLite single-connection behavior, logging, ping and open timing, the create-database flow and error wrapping.
6. Add table-driven factory tests for all engine names and invalid configuration.
7. Ensure `core/data` no longer imports `core/config` and no longer imports the four engine packages.

## Compatibility and non-goals

[...]

Issue #345, opened 2026-09-12
[Core refactor 14/29] Move account and domain validators owned by authserver


leodip:
## Goal

Move the database-backed email/profile validators and auth-only password/address/phone validation to authserver, leaving only validators genuinely used by both applications in core.

## Rationale

After dead admin interfaces are removed, adminconsole directly uses only identifier validation and the angle-bracket helper. Email uniqueness, profile persistence checks, password policy, canonical address, and phone validation run on the auth server behind its API. Treating all validation as inherently shared confuses a technical category with application ownership and keeps database and auth-only reference data in core.

## Implementation guidance

1. Re-run the production consumer census after #333; use it as the source of truth.
2. Move email, profile, password, address, and phone validators, input types, mocks, and tests to authserver.
3. Move core/phonecountries with phone validation if no admin production consumer remains.
4. Keep IdentifierValidator and ContainsAngleBrackets in core while both applications use them.
5. Preserve normalization, length limits, Unicode handling, reserved values, password rules, uniqueness queries, nullable fields, country/region rules, phone parsing, translated error codes, and validation order.
6. Keep validation at the same trust boundary: admin forms may still perform shared presentation checks, while authoritative validation remains in authserver API handlers.
7. Do not duplicate rule implementations between applications.

## Compatibility and non-goals

- This is a structural refactor only. Do not change business rules, OAuth/OIDC behavior, HTTP routes, status codes, response bodies, persistence behavior, configuration names, logging semantics, or security policy.
- Preserve public serialized representations, including JSON fields, session payloads, cookie names, token claims, database values, and error classification.
- Move or regenerate tests and mocks with the code they cover. Do not weaken assertions to make the move pass.
- Avoid temporary forwarding aliases unless a concrete compatibility consumer requires one; any temporary exception must name its removal issue.
- Keep every commit buildable where practical and finish with gofmt and tidy only in modules whose imports changed.

## Acceptance criteria

- The remaining core/validators package has no data/model dependency and every exported symbol has consumers in both applications.
- Authserver API validation responses and localization are unchanged.
- Adminconsole continues to surface the same field errors returned by the API.

- The unit tier passes for core, authserver, and adminconsole.
- The repository-wide gofmt and error/logging convention guards pass.
- Relevant race, data, and integration tests continue to pass.
- The PR description lists old and new package paths and confirms runtime behavior is unchanged.

## Dependencies

Depends on #333, #337, and #344.

## Epic context

[...]

Issue #231, opened 2026-08-17
Implicit flow response-mode support is inconsistent: form_post is advertised but rejected, and an error for an explicit response_mode=query is emitted in query


leodip:
Goiabada treats response-mode support as a single global set, but its actual capability depends on
the response type. The two facts do not line up, in both directions, and the seam runs through
request validation, discovery metadata, the success emitter, the error emitter, the tests and the
docs.

**1. `form_post` is advertised for implicit requests and refused for them.**

`HandleWellKnownOIDCConfigGet` advertises `response_modes_supported` as `["query", "fragment",
"form_post"]` unconditionally, and `validators.IsSupportedResponseMode` accepts the same three. But
`AuthorizeValidator.ValidateRequest` contains:

```go
// Per RFC 6749 4.2.2 and OIDC Core 3.2.2.5: implicit grant tokens MUST be in fragment
// If response_mode is explicitly set for implicit flow, it must be fragment
if isImplicitFlow && len(input.ResponseMode) > 0 && input.ResponseMode != "fragment" {
    return customerrors.NewErrorDetailWithHttpStatusCode("invalid_request",
        "Implicit flow requires response_mode=fragment or no response_mode (fragment is the default for implicit flow).",
        http.StatusBadRequest)
}
```

So `response_type=token&response_mode=form_post`, and the same for `id_token`, `id_token token` and
`token id_token`, is refused although the server advertises the mode. That refusal is not required by
either document the comment cites. OAuth 2.0 Form Post Response Mode section 4 says:

> "In particular, it is safe to return Authorization Response parameters whose default Response
> Modes are the query encoding or the fragment encoding using the form_post Response Mode."

and Appendix A of the same document demonstrates exactly
`response_type=id_token&response_mode=form_post`. OpenID Connect Core 3.2.2.5 uses the fragment
"unless a different Response Mode was specified", and OAuth 2.0 Multiple Response Type Encoding
Practices prohibits only the *query* encoding for these response types, not `form_post`.

**2. Even with that validation bypassed, the success emitter cannot honour the mode.**
`issueImplicitTokens` in `handler_auth_issue.go` has no response-mode dispatch. It ends in:

```go
http.Redirect(w, r, redirectURI+"#"+encodeResponseParams(params), http.StatusFound)
```

so an implicit success is always a fragment regardless of `AuthContext.ResponseMode`. The error
emitter `redirToClientWithError`, by contrast, does implement a `form_post` branch and renders it
[...]

PR #369, opened 2026-09-16
refactor(validators): move the provider and account validators to authserver

### LINKED ISSUE #123: Refresh token validation dereferences the user from GetUserBySubject without a nil check
## Problem

The `refresh_token` branch of `TokenValidator.ValidateTokenRequest` looks the user up by the token's `sub` claim, never checks the result for nil, and dereferences it about 45 lines later.

`src/core/validators/token_validator.go:501-505`:

```go
sub := refreshTokenInfo.GetStringClaim("sub")
user, err := val.database.GetUserBySubject(nil, sub)
if err != nil {
    return nil, err
}
// no nil check
```

`src/core/validators/token_validator.go:548`:

```go
userHasPermission, err := val.permissionChecker.UserHasScopePermission(user.Id, inputScopeStr)
```

`GetUserBySubject` returns `(nil, nil)` for a subject with no matching row. `commondb.getUserCommon` (`src/core/data/commondb/user.go:70`) only populates the struct when `rows.Next()` reports a row and returns a nil pointer otherwise. So an unresolvable `sub` would reach `user.Id` as a nil dereference.

## This is not reachable today

Worth stating up front, because it sets the priority. Three separate things stop it:

1. `codes.user_id` and `refresh_tokens.user_id` are both `REFERENCES users (id) ON DELETE CASCADE`, so deleting a user removes their codes and refresh tokens. `GetRefreshTokenByJti` returns nil and the request fails well before this point.
2. If the user row is missing for any other reason, `CodeLoadUser` and `RefreshTokenLoadUser` leave `Code.User` / `RefreshToken.User` at the zero value instead of erroring (`src/core/data/commondb/code.go:185`, `if user != nil { code.User = *user }`). `Enabled` is then false and the branch returns "The user account is disabled." earlier.
3. `users.subject` is assigned at creation and never updated anywhere in the codebase, so a live refresh token's `sub` always resolves to its row.

So this is latent, not a live crash. I could not construct a request that reaches it.

## Why fix it anyway

All three guards are incidental to this code and remote from it. Guard 2 is the fragile one: it is a load helper that succeeds silently on a missing row, so it sits one `if user != nil` away from letting a nil user through, at which point the refresh branch panics instead of returning an error.

The codebase already treats this lookup as nil-returning everywhere else:

- `src/core/validators/email_validator.go:81` and `:124`
- `src/core/validators/profile_validator.go:83`
- `src/authserver/internal/handlers/handler_userinfo.go:44`
- the account API handlers in `src/authserver/internal/handlers/apihandlers/`

The two sibling validators carry a comment naming this exact hazard: "An unresolvable subject is not a validation problem the caller can show to a user: it means the request carried a stale or forged subject. Surface it as an error rather than dereferencing nil below."

`token_validator.go` is also internally inconsistent. The `password` branch nil-checks its own user lookup at `:630`. The refresh branch is the only user lookup in the file that does not.

## Suggested fix

Match the sibling validators:

```go
sub := refreshTokenInfo.GetStringClaim("sub")
user, err := val.database.GetUserBySubject(nil, sub)
if err != nil {
    return nil, err
}
// An unresolvable subject on a live refresh token means the token and the users
// table disagree. Surface it rather than dereferencing nil below.
if user == nil {
    return nil, errors.WithStack(errors.New("subject not found: " + sub))
}
```

`errors.WithStack` (a 500) matches `email_validator.go` and `profile_validator.go`, and reflects that an unresolvable `sub` on a token this server issued is an internal inconsistency rather than client error. The alternative is a `customerrors` `invalid_grant` / 400 on the grounds that the token is unusable and the client should stop presenting it. Either is defensible; the in-package precedent points at the first.

## Impact

Low, and defensive only. No behavior change for any request that succeeds or fails today. Four lines in one file, plus a subtest if the branch is worth covering.

Found while reviewing #119, which refactored the code immediately below this lookup. Not caused by that PR.

### LINKED ISSUE #344: [Core refactor 9/16] Move provider and account validators to authserver

leodip:
Closes #344. Absorbs #230 and #124, both closed pointing here; #345 was closed and replaced by this.

## What this is for

`src/core/validators` is 12,811 lines across 19 files, and it is the single edge by which all four
database drivers and `go-sqlbuilder` reach the admin console's binary:

```
adminconsole/cmd/goiabada-adminconsole -> core/validators -> core/data -> core/data/<engine> -> driver
```

Inside the admin console's production closure `core/validators` is the only package that imports
`core/data`, and the only one that imports `core/phonecountries`. Yet the admin console names just
two symbols from it, `ContainsAngleBrackets` and `NewIdentifierValidator`, neither of which touches a
database.

This change moves the seven validators that belong to the auth server into
`authserver/internal/protocolvalidation` and `authserver/internal/accountvalidation`, and leaves
`core/validators` as a kernel package of two files — `identifier_validator.go` and
`angle_brackets_validator.go` — importing nothing but `strings`, `regexp` and `core/i18n`. 90
packages leave the admin console's production closure as a result, the four drivers and
`go-sqlbuilder` among them.

## The rule this establishes

A package in `core` is there because **both** applications need it, not because it was written before
the boundary existed. `ARCHITECTURE.md`'s two tables are the enforcement: the ownership row for
`core/validators` becomes `kernel` with no issue against it, and the five foreign-module rows that
say a driver is `reachable today` become `no`. Both directions are checked by `AssertArchitecture` in
all three module unit tiers, so the document cannot drift from the import graph in either direction.

Two defects are fixed on the way, because neither can be checked except against this move:

- **#230** — no test asserted the English sentence a user actually reads for 24 of the validators'
  messages. A catalog edit swapping "uppercase" for "lowercase" in the password policy message left
  6,929 tests green while the form told the user to fix the wrong thing. The message tables are what
  prove the move changed no message, so they are added first, and `src/core/i18n/error_codes.md` — a
  hand-maintained second copy of 60 of those sentences that nothing checked — is held to the catalog
  by a guard.
- **#124** — the same split-resolve-reject sequence for a `resource:permission` scope was written
  three times, in `ValidateScopes`, `validateClientCredentialsScopes` and `validateROPCScopes`. One
  `resolveScope` replaces all three while the three callers are still in one package, so the
  extraction is a refactor rather than a rename, and each caller keeps its own wording — the three
  sites' error descriptions are asserted byte for byte by the integration suite and none of them
  changes.

#123, a panic on a refresh token whose `sub` resolves to no user, is folded in as a commit before the
move. #137, #116 and #231 are moved as they stand; after this change they live at
`src/authserver/internal/protocolvalidation/token_validator.go` (#137, #116) and
`.../authorize_validator.go` (#231).

## Status

**Complete. All six stages have landed, the check suite is green on `51b2806b`, and the final review
came back clean.**

The thing this issue exists for is done: the admin console's production import closure no longer
contains `core/data`, any of the four database drivers, or `go-sqlbuilder`. 90 packages left it,
measured with `go list -deps -json` against a baseline captured before the move, and
`ARCHITECTURE.md`'s five foreign-module rows now read `reachable today = no` — asserted against the
real import graph in both directions by `AssertArchitecture`, in all three module unit tiers.

Every behaviour change this issue carries was in place and pinned after stage 3. Stages 4, 5 and 6
are the move itself, and `git diff -M` renders them as renames. Each stage's section below carries
its commit and its test totals.

### Stage 1: seven message tables, and the guard that holds `error_codes.md` to the catalog

Landed in `06f039f7`.

All 24 validator messages that no test asserted are now asserted against the English catalog, across
the seven groups that return an `*i18n.LocalizedError` — address, email, password, phone, profile,
identifier and angle brackets. `assertLocalizedErrorCode` gains a fourth parameter and becomes
`assertLocalizedError`; `assertLocalizedCode`, which differed from it only in name, is deleted and
its call sites repoint, so the package has one assertion helper rather than two that would land in
the same new package as duplicates.

`src/core/i18n/error_codes.md` gains the seven missing `validator.authorize.*` rows, taking it from
60 rows against 67 keys to 67 against 67, and `testutil.AssertErrorCodeDoc` now holds it to
`catalogs/active.en.toml` in both directions: exactly one row per `validator.*` and `handler.*` key,
that row's message equal to the catalog value byte for byte, and no row naming a key the catalog
lacks. It is written as a finder plus a reporting half, driven from a rule test through `RunGuard`,
and called from all three module unit tiers.

Unit green on all three modules, **6952 passing** against a 6929 baseline. Seven catalog mutations,
one per group, survived before this and are caught after, each by the group's own unit test when the
run is narrowed to it; three more on the guard itself, including blinding its `Errorf`, are caught by
its rule test. The data and integration tiers are not applicable — no query, handler, route or schema
is touched.

### Stage 2: a refresh token whose subject resolves to no user (#123)

Commit `a4d3906d`. Closes #123.

`GetUserBySubject` answers `(nil, nil)` for a subject that names no row, and the `refresh_token` arm
dereferenced `user.Id` in the permission re-check some sixty lines later, so an orphaned `sub`
panicked the process. The refusal is four lines, placed immediately after the lookup and above the
scope loop.

**The placement is the substance, not the nil check.** The dereference sits in a branch skipped for
an OIDC scope, `offline_access` and the injected `authserver:userinfo` scope, so a check written at
the panic site would refuse a resource-scope refresh and still mint a fresh access token for an
`openid`-only one. The two test rows are exactly that pair, and the resource row registers no
`UserHasScopePermission` expectation, so the strict mock proves the refusal precedes the loop.

**It answers a 500, per decision 8a**, via a single `errs.Errorf("subject not found: %v", sub)`. No
supported operation leaves a live refresh token whose subject resolves to nothing — `DeleteUser`
deletes the user's refresh tokens in the same transaction, and the authorization-code ones go with
`codes.user_id` CASCADE — so this is a disagreement between the tokens and the users table rather
than anything the client did, and it is what this arm already answers for its other
signed-but-impossible states (missing `jti`, missing `offline_access_max_lifetime`, invalid `typ`).
The classification was challenged at the plan gate on RFC 6749 5.2 and is recorded as provisional;
overturning it costs one branch and one test. Both rows assert through `errors.As` that the error is
**not** a client-facing `ErrorDetail`, so the choice is pinned either way.

Tiers: unit **6955 pass**, data (sqlite) **554 pass**, integration (sqlite) **1394 pass**. One
mutation, replacing the refusal with a fabricated user, **caught** by both rows.

### Stage 3: one `resolveScope` for the three scope sites (#124)
Commit `0ca1bcd0`.

`ValidateScopes`, `validateClientCredentialsScopes` and `validateROPCScopes` each carried a copy of
the same sequence - split on `:`, reject a malformed pair, resolve the resource, resolve the
permission on it, reject if absent - and the copies had already diverged: only the client credentials
one kept the resolved row, which is what the #104 fix needs to compare permission ids rather than
bare identifiers. `resolveScope` now owns that sequence, package-level and unexported.

It returns an outcome and never a message. The three sites answer the same conclusion with
deliberately different text and the integration suite asserts all of it verbatim, so each caller
keeps its own `switch` and its own `fmt.Sprintf`; `error` stays reserved for a genuine database
failure and is propagated unwrapped to a 500. Ownership stayed with the callers too: whether a
permission exists on a resource is the shared question, whether it is *this client's* is the client
credentials grant's own rule, and ROPC asks the same of the user.

New tests: a database-failure case per query at the authorize endpoint, pinning that a fault is a 500
and not an `invalid_scope` denial; and ROPC cases for an unknown resource and a permission absent on
a known resource, asserting both descriptions in full, which that site had no unit case for. Five
mutations, all caught - including #104 reintroduced by comparing bare identifiers, and the ROPC
wording replaced with the authorize endpoint's.

Tiers: unit 6961 passing, data and integration green on sqlite. No user-visible change: no endpoint,
status, response body or message moved.

### Stage 4: the five account validators become `authserver/internal/accountvalidation`
`af1602e0`

`email`, `profile`, `password`, `address` and `phone` leave `core/validators` for a new flat package
beside the auth server's other internal ones. Production and test travel together, and `git diff -M`
renders all ten as renames: nine have a one-line diff, the package clause, and
`address_validator.go` has thirteen lines, its import plus the five calls to the staying
`ValidateNoAngleBrackets` that it now qualifies. That is the single moving-to-staying code edge, and
the move being a path change rather than a rewrite is the reason the assertions stage 1 added still
mean what they meant.

Fifteen consumers repoint - nine production files, six test - and the `EmailValidator` and
`PasswordValidator` mocks regenerate into `accountvalidation/mocks` as `mocks_accountvalidation`,
with the two stale files under `core/validators/mocks` removed by hand, since mockery writes files
and never removes them.

Three smaller things land with it. `assertLocalizedError` is **copied** into
`angle_brackets_validator_test.go` rather than moved with its declaration: the staying test calls it
twice, and the two packages are in different modules now, so `core` cannot import the auth server to
share one. `core/testutil/fake`'s `TestEmail` hands its validator cross-check to the package that
owns the rule, as `TestValidateEmailAddress_AcceptsEveryGeneratedAddress` - the property is
unchanged and still proven against the real validator rather than a regex copied into core, which is
what the original comment was written to refuse. And `authserver/internal/accountvalidation` joins
`slogRequestPathDirs` though it writes no record today, so a plain `slog.Warn` added beside a
validator later is refused rather than silently uncovered.

**Tests.** Unit green, 6962 passing, 0 failing. Integration green on sqlite, 1394 passing - run
because the account and users API handlers are repointed. Data green on sqlite, 554 passing. One
mutation, capping the email local part below the twelve characters `fake.Email()` draws: caught, and
the new case is named in the failure list. A first attempt that dropped digits from the pattern also
reported caught, but by other tests only - `fake.Email()` draws no digits - which is why the failure
list rather than the exit code is what settles a mutation.

`core/validators` is left holding `authorize`, `token`, `identifier` and `angle_brackets`.

### Stage 5: `authorize` and `token` become `authserver/internal/protocolvalidation`, and the driver rows fall
`c4e88dca`

The last two validators leave `core/validators`, which is left holding `identifier_validator.go` and
`angle_brackets_validator.go`, importing `strings`, `regexp` and `core/i18n` and nothing else. Ten
consumers repoint, four production and six test. The `AuthorizeValidator` and `TokenValidator` mocks
regenerate into `protocolvalidation/mocks`, `core/validators/mocks/` is deleted, and no mock for a
validator is generated under `core/` any more.

**This is where the drivers leave the admin console.** The edge was
`adminconsole/cmd/goiabada-adminconsole -> core/validators -> core/data -> core/data/<engine> ->
driver`, and severing it removes 90 packages from the production import closure - `core/data`, all
four drivers, `go-sqlbuilder`, and incidentally `core/auditlog`, `core/oidc`, `core/rsautil`,
`core/urlutil` and `core/uuidutil`, which are other issues' ownership rows this quietly advances.
The count is checked rather than asserted: `go list -deps -json` against the finished tree, diffed
with the baseline captured before the move, reports exactly the 90 the grounding experiment
predicted and 0 arriving.

`ARCHITECTURE.md`'s five foreign-module rows therefore flip to `no`, and they are **kept at `no`
rather than deleted**, at `—` for `cleared by` the way the standing `pquerna/otp` row already is.
The epic exists to get those modules out of that binary; a deleted row would stop objecting if a
driver edge ever came back. The document's prose said the drivers leave at 13/16 via #353, on the
argument that `core/data/database.go` is the only production file in `core` importing an engine
package and so the single cut point. They left at 9/16, here, because severing the one edge that
reaches `core/data` does the same job from the other end - which is why `reachable today` is
asserted against the real closure and not derived from an argument about which issue owns the cut.

The `core/validators` ownership row becomes `kernel` with no issue, which turns kernel purity on for
the package. `core/oauth`'s package doc loses its stated reason for `ParseResponseType` staying in
core - the admin console no longer links `core/validators` for anything that reaches it - and the
conclusion now stands on the reason beside it, that `core` may not import the auth server at all.

**Tests.** Unit green, 6962 passing, 0 failing, unchanged from stage 4, which is what a move should
produce. Integration green on sqlite, 1394 passing - run because `handler_authorize.go` and
`handler_token.go` are repointed. Data green on sqlite, 554 passing. One mutation: flipping the
sqlite row back to `yes | #353` fails `TestArchitectureBoundaryHolds` with "nothing reaches it any
more", so the `no` is measured against the real closure rather than a table agreeing with itself.

### Stage 6: `core/phonecountries` becomes `authserver/internal/phonecountries`

Landed in `51b2806b`.

`core/phonecountries` had no admin console consumer at all: both callers are auth server
`apihandlers` that render a country list. It moves to `authserver/internal/phonecountries`, a
package of its own rather than a fold into `accountvalidation`, because `Get()` and `PhoneCountry`
are a UI data table and not a validation rule - merging them would make a validation package export
a country list. Both files move verbatim, with the package clause and the `core/countries` import
untouched, so `git diff -M` renders the pair as renames with zero changed lines. Three production
importers repoint.

`ARCHITECTURE.md` loses the `core/phonecountries` ownership row in the same commit, which the
architecture guard requires rather than merely permits: `checkTableHygiene` fails a row whose
package holds no production Go file.

Two comments in `core/countries` named the moved package and now reach across a module boundary.
The generator's carries more than a path: `maxCallingCodes = 5` mirrors the runtime cap in
`phonecountries.Get`, the two sides are in different modules now, so nothing compiles them together
and that comment is the whole of what holds them. `countries.go`'s named the symbol with no path,
so the planned fanout's grep missed it; it is one clause on a surface this commit already edited,
so it was fixed here rather than deferred.

**Tests.** Unit green, 6962 passing, 0 failing, unchanged - which is what a rename should produce,
and the tier that runs the architecture guard. Integration green on sqlite, 1394 passing, including
the 62 phone-endpoint assertions that check the repointing changed no wire output. Data green on
sqlite, 554 passing. One mutation, against the moved test file, because the risk in a rename is a
test that stopped running rather than one that broke: dropping the `+` from `callingCode` in `Get()`
is caught.

## Verification

Every tier ran. The split between what ran locally and what the check suite reports is deliberate:
each stage ran the unit tier and, where it touched a handler, route or query, the data and
integration tiers **on sqlite** locally, and deferred mysql, postgres and mssql to CI — which is the
only place a four-engine result exists.

**Locally, per stage** (inside the dev container, via `where.sh`):

| Tier | Final result | Baseline |
|---|---|---|
| Unit, three module tiers | **6962 passing**, 0 failing | 6929 at `940d0119` |
| Data, sqlite | **554 passing** | unchanged |
| Integration, sqlite | **1394 passing** | unchanged |

The unit count rises by 33 and never falls: stage 1 adds the message assertions and the
`error_codes.md` guard, stages 2 and 3 add the #123 and #124 cases, and stages 4, 5 and 6 leave it
flat at 6962 — which is what a move should produce and is itself part of the check.

**The check suite, on the final commit** — `51b2806b1d4afab741f3e4b4006830e18d9ce7fa`, green on all
thirteen jobs
([run](https://github.com/leodip/goiabada/actions/runs/35049957348)):

```
Build validation   Lint          Vulnerabilities   Versions      Unit / race
Unit / core        Unit / authserver               Unit / adminconsole         Unit / setup
Tests / sqlite     Tests / mysql Tests / postgres  Tests / mssql
```

That is where the four-engine data and integration result lands, plus the race leg, the lint tier
(golangci-lint, `sloglint`, and the Tailwind and mockery regeneration checks), the vulnerability scan
and the version pins. No stage's deferral is outstanding.

**Mutations, not just green tiers.** A refactor that changes no behaviour is exactly the change where
a passing suite proves least, so each stage ran mutations and recorded which test caught each one —
seven catalog rewordings for stage 1 (one per message group, all of which *survived* before the stage
and are caught after), three on the new guard including blinding its `Errorf`, one on the #123
refusal, five on `resolveScope` including reintroducing #104 by comparing bare identifiers, one on
the email generator's length, one flipping an `ARCHITECTURE.md` driver row back to `yes`, and one on
the moved phone-country table. All caught, each by a named test. The final reviewer re-ran its own
scratch-copy mutations across the same surfaces independently.

Two, both found by review, both corrected in flight rather than carried:

1. **Section 2's sentence "re-running `probe/adminconsole_closure.py` against the finished tree is the
   check" is false**, and the probe as sealed would have reported 0 packages leaving instead of 90
   once the move made its simulation real. The probe was extended with `--mode baseline` and
   `--mode final`, and stage 5's acceptance check names the latter. Section 2's *criterion* — no
   driver, no `go-sqlbuilder`, no `core/data` in the closure — is untouched and is now directly
   asserted. Detail in the review comment, finding 3.
2. **The plan scheduled the `assertLocalizedError` copy one stage too late**, which would have left
   stage 4's tree failing `vet`. Fixed in the plan before stage 4 ran. Detail in the review comment,
   finding 1.

Neither changes what this PR delivers. Section 2 is frozen at the seal, so both corrections live in
the stage records and the log rather than being edited into the agreement.

Also worth naming, because the agreement predicted it: **the acceptance criterion "every symbol
`core/validators` still exports has production consumers in both applications" is false as written,
deliberately.** `ValidateNoAngleBrackets` has no admin console consumer. Decision 2 recorded this
rather than working around it — the wrapper is four lines over `ContainsAngleBrackets`, which both
applications genuinely call, and the guard that actually fires on the package now that its ownership
row reads `kernel` is rule 2, kernel purity, which passes. Moving the wrapper instead would cost eight
`apihandlers` files a new import of an account-validation package for a bracket check on clients,
groups, resources and settings.

## Review

Reviewed by `openai-codex/gpt-5.6-sol` at effort `xhigh`. Two gates, one round each. No stage was
reviewed on its own (`GBD_REVIEW_MODE=end`), so the final round read the whole branch.

|---|---|---|---|---|
| plan | 1 | 3 | 1 | 2 fixed, 1 refuted |
| final | 1 | 0 | 0 | clean, all three axes reviewed, nothing unchecked |

- **blocking, conformance:** stage 4 would have moved the only `assertLocalizedError` declaration
  away from the staying test that calls it, failing `vet`. Fixed in the plan; the copy lands in
  `af1602e0`.
- **significant, quality:** the closure probe simulated the post-move graph by an overwrite that
  becomes a no-op once the move is real, so it would have reported 0 packages leaving rather than 90.
  Fixed in the probe, proven failing against the pre-move tree and passing against the post-move one;
  stage 5's check names `--mode final`.
- **significant, conformance — refuted:** that RFC 6749 §5.2 *requires* HTTP 400 `invalid_grant` for
  an unresolvable refresh-token subject. §5.2's opening sentence carries no RFC 2119 keyword and is
  qualified "unless specified otherwise", and its six token-endpoint error values contain no code for
  an unexpected internal condition. One part of the finding was right and is in the shipped code: the
  nil check must precede the scope loop, not sit at the dereference. The classification remains open
  as this PR's one deferred decision.

Full account, with the evidence and the reviewer's stated execution, in
[this comment](https://github.com/leodip/goiabada/pull/369#issuecomment-5691580918).

## What needs your attention

- **[One deferred decision](https://github.com/leodip/goiabada/pull/369#issuecomment-5691565455)** —
  whether an unresolvable refresh-token subject answers 500 (built) or 400 `invalid_grant`. It settles
  four sites at once, and reversing costs one branch and one test.
- **[Follow-ups, fold-ins and ceilings](https://github.com/leodip/goiabada/pull/369#issuecomment-5691568470)**
  — nothing to file, nothing annexed mid-run, no `ceiling:` markers. The one follow-up candidate was
  declined with its reason, and comes back into existence only if you answer 400 above.
- **[The review account](https://github.com/leodip/goiabada/pull/369#issuecomment-5691580918)** — every
  round, every finding, what answered it.

No release-notes entry is owed: this change asks an operator nothing on upgrade. No endpoint, status,
response body, configuration key, schema or user-visible string moves. The one behaviour change a
deployment could observe is that a refresh token whose subject resolves to no user now returns an
error instead of crashing the process.


leodip:
## Deferred decisions — one, and it is yours to settle

This run met one question it judged safe to postpone. It did not halt the run: the code was built on
the assumption below, and reversing it is contained. Nothing else was deferred.

---

### 1. What should `POST /auth/token` answer when a refresh token's subject resolves to no user?

**Status: Open.**

A refresh token carries a `sub` claim naming the account it belongs to. Today the code does not check
whether that account exists, and the process panics — that is #123, fixed in this PR at `a4d3906d`.
The question the fix had to answer is what it returns instead.

**Assumed, and what is built:** HTTP **500**, via a single `errs.Errorf("subject not found: %v", sub)`.
This is decision 8a as sealed in the agreement.

**The alternative, argued at the plan gate:** HTTP **400** with the OAuth error code `invalid_grant`,
which is what the same endpoint already returns when a refresh token is revoked, expired, or used by
the wrong application.

**Why the 500 was kept.** Three things, all checked rather than recalled:

- **RFC 6749 §5.2 does not force the 400.** Its opening sentence carries no RFC 2119 keyword and is
  explicitly qualified — "The authorization server responds with an HTTP 400 (Bad Request) status
  code (unless specified otherwise)" — and its list of token-endpoint `error` values
  (`invalid_request`, `invalid_client`, `invalid_grant`, `unauthorized_client`,
  `unsupported_grant_type`, `invalid_scope`) contains no code for an unexpected internal condition.
  `server_error` belongs to §4.1.2.1's authorization-endpoint list, a different set. The spec is
  silent on the classification, not prescriptive about it.
- **No supported operation can produce the state.** `DeleteUser` removes the user's refresh tokens in
  the same transaction as the user row, and `codes.user_id` CASCADE plus `refresh_tokens.code_id`
  CASCADE removes the authorization-code ones. A live refresh token with an unresolvable subject
  means the tokens table and the users table disagree.
- **It matches the line the file already draws.** The refresh arm answers its three other
  "this server issued a token it cannot make sense of" states with a plain error and a 500 — no
  `jti` (`token_validator.go:565`), no `offline_access_max_lifetime` (`:750`), invalid `typ`
  (`:792`) — while reserving `invalid_grant` for states a client, a user or an administrator can
  actually cause.

**What the answer buys, either way.** A 400 tells a well-behaved client to discard the token and
re-authenticate, which is friendlier to the client. A 500 is louder for the operator, whose data is
genuinely inconsistent when it fires. Nothing a client can do reaches the state either way, so this
is a choice about who is better served by the noise — not a security or conformance question.

**This settles four sites, not one.** If the 500 is wrong here it is wrong at all four listed above.
Answering "400 `invalid_grant`" converts all four; answering "500" leaves all four as they are. No
issue has been filed for any of them pending your answer.

**Cost to reverse:** one branch and one test for the #123 site; one line plus a test for each of the
other three. Both test rows already assert through `errors.As` that the error is **not** a
client-facing `ErrorDetail`, so the choice is pinned in the suite either way and flipping it fails
loudly rather than silently. Nothing in any later stage depends on it. Reversing after merge is the
same work in a new issue.

leodip:

This run's complete account of everything it found next to the work. Three of the four blocks are
empty, and that is the finding rather than an omission.

### Follow-ups to file

**None.** Nothing was drafted, so there is nothing here for you to `gh issue create`.

One candidate was raised and declined. The plan-gate reviewer proposed filing *"return `invalid_grant`
for malformed required refresh-token claims"*, covering the missing-`jti` and invalid-`typ` sites in
`token_validator.go`. It was verified as real code and not already tracked, but it was **not drafted**,
because it rests entirely on the premise that the sealed 500 for these states is wrong — and that
premise was refuted with the RFC text in plan-gate round 1. With decision 8a standing, those 500s are
the consistent answer rather than a defect. It is folded into the deferred-decision note instead, as
one class of four sites, so your single answer there settles all of them at once. If you answer
"400 `invalid_grant`", this follow-up comes back into existence and should be filed then.

Re-checked at the finish, as the closing pass requires: `gh issue list --state all --search
"invalid_grant refresh token claims"` and `--search "refresh token jti typ"` return only #244, #128,
#239, #42 and #127, none of which reaches these sites. Nothing was filed while the run was working.

### Folded in

**No adjacent open issue was annexed mid-run.** #123 is in this PR, but it was folded in at spec time
by decision 8 and is named in the agreement's section 2 goal — it is scope you approved at the seal,
not scope the run widened.

Two in-scope corrections did land outside their stage's planned fanout. Both are one-line
consequences of the work rather than new scope, and both are named in their stage's log entry and
commit message; they are listed here so no hunk in the diff is unexplained:

- **`06f039f7`** — `assertLocalizedCode` deleted and its call sites repointed to
  `assertLocalizedError`. The two helpers differed only in name. Left alone, they would have landed
  in the same new package as duplicates when the validators moved two stages later.
- **`51b2806b`** — the `core/countries/countries.go` comment naming `phonecountries` by symbol rather
  than by path. The fanout's grep was path-shaped and missed it. One clause, on a file that commit
  was already editing.

**None.**

```
$
```

No `ceiling:` marker was added on this branch. This change ships no simplification with a known limit
— no global lock, no linear scan over something currently short, no heuristic standing in for a rule.
The move is a path change, and the three behaviour fixes (#123, #124, #230) are each the whole of
their fix rather than a first step.

leodip:
## Review account

The complete record of every review round this change got, what each found, and what happened to each
finding. Pasted in full rather than linked: the issue folder this comes from lives on one machine and
is pushed nowhere, so without this comment there is no way to learn what prompted the fix commits,
what was refuted, or what was deliberately left.

Two gates ran: the **plan gate** (one round, three findings) and the **final gate** over the whole
branch (one round, clean). No stage was reviewed on its own — `GBD_REVIEW_MODE=end`.

|---|---|---|---|---|---|
| plan | 1 | `openai-codex/gpt-5.6-sol` @ `xhigh` | 3 | 1 | 2 fixed in the plan and probe, 1 refuted |
| final | 1 | `openai-codex/gpt-5.6-sol` @ `xhigh` | 0 | 0 | clean |

---

## Plan gate, round 1

**Reviewer:** `pi openai-codex/gpt-5.6-sol`, effort `xhigh`, fresh session. Request
`344-20260916T000322Z`, `type: plan-review`, round 1 of 2. Verdict `findings`, three of them. All
three axes — `conformance`, `quality`, `security` — read `reviewed`. `unchecked` held three entries,
all three answered in `log/plan.md`: two dismissed as plan-gate-by-construction, one covered in part
by an addition to stage 2's sketch.

This round read the plan, not code — at this gate the branch is `origin/main` plus one empty
PR-opening commit. It ran `check-anchors.sh` (19/19), `check-fanout.sh` (17 rows, 0 gaps, 0
unchecked), fetched RFC 6749 §§3.3, 4.1.2.1 and 5.2 and RFC 9700 §4.14, wrote three scratch
experiments of its own, and checked the baseline commit's CI (`940d0119`, green on unit, race, lint,
vulnerabilities, build and all four database jobs) rather than re-running tiers.

### Finding 1 — stage 4 strands the staying validator tests without their assertion helper

`conformance`, **blocking**. **Confirmed, and fixed in the plan.**

Stage 1 left the single `assertLocalizedError` declaration in `profile_validator_test.go`. Stage 4
moves that file into `accountvalidation`, while `angle_brackets_validator_test.go` stays in
`core/validators` and calls the helper twice — and stage 5 was the first stage to give the staying
package a copy. The stage 4 tree would therefore fail `vet: undefined: assertLocalizedError` and
could not pass the unit tier stage 4's own `Tiers:` line requires.

The call census reproduces it exactly: `git grep -c` gives 2 calls in the staying test and 18 in the
moving one, and `probe/split_coupling.out` had already named this precise `STAY -> ACCOUNT` edge —
the plan simply scheduled the copy one stage too late. The copy now lands in stage 4's commit
(`af1602e0`). No decision was involved; the compiler forces the answer.

### Finding 2 — the planned 500 for an unresolvable refresh subject

`conformance`, **significant**, with a populated `forced_answer` of `invalid_grant` + HTTP 400.
**Refuted, with the demonstration below — and recorded as a deferred decision so you can overturn it.**

The finding's load-bearing claim is that RFC 6749 §5.2 *requires* 400, which would settle the question
by normative text and leave nothing to decide. Fetched from `https://www.rfc-editor.org/rfc/rfc6749.txt`
rather than recalled, §5.2 opens *"The authorization server responds with an HTTP 400 (Bad Request)
status code (unless specified otherwise)"* — no RFC 2119 keyword, and an explicit escape clause — and
enumerates six token-endpoint `error` values, none covering an unexpected internal condition.
`server_error`, which the finding treats as the alternative the RFC excludes, belongs to §4.1.2.1's
authorization-endpoint list, a different set. The text is silent on the classification, which is
exactly what decision 8a decided.

Two code facts back the sealed choice. `DeleteUser` at `src/core/data/commondb/user.go:599` removes
the user's refresh tokens in the same transaction as the user row, and `codes.user_id` CASCADE plus
`refresh_tokens.code_id` CASCADE removes the authorization-code ones — so no supported operation
leaves a live refresh token with an unresolvable subject. And the refresh arm already answers all
three of its other signed-but-malformed states with a plain error and a 500, while reserving
`invalid_grant` for states a client, user or administrator can cause. The finding read that arm as
uniformly `invalid_grant`; it is not.

**One part of the finding was right, and it is in the shipped code.** The nil check must sit
immediately after the lookup and above the scope loop, because `user.Id` is dereferenced only inside
the loop's permission re-check — and that branch is skipped for OIDC scopes, `offline_access` and the
injected `authserver:userinfo` scope. Placed at the dereference, the fix would refuse a resource-scope
refresh and silently admit an `openid`-only one. Stage 2 pins both cases with a test row each, and the
resource row registers no `UserHasScopePermission` expectation so the strict mock proves the refusal
precedes the loop.

The classification itself remains open and is posted as this PR's one deferred decision.

### Finding 3 — the stage 5 closure probe cannot produce its promised result after the move

`quality`, **significant**. **Confirmed, and fixed in the probe.**

`probe/adminconsole_closure.py` simulated the post-move graph by *unconditionally* overwriting
`core/validators`' imports. Once stage 5 makes that real the overwrite is a no-op, before and after
are the same graph, and it reports 0 packages leaving instead of 90 — the acceptance check for the
whole issue would have passed by measuring nothing.

The probe was extended rather than replaced, and proven both ways in
`probe/adminconsole_closure_final.out`:

| mode | what it does | result |
|---|---|---|
| `--mode simulate` | the original overwrite | reproduces the original byte for byte: 90 leaving, five rows stale |
| `--mode baseline` | captures the pre-move closure | 365 packages, to `probe/adminconsole_closure_baseline.txt` — capturable only at this gate |
| `--mode final` | reads the closure as it actually is | against the pre-move tree **FAILs**, 7 must-be-absent packages still reachable, exit 1; against the post-move graph **PASSes**, 90 left, 0 arrived, exit 0 |

Stage 5's acceptance check names `--mode final`, and `c4e88dca` reports exactly the predicted 90 and
0 against the saved baseline.

This falsifies one sentence in the agreement's section 2 — *"re-running it against the finished tree
is the check"* — while leaving section 2's acceptance criterion (no driver, no `go-sqlbuilder`, no
`core/data` in the admin console's production closure) intact and now directly asserted by the
replacement. Section 2 is frozen at the seal, so the correction lives in stage 5 and the log.

### Follow-up offered and declined

Covered in the follow-ups comment. In short: the reviewer proposed filing "return `invalid_grant` for
malformed required refresh-token claims"; it rests on the premise finding 2 refuted, so it was folded
into the deferred decision as one class of four sites rather than drafted.

### Round budget

One round, which is what the plan gate funds. Two findings fixed, one refuted with evidence, three
`unchecked` entries answered, one follow-up declined with its reason. No round 2 requested: nothing
contested remained that a second plan review could settle.

### Correction to this entry, recorded rather than edited in

The entry as first written cited the refresh arm's three plain-error sites as `token_validator.go:565`,
`:660` and `:793`. Only the first is right. Verified by grepping the file for the
`errs.New` refusal, they are **565** (no `jti`), **750** (no `offline_access_max_lifetime`) and
**792** (invalid `typ`). The two wrong numbers were read off a `sed` window offset rather than off the
file. Nothing else changes: all three sites exist, all three return a plain `errs.New` and therefore a
500, and that is the line finding 2 was refuted against.

---

## Final gate, round 1 — the whole branch

**Reviewer:** `openai-codex/gpt-5.6-sol` at effort `xhigh`, fresh session, gate `final`, round 1 of
max 5. Request `344-20260916T030608Z`, started 2026-09-16T03:08:10Z.

**Verdict: `clean`. Zero findings.** Nothing to fix, nothing to refute, nothing deferred.
`axis_coverage` reports `conformance`, `quality` and `security` all `reviewed`, and `unchecked` is
empty — which asserts full coverage rather than merely reporting no hits.

A clean verdict is only worth its stated execution, so here is what the reviewer reports having run:

- Wrote and exhausted a five-part review plan covering contract evidence, conformance, quality and
  coverage, security, and request reconciliation
- Read `agreement.md` in full, `plan.md`, `log/closing.md`, the final agreement digest, and the
  security reference
- Ran `review-diff.sh origin/main`, then inspected semantic stage diffs and **normalized pre-move
  versus post-move Go blobs** — which is the check that matters for a move: it is how "the files are
  unchanged apart from the package clause" stops being a claim
- Ran `check-anchors.sh`, `check-fanout.sh`, `git diff --check`, stale-path and caller greps,
  `tree-hash.sh --verify`, and `where.sh ci HEAD`
- Ran `go list -deps -json` for the admin console and `adminconsole_closure.py --mode final` against
  the saved baseline
- Compared `error_codes.md` bidirectionally with `active.en.toml`, checked English and Portuguese
  catalog key parity, and checked `AGENTS.md` against `CLAUDE.md`
- Read RFC 6749 §§3.3, 4.1.2.1 and 5.2, plus RFC 9700 §4.14
- Ran **scratch-copy mutations** for the nil-subject guard, the resource-scoped permission-id check,
  all seven localized-message groups, and the error-code document guard — every final targeted
  mutation was caught
- Ran `go list -json` for the three new authserver packages and the remaining `core/validators`, and
  verified the checkout stayed byte-identical throughout

**Reviewer's note:** *"Clean final review. The package moves preserve behavior, current mutations
exercise the changed guarantees, the documented dependency closure is real, and exact-HEAD CI is
green on all supported databases."*

**Recorded by the driver, not by a session.** The verdict met the clean definition exactly — no
findings, every axis reviewed, nothing unchecked — so no session was spent to say so, and nothing in
the tree was touched. Tree hashes at ingest:

```
production         a580b54a125b
unit-tests         40410622dfcb
data-tests         10910c50d814
integration-tests  cfae00042bba
docs               85fceb67a8f3
all                7378cc7b9e2e
```

No round 2: a clean round with full axis coverage is where the final gate stops.

Issue #359, opened 2026-09-12
[Core refactor 15/16] Move persistence contracts, models, common SQL, and startup data services to authserver


leodip:
## Goal

Move the remaining core persistence boundary into authserver: the `Database` interface, the `core/data` root and `commondb`, persistence models and mocks, seeding and startup tasks, and database-backed audit support.

## Rationale

After the API boundary, the adapters and the tooling have moved, no legitimate adminconsole consumer remains for persistence models or database services. This is the coordinated cutover that cannot be split further: the `Database` interface, the common SQL implementations and the model types refer to one another, so moving only one would create a forbidden core-to-authserver edge or a temporary duplicate set of domain models.

## Why this is one issue and not several

**This issue was kept separate deliberately**, for the reason its own rationale gives. It is the one remaining piece where the coupling is real rather than incidental: `core/data` (1246 production lines), `core/data/commondb` (11313), `core/models` (30 files) and `core/auditlog` are mutually referential, and any cut between them is a forbidden edge.

## Implementation guidance

1. Run and attach an import census proving `core/api` and adminconsole no longer import `core/models` or `core/data`.
2. Move the `Database` interface, the transaction runner and contract, the data root helpers, `commondb`, the database mocks, the persistence models and their tests, the seeder, the startup, backfill and key-rotation services, and the database-backed audit helpers into authserver.
3. Update the four moved engine adapters to local contract, model and common imports in one coordinated change.
4. Preserve every interface method signature and transaction rule until all call sites compile. Narrowing the interface or redesigning the repository is out of scope.
5. Preserve struct and database tags, scan and value behavior, `dont-update` fields, `sql.Null*` semantics, model helper results and encryption hooks.
6. Preserve `RunInTransaction` retry bounds and classification, statement order, audit-after-commit behavior, startup idempotency, backfills, seed defaults and rotation behavior.
7. Move the data tests and their four-engine matrix without weakening the shared conformance assertions, and move the `core/data` lint tests — begin-transaction, benign-sentinel and page-offset — with the package they guard.
8. Delete the old packages rather than leave aliases once all imports move.

## Compatibility and non-goals

- This is a structural refactor only. Do not change business rules, OAuth/OIDC behavior, HTTP routes, status codes, response bodies, persistence behavior, configuration names, logging semantics, or security policy.
- Preserve public serialized representations, including JSON fields, session payloads, cookie names, token claims, database values, and error classification.
[...]

PR #370, opened 2026-09-16
refactor(core): the six application-service and leaf packages move to the auth server

### LINKED ISSUE #198: StartNewUserSession commits the new session before the work that can still fail, so a late error leaves a row nobody can reach
`UserSessionManager.StartNewUserSession` (`src/core/user/usersession_manager.go`) opens a transaction, inserts the `user_sessions` row and its `user_session_clients` rows, and commits. Only after that commit does it read the user's other sessions, delete the ones matching this device and IP, load the cookie session, and write the new identifier into it. Each of those four steps returns `nil, err` on failure and none of them undoes the commit, so a failure after that point leaves a session row that no browser holds a cookie for, and it may already have deleted the session the browser was using.

Verified by reading the function: `CommitTransaction` is followed by `GetUserSessionsByUserId`, a `DeleteUserSession` loop, `sessionStore.Get` and `sessionStore.Save`, with an early `return nil, err` at each. The caller sees only the error and cannot tell that a session was created, so it returns 500 and the row stays.

The visible effect is an orphaned session row that no cookie points at, counted by the admin console's session list and holding its own idle timeout and maximum lifetime, plus the possible loss of the session the browser did have. Nothing is issued against the orphan, because the ceremony stops, so this is a consistency and hygiene defect rather than an access-control one.

Decide between compensating cleanup (delete the row before returning the error) and moving the commit boundary below the sibling sweep, then pin it with tests that assert database state through the real manager rather than a nil return: today's coverage checks the returned error only, so the leftover row is invisible to it.

Found while working #133 (PR #196). Related in shape but not the same site: #160 (partial commit in `issueAuthCode`).

### LINKED ISSUE #346: [Core refactor 10/16] Move the auth server's application services and leaf packages

leodip:
Closes #346. Closes #198.

`core/user`, `core/useragent`, `core/communication`, `core/otp`, `core/imaging` and
`core/uithemes` are application services and leaves of the auth server that have been sitting in the
shared kernel. Nothing in the admin console reaches any of them — all 35 files outside the six
directories that name one are under `src/authserver/` — so the only thing keeping them in `core` was
that nobody had moved them. This moves them, and the rule it establishes is the epic's:
a package in `core` is there because both binaries compile it, not because it started there.

## What it does

Eight packages appear under `src/authserver/internal/`, flat beside `ceremony`, `issuance`,
`signingkeys` and the rest, and six directories leave `core`:

| new package | from | why |
|---|---|---|
| `permissions` | `core/user/permission_checker.go` | the three user services have distinct responsibilities and do not share a catch-all |
| `usercreation` | `core/user/user_creator.go` | " |
| `usersession` | `core/user/usersession_manager.go` | " |
| `useragent` | `core/useragent` | two consumers now, `usersession` and `issuance`, so it cannot live inside either |
| `emaildelivery` | `core/communication` | `email` is unusable: two handlers declare a local `email` that shadows it |
| `otp` | `core/otp` | leaf |
| `imaging` | `core/imaging` | leaf |
| `uithemes` | `core/uithemes` | leaf |

Every moved file is a rename: `git diff -M` renders it as one, and the package clause, the import
block and the qualifiers at its call sites are the only edits. The six interfaces the auth server
declares and mocks — `PermissionChecker`, `UserCreator`, `UserSessionManager`, `EmailSender`,
`OtpSecretGenerator` and `TokenParser` — now write their mocks into
`authserver/internal/handlers/mocks`, beside the interfaces that declare them, leaving `HttpHelper`
as the only mock the auth server's config still generates under `core/`.

`ARCHITECTURE.md` loses six ownership rows and gains no exception: every core package the movers
import is already authserver-owned or kernel. `core/go.mod` stops requiring `github.com/pquerna/otp`
and `golang.org/x/image`.

## How it is checked

The 120 tests that travel with the code are the fixture. They own everything the move preserves —
session validity, bump and create, ACR maximum, AMR ordering, `OtpConfigGeneration` snapshots, user
creation defaults and permissions, scope filtering, SMTP dialling and mechanism choice, TOTP
algorithm and skew, image content detection, theme ordering, Client Hints parsing — and not one
assertion is weakened. The three module unit tiers carry the guards that would notice a boundary
mistake: `AssertArchitecture` in both directions, `AssertGofmted`, `AssertSlogConvention`,
`AssertNoLegacyErrors`, `AssertAgentDocs` and `AssertGeneratedMocksArePinned`.

CI's mailpit service moves with the one test that needs it, from `Unit / core` to
`Unit / authserver`, in the same commit.

## What rides along

- **#198**, in its own commit after the move: a user session row committed by a login whose later
  steps failed outlives the request, holding no cookie and issuing nothing, until its idle timeout
  expires — and may already have deleted the session the browser did have. The sibling sweep joins
  the transaction, the browser-session read is hoisted above it, and the one call left after the
  commit is compensated.
- **`src/adminconsole/go.mod` and `go.sum`**, tidy at last: 22 indirect requirements and 107
  `go.sum` lines that nothing in that binary's build graph needs, the four database drivers and
  `go-sqlbuilder` among them, left behind when #344 severed the edge. `ARCHITECTURE.md` stops
  asserting they cannot be removed.
- **`testutil.ClearMailpit`**, deleted: declared, called from nowhere, and invisible to `unused`
  because it is exported test support.

## Status

**Complete.** All four stages landed, the whole change was reviewed once at the final gate and came
back clean, and the check suite is green on the final commit.

Two things happened after that. The single follow-up the closing pass wrote up was folded in as
`e66b4408`, then **reverted in `55548b97`** once two findings against it held: the compensation it
added could not be conclusive, and the defect it addressed already had a backstop. The window is
recorded as a ceiling instead, and a second hazard that review surfaced — pre-existing, and worse —
is filed as #371. The record comment carries the whole account.

### Review

Reviewed by `openai-codex/gpt-5.6-sol` at effort `xhigh`, fresh session each round. `GBD_REVIEW_MODE=end`:
no stage was reviewed on its own, so the final gate is the only reading the code got. The plan gate
is counted here because its findings changed what the stages built.

| Round | Scope | Findings | Blocking | Outcome |
|---|---|---|---|---|
| Plan gate, 1 | plan + agreement at `2992a5bd` | 4 (all minor, conformance) | 0 | all fixed, `4797cff8`, `bda56f75`, `6a9843b8` |
| Final gate, 1 | whole branch vs `main`, 82 files | 0 | 0 | clean; all three axes reviewed |

- **minor, conformance:** `Unit / race`'s mailpit comment pointed at `Unit / core`, the tier the
  service leaves. Fixed in `4797cff8`: comment repointed, service kept.
- **minor, conformance:** `ARCHITECTURE.md:209`'s admin console `go.mod` claim is false in both
  clauses after decision 6's tidy, not only the second. Fixed in `bda56f75`: whole claim rewritten,
  shorter, no sentence added.
- **minor, conformance:** `usersession_manager.go:132` said the same-device sweep runs after the
  commit, which decision 9 inverts. Fixed in `6a9843b8`.
- **minor, conformance:** `run-tests.sh:490` attributed `TestSendEmail` to core. Fixed in
  `4797cff8`: owner repointed, cache argument kept.

The final round ran four independent scratch mutations of its own at the #198 seam and caught all
four. Full account, with the evidence, in
[this comment](https://github.com/leodip/goiabada/pull/370#issuecomment-5697752857).

### Tiers

- **Locally**, per stage in the dev container: unit on all three modules (6966 passing at stage 4),
  `go vet` in all four modules, data on sqlite (554), integration on sqlite (1394), lint including
  the mock-regeneration check, and `golangci-lint` on the touched packages. Plus the mutations —
  one per behaviour each stage claimed to make observable, all caught.
- **The check suite**, which is the only place the four-engine data and integration tiers run:
  green on `6a9843b8` across all thirteen jobs — `Unit / core`, `Unit / authserver`,
  `Unit / adminconsole`, `Unit / setup`, `Unit / race`, `Tests / sqlite`, `Tests / mysql`,
  `Tests / postgres`, `Tests / mssql`, `Lint`, `Vulnerabilities`, `Build validation`, `Versions`
  ([run](https://github.com/leodip/goiabada/actions/runs/35095397663)). That matters most for
  stage 4: SQLite has one connection and cannot deadlock, so the retry path the widened transaction
  leans on is exercised on the other three engines and nowhere locally.

### Deviations from the agreement

Three, each in a decision rather than inferred, plus one silence filled while building.

- `core/useragent` became its own package rather than travelling inside `UserSessionManager`'s,
  because #339 left it with two authserver consumers rather than one (decision 1).
- `core/communication` became `emaildelivery` rather than anything containing `email` alone, because
  two handlers shadow that identifier (decision 2).
- The admin console's `go.mod` was tidied although its imports did not change, which the issue's
  "tidy only in the modules whose imports changed" excludes (decision 6).
- Decision 9 is silent on a compensating delete that itself fails; the cleanup error is `errs.Join`ed
  onto the original rather than dropped, so `errors.Is` still finds the reason the ceremony failed.

### The record

- [Deferred decisions](https://github.com/leodip/goiabada/pull/370#issuecomment-5697738164) — none;
  nothing was built on an unanswered question.
- [Follow-ups, fold-ins and ceilings](https://github.com/leodip/goiabada/pull/370#issuecomment-5697744474)
  — one filed (#371), four things folded in, two ceilings.
- [Review account](https://github.com/leodip/goiabada/pull/370#issuecomment-5697752857) — every
  round, what it ran, and what happened to each finding.

### Stage 1: the six packages move to the auth server

Landed in `4797cff8`. 76 files changed, +814 / -791; 29 of them are renames under `git diff -M`,
and the one added file is the copy of `run_in_transaction_stub_test.go` that the `core/user` split
requires in both `usercreation` and `usersession`.

Eight packages appear under `src/authserver/internal/`: `permissions`, `usercreation`,
`usersession`, `useragent`, `emaildelivery`, `otp`, `imaging`, `uithemes`. Six mocks move to
`handlers/mocks` as `mocks_handlers` — the five from the moved packages plus `TokenParser` — and
are the pinned generator's output rather than hand-moved copies. `ARCHITECTURE.md` loses the six
`#346` ownership rows and no exception replaces them. `core/go.mod` loses `github.com/pquerna/otp`
and `golang.org/x/image`; `authserver` promotes `x/image` to a direct requirement. CI's mailpit
service moves from `Unit / core` to `Unit / authserver`, following the one test in either tier that
needs it.

Checked: unit green in all three modules (6962 passing, all eight new packages ran); `go vet ./...`
clean in all four, which is the only thing here that compiles `tests/data` and `tests/integration`;
integration green on sqlite (1394 passing); lint green including the mock-regeneration check.
A mutation narrowing the OTP skew window is caught by the moved `otp` tests, so they assert in
their new home.

Worth knowing: `communication.EmailSender` is not among the moving qualifiers after all — it is the
interface `handlers` declares, not a type in the moved package. The CI comment edit came to four
comments rather than the two the plan scoped, because the `services:` block carries its own reason
between jobs and the "no services" note has to be transferred to `Unit / core` rather than dropped.
The admin console's `go mod tidy` was deliberately left for stage 3.

### Stage 2: the eight new directories join the slog request-path list
Landed in `50b6b7d1`. The eight packages stage 1 created join `slogRequestPathDirs` in
`src/core/testutil/slog_lint.go`, so rule 5 of the slog convention — no plain
`slog.Debug`/`Info`/`Warn`/`Error` in a directory a request runs through, since a record with no
context cannot carry `request_id` — now reaches them. Their six predecessors were never listed, so
this loses no coverage; it closes a gap while the file is open, which is the position
`accountvalidation` was in when #344 listed it. All eight run only while serving a request and none
writes a record today, so no moved code is edited. The comment above the list and pattern 8 of
`CLAUDE.md` and `AGENTS.md` both enumerate the list and are corrected in place, and the agent docs'
Auth Server directory list gains the eight packages on one line.

Checked: unit green on all three module tiers, 6962 passing, 0 failing. No test is added — the
stage's whole product is coverage the existing guards apply, and a test asserting the list holds
eight strings would assert the list against itself. Two mutations are the evidence instead, both
caught: a plain `slog.Warn` planted in `internal/uithemes` is refused with "a plain slog.Warn in a
request-path package", a message only the new entry can produce, and a misspelt path in the list is
refused by `TestSlogConvention_EveryRequestPathDirExists`. Data and integration not applicable: no
query, migration, schema, handler or route is touched.

Worth knowing: all three prose edits were drafted over their stated size and cut back before the
commit rather than shipped — the guard comment's clause from 29 words to 15, to match the `#344`
clause it was told to be written in the voice of, and the agent-doc line from 20 words to 16.

### Stage 3: the folded-in cleanups
Landed in `bda56f75`. Two cleanups folded in because each rides on a surface this change already
touches and costs less than the issue, agreement and review cycle filing it would take.
`go mod tidy` in `src/adminconsole` drops 23 indirect requirements nothing in its build graph needs
— the four database drivers and `go-sqlbuilder` among them — and takes `go.sum` from 125 lines to
18. That staleness is #344's, not this issue's: it reproduces on an untouched copy of `main`, and
nothing checks it, because no CI job and no tier runs `go mod tidy`. `testutil.ClearMailpit` is
deleted, declared and called from nowhere; being exported test support, `unused` cannot see it and
this tree deliberately does not enable `exported-is-used`, which is how it survived looking like the
way to reset mailpit between tests.

`ARCHITECTURE.md` asserted the opposite above its foreign-module table, that `adminconsole/go.mod`
"still lists all four database drivers as indirect dependencies, and will until `core` stops
requiring them". The tidy falsifies both clauses rather than only the second, so the whole
current-version claim is rewritten — shorter than it was, adding no sentence, and leaving the
argument that follows it untouched, because that one is still true and is the one that matters: a
`go.mod` line is what the module graph permits rather than what the linker pulls in, and the import
closure is what the guard measures.

Checked: unit green on all three module tiers, 6962 passing, 0 failing, which is where
`AssertArchitecture` reads that same table; data green on sqlite, 554 passing, run beyond what the
stage claimed because it is nearly free and `core/testutil` compiles into it. Integration not
applicable: no handler, route or schema is touched. The claim the tidy could actually have broken is
the epic's measurable one, so it was verified rather than assumed — `go list -deps ./cmd/...` in
`src/adminconsole` returns 275 packages, and no driver, `go-sqlbuilder`, `pquerna/otp`, `x/image` or
`go-chi/cors` is among them. No mutation is owed: the stage adds no test and makes no behaviour
observable.

### Stage 4: #198 — the user session row outlives the request that failed

Landed in `6a9843b8`. 6 files changed, +358 / -129.

`StartNewUserSession` committed the session row and its client associations, then did four more
things that could each fail: read the user's other sessions, delete the ones from the same device,
read the browser session, and write the new identifier into it. Any of those returning an error
left a committed row that no browser held a cookie for, having possibly already deleted the session
the browser did have.

Decision 9's shape, applied. **Three of the four windows close outright**: the browser-session read
moves above the transaction (a memoised per-request read `/auth/completed` has already resolved, so
it consults no backend and writes nothing, and taking it first turns an unreadable cookie into a
refusal with an empty database), and the sibling read and the same-device sweep move inside it,
which they can with no interface change because `GetUserSessionsByUserId` and `DeleteUserSession`
already take a `*sql.Tx`. **The fourth cannot**: `RunInTransaction` reruns its body on a deadlock
and a rerun of `Regenerate` or `Save` would write `Set-Cookie` twice, so the browser-store write
stays after the commit and is compensated — `abandonUserSession` deletes the row and returns the
original error. It carries a `ceiling:` marker, because the sweep commits with the row and undoing
the row cannot bring back the sessions it superseded.

One departure from decision 9, which is silent on a cleanup that itself fails: the cleanup error is
`errs.Join`ed onto the original rather than dropped, so `errors.Is` still finds the reason the
ceremony failed and nothing anywhere loses the fact that the row survived.

**The seam needed one change before it could observe any of this.**
`run_in_transaction_stub_test.go` handed the body a nil `*sql.Tx` and every assertion matched with
`mock.Anything`, so a call inside the transaction was indistinguishable from one after the commit.
It now hands a non-nil `txSentinel` the mock never dereferences, and the expectations match on it,
so a sweep moved back outside fails on the pointer rather than passing on a call count. Both copies
of that file changed and stay byte-identical apart from the package clause.

Four new tests cover the post-commit arm, which needed a store implementing
`sessionstore.Regenerator`: the generated mock implements `Store` only, so **the rotation arm had
no coverage in this package at all** — every test took the `Save` arm. Existing cases in
`TestStartNewUserSession_ErrorsPropagate` gained the rollback and transaction-membership
assertions, and the store-read case now asserts an empty database.

Tiers: unit (three modules, 6966 passing), data on sqlite (554), integration on sqlite (1394, run
because seam 5 — a completed login through `GET /auth/authorize` including the same-device sweep —
is what says a client sees no change; it does not), `golangci-lint` clean on both touched packages.
Five mutations, one per behaviour the stage claims to have made observable, **all caught**.

### After the gate: the ambiguous commit, tried and reverted

Two commits after the final gate, both outside the four stages, and the net effect on the tree is
one comment.

`e66b4408` folded in the one follow-up the closing pass had written up. `RunInTransaction`
documents that a commit failing for a reason the engine did not declare may have landed anyway, and
`StartNewUserSession` treated that error as "nothing was written", so a commit that did land left a
session row no cookie names — #198's shape, through the helper's contract rather than through a step
of the ceremony. The commit tagged that failure with a new `ErrIndeterminateCommit` sentinel and had
the one caller reconcile: read by `SessionIdentifier`, delete the row if it is there.

`55548b97` reverts it. Two findings against it, both of which hold.

**The compensation could not be conclusive.** It treated a read that found no row as proof the
commit had rolled back. It is not. The connection can fail while the server is still processing the
`COMMIT`, and a read issued on another connection then sees nothing under MVCC and returns "did not
land" while the original commit goes on to finish. The comment asserting otherwise claimed a proof
it did not have, and **the tests could not have caught it**: they modelled only settled outcomes,
row present and row absent, so neither they nor the six mutations ever reached the in-flight case.
Making it sound needs bounded polling and a durable deferred cleanup.

**The defect was smaller than it looked, because the backstop already existed.** The background
worker sweeps idle sessions on its poll cycle (`background_worker.go`, `DeleteIdleSessions`) and
expired ones by max lifetime. An orphaned row was never permanent — it costs an entry in the admin
console's session list until the idle timeout, with nothing ever issued against it. The
reconciliation only made the common case tidy sooner, which does not buy a sentinel, a contract
change, and a compensation function whose signature — an error in, the same error out unless the
cleanup also failed — is hard to read at the call site.

**And the contract claim was too strong.** It asserted that `StartNewUserSession` is the only caller
that owes anything, on a survey of one failure mode: rows left unreachable. There is a second, and
it is worse. `SigningKeyRotator.Rotate` commits but reports the error, the admin API answers a
generic 500 with no audit event, and an operator who retries rotates twice — the second pass deleting
a previous key that is still validating live tokens. That predates `e66b4408` (the same 500 and the
same retry existed before the tag), it has no backstop, and it is **filed as #371** rather than
fixed here.

What stays is a ceiling comment where a reader would ask the question: the outcome can be unknown, a
read afterwards would not settle it, and the idle sweep is what bounds the window. Unit back to
**6966** passing, lint clean; no test is added, because no behaviour changed.


leodip:
## Deferred decisions

**None.** Nothing in this run was built on an assumption the user has not already signed off.

All nine decisions in the agreement were settled at interview time and none was reopened: the run
raised no new decision, deferred none, and escalated none to the pull request. The guard reports
`9 total, 0 open, 0 raised by the run` at the closing pass, and no stage recorded an assumption
taken in lieu of an answer.

The three judgement calls made while building are recorded in their stage entries in the body
rather than here, because each is a detail of executing a decided design rather than a question
left open:

- **Stage 1** — `communication.EmailSender` turned out not to be a moving qualifier (it is the
  interface `handlers` declares, not a type in the moved package), and the CI comment edit came to
  four comments rather than the two section 2 scoped, because the `services:` block carries its own
  reason between jobs.
- **Stage 4** — decision 9 is silent on a compensating delete that itself fails. The cleanup error
  is `errs.Join`ed onto the original rather than dropped, so `errors.Is` still finds the reason the
  ceremony failed and nothing loses the fact that the row survived. Reversing that costs one line.

leodip:

### Filed — 1

**#371 — an indeterminate commit in signing-key rotation invites a retry that deletes a key still
validating live tokens.** Not the follow-up this run first wrote up; a second, worse hazard that
surfaced while trying to fix the first, and one that predates this branch.

`SigningKeyRotator.Rotate` runs the rotation in one transaction. When its commit fails for a reason
the engine did not declare and the rotation in fact landed, `Rotate` reports the error, the admin
API answers a generic 500 and writes no `AuditRotatedKeys` event, and an operator who reads that as
"it did not work" rotates a second time — deleting the key that was `current` before the first pass,
now `previous`, and still validating live tokens. The handler already knows how to say "do not
retry" (`ErrRotationInProgress`, answered 409), and this case gets the one status that invites one.

Filed rather than fixed here because it predates this work, has no backstop, and fixing it properly
needs a decision about whether `RunInTransaction` should report an indeterminate commit as a
matchable condition — which is exactly the decision the reverted attempt made unilaterally.

### The follow-up that was folded in, then reverted

The one follow-up the closing pass wrote up — *an ambiguous transaction commit can still leave a
user session row nobody can reach* — was folded in as `e66b4408` and reverted in `55548b97`.
It is not filed, because after the revert it is recorded as a ceiling instead.

**The defect, as first described.** `RunInTransaction`'s contract says a commit that fails for a
reason other than a deadlock has an outcome the client cannot know. `StartNewUserSession` answered
that error with a bare `return nil, err`, so a commit that succeeded while only its acknowledgement
was lost left a session row and its committed same-device deletions behind while the ceremony
answered 500 — #198's shape, through a door its four windows do not cover.

**Why the write-up said to file it.** Two of the fold-in tests failed. *Nothing to settle*: the
compensating delete cannot reuse `abandonUserSession`'s handle, because `userSession.Id` comes from
`LastInsertId()` inside the transaction, so on an indeterminate commit it may name nothing or, on an
engine that reuses rowids, someone else's session. *The whole of the fix*: the exposure belongs to
`RunInTransaction`'s contract, shared by 15 production call sites, not to this caller.

**What the fold-in did, and why it did not survive.** It tagged the commit failure with a new
sentinel and had `StartNewUserSession` reconcile by `SessionIdentifier`. Both objections came back,
and a third arrived with them:

- **A miss is not proof.** The reconciliation treated a read that found no row as the commit
  reporting that it did not land. The connection can fail while the server is still processing the
  `COMMIT`; a read on another connection then sees nothing under MVCC and returns "did not land"
  while the commit goes on to finish. The tests modelled only settled outcomes — row present, row
  absent — so neither they nor the six mutations reached the in-flight case. Soundness needs bounded
  polling and durable deferred cleanup.
- **The backstop already existed.** The background worker sweeps idle sessions on its poll cycle
  (`background_worker.go`, `DeleteIdleSessions`) and expired ones by max lifetime. The orphan was
  never permanent. The reconciliation bought a tidier common case, not a correctness guarantee — and
  that does not pay for a sentinel, a contract change and a compensation function taking an error
  and returning the same error unless its own cleanup failed.
- **The contract claim was too strong.** It asserted `StartNewUserSession` is the only caller that
  owes anything, on a survey of one failure mode. #371 is the second, and the worse one.

**Duplicate search, from the original write-up**, kept because it is what established the defect was
neither tracked nor settled: `--search "commit acknowledgement"` and
`--search "indeterminate commit RunInTransaction"` both return empty; `--search "RunInTransaction"`
returns #359, #346 and #339, all Core-refactor moves; `--search "orphaned session row"` returns #198
itself, #246 and #109. #301 settles the *deadlock rerun* rule and deliberately documents the
non-deadlock commit as unknown, deciding nothing about what a caller owes.

### Folded in rather than filed — 4

Each rides on a surface this change already touches, and each is in the diff with its own commit.

- **#198 itself** (decision 5), `6a9843b8`. Its one file is a file this issue moves, so filing it
  would have reviewed the fix against a path about to change, or landed it weeks later against the
  moved file. Its own commit after the move, so `git diff -M` still reads the move as renames.
- **`src/adminconsole/go.mod` and `go.sum` tidied** (decision 6), `bda56f75`. 23 indirect
  requirements and 107 `go.sum` lines nothing in that binary's build graph needs — the four database
  drivers and `go-sqlbuilder` among them — left behind when #344 severed the edge, and reproducing
  on an untouched copy of `main`. One command, and `ARCHITECTURE.md` was being edited anyway to stop
  asserting the opposite.
- **`testutil.ClearMailpit` deleted** (decision 7), `bda56f75`. Declared and called from nowhere;
  invisible to `unused` because it is exported test support. Five lines, in a file this change's one
  mailpit test already put in the neighbourhood.
- **`core/oauth/mocks/token_parser_mock.go` moved** (decision 8), `4797cff8`. An authserver-only
  mock generated into `core` by `authserver/.mockery.yaml`, for an interface `handlers` declares,
  with all three consumers in the auth server — which is exactly the thing this epic removes, and
  this change was already editing the config that writes it. `core/oauth` itself stays kernel.

### Ceilings left by this change — 2

1. **The browser-store write.** `abandonUserSession` — the same-device sweep commits with the row,
   so undoing the row cannot bring back the sessions it superseded; a user whose cookie write fails
   is left with neither the session this ceremony created nor the one it replaced, and signs in
   again. **Revisit when** the session store gains a two-phase write that can be prepared before the
   commit and completed after it.
2. **The commit's own outcome.** `StartNewUserSession`, at the transaction's error — a commit that
   fails for a reason the engine did not declare may have landed, leaving a row no cookie names.
   Nothing at this layer closes it, and a compensating read does not settle it, because the commit
   may still be completing when the read runs. The background worker's idle sweep bounds the window;
   the cost is an entry in the admin console's session list until the idle timeout, with nothing
   ever issued against it. `55548b97` is the attempt at more than that, and why it was withdrawn.

leodip:
## Review account

`GBD_REVIEW_MODE=end`: no stage was reviewed on its own, so the final gate is the only reading the
code got. The plan gate is recorded first because its four findings changed what the stages built.

### Plan gate, round 1 — `openai-codex/gpt-5.6-sol`, effort xhigh, fresh session

Request `346-20260916T103522Z`, judged against the branch at its one empty commit `2992a5bd`, so
this round read the plan and the agreement rather than any code. **Four findings, all conformance,
all minor, none needing a human; all four confirmed against the file before being accepted; none
refuted, none deferred.** Every one was the same kind of thing — a sentence this change makes false
that section 2's documentation sweep did not reach.

| # | Site | Claim | Outcome |
|---|---|---|---|
| 1 | `.github/workflows/check.yml:473` | `Unit / race`'s comment explains its mailpit service as "the same reason as in `Unit / core`", the tier the service is leaving | Folded into stage 1 step 10 (`4797cff8`). The race job keeps the service |
| 2 | `ARCHITECTURE.md:209` | decision 6's tidy falsifies **both** clauses of the admin console `go.mod` sentence, not only the second one section 2 named, because the tidy removes the driver requirements outright | Stage 3 (`bda56f75`) rewrites the whole claim, shorter, adding no sentence |
| 3 | `src/core/user/usersession_manager.go:132` | "The same-device sweep and the cookie write below run only after the commit" is exactly what decision 9 inverts for the sweep | Stage 4 (`6a9843b8`) rewrites it; the stage's `Docs:` line stopped reading "none, internals only" |
| 4 | `src/authserver/run-tests.sh:490` | the `-count=1` rationale attributes `TestSendEmail` to core | Folded into stage 1 step 10 (`4797cff8`), owner repointed, cache argument kept |

The reviewer also independently compiled decision 8's `TokenParser` relocation in a container
scratch copy — the one part of the design no `probe/` program had covered — reproducing the compile
failure without stage 1 step 3's duplicate-import cleanup, then compiling `internal/handlers` and
`internal/protocolvalidation` with it.

One note for the next fanout reading: finding 4's file sits in no section 4 row, because the
`CI unit jobs` row's recount is scoped to `.github/workflows/check.yml`, which is how
`check-fanout.sh` reported `gaps=0` with that sentence standing. The row is narrower than the fact
it describes rather than wrong about its own file.

### Final gate, round 1 — `pi openai-codex/gpt-5.6-sol`, effort xhigh, fresh session

Request `346-20260916T122956Z`, completed 2026-09-16T12:43:13Z, round 1 of a 5-round budget.
Validated by `ingest-verdict.sh` before anything was read: request id matched and the request was
not already ingested. Scope was the whole branch against `main` — commits `4797cff8`, `50b6b7d1`,
`bda56f75`, `6a9843b8`; 82 files, +1179/-1059 with `-M`; working tree clean at `6a9843b8`.

**Verdict: clean.** `findings: []`, and all three axes read `reviewed` — conformance, quality,
security. That is clean rather than merely empty: the axes are covered and nothing in `unchecked`
bears on what the gate depended on. Nothing to verify, refute or resolve, so no production code
moved this round and no tier was re-run.

**What it ran**, recorded so the absence of findings has a stated scope: `review-diff.sh` against
`origin/main` with the 82-file diff captured and classified; `tree-hash.sh`, `check-anchors.sh`,
`check-fanout.sh` and `where.sh ci HEAD`; production and configuration diffs, `git diff --check`,
old-path fanout searches and a stage 1 word-diff fragment census;
`where.sh test --type internal --run TestStartNewUserSession`; `go mod tidy -diff` in all three
modules and `go list -deps` over the admin console's production closure; and **four independent
scratch mutations of its own, all caught** — removing the save-failure compensation, moving the
sibling read outside the transaction, moving sibling deletion outside the transaction, and ignoring
the browser-session read error. It confirmed the real checkout stayed byte-identical afterwards.

That last item is the part worth keeping. The one soft spot in this run's record was that stages 1
to 3 ran their mutations against `production` trees that have since moved, so that evidence was a
claim about a tree no longer on disk:

```
current   production e40be57b5800   unit-tests bf18a55f4d71   data-tests 10910c50d814
          integration-tests 6f32b236a8e2   docs 0f7b18a141f2   all 6622a3f56ef6

stage 1  unit stale   data ?       integration ?       docs stale   moved: production, unit-tests, docs
stage 2  unit stale   data ?       integration ?       docs stale   moved: production, unit-tests, docs
stage 3  unit stale   data ?       integration ?       docs live    moved: production, unit-tests
stage 4  unit stale   data stale   integration stale   docs live    moved: production
```

Every *tier* claim in those entries is re-established by the check suite, which is green across all
thirteen jobs at `6a9843b8` on four engines. The *mutation* evidence is what CI cannot re-establish,
and the reviewer narrowed that gap with independent evidence rather than assertion, by writing and
running its own mutations against the current tree at the #198 seam — the only place in this branch
where behaviour changes.

**`unchecked`, both entries answered:**

1. *"No driver-level fault injection simulated a transaction that committed but lost its Commit
   acknowledgement."* — **Tried in `e66b4408`, reverted in `55548b97`; now a ceiling, and a second
   hazard it surfaced is filed as #371.** Same defect as the reviewer's `## Follow-ups`
   entry. It was first written up as the one follow-up to file, in the comment above: verified
   against the code rather than accepted from the verdict, checked against every issue state, and
   shown to fail two of the fold-in tests. It was then folded in after the closing pass at the
   user's direction. It did not survive review: the reconciliation it added treated a read that
   found no row as proof the commit rolled back, which it is not — the commit may still be
   completing — and the orphan it removed was already bounded by the background worker's idle
   sweep. The reviewer was right that this needed filing rather than folding in; the comment above
   carries the full account, and the window is now recorded as a ceiling. First checked against the agreement to be
   sure it was not required work arriving disguised as out of scope: §2's goal names #198's four
   windows and decision 9 enumerates them, and the ambiguous-commit case is none of the four.
2. *"No interactive browser pass was performed."* — **Dismissed**, premise verified rather than
   trusted. The branch changes no user-facing surface at all:
   `git diff --name-only origin/main...HEAD` filtered for
   `web/|template|.css|.js|.html|catalog|locale|openapi` returns nothing, and the only non-Go,
   non-Markdown files in the diff are `check.yml`, `.golangci.yml`, the two `.mockery.yaml`, five
   module files and `run-tests.sh` — the last a one-word comment correction naming `emaildelivery`.
   `routes.go` is in the diff but adds and removes no route: its 9 insertions and 7 deletions are
   import paths and the five constructor qualifiers decision 1 requires.

**Bookkeeping:** the reviewer notes `check-fanout.sh` reports the five known post-move path hits
already recorded by the run, and that its own independent old-path and caller searches found no
omitted live consumer. Round 1 of 5 used; nothing open, so no round 2 was written.

leodip:
## After the gate: the follow-up was folded in, not filed

`e66b4408` — the one follow-up the closing pass wrote up is now fixed on this branch, so there is
nothing left to file. The [follow-ups comment](https://github.com/leodip/goiabada/pull/370#issuecomment-5697744474)
keeps the analysis that said to file it alongside what changed that answer, and the
[review account](https://github.com/leodip/goiabada/pull/370#issuecomment-5697752857)'s disposition
is corrected from *Filed* to *Fixed*. The PR description carries the full write-up.

**What it was.** `RunInTransaction` documents that a non-deadlock commit failure may have landed
anyway, and `StartNewUserSession` treated that error as "nothing was written". When such a commit
had in fact landed, the session row, its client association and the same-device sweep's deletions
were all committed while the browser never received the identifier — #198's shape, reached through
the helper's contract rather than through a step of the ceremony.

**What unblocked it.** The write-up rejected the fold-in on two counts, and splitting the problem
answered both. The *hazard* is the contract's, so it gets a name there: `ErrIndeterminateCommit`,
applied once, at the commit, and never to a declared deadlock, whose rollback is known. The
*obligation* is not shared by all 15 call sites — only a caller whose committed rows are reachable
*only through a step after the commit* owes a reconciliation, and `StartNewUserSession` is the one
whose row is long-lived. The three questions left open answer themselves from there: reconcile on
the tag alone, try once, and do not change the request's answer.

The unreliable `userSession.Id` the write-up flagged is exactly as unusable as it said, so the
reconciliation reads by `SessionIdentifier` — minted before the transaction opens. The test fixture
makes the insert report `99` while the row that exists is `4242`, so a compensation keyed on the
wrong name is caught rather than argued about.

**Checked:** unit 6971 passing (was 6966), data 554 and integration 1394 on sqlite, lint clean.
Six mutations against the new behaviour, all caught. The check suite is running on `e66b4408`.

leodip:
Check suite green on `e66b4408`, all thirteen jobs ([run](https://github.com/leodip/goiabada/actions/runs/35108915978)) — including the four-engine data and integration tiers, which is where this commit needed it: SQLite has one connection and cannot deadlock, so the tag's deadlock exclusion is only exercised against MySQL, PostgreSQL and SQL Server.

leodip:
## `e66b4408` reverted in `55548b97` — the fold-in did not hold up

Two findings against that commit, both correct, and a third thing it was wrong about. Reverted
rather than patched: once the compensation cannot be sound and the defect already has a backstop,
what is left does not pay for the machinery.

**A missing read does not prove the commit rolled back.** The reconciliation read by
`SessionIdentifier` and treated a miss as proof. The connection can fail while the server is still
processing the `COMMIT`; a read issued on another connection then sees nothing under MVCC and
returns "did not land" while the original commit goes on to finish — leaving exactly the orphan the
change meant to remove. **The tests could not have caught this**: they modelled only settled
outcomes, row present and row absent, so neither they nor the six mutations ever reached the
in-flight case. Soundness needs bounded polling plus durable deferred cleanup, which is more
machinery than the defect is worth.

**The defect was smaller than it looked, because the backstop already existed.** The background
worker sweeps idle sessions on its poll cycle (`background_worker.go`, `DeleteIdleSessions`) and
expired ones by max lifetime. The orphaned row was never permanent — it costs an entry in the admin
console's session list until the idle timeout, with nothing ever issued against it. The
reconciliation only made the common case tidy sooner. That does not buy a sentinel, a contract
change, and a compensation function taking an error and returning the same error unless its own
cleanup also failed — which is the signature that made the call site hard to read.

**And the contract claim was too strong.** It asserted `StartNewUserSession` is the only caller that
owes anything, on a survey of exactly one failure mode: rows left unreachable. There is a second,
and it is the worse of the two — `SigningKeyRotator.Rotate` commits but reports the error, the admin
API answers a generic 500 with no audit event, and an operator who retries rotates twice, the second
pass deleting a previous key still validating live tokens. That hazard **predates `e66b4408`**: the
same 500 and the same retry existed before the tag. It has no backstop, and it is filed as **#371**
rather than fixed here, because fixing it properly needs the same `RunInTransaction` decision the
reverted attempt made unilaterally.

**What stays** is a ceiling comment at the point a reader would ask the question: the outcome can be
unknown, a read afterwards would not settle it, and the idle sweep is what bounds the window. Unit
back to **6966** passing, lint clean. No test is added, because no behaviour changed.

The branch is back to `6a9843b8`'s surface plus that comment. The
[follow-ups record](https://github.com/leodip/goiabada/pull/370#issuecomment-5697744474) carries the
full account, including the original write-up's reasoning — which was right that this should have
been filed rather than folded in.

PR #374, opened 2026-09-17
feat(api)!: the session read surface publishes no presentation

### LINKED ISSUE #373: The admin session read surface: a view model published as the API contract, and a client method with no caller
Two connected defects on the admin console's session read surface, filed together because the fix
for one is the safety net for the other. **Part 1** is the wire type; **Part 2** is the client method
beside it. The connection is set out under *How the two connect*, and the recommendation is a
sequence rather than a choice between them.

This issue is investigate → decide → fix. The decisions are which option to take in each part, and
they need settling before the fix, because two of Part 1's three options change a published contract.

---

# Part 1 — `EnhancedUserSessionResponse` publishes a view model as the API contract

`EnhancedUserSessionResponse` is the admin console's 2023 Go template view model, published as the
public API contract. Of its seven "enhancement" fields, one carries information a client cannot get
otherwise; four are English, Go-formatted display strings duplicating timestamps already in the same
payload; one is a constant; and one is truthful on only one of the three endpoints that return it.

## Provenance

`git log -S` places it exactly.

| Commit | Date | What happened |
|---|---|---|
| `b50379c5` | 2023-11-05 | Monolith era, one server, server-rendered admin pages on direct database access. `handleAdminClientUserSessionsGet` declares a **function-local struct, `sessionInfo`**: `StartedAt` (`us.Started.Format(time.RFC1123)`), `DurationSinceStarted` (`time.Now().UTC().Sub(us.Started).Round(time.Second).String()`), `LastAcessedAt` (sic), `IsCurrent`. Pure presentation — timestamps pre-chewed in Go because Go templates cannot do date math |
| `31557ced` | 2025-09-06 | The split into authserver + adminconsole needs an admin API. The row becomes `UserSessionResponse`; the view model becomes that struct **plus those fields**, named `EnhancedUserSessionResponse` |
| `459b292b`, `9bdcb419` | 2025-09-08/09 | The client-sessions and account-sessions endpoints copy the same literal |
| `27bc794c` | 2026-09-12 | #281 adds `userAgent` to the base type |

The name describes an operation performed on another struct, not a thing in the domain.
`openapi.yaml:4208` says so literally: the schema is `allOf: [UserSessionResponse, {seven extra
properties}]`. There was no name for what the extras *are*, because they are not one idea.

## What the seven fields actually are

`src/core/api/responses.go:388-410`. All seven are `required: true` in the published schema.

| Field | What it is | Verdict |
|---|---|---|
| `startedAt` | English RFC1123 rendering of `started` — **which is in the same response** as a real timestamp | presentation |
| `durationSinceStarted` | `"72h3m0.5s"`, Go's `time.Duration.String()`. No other language parses this | presentation |
| `lastAccessedAt` | English RFC1123 rendering of `lastAccessed`, also already in the response | presentation |
| `durationSinceLastAccessed` | Go duration format again | presentation |
| `isValid` | **Always `true`.** All three producers compute it, `continue` if false, then assign it. No session with `isValid: false` has ever reached the wire | constant |
| `isCurrent` | Set only by `handler_api_account_sessions.go:89`. `handler_api_users_sessions.go` and `handler_api_client_sessions.go` never assign it, so it is always `false` there | endpoint-dependent |
| `clientIdentifiers` | A real join over `user_session_clients` → `clients` that a client cannot perform | **legitimate** |

## The four problems

**1. The contract publishes presentation.** `startedAt` and `durationSinceStarted` encode the admin
console's rendering decisions — a Go-specific duration serialization, an English month-and-weekday
date — into a contract every other consumer must accept. Both duplicate values already present as
proper RFC3339 timestamps (`started`, `lastAccessed`). They are computed with `time.Now().UTC()` at
response time, so they are stale on arrival and meaningless in anything cached or replayed.

**2. It breaks localization on a localized page.** The console renders these pages in pt-BR — there
are render tests for it — and `admin_users_sessions.html:101` does:

```
<td>{{.StartedAt}}<br />{{.DurationSinceStarted}}{{ T $.ctx "adminconsole.admin_users.sessions.duration_ago_suffix" }}</td>
```

A **translated** "ago" suffix glued to an **untranslated** Go duration literal, next to an English
RFC1123 date. The console holds `started` and `lastAccessed` as real timestamps and has the i18n
machinery to format them, and cannot, because it renders the strings the server chose.

**3. `isCurrent` means different things per endpoint, and the console already knows.**
`handler_account_user_sessions.go:50` trusts `es.IsCurrent`. Both admin handlers ignore it and
recompute locally — `handler_admin_user_sessions.go:80-82` and
`handler_admin_client_sessions.go:94` compare `es.SessionIdentifier` against the context's session
identifier. One `required` json key, truthful on one of three endpoints, with nothing in the spec or
the docs saying which. A third-party client generated from `openapi.yaml` would trust
`isCurrent: false` from an admin endpoint and lose the "you are about to end your own session"
affordance the console keeps only by working around the field.

**4. It round-trips.** The console decodes `EnhancedUserSessionResponse` and immediately converts it
back into `SessionInfo` (`adminuserhandlers/types.go:30-45`) — a console-side view model with the
same fields, **including the 2023 misspelling `LastAcessedAt`**, still in the struct and in both
templates. So: 2023 view model → lifted to the wire → decoded → converted back into a view model →
rendered.

That last one is why this surfaced during #350. #350's complaint is *"the admin console decodes a
response and then converts it back into the row type it came from."* This is the identical move with
a view model instead of a row — the same disease, one layer over, and not covered by #350's scope.

## Two documentation defects that follow

- `site/src/content/docs/integration/rest-api.mdx:524` — "Returns enhanced session details including
  device info, IP address, and **validity status**." The validity status is a constant, and invalid
  sessions are silently omitted rather than reported.
- The silent filtering is undocumented everywhere. An expired-but-not-yet-deleted session is
  invisible to the admin API. An operator auditing sessions cannot see it and has no way to ask for
  it, while `DELETE /api/v1/admin/user-sessions/{id}` can still remove one whose id they cannot
  discover through the API.

### Option 1A — Document the behaviour, change nothing

Write down that `isValid` is always true, that invalid sessions are filtered, that `isCurrent` is
only meaningful on the account endpoint, and that the four strings are English and Go-format.

*For:* zero API change, smallest diff.
*Against:* enshrines a probable accident as intent, leaves the i18n hole permanently, and leaves a
`required` constant and a per-endpoint-truthful boolean in a published contract. It is the option
that makes every later fix harder.

### Option 1B — Repair the semantics, keep the shape (non-breaking)

- Set `isCurrent` on both admin endpoints from the access token's `sid` claim. The mechanism already
  exists three lines away: `handler_api_account_sessions.go:61` reads
  `jwtToken.GetStringClaim("sid")`, and `token_issuer.go:718,856` put it there. The console then
  deletes its two local recomputations.
- Resolve `isValid`: either stop filtering and report it truthfully — richer, and arguably what an
  admin API should do — or keep filtering and drop the field, which is breaking.
- Fix the docs.

*For:* no consumer breaks; fixes the one field that can actually mislead a client.
*Against:* the four presentation strings stay in the contract, so problems 1, 2 and 4 remain.

### Option 1C — Take presentation out of the contract (breaking)

Everything in 1B, plus: remove `startedAt`, `durationSinceStarted`, `lastAccessedAt` and
`durationSinceLastAccessed`. The console formats `started` and `lastAccessed` — which it already
receives as RFC3339 — in the user's locale through the i18n machinery it already has, closing the
i18n hole as a side effect. Collapse the type: `UserSessionResponse` plus `isValid`, `isCurrent` and
`clientIdentifiers`, under a name that names a thing (`UserSessionDetailResponse`, or fold the three
fields into `UserSessionResponse` and retire the second type). Fix `LastAcessedAt` while the
templates are open.

*For:* the contract describes sessions instead of describing a Go template's needs; the i18n hole
closes; the mapping into `SessionInfo` becomes a genuine view-model mapping rather than a copy; and
the type name stops being an operation.
*Against:* removes four `required` fields from a published schema. Breaking for any third-party
client generated from `openapi.yaml`.

---

# Part 2 — `ApiClient.GetUserSession` is implemented, tested and called by nothing

`src/adminconsole/internal/apiclient/auth_server_client.go:31` declares
`GetUserSession(accessToken, sessionIdentifier string) (*api.UserSessionResponse, error)` and
`user_session_client.go:134` implements it against `GET /api/v1/admin/user-sessions/{identifier}`.
No production file in `src/adminconsole/` calls it. The only caller is `session_client_test.go`.

**It never had a caller.** `git log --all -S "GetUserSession(" -- src/adminconsole/internal/handlers`
returns zero commits. It was born dead in `9d4104a4` (2025-09-06), the batch that mirrored the auth
server's new session endpoints into the console client; only the list endpoints were ever wired to
pages.

**The endpoint stays either way.** It has consumers independent of the console: `openapi.yaml:895`,
`HandleAPIUserSessionGet` at `routes.go:211`, and `TestAPIUserSessionGet` at
`api_users_auth_test.go:341` driving 200, nonexistent and a random UUID. Nothing in Part 2 touches
the auth server.

**No page was meant to call it.** There is no single-session page. The three list pages use
`GetUserSessionsByUserId`, `GetClientSessionsByClientId` and `GetAccountSessions`, all returning
`[]api.EnhancedUserSessionResponse` — a strict superset of `UserSessionResponse`, every one of its
fifteen fields including `UserAgent`. `TestRender_SessionPagesTooltipTheRawUserAgent` shows the raw
header already rendering in the Device tooltip, so #281's surface was delivered through the lists
and there is no gap this method would fill.

**The cost of leaving it is small.** The stubs *embed* `apiclient.ApiClient` and override only what
they need, so an unused method costs zero lines per stub, and there is no generated mock at all —
`src/adminconsole` has no `.mockery.yaml`. The deletion is one interface line, ~32 implementation
lines and two tests.

## The catch, and it is the whole reason Part 2 is here rather than in a one-line cleanup

`TestAuthServerClient_GetUserSessionCopiesTheRawUserAgent` and
`TestAuthServerClient_GetUserSessionAcceptsALegacyEmptyUserAgent` are **the console's only
client-side decode tests for a session response**, and the first pins the `userAgent` json tag that
#281 decision 6 pinned deliberately. The three live methods appear in tests only as hand-written
stubs returning canned structs, so their real HTTP + `json.Unmarshal` path never executes. Everything
else in `session_client_test.go` is `SessionTokenSource`, unrelated.

Deleting the pair as a cleanup would silently drop the console's only proof that a tag renamed in
`core/api` gets caught.

### Option 2A — Delete the method, the implementation and both tests

*For:* smallest diff, removes an unreachable method.
*Against:* takes the console's only session-decode coverage with it, leaving all three live methods
with none.

### Option 2B — Delete the method and the implementation, **move** both tests to a live method

Re-point the two cases at `GetUserSessionsByUserId` (`/api/v1/admin/users/{id}/sessions`, decoding
`api.GetUserSessionsResponse`). All three live methods decode that same envelope, so one test there
pins the tags for the type they share.

*For:* same removal, and the coverage moves from an unreachable path to a reachable one. Net: −1
interface method, −32 lines, and a test that currently guards nothing starts guarding three live
call sites.
*Against:* slightly more than a delete.

### Option 2C — Keep it and build the page it was meant for

*Against:* no evidence such a page was ever intended, and the list pages already carry every field
it would fetch.

### Option 2D — Leave it alone

*Against:* keeps an unreachable method on a 106-method interface and leaves the decode coverage
pointed at it.

---

# How the two connect

**1. Same surface.** Both live in `user_session_client.go` and the session DTOs in `core/api`.
`GetUserSession` is the only session read that returns the base `UserSessionResponse`; the other
three return `EnhancedUserSessionResponse`.

**2. Part 1 forces Part 2's question.** Option 1C collapses or renames those two types. It cannot be
done without deciding what the single-session client method returns — so Part 1 cannot be finished
while Part 2 is open.

**3. The real link is safety, and it sets the order.** Option 1C removes four `required` fields from
the session wire types. The console's *only* decode test for a session response is attached to the
method nobody calls. Doing Part 1 first therefore means changing the session wire shape with **zero
client-side decode coverage on the three live paths** — exactly the mistake a decode test exists to
catch. Moving those tests first is the safety net.

`userAgent` survives Option 1C untouched — it is in the base type and legitimate (#281), not one of
the four presentation fields — so the moved test stays meaningful afterwards.

---

# Recommendation

**Do Option 2B first, then Part 1 Option 1C.**

2B is small, independent and safe on its own: it removes an unreachable method and puts the console's
session-decode coverage on a path that actually runs. That coverage is then in place before 1C
changes the shape those methods decode, which is the order that makes 1C safe rather than hopeful.

For Part 1 specifically, **1C, and preferably in the same release as #350 / PR #372.** The reasoning
is timing rather than size: PR #372 already ships breaking wire changes under its decision 1 — five
nested shapes whose keys go from `GroupIdentifier` to `groupIdentifier` and whose `NullTime` objects
become `null`. A client generated from the published contract must be regenerated for that release
regardless. Removing four fields in the *same* release costs those consumers one regeneration;
deferring it costs them two, on two different releases, for the same area of the API.

If a breaking change is not acceptable in this release, take **1B** as the floor — it is cheap, it is
the field that can actually mislead a caller into terminating the wrong session, and it does not
foreclose 1C. **2B is worth doing either way**, and does not depend on which Part 1 option wins.

**Option 1A is not recommended in any form.** `isValid` being a `required` constant and `isCurrent`
being true on one endpoint of three are defects, and documenting a defect makes it permanent.

# Security assessment

Checked deliberately across both parts, and **there is no vulnerability here.** Recorded so nobody
has to re-derive it:

- **`sessionIdentifier` on the wire is not a bearer credential.** Since #266 the browser carries a
  sealed, opaque handle (version byte, 24-byte nonce, 64 characters of identifier, 16-byte tag,
  AEAD) to a server-side session row; `user_sessions.session_identifier` is a value *inside* that
  row, not the cookie. Setting it as a cookie authenticates nobody. The admin endpoints that expose
  it are behind `authserver:manage`.
- **`userAgent` is safe at both ends.** `useragent.Raw` applies `strings.ToValidUTF8` before a
  rune-aligned 512-byte cut (`useragent.go:606-627`), and the templates escape it — pinned by
  `TestRender_SessionPagesEscapeTheDeviceLabelIntoTheModal`.
- The `security` label is deliberately **not** applied. Part 1's problem 3 is a correctness defect
  with a safety consequence for third-party clients, not a vulnerability.

Any fix must preserve all three of the above. In particular, do not "simplify" the `userAgent`
handling while removing the presentation fields, and do not drop the `userAgent` decode assertion
when moving it under Option 2B.

# What closing it takes

- `TestOpenAPI_SchemaPropertiesMatchTheAPIStructs` (`openapi_contract_lint_test.go:922`) holds the
  schema to the Go structs, so the DTO and `openapi.yaml` cannot drift apart — the change is forced
  to be complete.
- `TestOpenAPI_ResponseRequirednessMatchesOmitempty` covers the `required` list.
- The three producers: `handler_api_users_sessions.go`, `handler_api_client_sessions.go`,
  `handler_api_account_sessions.go`.
- The three consumers: `adminuserhandlers`, `adminclienthandlers`, `accounthandlers`, plus
  `SessionInfo` and the two templates.
- `site/src/content/docs/integration/rest-api.mdx:524` and `:1543`.
- Under 1B or 1C, the console's two local `isCurrent` recomputations are deleted, and a case should
  pin that the field arrives correct from all three endpoints rather than being rebuilt.
- Under 1C, a `rendertest` case should pin that the two timestamp cells render in the page's locale,
  which is the regression the current shape makes impossible to write.
- Under 2B, `auth_server_client.go:31`, `user_session_client.go:134` and the two cases in
  `session_client_test.go:345,391`.

# Related

- Both parts were found while assessing follow-ups from #350 (PR #372), and neither is covered by
  #350, whose subject is `models` in `core/api`. Part 1 is a *view model* in `core/api`, a different
  axis; Part 2 is a client method #350 converted rather than deleted, because removing an interface
  method is a scope question a refactoring stage should not settle on its own.
- The assessment behind Part 2 is written up at
  https://github.com/leodip/goiabada/pull/372#issuecomment-5713882640.

leodip:
Closes #373.

## What this is for

`GET /api/v1/admin/users/{id}/sessions`, the client equivalent and `GET /api/v1/account/sessions`
publish `EnhancedUserSessionResponse`: 22 flat fields, of which seven sit past the base type. Four of
those seven are presentation — `startedAt` and `lastAccessedAt` are `time.RFC1123` renderings of
timestamps already in the same payload, `durationSinceStarted` and `durationSinceLastAccessed` are
`72h3m0s` strings that are stale the moment they are serialized. A fifth, `isValid`, is a constant:
all three producers compute it, `continue` on false, then assign `true`. A sixth, `isCurrent`, is
assigned on one endpoint of the three and is silently `false` on the other two.

The admin console then renders those English strings under a translated suffix — pt-BR shows
`72h3m0s atrás` beside an English date — because it has no date or duration formatter of its own.

## The rule this establishes

**The session endpoints publish no presentation.** A field earns its place on the wire only if the
caller cannot work it out for itself. Two survive that test: `clientIdentifiers`, which is a join,
and `isCurrent`, which needs a claim from a token an API caller may not be able to read
(RFC 6749 1.4). The four rendered strings and the constant do not, and they go.

`UserSessionDetailResponse` embeds `UserSessionResponse` and adds the two. It is built by exactly one
function and is identical on all three endpoints, so no endpoint can quietly disagree with another
about a field again. The three duplicated forty-line producer loops — which are *why* `isCurrent` was
set in one place and not the others — become one shared helper, which also collapses a per-session
query into one batch read.

The console gets what it was missing: one localized formatter in `core/i18n`, two template
functions, and with it every date the admin console renders is localized rather than only the
session pages.

**Breaking**, in a release already carrying 13 breaking commits since v1.6.0: a schema rename, five
removed properties, and a new envelope on the client-sessions endpoint.

### Stage 1: the console's session decode table
Commit `588a6992`. Part 2 of the issue, decision 5's widened form.

`ApiClient.GetUserSession` and its implementation are deleted: no production caller has ever existed, the console has no generated mock, and every handler stub embeds the interface, so nothing else moved. The auth server endpoint it fetched, `GET /api/v1/admin/user-sessions/{sessionIdentifier}`, is untouched and keeps its own consumers and tests.

The two single-session decode tests that method existed for become one table over the three live methods — `GetUserSessionsByUserId`, `GetClientSessionsByClientId`, `GetAccountSessions` — each row asserting the request target the method builds, the `Authorization` header and every field the console binds, against a body written as literal bytes rather than marshalled from the `api` structs. None of the three had ever had its real request path executed before this: the console's handler tests reach them only through hand-written stubs returning canned structs. A sub-table covers the hand-rolled `?page=&size=` assembly, which had no test at all, over page-and-size, page only, size only and neither.

This is first on purpose: the console's proof that a renamed json tag is caught has to exist before stage 4 renames anything.

**Tiers.** Unit green, 7133 passing across all four modules. Data and integration not applicable — no data-layer method, handler, route or schema moves. Three mutations, all caught: a renamed `clientIdentifiers` tag in `core/api`, the query separator made unreachable, and the account sessions path.

**Worth knowing.** Section 0 of the agreement was re-swept for the code this stage deleted: the two test anchors are repointed at their replacements and the two dead-code anchors are removed, with a note recording it.

### Stage 2: the localized date and duration formatter
Commit `e62911de`. Decision 3, and the machinery the issue assumed already existed.

The issue says the console "has the i18n machinery to format them". It does not, and Go does not supply it: `time.Format` has no locale, its month and weekday names come from a package-level English table, so RFC1123 is English under every layout Go offers and only a numeric layout carries no English — while the numeric field order is itself what differs per locale. `golang.org/x/text`'s `date` package exports a CLDR version constant and nothing else.

So `core/i18n` gains two functions. `FormatDateTime` reads the layout from the catalog, which is what makes the *format* the translatable thing: `01/02/2006 3:04 PM` for en, `02/01/2006 15:04` for pt-BR. `FormatSince` picks the largest whole unit among day, hour, minute and second and reads a per-unit, per-plural key holding the whole phrase rather than a number with a suffix after it, so word order belongs to the translator — `há 3 dias` against `3 days ago`, where the old pages could only ever produce `72h3m0s atrás`. Nil and zero instants render empty; under a second, which is also where an instant in the future from clock skew lands, renders `just now`. A layout key missing from every catalog falls back to `2006-01-02 15:04`, because `time.Format` would otherwise take the key itself as a layout and render it verbatim into a table cell.

Two limits ship marked as `ceiling:` in the source, each with the event that would make it worth revisiting: day is the largest unit, and the plural form is chosen by `count == 1`, which is the rule for both locales this tree ships but not for a locale with a few/many form.

Nothing consumes it yet. The `DateTime` and `Since` `templateFuncMap` entries exist and no template calls them; stage 3 binds the three session pages and retires the three `duration_ago_suffix` keys they still read.

**Tiers.** Unit green, 7162 passing across all four modules, including 40 new cases: every unit boundary, both plural forms, both locales, nil, zero, sub-second and future instants, and the missing-layout fallback. Data green on sqlite, 554 passing — not owed by the stage, run because the change is in `core`. Integration not applicable: no handler, route or schema. Five mutations, all caught.

**Worth knowing.** One case walks all twelve months in both locales and asserts the output carries none of the twelve English month abbreviations or seven weekday abbreviations. That is decision 3's pin: a layout regressing to `time.RFC1123` has to fail rather than merely look different to a reader who does not speak pt-BR, and mutating the pt-BR layout to RFC1123 confirms it does.

**Deviation.** The three `duration_ago_suffix` keys were owed retirement in this stage and stay one stage longer: the three session templates still read them, and deleting a key renders the key itself into the cell. They go with the templates in stage 3.

### Stage 3: the console session pages render localized dates
Commit `e59d7c82`.

The three `SessionInfo` view models stop carrying `StartedAt`, `DurationSinceStarted`,
`LastAcessedAt` and `DurationSinceLastAccessed` and carry `Started` and `LastAccessed` as
`*time.Time` instead — which deletes the 2023 `LastAcessedAt` misspelling in all three copies by
construction rather than by choosing to rename it. The three handlers copy the instants the wire
already carries, and the three templates render each cell as
`{{ DateTime $.ctx .Started }}<br />{{ Since $.ctx .Started }}`. pt-BR now reads `14/09/2026 21:03`
over `há 3 dias` where it read `Mon, 14 Sep 2026 21:03:00 UTC` over `72h3m0s atrás`.

The three `duration_ago_suffix` keys retire from both catalogs here, which is the deviation stage 2
recorded: they could not go until something stopped reading them.

**Tiers.** Unit green, 7170 passing, 8 new cases. Data green on sqlite, 554 passing. Integration not
applicable: the tier drives the auth server, and the console is not reachable from it — which is why
`rendertest` exists at all.

**Worth knowing.** The three render cases assert both cells byte for byte in pt-BR, with the layout
literal written out in the test as the thing the catalog is held against. Six mutations, all caught,
three of them one per hand-built `SessionInfo` literal — the shape where a copy lands at one of three
sites and the other two ship broken.

The four pre-rendered strings stay on the wire until stage 4, and the two local `isCurrent`
recomputations with them.

### Stage 4: the collapsed type, the shared loop, and `isCurrent` on all three endpoints
`429d36be`. The breaking one, and the stage the fanout called inseparable: `core/api` changes and both modules compile against it.

`api.EnhancedUserSessionResponse` is deleted. `api.UserSessionDetailResponse` takes its place — `UserSessionResponse` embedded, plus `isCurrent` and `clientIdentifiers` — and `startedAt`, `durationSinceStarted`, `lastAccessedAt`, `durationSinceLastAccessed` and `isValid` leave the Go type, the OpenAPI `allOf`'s second member and its `required` list. The first four were an English RFC1123 date and a Go duration string computed at the server from instants already in the same payload; the three console pages now format those instants themselves, in the viewer's language, through the formatter stage 2 landed. The fifth was a constant: every producer computed it, skipped every session for which it was false, then assigned `true`. The endpoints still list only active sessions and the docs now say that instead of calling it a validity status.

`isCurrent` means the same thing on all three endpoints for the first time. It was assigned at the account endpoint alone, so both admin endpoints published a constant `false` and nothing failed. `apimapping.ToUserSessionDetailResponse` decides it once, from the `sid` claim of the caller's access token, which the two admin handlers now read through `middleware.GetValidatedToken` — no routing or scope change, the middleware already put it on the context. A token with no `sid`, client credentials or an offline grant, correctly gets `false` throughout. The console's two local recomputations are deleted and all three pages read the field.

A package-private `buildSessionDetails` replaces three duplicated forty-line loops. That duplication is what let one producer set `isCurrent` while the other two did not, so sharing it is the structural half of the fix. It also hoists the per-session `UserSessionClientsLoadClients` into one `GetClientsByIds` over the union of client ids, turning a fifty-session page from fifty queries into one; the method itself stays, since the two logout handlers call it on a single session.

**Also deleted, and named because the goal does not mention it:** the admin console's `SessionIdentifierToContext` middleware, its test and its three chain entries. The two recomputations above were its last readers, so it was left lifting a claim onto a context value nothing reads on every authenticated request. The auth server's middleware of the same shape is untouched.

**Tiers.** Unit green, 7193 passing. Integration green on sqlite, 1419 passing. Seven mutations, all caught. New coverage: 7 cases on the shared loop including the query-count claim, 4 on the three handlers' `sid` read, three endpoint tests asserting the five keys absent from the **raw body** (decoding cannot see a key the struct no longer declares) and `isCurrent` true on all three, the mapper's own table, three wire-bytes rows and a bind case per console page.

### Stage 5: the client-sessions users envelope
`6463221b`. The second breaking one, and the smaller of the two: one endpoint's response type changes.

`GET /api/v1/admin/clients/{id}/sessions` returns `api.GetClientSessionsResponse` — the same `sessions` array as before, plus `users`. It is the only session list that spans users, and the admin console had no way to name a session's owner except to read each one back: one `GetUserById` per row, up to 50 HTTP round trips to fill the email and name columns of a single page. The owners ride along now, normalized so a person holding several sessions appears once, and the console resolves each row's `userId` against a map. `UserSessionDetailResponse` itself is byte-for-byte identical on all three endpoints, which is the property the issue's problem 3 rests on.

**The element is deliberately not `api.UserResponse`.** This route is guarded by the clients scopes — `admin-read`, `manage-clients` or `manage` — while every users route needs the users scopes, so a token holding `manage-clients` alone reaches this list and reaches no users route. Full profiles here would hand it the subject, birth date, phone number, postal address, locale and `otpEnabled` of everyone holding a live session on a client it manages, through a door this change would be opening for the first time. `api.SessionOwnerResponse` is the five fields a session page shows, and the wire-bytes case asserts that exact key set so a later widening fails rather than merely returning more.

Both arrays are required and non-nullable, so an empty page emits `[]` and never `null`; the endpoint refuses a session whose `userId` names no row rather than answering 200 with the owner silently missing, which is the choice the client join beside it already makes. `handlers.SessionOwnerFullName` joins the owner's name parts, and it and `UserFullName` now share one helper rather than becoming a fourth copy of that join.

**Tiers.** Unit green, 7220 passing. Integration green on sqlite, 1420 passing. Four mutations, all caught. New coverage: a six-case table on `sessionOwners` (normalization with the ids asked for asserted, first-seen order, the empty page asking nothing, an unresolvable id refused), three console bind cases including one asserting zero `GetUserById` calls, both name joins driven through one table, two client decode cases, three wire-bytes rows, and two endpoint cases — one on the raw body for `"sessions":[]` and `"users":[]`, which a decode cannot see, and one asserting eight profile keys absent from the response.

### Stage 6: the other seven dates and the audit log instant
`9390138b`. The last one, and the one that takes the goal past the session pages: every date the admin console renders is localized now.

Seven sites, three of them formatted in a handler with `time.RFC1123` or `02 Jan 2006 15:04:05 MST` and four inside a template with the same layouts — the two consent pages, the signing keys page, the admin user details and delete pages, and the audit log viewer, which did not format its date at all and printed the raw ISO 8601 string. All seven reach their template as the instant and go through `DateTime`, which reads its layout from the viewer's catalog. `time.RFC1123` and `02 Jan 2006 15:04:05 MST` are gone from the admin console entirely: the one `.Format` call left in its production code is an HTTP `Expires` header.

**`api.AuditLogResponse.CreatedAt` is a `time.Time`**, not a string the handler built by calling `Format("2006-01-02T15:04:05Z07:00")` on the model's instant. That is what gives the viewer something to format rather than a string to print. A value and not a pointer, because `audit_logs.created_at` is `NOT NULL` on every engine and the schema declares `createdAt` required and not nullable, so a pointer would publish a null that cannot occur. `encoding/json` marshals a `time.Time` as RFC3339, which is what the hand-rolled layout spelled out, so the wire is unchanged but for the sub-second precision that layout truncated — and `type: string, format: date-time` in the spec covers both, so no schema edit was owed. Breaking for a generated client that models `createdAt` as a bare string.

**Two things stay machine-format, deliberately, and are now pinned as staying.** The `dateOfBirth` inputs on the two profile pages are form values the server parses back with `2006-01-02`, so localizing them would break the round trip silently — `02/01/2026` parses as neither a date nor an error a user would understand. And the audit row's `<time datetime="...">` attribute is the machine value the HTML element defines; the text beside it is the one a reader sees, and only that one is localized. Both look like an oversight to anyone sweeping this console for unlocalized dates, which is exactly why there is now a case asserting each.

The signing keys page loses its `MST` suffix with its old layout. The catalog layout carries no zone, the values are UTC as they have always been, and no other page in this console has ever shown one.

**Tiers.** Unit green, 3684 passing. Data and integration green on sqlite. Five mutations, all caught: the formatter dropped from two of the templates, the `dateOfBirth` layout localized, `instantOf`'s value arm returning nil, and the `createdAt` json tag renamed. New coverage: a render case per page in pt-BR asserting the localized cell and the absence of the retired layout, the blank cell a nil instant renders, the `dateOfBirth` exclusion on both profile pages, three wire-bytes rows on the audit family including one on sub-second precision, and the funcmap adapter's four answers.

### Folded in after the review: the three session DELETE handlers read `isCurrent`
`6f254543`. Not one of the six planned stages. It was draft 1 of the follow-ups comment, and folding it in is what that comment now records instead.

Stage 4 made `isCurrent` true on all three list endpoints, computed server-side in one mapper from the `sid` of the very token the console forwards, and deleted the two GET-side recomputations together with the middleware feeding them. The three session pages also *delete* a row, through an AJAX POST, and each of those handlers was the same comparison surviving on the POST side: lift `sid` off the console's own access token, re-fetch the page of sessions, compare `sessionIdentifier` per row, and decide whether deleting this row must send the browser through the logout flow rather than simply remove it. Each is now `es.Id == int64(userSessionId) && es.IsCurrent`, and the claim has no reader left anywhere in the admin console.

**The list read is no longer swallowed.** `if err == nil` was harmless while the read only fed a comparison against a claim the handler already held; it decides the answer now, and deleting without it would destroy the caller's own session while telling the browser it had deleted somebody else's — no logout, and a console holding a session the auth server has already forgotten. All three forward it to `HandleAPIErrorJson`, the classifier every other `apiClient` failure in those files reaches (pattern 7).

**None of the three POST handlers had a test**, which is why this was drafted rather than folded in the first place. It has them now: a five-row table per handler and a failed-read case beside it. Two rows per table exist to pin that the *field* is what is read — a row whose `sessionIdentifier` matches the console's own claim is not current unless the field says so, and a row the field marks current is current even when the claim names another — and a handler that went back to the claim answers both backwards.

`handlertest` gains `ExpectEncodeJson` and `Encoded`, the AJAX halves of `ExpectRender` and `Bind`. `Encoded` marshals the answer and reads it back as a map, because these handlers answer anonymous structs with no tags: the names the console's own scripts read are the Go names, and absence is part of the answer, since a handler answering `Success` alone leaves `IsCurrentSession` out of the object rather than sending it false.

**Tiers.** Unit green, 7259 passing. Lint green. Nine mutations, all caught: per handler, every row read as current, the identifier read in place of the field, and the failed-list guard made unreachable. Data and integration not applicable: no handler outside the admin console, no route, no schema.

## What the review changed

Three rounds over the whole branch, four findings, all four confirmed and fixed. The full account,
with the evidence, is in [this comment](https://github.com/leodip/goiabada/pull/374#issuecomment-5719505905).

### Round 1: two findings, both confirmed
Commit `cbedb454`. Reviewed by `openai-codex/gpt-5.6-sol` at `xhigh` effort, over the whole branch against `main`.

**An unbounded client-id union answered a valid request with HTTP 500.** Stage 4 replaced the per-session client load with one `GetClientsByIds` over the union of client ids across the page, which is what removed up to 50 queries per render. Every id in an `IN` list is a bound parameter, though, and SQL Server refuses a statement carrying more than 2,100 of them with error 8003 — the lowest ceiling of the four engines, against 32,766 for modernc.org/sqlite, 65,535 for PostgreSQL, and a statement-size bound rather than a parameter count for MySQL. The union is bounded by the deployment's client count and by nothing else, and the users and account endpoints do not paginate their session lists at all, so a large enough deployment got a 500 on a request that was entirely valid. The reviewer reproduced it by calling the real data method with 2,101 ids against a live SQL Server.

`GetClientsByIds` now reads a long id list in several statements and merges the rows, under `maxIdsPerStatement = 1000`. The bound lives in `commondb` rather than in the auth server helper because the ceiling is an engine property and `commondb` is the only layer holding the flavor. Its signature does not change, so no mock and no schema moves. What the choice costs is recorded as a `ceiling:` beside it: one budget for all four engines means a list of 1,001 to 2,100 ids costs two statements on the three engines that would have taken it in one, and a list read in several statements is no longer one snapshot. A per-flavor budget is the next shape if a deployment is ever measured paying for the extra round trip.

Only that one lookup is bounded. Twelve more `IN` expansions in the same package still send whatever they are given — including `GetGroupsByIds`, which takes its ids straight from a request body, so a `manage-users` caller can post 2,101 group ids and get a 500 instead of a validation error. That is thirteen call sites, a budget that wants one home, and a decision this change deliberately did not take, so it is drafted as a follow-up rather than typed here.

**The REST API guide promised `isCurrent` on an endpoint that does not return it.** The sentence read "Every session endpoint marks the session the calling token was issued through with `isCurrent`". `GET /api/v1/admin/user-sessions/{sessionIdentifier}` returns the plain `UserSessionResponse` by design, which has no such property, so the sentence now names the three list endpoints it is true of.

**Tiers.** Unit green, 7233 passing. Data green on **all four engines**, 2312 passing, because the round touched `src/core/data/` and because SQL Server is the only engine on which the defect is observable. Integration green on sqlite, 1420 passing. Lint green. New coverage: a data tier case asking for 2,101 ids, one past SQL Server's ceiling, and for 2,000, an exact multiple of the budget where an off-by-one drops the final batch instead of failing loudly, with real clients in different batches including the last slot. Two mutations, both caught: raising the budget to 100,000 reproduces error 8003 on SQL Server, and dropping the final partial batch loses a client on sqlite.

The reviewer's own four mutations against the branch — forcing `isCurrent` false, retaining invalid sessions, adding `subject` to the owner wire shape, and replacing the pt-BR date layout with the English one — were all caught by the tests the stages wrote.

### Round 2: round 1's own fix had turned a set read into a multiset read
Commit `fff211af`. One finding, minor, confirmed.

`GetClientsByIds` batched the id list without deduplicating it first, so an id appearing on both sides of a batch boundary came back as two copies of the same row. A single `IN` list collapses a repeated id to one, so this was the method's answer changing under round 1's fix, and only for a list long enough to be split — nothing before the bound could have shown it. The reviewer demonstrated it with a scratch sqlite assertion putting one real id at positions 0 and 1000 of a 1,001-id input and getting two clients back.

Fixed by deduplicating the ids before they are batched rather than deduplicating the merged rows: that is the set the single statement used to return, and a caller passing duplicates now spends fewer statements rather than more. **No endpoint was answering wrongly** — `loadSessionClients` deduplicates its union before the call and both `commondb` loaders map the result by id — so this is the `Database` interface's documented behaviour restored, not a response repaired.

**Tiers.** Data green on sqlite, 558 passing; unit green on all four modules, 7233 passing. Integration not re-run: the change is inside one data-layer method and no handler, route or schema moved. The mutation replacing the seen check with `if false` fails the new case.

### Round 3: round 2's regression test did not test round 2's fix
Commit `ee002b48`. One finding, significant, confirmed. **No production code moved.**

`TestGetClientsByIds_ARepeatedIdOnEitherSideOfABatchBoundary` put both copies of the repeated id at positions 999 and 1000 — adjacent, either side of the boundary — which does not establish the guarantee the fix implements. A read that drops only a repeat of the id it just kept answers that input with one row as well, because collapsing the two adjacent copies leaves 1,000 ids and a single statement. The case was green against deduplication that is not deduplication of the list, which is the one thing it exists to refuse. The reviewer demonstrated it rather than asserting it: adjacent-only deduplication in a scratch tree left both checked-in batching tests green, and a scratch assertion with the same id at positions 0 and 1000 returned two clients.

Fixed by making the case a table of two placements rather than replacing the one it had. Both are load-bearing and the comment says which does what: the near pair, 999 and 1000, is the tightest straddle of the boundary and catches a batch bound off by one slot; the far pair, 0 and 1000, has nothing adjacent to collapse, so only a read that remembers every id it has already seen answers one row. Mutating the seen check to compare against the previous id alone is **caught** by the far-pair subtest and passed by the near-pair one, which is the claim reproduced in the checked-in suite.

**Tiers.** Data green on sqlite, 560 passing; unit green on all four modules, 7233 passing, which is where the gofmt guard reads the edited file. Integration not re-run: one test file moved and no production line.

### Review

Reviewed by `openai-codex/gpt-5.6-sol` at effort `xhigh`, 3 rounds over the whole branch against
`main`. **This run reviews once at the end, so no stage was reviewed on its own** — these three
rounds are the whole of the code review.

| Round | Findings | Significant | Outcome |
|---|---|---|---|
| 1 | 2 | 1 | both fixed, `cbedb454` |
| 2 | 1 | 0 | fixed, `fff211af` |
| 3 | 1 | 1 | fixed, `ee002b48` (test only) |

- **significant, quality (round 1):** a valid, authorized request to any of the three session list
  endpoints answered HTTP 500 once the union of client ids on the page exceeded 2,100 bound
  parameters, which is SQL Server's statement ceiling. Fixed in `cbedb454`: `GetClientsByIds` now
  reads a long id list in several statements under `maxIdsPerStatement = 1000`, with the
  single-budget tradeoff recorded as a `ceiling:` in the code and the other twelve unbounded lookups
  in `commondb` drafted as a follow-up. Its `forced_answer` was empty, which ordinarily escalates;
  it was fixed here instead under the crash carve-out, because a verified defect that loses a
  request on a surface this change already touches has no option worth weighing.
- **significant, quality (round 3):** round 2's regression test placed the repeated id adjacently,
  so it passed against deduplication that is not deduplication of the list. Fixed in `ee002b48` by
  making the case a table of two placements, near and far; no production code changed.
- **minor, quality (round 2):** round 1's batching had turned `GetClientsByIds` from a set read into
  one returning an id twice when it straddled a batch boundary. Fixed in `fff211af` by
  deduplicating before batching. No endpoint was answering wrongly.
- **minor, conformance (round 1):** the REST API guide claimed `isCurrent` on all four session read
  routes, one of which returns the base type by design. Fixed in `cbedb454`, one sentence.

Nothing was refuted, nothing was deferred, and round 3's `unchecked` was empty. The gate closed at
three rounds of a budget of five: rounds 1 and 2 each changed production code, round 3 changed none.

Full account, with the evidence, mutation commands and per-round tier numbers, in
[this comment](https://github.com/leodip/goiabada/pull/374#issuecomment-5719505905).

## Tiers

Green on the final commit, `ee002b48`, on **all thirteen check suite jobs**: `Tests / sqlite`,
`Tests / mysql`, `Tests / postgres`, `Tests / mssql`, `Unit / core`, `Unit / authserver`,
`Unit / adminconsole`, `Unit / setup`, `Unit / race`, `Lint`, `Vulnerabilities`, `Versions` and
`Build validation`. That is one run of every tier against the finished tree, which is the claim of
record.

Locally, per stage and per review round: unit on all four modules every time, data on sqlite for
every stage touching `core` and **on all four engines** for the round that changed
`src/core/data/`, integration on sqlite for the stages that moved a handler or a route, and lint at
the gate. The four-engine data and integration results are the check suite's, as the process
intends — those tiers no longer run locally for all engines. 32 mutations across the six stages and
the three review rounds, every one caught, listed per stage above.

**Deviation, in force for every local tier.** As `where.sh` reports each time it runs, the dev
container was not created by the devcontainer CLI, so it is missing all 43 `containerEnv` values and
the script supplies them on the exec instead. The local tiers therefore ran against the declared
configuration, but through a workaround rather than at the source. The check suite is unaffected.

**Deviation, stage 2.** The three `duration_ago_suffix` catalog keys were owed retirement in stage 2
and stayed one stage longer, because the three session templates still read them and deleting a key
renders the key itself into the cell. They went with the templates in stage 3.

## For the release notes

This release changes the admin and account session API in ways a generated or hand-written client
will notice, and one of them silently.

**Five properties are gone** from the session list responses of
`GET /api/v1/admin/users/{id}/sessions`, `GET /api/v1/admin/clients/{id}/sessions` and
`GET /api/v1/account/sessions`: `startedAt`, `lastAccessedAt`, `durationSinceStarted`,
`durationSinceLastAccessed` and `isValid`. Read `started` and `lastAccessed`, the RFC3339 instants
already in the same payload, and format or subtract them yourself. `isValid` was always `true` —
these endpoints list only active sessions, and they still do — so code branching on it can be
deleted.

**The schema is renamed.** `EnhancedUserSessionResponse` becomes `UserSessionDetailResponse`.
Regenerate your client.

**`GET /api/v1/admin/clients/{id}/sessions` returns a different envelope.** It was
`{"sessions": [...]}` and is now `{"sessions": [...], "users": [...]}`, where `users` carries the
owner of each listed session — `id`, `email`, `givenName`, `middleName`, `familyName` — normalized,
so a person holding several sessions appears once. Resolve a session's `userId` against that array
instead of calling `GET /api/v1/admin/users/{id}` per row. The other two session endpoints keep the
bare `{"sessions": [...]}` shape.

**`isCurrent` now means what it says on all three endpoints.** It was set on the account endpoint
alone; both admin endpoints published a constant `false`. If you worked around that by reading `sid`
out of your own access token and comparing it yourself, you can stop.

**`createdAt` on `GET /api/v1/admin/audit-logs` is unchanged on the wire but for precision.** It was
built by formatting with `2006-01-02T15:04:05Z07:00`, which truncated sub-second digits, and is now
marshalled from a `time.Time`, which does not. Still RFC3339, still `type: string,
format: date-time`. A client parsing it as RFC3339 needs no change; a client comparing the string
literally, or assuming a fixed length, does.

**Nothing to do on the server.** No configuration setting, database migration, schema change or
session-validity rule moves, and the background session sweep is untouched. One cosmetic change in
the admin console: dates are now formatted in the viewer's language — pt-BR reads `14/09/2026 21:03`
over `há 3 dias` where it read `Mon, 14 Sep 2026 21:03:00 UTC` over `72h3m0s atrás` — and the
signing keys page no longer prints an `MST` suffix. Values are UTC, as they have always been.

## The record

- [Deferred decisions](https://github.com/leodip/goiabada/pull/374#issuecomment-5719470603) — none;
  the one escalation, decision 12, and how your answer landed.
- [Follow-ups, fold-ins and ceilings](https://github.com/leodip/goiabada/pull/374#issuecomment-5719491763)
  — one issue drafted with the `gh issue create` command that files it, three things folded in
  rather than filed, and the three `ceiling:` markers this branch adds with their revisit triggers.
  Updated after `6f254543`: what was draft 1 moved from the first section to the second.
- [The review account](https://github.com/leodip/goiabada/pull/374#issuecomment-5719505905) — three
  rounds, four findings, the evidence for each fix and the mutations that prove the new tests fail
  without it.

## Status

**Complete.** All six stages landed, the final review closed at three rounds with every finding
confirmed and fixed, and the check suite is green on all thirteen jobs at `ee002b48`. Ready for
review.

One commit landed after that, `6f254543`: draft 1 of the follow-ups comment, folded in rather than
filed, described above. The check suite is green on all thirteen jobs at that commit too.


leodip:
## Decision 12: should the client Sessions API hand a "Manage clients" integration the profiles of the people signed in?

Goiabada's Admin API is reached with granular scopes, so an integration can be given one domain
rather than the whole console. Two of them matter here: **Manage users**, documented as covering
"Users, groups, permissions, sessions, consents", and **Manage clients**, documented as "OAuth2
clients and their configurations". The docs say plainly that these exist "to follow the principle of
least privilege".

You meet this in the console under **Clients → a client → Sessions**, the page listing who currently
has a live session with that app, and behind it `GET /api/v1/admin/clients/{id}/sessions`. That
route is gated on the clients scopes; the equivalent user pages are gated on the users scopes.

Today that endpoint answers one row per active session, and the only thing a row says about the
person is a numeric user id. The console then fetches each of those people separately, one HTTP call
per session, using its own full-access token. Removing those extra calls is one of the things this
change is for, and the way it does it is to have the endpoint return the session owners alongside
the sessions, so the console reads them from one response.

That is the part I want you to settle, because it widens what the endpoint tells a caller who was
only ever granted the clients domain. Nobody is stuck today and nothing is broken; this is a new
door, and it is cheaper to decide where it opens now than after it ships.

### A worked example

Say you have a provisioning integration that manages your OAuth apps, granted **Manage clients**
only, and you look at a client with 50 live sessions belonging to 30 different people.

- Today, and under **D**, it learns 30 numeric user ids and nothing else about those people.
- Under **A** it also receives, for each of the 30: email, given/middle/family name, nickname,
  gender, date of birth, phone number, the full postal address, locale, and whether they have
  two-factor enabled.
- Under **B** it receives their names and email addresses, which is exactly what the Sessions page
  displays.
- Under **C** the call fails with 403 unless you also grant that integration **Manage users**.

### The options

The axis is what a clients-only integration learns about the people using a client, against what it
costs to build.

| | What a clients-only caller learns | Cost |
|---|---|---|
| **A.** Return full user records | Everything on the profile: email, names, birth date, phone, address, locale, two-factor status | Nothing extra; this is what was already planned |
| **B.** Return five fields: id, email, given/middle/family name | Name and email, the two things the page shows | One small response type, one mapper, one helper |
| **C.** Full records, but require both Manage clients and Manage users | Nothing new, unless you also grant it the users domain | Small, but it takes access away from an existing clients-only caller of a route that ships today |
| **D.** Don't return owners at all | Nothing new | The console keeps making up to 50 extra calls to draw one page |

What does not differ: A, B and C all remove those 50 calls, all keep the session row itself identical
on all three session endpoints, and all land in one stage that nothing else depends on. Any of the
four is reversible in that same stage. The admin console itself is unaffected by every option, since
it signs in with full access and only ever shows the owner's name and email.

**My recommendation is B.** It still removes the extra calls, it keeps the boundary where the
documentation already draws it, and it costs one small type rather than a judgement call about which
profile fields a clients-only caller should see.

The reason I am asking rather than just doing it: the plan already chose A, on the grounds that the
full record was free and a smaller one meant a new type. That reasoning weighed payload size and did
not weigh who can call the endpoint, which is what changed my mind. If you would rather the clients
domain simply included the people using those clients, A is a coherent answer and worth saying out
loud, because it also settles how the next endpoint like this one should behave.

### The code, for reference

- `src/authserver/internal/server/routes.go:263` gates the route on `scopesClientsRead`
  (`admin-read`, `manage-clients`, `manage`); `:210` and `:211` gate the user session routes on
  `scopesUsersRead` (`admin-read`, `manage-users`, `manage`).
- `api.UserResponse` in `src/core/api/responses.go` is the 29-field record option A would return.
- `site/src/content/docs/integration/rest-api.mdx:26-32` is the documented scope split quoted above.

Found by the plan review (round 1, blocking, security). The run is stopped until this is answered;
the other five stages do not depend on it.

leodip:
12: B

leodip:
## Deferred decisions: none

This run deferred nothing. All 12 decisions in the agreement are `Decided`, and no stage proceeded
past an open question on a stated assumption. There is nothing here for you to settle.

One decision was escalated to you rather than deferred, and you answered it before any code was
written, so it is recorded for completeness rather than as an open item:

**Decision 12 — what a clients-only caller learns about the people using a client.** Decision 10 had
put a `Users` array of full `api.UserResponse` records on the client-sessions response, to remove the
console's per-session `GetUserById`. It weighed payload size, a new type and a changed `UserFullName`
signature, and did not weigh authorization: `GET /api/v1/admin/clients/{id}/sessions` is guarded by
the clients scopes, while every users route needs the users scopes, so a token holding
`manage-clients` alone reaches this list and reaches no users route. Under decision 10 it would have
learned the subject, birth date, phone number, full postal address, locale and `otpEnabled` of
everyone holding a live session on a client it manages.

You answered `12: B` on 2026-09-17T14:12:54Z. The envelope therefore carries
`api.SessionOwnerResponse` — `id`, `email`, `givenName`, `middleName`, `familyName`, the five fields
a session page shows — and a wire-bytes case asserts that exact key set, so a later widening back to
`UserResponse` fails rather than merely returning more. Decision 9's console fix survives intact:
the up-to-50 per-render `GetUserById` calls are gone.

Nothing had been implemented when the answer arrived, so no built code contradicted it and there was
nothing to unpick.

leodip:

One issue is drafted and not filed, three things were folded in rather than filed, and three
`ceiling:` markers were added to the code. Duplicate searches were re-run just now, at the finish:
nothing in the tracker covers the draft, open or closed.

**Updated after `6f254543`.** What was draft 1 is folded in rather than filed: the three session
DELETE handlers now read `isCurrent`, and the entry has moved to the second section below. Draft 2
is renumbered and unchanged.

### Drafted, ready to file

**1. The twelve id-list lookups in `commondb` still expand without a bound, and one takes its
ids from a request body**

Found during the final review of this PR, as the family the round 1 fix belongs to. #373 bounded
`GetClientsByIds` alone, because that is the lookup its own change made reachable. Twelve more `IN`
expansions in the same package build one bound parameter per element with no bound at all
(`grep -rn "\.In(" src/core/data/commondb/*.go`, excluding tests): `group.go:122`,
`group_attribute.go:126`, `group_permission.go:147`, `permission.go:187`, `refresh_token.go:617`,
`resource.go:144`, `user.go:108`, `user_group.go:130`, `user_permission.go:130`,
`user_session_client.go:138`, `user_session_client.go:208`, `email_backfill.go:279`.

Two differ in kind. **Caller-steerable:** `handler_api_user_groups.go:112` passes
`request.GroupIds`, decoded straight from the request body, to `GetGroupsByIds`, so a caller holding
`manage-users` can post more than 2,100 group ids and get a 500 rather than a validation error —
which unlike the #373 case needs only an unusual request, not an unusual deployment. **Unbounded by
data volume:** `UserSessionsLoadClients` is called by all three session list handlers with every
session in the list, and the users and account endpoints do not paginate, so a user with more than
2,100 live sessions reaches the same ceiling. That shape pre-dates #373 and is untouched by it.

**Not folded in:** thirteen call sites across eleven files, a budget that wants one home rather than
a constant per file, and a decision #373 deliberately did not take — whether the budget is one
number for all four engines or each engine's own. Validating the request-body array is also a
different remedy from batching, and it belongs with the endpoint rather than the data layer.

File it with:

~~~bash
gh issue create \
  --title 'Unbounded id lists in `commondb` become unbounded bound parameters, and SQL Server refuses past 2,100' \
  --label bug --label go \
  --body-file - <<'BODY'
Every id in an `IN` list built by `commondb` becomes one bound parameter, and SQL Server refuses a
statement with more than 2,100 of them (error 8003). #373 bounded `GetClientsByIds` for the session
list endpoints, with `maxIdsPerStatement = 1000` and a `ceiling:` comment recording that the budget
is one number for all four engines; commit `cbedb454` is the one that stopped that 500.

Twelve more expansions in the same package are still unbounded
(`grep -rn "\.In(" src/core/data/commondb/*.go`, excluding tests): `group.go:122`,
`group_attribute.go:126`, `group_permission.go:147`, `permission.go:187`, `refresh_token.go:617`,
`resource.go:144`, `user.go:108`, `user_group.go:130`, `user_permission.go:130`,
`user_session_client.go:138`, `user_session_client.go:208`, `email_backfill.go:279`.

Two are worth naming because they differ in kind:

- `GetGroupsByIds` takes its ids from a request body — `handler_api_user_groups.go:112` passes
  `request.GroupIds` straight through — so a `manage-users` caller can trigger the 500 deliberately
  and should be getting a validation error instead.
- `UserSessionsLoadClients` is called with every session of an unpaginated list, so a user with more
  than 2,100 live sessions reaches the ceiling on data volume alone.

Decide first whether the budget is one number across the engines or each engine's own, then give it
one home in `commondb` and apply it to every id-list lookup. Bound the request-derived arrays at the
handler as well, so an oversized `groupIds` is a 400 rather than a batched read of ids that cannot
exist.

Each site owes a data tier case past 2,100 ids that passes on SQL Server and still returns every
existing row, in the shape of `TestGetClientsByIds_MoreIdsThanOneStatementCanCarry`.

Engine ceilings, for the budget decision: SQL Server 2,100 parameters; modernc.org/sqlite 32,766;
PostgreSQL 65,535; MySQL bounded by statement size rather than parameter count.

#205, "The server accepts unbounded input before anything bounds it", is the nearest existing issue
and is a different axis: request size before authentication, not how many parameters an id list
becomes.
BODY
~~~

**The admin console's `SessionIdentifierToContext` middleware, deleted.** `429d36be`, stage 4. Its
test file and its four lines in the console's `routes.go` went with it. Stage 4 deleted the two
handlers that compared each session row's `sessionIdentifier` against the console's own `sid`, and
they were the only readers of `constants.ContextKeySessionIdentifier` in the admin console. Left
standing, the middleware would have gone on casting the JWT info, reading the claim and writing a
context value on every authenticated request the console serves, for no reader at all — dead
machinery this stage created rather than found. The auth server's middleware of the same shape is a
different package with seven live readers and is untouched.

**The client-id union sent to `GetClientsByIds` is bounded.** `cbedb454`, found by the final review
as finding 2 and fixed under the crash carve-out: a valid, authorized request to any of the three
session list endpoints answered HTTP 500 once the union of client ids exceeded what one statement can
bind. It rides on `session_details.go`, which stage 4 wrote, and on the data method that helper
calls; the 500 itself had no option worth weighing, and the option that did exist, where to put the
bound, is settled by the flavor living only in `commondb`. The strategy question the reviewer flagged
as needing a human is recorded as a `ceiling:` in the code, and the family it belongs to is the draft
above.

**The three session DELETE handlers read `isCurrent`.** `6f254543`, folded in after this comment
was first written. Each of `HandleAccountSessionsEndSesssionPost`, `HandleAdminUserSessionsPost` and
`HandleAdminClientUserSessionsPost` lifted `sid` off the console's own access token and compared it
against every row's `sessionIdentifier`, to decide whether deleting this row must send the browser
through the logout flow rather than simply remove it. Stage 4 deleted the two GET-side
recomputations and the middleware feeding them; these three were the same comparison surviving on
the POST side, where the field they would read is already on every row they already fetch. Each is
now `es.Id == int64(userSessionId) && es.IsCurrent`, which leaves no reader of that claim anywhere
in the admin console.

The argument for filing it was fold-in test 3: three handlers in three packages, and stage 5
touched only one of them, so fixing that one alone would have been the shape #373 exists to
complain about. Doing all three in one stage answers that instead, and the work is the same three
edits either way. The reason it was drafted rather than folded — that none of the three had a test — is what the
stage actually cost: five table rows and a failed-read case per handler, eighteen new cases, and
two new `handlertest` helpers with four cases of their own.

One thing changed beyond the comparison. The list read was swallowed with `if err == nil` at all
three, which was harmless while the read only fed a comparison against a claim the handler already
held. It decides the answer now, so a swallowed failure would delete the caller's own session while
telling the browser it had deleted somebody else's: no logout, and a console holding a session the
auth server has already forgotten. All three forward it to `HandleAPIErrorJson`, the classifier
every other `apiClient` failure in those files reaches (pattern 7). Nine mutations, all caught.

Not a ceiling, but worth naming: the client handler reads back page 1 at size 50, which is the page
its own GET renders, so the two agree by construction. A client holding more than 50 live sessions
shows and deletes only the first 50 from the console, and that is the page's own limit, unchanged
by this run.

Three markers, each with its revisit trigger.

- `src/core/data/commondb/client.go:314` — one budget for all four engines rather than each engine's
  own, so a list of 1,001 to 2,100 ids costs two statements on the three engines that would have
  taken it in one, and a list read in several statements is no longer one snapshot: a client deleted
  between two of them is absent from the result. Revisit when a deployment is measured spending real
  time on the extra round trip; a per-flavor budget is the next shape, and it owes a data tier case
  per engine.
- `src/core/i18n/timefmt.go:68` — day is the largest unit `FormatSince` uses, so a year-old consent
  reads "412 days ago". Acceptable because every page showing the phrase shows the absolute date
  beside it, and because a session's age is bounded by the max lifetime setting. Revisit when a page
  shows the relative line alone; month and year keys beside these are the next shape.
- `src/core/i18n/timefmt.go:83` — the plural form is chosen by `count == 1`, which is the rule for
  both locales this tree ships, en and pt-BR. A locale with a few or many form (ru, pl, ar) would
  take the "other" key for counts its grammar treats differently. Revisit when a third locale is
  added; CLDR plural rules keyed by the locale tag are the next shape.

leodip:
## The review account

This run reviewed once at the end rather than per stage, so **this is the whole of the code review**:
no stage was reviewed on its own. Three rounds ran over the entire branch, `origin/main...HEAD`, by
`openai-codex/gpt-5.6-sol` at effort `xhigh`. Four findings in all, all four confirmed, all four
fixed. Nothing was refuted, nothing was deferred, and the last round left nothing behind.

The gate closed at three rounds of a budget of five. Rounds 1 and 2 each changed production code,
which is what bought round 3; round 3 changed none, and its own summary reads "The production repair
is correct and CI is green. Its regression test is too narrow." A gate has had its answer at the
point where the reviewing stops moving the product.

---

### Round 1 — two findings, both confirmed, both fixed in `cbedb454`

Request `373-20260917T170418Z`, fresh session, started 2026-09-17T17:05:33Z, answered 17:21:00Z.
Verdict `findings`, 2 findings, all three axes `reviewed`.

**What it ran**, in its own account: the whole contract, plan, closing record and the full
`origin/main...HEAD` diff; `check-anchors.sh`, `tree-hash.sh` and the green CI result for `9390138b`;
targeted mapping, handler, i18n, template-helper, API-client, display-helper and render tests; a
12-test sqlite integration pass over all three list endpoints; **four scratch mutations of its own —
forcing `isCurrent` false, retaining invalid sessions, adding `subject` to the owner wire shape, and
replacing the pt-BR date layout with the English one — all four caught** by the tests the stages
wrote; a scratch SQL Server data-tier assertion calling the real `GetClientsByIds` with 2,101 ids;
and OpenAPI plus catalog parity checks. That last one is the round's value: it wrote its own program
against the real engine rather than re-running a tier the record already claimed.

#### Finding 1 — the REST API guide promised `isCurrent` on an endpoint that does not return it

`conformance`, **minor**, confidence high. `judgement: false` with a populated `forced_answer`, so a
mechanical repair rather than a choice.

**Confirmed.** `rest-api.mdx:1545` read "Every session endpoint marks the session the calling token
was issued through with `isCurrent`". `GET /api/v1/admin/user-sessions/{sessionIdentifier}` returns
the plain `UserSessionResponse` through the base mapper, which has no such property, and section 2 of
the agreement puts that endpoint out of scope on purpose (decision 4). The sentence was false as
written about one of the four session read routes.

**Fixed** in `cbedb454`: the sentence now reads "The account, user and client session list endpoints
mark ...". One sentence, same passage, no change to the singular endpoint's contract. The passage
stays two sentences, its stated size: 46 words before, 51 after.

#### Finding 2 — an unbounded client-id union answered a valid request with HTTP 500

`quality`, **significant**, confidence high. `judgement: true` with an **empty** `forced_answer`,
which is ordinarily a choice that blocks; **confirmed and fixed here rather than escalated**, under
the crash carve-out in agreement section 9 — a verified defect that loses a request, on a surface
this change already touches, is fixed in the change that touches it, narrowly, with the behaviour
question around it drafted.

**Confirmed independently of the reviewer's own run:**

- `loadSessionClients` in `session_details.go`, which stage 4 wrote, collects the union of client ids
  across the whole page and hands all of them to `GetClientsByIds`.
- `commondb/client.go:GetClientsByIds` expanded that list into one `IN` clause, one bound parameter
  per id, with no bound.
- All three list handlers turn that error into `writeInternalServerError`, so it is an HTTP 500. Two
  of the three, the users and account endpoints, call `GetUserSessionsByUserId` with no pagination at
  all, so the union there is bounded by nothing but the deployment's client count.
- **Reproduced at this tree**, not taken on trust: raising the budget to 100,000 and running the new
  data case on SQL Server gives `too many parameters. The server supports a maximum of 2100
  parameters. Reduce the number of parameters and resend the request. (8003)`.

**The fix:** `maxIdsPerStatement = 1000` in `commondb/client.go`, with `GetClientsByIds` reading a
longer list in several statements and merging the rows. The number is below the lowest of the four
engines' ceilings — SQL Server 2,100, modernc.org/sqlite 32,766, PostgreSQL 65,535, MySQL bounded by
statement size rather than parameter count. The data layer is where it belongs rather than the
authserver helper: the ceiling is an engine property, `commondb` is the only layer holding the flavor,
and the reviewer's own recommendation independently said the same. The signature does not change, so
no mock moves; no schema changes, so no migration and no `schema.golden`.

**What the fix costs, recorded as a `ceiling:` in the code** rather than left to be discovered: one
budget for all four engines means a list of 1,001 to 2,100 ids now costs two statements on the three
engines that would have taken it in one, and a list read in several statements is no longer one
snapshot. Revisit when a deployment is measured spending real time on the extra round trip. That is
the "cross-engine batching strategy" the reviewer flagged as needing a human: the strategy question
is recorded where the next reader meets the code and the family it belongs to is drafted as a
follow-up, but the 500 itself had no option worth weighing.

**Tests**, at the seam the defect lives in: `TestGetClientsByIds_MoreIdsThanOneStatementCanCarry` in
`tests/data/client_test.go`, a table over two lengths chosen for what each can catch. 2,101 is one
past SQL Server's ceiling. 2,000 is an exact multiple of the budget, where an off-by-one in the loop
bounds drops the final batch instead of failing loudly. Real clients sit in different batches in both,
including the last slot.

#### `unchecked`, all three answered

- *No interactive browser pass over the three console session tables or the audit viewer.*
  **Dismissed.** Decision 3 rejected client-side rendering, so there is no second rendering path to
  observe, and section 5 already rejects a browser seam for the dates on that ground. The render tier
  asserts the emitted cells in pt-BR, which is what decision 3's pin needs.
- *The 2,101-client case was reproduced at the data tier rather than through a live HTTP endpoint.*
  **Covered.** The defect is in a data method and is fixed there, which is where the reviewer's own
  recommendation put the case. The path from that error to a 500 was re-verified by reading the three
  handlers: each wraps `buildSessionDetails` in `writeInternalServerError`. An integration case would
  have to create 2,101 clients to assert what one data case asserts.
- *Already green full tiers were not re-run, their recorded hash matching the reviewed tree.*
  **Covered**, and superseded: this round changed `src/core/data/`, so the data tier ran on all four
  engines locally rather than on sqlite alone.

#### Tiers and mutations at this round's tree

|---|---|---|---|
| unit | `where.sh test --type modules` | n/a | 7,233 pass, 0 fail, ~1m |
| unit, re-run after the last comment edit | `where.sh test --type internal` | n/a | 3,665 pass, 0 fail |
| data | `where.sh test --type data` | **mysql, postgres, mssql, sqlite** | 2,312 pass, 0 fail, ~6m |
| integration | `where.sh test --type integration --db sqlite` | sqlite | 1,420 pass, 0 fail |
| lint | `where.sh test --type lint` | n/a | green, Tailwind and mocks unchanged |
| anchors | `check-anchors.sh` | n/a | all 29 rows resolve |

|---|---|---|---|
| `commondb/client.go` | `maxIdsPerStatement = 1000` to `= 100000` | data, mssql | **caught**, with error 8003 in the log: the original defect reproduced and refused |
| `commondb/client.go` | loop condition to `start+maxIdsPerStatement <= len(clientIds)` | data, sqlite | **caught**: the final partial batch is dropped and a client goes missing |

`where.sh mutate` restored both from its own backup and checked by hash, so no mutation survives in
the tree.

**A correction to this round's own bookkeeping**, recorded because the closing log is append-only and
the entry cannot be edited: the tier and mutation results above are recorded at tree `4ff6bb773108`,
which is **one comment behind the commit**. One edit landed after the tiers finished — a comment added
to `loadSessionClients` saying which layer now decides how many ids a statement binds — moving
`production` to `1d89a35021a9`, which is what `cbedb454` carries. The unit tier *was* re-run after that
edit (3,665 passing, the module holding the file that moved); the data, integration and lint claims were
not, on the judgement that the edit changed no statement in any language the engines see. That judgement
is stated here so it can be disagreed with rather than discovered. The lesson is ordering: hash the tree
after the last edit, not after the last tier.

---

### Round 2 — one finding, confirmed, fixed in `fff211af`

Request `373-20260917T175213Z`, session resumed. Verdict `findings`: 1 finding, all three axes
reviewed. Whole branch in scope at `cbedb454`, with round 1's repair diff read separately.

#### Finding 1 — round 1's batching turned a set read into a multiset read

`quality`, **minor**, confirmed.

`GetClientsByIds` batched an id list without deduplicating it, so an id appearing on both sides of a
batch boundary came back as two copies of the same row. A single `IN` list collapses a repeated id to
one, so this was the method's answer **changing under round 1's own fix**, and only for a list long
enough to be split: nothing before the bound could have shown it. The reviewer demonstrated it with a
scratch sqlite data-tier assertion putting one real id at positions 0 and 1000 of a 1,001-id input
and getting two clients back.

**Fixed** by deduplicating the ids before they are batched, rather than deduplicating the merged rows:
that is the set the single statement used to return, and a caller passing duplicates now spends fewer
statements rather than more.

**No endpoint was answering wrongly.** `loadSessionClients` deduplicates its union before the call and
the two `commondb` loaders map the result by id, which the reviewer checked and recorded in
`siblings_checked`. This is the `Database` interface's documented behaviour restored, not a response
repaired.

**The reviewer's falsifications of the round 1 fix that did *not* produce a finding** are worth
recording: the batch merge mutated from append to assignment fails the checked-in oversized-id case,
and forwarding `nil` instead of the caller's transaction is caught by an enlistment assertion, though
the full sqlite data suite stays green against it.

#### `unchecked`, both answered

- *No concurrent client-deletion stress test between batches.* **Dismissed.** A list read in several
  statements is no longer one snapshot, and that is recorded in the code as a `ceiling:` on
  `maxIdsPerStatement` with its revisit trigger, weighed against the same delete already racing the
  gap between loading a session and loading its clients. A known accepted limit rather than an
  unpinned property.
- *The four-engine data tier was not re-run locally.* **Covered.** CI was green on all thirteen jobs
  at `cbedb454`; `fff211af` owes its own suite, which is green (see below).

**Tiers at `fff211af`:** data on sqlite, 558 passing, ~18s; unit on all four modules, 7,233 passing.
Integration was not re-run: the change is inside one data-layer method, and no handler, route or
schema moved. mysql, postgres and mssql are the check suite's, on this commit.

**Round 3 requested:** rounds 1 and 2 had each produced a confirmed finding, and the function round 2
found this in had by then been changed twice.

---

### Round 3 — one finding, confirmed, fixed in `ee002b48`

Request `373-20260917T182100Z`, session resumed. Verdict `findings`: 1 finding, all three axes
reviewed, **`unchecked` empty**. Whole branch in scope at `fff211af`, with rounds 1 and 2's repair
diffs read separately. The reviewer read `where.sh ci fff211af` twice and recorded all thirteen jobs
green.

#### Finding 1 — round 2's regression test did not test round 2's fix

`quality`, **significant**, confirmed. `judgement: false` with a populated `forced_answer`, so the
code settles it and it was fixed here rather than escalated.

**No production code was wrong.** `TestGetClientsByIds_ARepeatedIdOnEitherSideOfABatchBoundary` put
both copies of the repeated id at positions 999 and 1000 — adjacent, either side of the boundary — so
it did not establish the guarantee the fix actually implements. A read that drops only a repeat of the
id it just kept answers that input with one row as well: collapsing the two adjacent copies leaves
1,000 ids and a single statement. The case was green against deduplication that is not deduplication
of the list, which is the one thing it exists to refuse.

**Demonstrated, not asserted.** The reviewer replaced the seen-map block with adjacent-only
deduplication in a scratch tree, watched both checked-in batching tests stay green, then added a
scratch assertion with the same real id at positions 0 and 1000, which returned two clients. Confirmed
against the code before fixing: under adjacent-only deduplication the 999/1000 pair collapses to 1,000
unique ids and one statement, so one row comes back and the assertion holds, while a 0/1000 pair
collapses nothing and is split into a 1,000-id statement and a 1-id statement, each matching the same
row.

**Fixed** by making the case a table of two placements rather than replacing the one it had. Both are
load-bearing and the comment says which does what: the near pair, 999 and 1000, is the tightest
straddle of the boundary and is what would catch a batch bound off by one slot; the far pair, 0 and
1000, has nothing adjacent to collapse, so only a read that remembers every id it has already seen
answers one row. Taking the near pair out to "simplify" would lose the off-by-one, which is why it is
written down beside it.

**The mutation, this round's evidence.**

```
where.sh mutate src/core/data/commondb/client.go \
  --replace 'if _, alreadySeen := seen[clientId]; alreadySeen {' \
  --with 'if len(uniqueIds) > 0 && uniqueIds[len(uniqueIds)-1] == clientId {' \
  --type data --db sqlite
```

**MUTATION CAUGHT**, and the failure is precisely the new subtest:
`.../one_slot_in_each_statement,_nothing_adjacent` reports `Expected 1 client for an id repeated
across a batch boundary, got 2`, while `.../the_two_slots_the_boundary_sits_between` passes. So the
reviewer's claim is reproduced in the checked-in suite and the gap is closed rather than argued about.
Production restored and proven gone by hash — `production 04e0f108ddb3`, unchanged from round 2 — and
`git status` shows the commit touching one `_test.go` and nothing else.

**Tiers at `ee002b48`:** data on sqlite, 560 passing, 17.6s; unit on all four modules, 7,233 passing,
which is where the gofmt guard reads the edited file. Integration was not re-run: the change is one
test file and no production line, handler, route or schema moved. `check-anchors.sh`: all 29 rows
resolve. mysql, postgres and mssql are the check suite's — **green on all thirteen jobs at
`ee002b48`**, which is `HEAD`.

---

### What the three rounds bought

An HTTP 500 on SQL Server for any request whose session list spans more than 2,100 clients; a set read
that had quietly become a multiset; and a regression test that would not have caught the second one
coming back.

### Deviation, in force for every local tier in every round

As `where.sh` reports each time it runs: the dev container was not created by the devcontainer CLI, so
it is missing all 43 `containerEnv` values and the script supplies them on the exec instead. The tiers
therefore ran against the declared configuration, but through a workaround rather than at the source.
The check suite is unaffected and is the four-engine result of record.

Issue #352, opened 2026-09-12
[Core refactor 21/29] Partition constants by contract and application ownership


leodip:
## Goal

Move authserver-only and adminconsole-only constants to their owning modules while retaining only true cross-process contracts in core.

## Rationale

Core/constants is imported everywhere, but that does not mean every symbol is shared. It mixes HTTP context keys, cookie/session names, resource and permission identifiers, audit event names, OAuth policy values, and UI/application details. A symbol-by-symbol partition prevents a constants package from becoming a loophole around the architecture boundary.

## Implementation guidance

1. Produce a symbol-level consumer census, not only a package-level import count.
2. Move constants referenced only by authserver to a focused authserver package and admin-only constants to adminconsole.
3. Keep values that are genuinely part of the cross-process HTTP/API contract in core, preferably next to the contract they describe.
4. Treat audit event string values as persisted/external data: their exact strings must not change. If adminconsole needs the filter catalog, expose the same catalog as a shared wire value without moving auth-only emission logic back into core.
5. Preserve resource identifiers, permission scopes, context keys, session/cookie names, header names, and ordering of any exported catalogs.
6. Update imports in cohesive groups and add an architecture guard that rejects new application-only constants in core.
7. Do not rename symbols and string values in the same PR; move first, preserving aliases only when a real serialized/API compatibility need exists.

## Compatibility and non-goals

- This is a structural refactor only. Do not change business rules, OAuth/OIDC behavior, HTTP routes, status codes, response bodies, persistence behavior, configuration names, logging semantics, or security policy.
- Preserve public serialized representations, including JSON fields, session payloads, cookie names, token claims, database values, and error classification.
- Move or regenerate tests and mocks with the code they cover. Do not weaken assertions to make the move pass.
- Avoid temporary forwarding aliases unless a concrete compatibility consumer requires one; any temporary exception must name its removal issue.
- Keep every commit buildable where practical and finish with gofmt and tidy only in modules whose imports changed.

## Acceptance criteria

- Every constant left in core has production consumers in both applications or is documented as a stable shared contract.
- Audit filters, authorization scopes, cookie/session behavior, and API values are unchanged.
- Core/constants no longer provides authserver as a general dumping ground.

- The unit tier passes for core, authserver, and adminconsole.
- The repository-wide gofmt and error/logging convention guards pass.
- Relevant race, data, and integration tests continue to pass.
- The PR description lists old and new package paths and confirms runtime behavior is unchanged.

## Dependencies

[...]

PR #375, opened 2026-09-17
refactor(core)!: split configuration and constants by process ownership

### LINKED ISSUE #351: [Core refactor 12/16] Split configuration and constants by process ownership
## Goal

Move configuration loading and the application-only half of `core/constants` into the processes that own them, preserving every environment variable, flag, constant value and audit event string.

## Rationale

`core/config` combines two process configurations, database and bootstrap secrets, and a process-global singleton, so each executable imports configuration it cannot use while auth-only database and encryption configuration sits in the shared module. `core/constants` is imported by nearly everything, but that does not make every symbol shared: it mixes HTTP context keys, cookie and session names, resource and permission identifiers, audit event names, OAuth policy values and UI details. Configuration belongs at the composition root, and a constants package must not become a loophole around the boundary.

## What this issue absorbed

This issue replaces the former #352 (partition constants by contract and application ownership), and absorbs **#209**, the missing completeness check over `AuditEventTypes` — see *Work folded in from other issues* below. Both are closed and point here.

These are the same job — partitioning a shared package by process ownership — and they are the two tree-wide import sweeps left in the epic: `core/config` has 19 importer packages and `core/constants` has 34, overlapping almost completely. Run separately, the same files are edited twice.

They also overlap in content. The admin console's production code uses exactly 21 symbols from `core/constants`, and deciding where `SessionKeyJwt`, `AdminConsoleSessionName` and `AdminConsoleClientIdentifier` live is a configuration-ownership question, not a separate one. The former #352 already depended on #351 for this reason.

This is the loosest of the regrouping's merges: it makes one diff bigger in exchange for one fewer full-tree sweep. If the diff proves unmanageable in practice, splitting it back out along the config/constants line is legitimate — but split it at implementation time, with the census in hand, not up front.

## Implementation guidance

1. Add behavior-locking tests for the current environment-variable and flag matrix, and a symbol-level consumer census for `core/constants`, before moving anything. A package-level import count is not enough.
2. Create authserver-owned configuration for listeners, public and internal URLs, database, AES keys, bootstrap credentials and output, SQL logging, rate limiting, uploads, assets and sessions.
3. Create adminconsole-owned configuration for listeners, URLs, the OAuth client secret, assets, logging, proxy trust and sessions.
4. A process may carry the other process's public or internal endpoint as a value; it must not load the other process's listener or secret configuration.
5. Preserve every `GOIABADA_*` name, default, precedence rule, trimming and conversion rule, validation, secure-cookie derivation and flag behavior. Preserve initialize-once production semantics while keeping tests isolated from global flag registration.
6. Move constants referenced only by authserver to a focused authserver package, and admin-only constants to adminconsole. Keep values that are genuinely part of the cross-process HTTP/API contract in core, preferably next to the contract they describe.
7. Treat audit event string values as persisted external data: the exact strings must not change. If the admin console needs the filter catalog, expose the same catalog as a shared wire value rather than moving auth-only emission logic back into core.
8. Preserve resource identifiers, permission scopes, context keys, session and cookie names, header names, and the ordering of any exported catalog.
9. Do not rename symbols and change string values in the same PR. Move first, preserving an alias only where a real serialized or API compatibility need exists.
10. Update main packages, server construction, handlers, tests, docs and examples. Do not use this issue for a whole-application dependency-injection rewrite.
11. Prepare database configuration for the authserver-owned factory in #353, and add a guard that rejects new application-only constants in core.

**#209 — nothing checks that every audit constant is registered in `AuditEventTypes`.** Closed into
this issue. `AuditEventTypes` in `src/core/constants/constants.go` describes itself as "the
canonical list of all audit event type strings" and is hand-maintained beside the `Audit*`
constants. `constants_test.go` asserts uniqueness and non-emptiness of the slice's contents, and
nothing asserts it is *complete* with respect to the constants. The two agree today, 97 and 97. An
event added to the const block and forgotten in the slice vanishes from the admin UI's filter
dropdown with no test failing.

**Add the completeness check before partitioning anything.** Point 7 of this issue requires that
audit event strings be treated as persisted external data whose exact values must not change, and
point 8 requires preserving the ordering of exported catalogs — while this is the issue that moves
those constants between modules and possibly splits the catalog from the emission sites. A
hand-maintained list with no completeness check is precisely what a partition breaks silently.
Parsing `constants.go` for `Audit*` declarations and asserting each appears in the slice would do
it, as would generating the slice; either way it must survive the partition.

Note the `[A-Za-z0-9]` character class matters when matching the names — `AuditUpdatedClientOAuth2Flows`
contains a digit, and a letters-only pattern truncates it and reports a spurious difference.

## Acceptance criteria

- `core/config` is removed or contains only a demonstrably shared, dependency-free helper, and each executable loads only configuration it owns plus explicit peer endpoint values.
- Every constant left in core has production consumers in both applications or is documented as a stable shared contract.
- The existing configuration test matrix and startup behavior pass unchanged.
- Audit filters, authorization scopes, cookie and session behavior, and API values are unchanged.

## Dependencies

#350.

## Architecture guard

**Package-ownership rows this issue must resolve.**

|---|---|---|
| `core/config` | split | #351 |
| `core/constants` | split | #351 |

Both are `split`, so no edge rule applies to them until this issue lands. Whatever remains of each package afterwards becomes a `kernel` row with no issue, and whatever is left of neither gets its row deleted. A kernel row that still names an issue fails table hygiene.

## Epic context

<details>

- [ ] [#351 — Split configuration and constants by process ownership](https://github.com/leodip/goiabada/issues/351) (this issue)

</details>

leodip:
Closes #351. Absorbs #352 and #209, both closed pointing here.

## What this is for

`core/config` holds both processes' configuration behind one package-global, and `core/constants`
holds 139 exported symbols of which 108 only the auth server ever references. Every module imports
both, so "shared" has come to mean "imported by more than one module" rather than "meant for more
than one module", and that is how a package reaches 139 symbols without anybody deciding it should.

The rule this establishes: **a process loads only the configuration it owns, and a constant lives in
core only if it can say why.** Each binary gets its own `internal/config`, reading its own variables
and carrying the other's endpoints as values. What survives in `core/constants` gets a row in a new
`ARCHITECTURE.md` table justified `kernel`, `both-apps` or `contract`, checked in both directions, so
the next auth-server-only constant that tries to land in core fails the tier instead of settling
there.

## What an operator sees

Almost nothing. The same variables with the same names, defaults, precedence, trimming and
validation, asserted by a table-driven matrix that runs against the old package and both new ones.
Two things do change, and both are deliberate:

- **Each binary accepts only its own flags.** Both accept the identical 45 today, so
  `goiabada-adminconsole -db-type=mysql` parses now and is ignored; afterwards it is refused. A flag
  the binary cannot act on reads as having configured something.
- **Startup refuses, once, on an email-case collision.** The Go email backfill that lowercased stored
  addresses at startup — disabling accounts and revoking credentials with nobody asking — is replaced
  by a SQL migration and a pre-flight that names the offending rows and stops instead.

## How it works

The change is mostly a move, which is what makes it safe to do in one pass: a behaviour-locking
matrix lands first, covering every live `GOIABADA_*` variable's default and environment value, every
flag beating its variable, and the registered flag set of each binary. The same assertions then run
against `authserver/internal/config` and `adminconsole/internal/config`, so the move either preserves
every one of them or goes red.

The audit event names go with them. All 100 move to the auth server, which serves the catalog from
`GET /api/v1/admin/audit-logs/event-types`; the admin console fills its filter dropdown from that
response and compiles no audit string at all. A guard parses the declarations and holds the catalog
to them, which is #209's completeness check finally sitting beside the code it checks — today three
hand-maintained lists agree with each other and none of them reaches the declarations, so an event
added and forgotten passes the suite and vanishes from the dropdown.

## Status

**Complete, and ready for review.** All eight stages have landed, the whole change has been reviewed
over two rounds, and every finding is resolved. `core/config` no longer exists, and `core/constants`
holds 23 symbols with a row apiece saying why. Each stage landed as one commit with its own test run,
and the entries below were filled in as they did.

**The one follow-up this run drafted is fixed here rather than filed** — `3456c817`, the audit
documentation, described below the stages. Nothing is left to open against this change.

### Tests

| tier | where it ran | result |
|---|---|---|
| unit, three modules + setup | local, `where.sh test --type modules`, and the check suite | 7750 passing |
| data, sqlite | local, `where.sh test --type data --db sqlite` | 598 passing |
| data, mysql / postgres / mssql | the check suite only | pass |
| integration | the check suite only | pass |
| lint, four modules | local, `where.sh test --type lint`, and the check suite | 0 issues; mocks and Tailwind CSS identical to what is committed |
| race, vulnerabilities, build, versions | the check suite only | pass |
| anchors | local, `check-anchors.sh` | 26 of 26 resolve |
| docs site | local, `docker build` of `site/` | 43 pages, no warnings |

The check suite is **green on `3456c817`, the branch head** —
[run 35353271233](https://github.com/leodip/goiabada/actions/runs/35353271233): twelve jobs passing
and `Build validation` skipped, which is how it resolves on every run of this branch. That is what
stands behind the four-engine data and integration claims; no local run ever executed them. The
figures above are `2c628c8b`'s, unchanged: the head commit is documentation only and compiles
nothing.

The docs site is not part of that suite. The `Docs` workflow builds `site/` on pushes to `main` and
not on pull requests, so the last row above was run by hand.

### Review

round at the plan gate before any code landed. `GBD_REVIEW_MODE` was `end`, so no stage was reviewed
on its own.

|---|---|---|---|
| plan gate | 4 | 3 | 1 escalated to the owner as decision 18, 3 fixed in the plan |
| final 1 | 4 | 0 | all fixed, `58fa25fe`; one reviewer claim refuted with evidence |
| final 2 | 3 | 0 | all fixed, `2c628c8b` |

- **blocking, security (plan gate):** a self-registration landing between the email-case pre-flight's
  scan and migration 000047's `UPDATE` can fail the migration and leave the schema dirty. Escalated
  as decision 18; the owner answered `C` — require downtime, build no machinery — which is the third
  release-note bullet below.
- **blocking, conformance (plan gate):** `goiabada-authserver migrate to 47` reached the migrator
  without the pre-flight. Fixed in the plan; stage 2 guards both call sites.
- **blocking, conformance (plan gate):** SQL Server's `LOWER()` disagrees with Go's on U+1E9E and
  U+212A, so the problem was not SQLite-only. Measured, then fixed: the pre-flight compares each
  engine's own `LOWER()` against Go's, per row.
- **significant, quality (final 1):** `ScanEmailCase`'s three storage-failure exits could all be
  neutralized with the core tier green. Fixed in `58fa25fe`: five cases against the scripted driver;
  all three of the reviewer's mutations now caught.
- **significant, quality (final 1):** the audit catalog guard parsed one file, so an event declared
  elsewhere in the package escaped it — the #209 defect surviving inside the guard written to close
  it. Fixed in `58fa25fe` by a sibling scan over the package.
- **significant, quality (final 2):** three holes in that new guard, the real one being that
  `var AuditFoo = "foo"` in `events.go` was in neither list while the guard's own advice told authors
  to put it there. Fixed in `2c628c8b`: an event name is a `const` in `events.go`, checked on both
  sides.
- Two smaller conformance findings — this description's missing downtime instruction, and
  `concepts/audit-log.mdx` still pointing at the emptied `core/constants/constants.go` — both fixed.

Full account, with the mutations and the refutation,
[in this comment](https://github.com/leodip/goiabada/pull/375#issuecomment-5730379870).

### Deviations from the agreement

Two things landed that the agreement did not name in advance, both recorded: stage 4 pinned thirteen
audit events' stored values after a mutation showed nothing did (`f706fae7`), and decision 18's
answer narrowed decision 16's promise from "that state never occurs" to "the release note keeps the
operator out of it".

A third landed after the run finished, at the owner's direction, and it reverses the agreement
outright. `3456c817` folds in the audit documentation follow-up the run had drafted as an issue, and
part of it — documenting `GET /api/v1/admin/audit-logs/event-types` on `concepts/audit-log.mdx` and
`integration/rest-api.mdx` — is what section 2 explicitly rejected, reasoning that an API consumer
has little reason to call an endpoint that exists to fill a dropdown. What changed the answer is a
census the agreement did not have: `rest-api.mdx` describes 56 of the 57 admin paths `openapi.yaml`
registers, and the one it omits is this one. The full account is under **What has landed**, below
the stages.

Deferred decisions:
[none, and why](https://github.com/leodip/goiabada/pull/375#issuecomment-5730353959).
Follow-ups, fold-ins and ceilings:
[none, two and one](https://github.com/leodip/goiabada/pull/375#issuecomment-5730366965).

### Stage 1: the behaviour-locking matrix, against `core/config`'s `loadFrom`
`19a78375`. Nothing moved: this stage writes the lock that makes the rest safe, and changes no
production line at all.

`src/core/config/load_matrix_test.go` carries one row per live `GOIABADA_*` variable — 24 auth
server, 19 admin console, 8 database, 5 top-level, 56 in all, 45 of them with a flag — each with its
default, an environment value that differs from it, a flag argument that differs from that, and a
reader that takes the landed value **through the exported accessor** rather than off the
package-global `cfg`. Three tests drive the rows (defaults, the environment, the flag beating the
environment) and a fourth walks the flag set `loadFrom` was handed, which is the "before" half of the
narrowing in decision 3. The same table is carried into each process's own config package and
narrowed there in stages 6 and 7, which is what turns it from a description into a lock.

Two guards keep it honest. One reads `config.go` and compares its `GOIABADA_*` names and its
`fs.*Var` registrations against the table in both directions, carrying the four names that are not
live configuration as a named exception, so a variable or flag added later cannot land uncovered. The
other holds each row to the discipline its own cases depend on. `config_test.go` gains the four
accessor rules that had no owner: `getEnv`'s trimming, `getEnvAsInt` and `getEnvAsInt64` falling back
on anything `strconv` refuses, and `getEnvAsBool` answering false for an unparseable value.

Tiers: unit green, 7536 ok 0 fail, against the 7278 baseline. Data green on sqlite, 594 ok 0 fail.
Integration not applicable — no handler, route or schema is touched. Three mutations against
`config.go` all caught: a changed default, a dropped flag registration, and an accessor returning the
wrong field, the last of which is what proves the rows read through the exported surface the later
stages move.

### Stage 2: migration 000047 and the startup pre-flight replace the Go email backfill

Commit `0fcf6c39`.

`commondb.BackfillLowercaseEmails` is gone, with its three test files, its `Database` interface
entry, its four engine wrappers, its generated mock and `RevocationReasonEmailCollisionBackfill`.
It lowercased every stored address at startup and, where two differed only by case, disabled
accounts and destroyed their credentials without anyone asking. That collision cannot pre-exist on
the two engines that need the repair: `idx_email` has been UNIQUE since `000001`, and the folding
collation MySQL and SQL Server carried before `000040` cannot hold two case variants. On SQLite and
PostgreSQL, which compare byte-wise, such a row already cannot sign in.

In its place:

- **Migration `000047` on all four engines**, above `000040` so the predicate means something:
  `UPDATE users SET email = LOWER(email) WHERE email <> LOWER(email)`. The down is a no-op — the
  original casing is recorded nowhere. Goldens regenerated to `migrated=47`.
- **A startup pre-flight that refuses rather than repairs.** It scans once and stops, naming every
  offending row by id and address, when two addresses share a Go lowercase form, or when this
  engine's own `LOWER()` does not reduce an address the way `strings.ToLower` does. The database is
  left not migrated and not dirty, which matters because a dirty schema refuses to start and
  `migrate` has no force verb.

The engine comparison is not a SQLite special case, and this is the finding worth carrying out of
the stage: **SQL Server leaves U+1E9E and U+212A unchanged** at the collation `000040` installs,
measured against the dev stack, where Go maps them to `ß` and `k`. SQLite maps ASCII only; MySQL and
PostgreSQL agree with Go on both. Written as a comparison rather than a character list, the check
also covers whatever a future engine or collation does.

The check guards **both** migration entry points. `migrate to` reaches `OpenDatabase` directly, so
that a downward step is possible, and never comes through `NewDatabase`; without a check there,
`migrate to 47` on a colliding database would trip `idx_email` half way up the chain.

Core now emits no audit event at all, which is what stage 4 needs before the audit names can move.

**Tiers.** Unit green (7507). Data green on mysql, postgres, mssql and sqlite (2449) — the tier that
pins the per-engine `LOWER()` divergence in both directions. Integration green on sqlite (1423).
Migration rules green. Nine mutations against the new cases, all caught, including neutralising
`LOWER(email)` in the scan and the migration's own predicate.

**Worth knowing.** `migrateTo` now refuses a failed schema-version read rather than skipping the
check: `Version()` answers `NilVersion` on every error, so passing it through would read a broken
database as a fresh one. The package's `scriptedDriver` was extracted to its own file rather than
deleted with the tests it arrived in, since `RunInTransaction`'s contract tests depend on it.

**CI was red on this commit, and `1c24a943` fixes it.** Every job passed but `Lint`, which found
five issues in core, both causes this stage's own: four `staticcheck QF1012` in the new pre-flight
(`WriteString(fmt.Sprintf(...))` for `Fprintf`), and `insertAuditLogWithoutId` left unused, #283
having written it for the one caller this stage deleted. The helper is deleted rather than
suppressed, and the guard that cited it as its worked example no longer names a function that is
gone. Lint is now clean on all four modules; unit, four-engine data and sqlite integration were
re-run green on the fix.

### Stage 3: the catalog endpoint, and the admin console stops compiling audit strings

Commit `8d85dc50`.

`GET /api/v1/admin/audit-logs/event-types` serves the audit event catalog, under the same read
scope as `GET /api/v1/admin/audit-logs`, and the admin console's audit log viewer fills its filter
dropdown from that response. It compiled `constants.AuditEventTypes` into its own binary before,
which made a 104-entry list in `core` a build-time coupling between two processes: a new audit
event could not reach the dropdown without rebuilding the binary that never writes one. The admin
console now names no `Audit*` symbol at all, which is what lets the names themselves move next.

A dedicated route rather than a field on the existing response, because `GET /audit-logs` is paged
and filtered: a fixed catalog on it would ship constant data on every page of every query.

Four audit names nothing has ever emitted are deleted -- `auto_refreshed_token`,
`sent_phone_verification_message`, `updated_sms_settings`, `verified_phone`. All four were already
declared and already unemitted at `v1.6.0`, so they have shipped as filter options that can never
return a result. No `audit_logs` row can reference a name nothing ever wrote, so no stored data is
affected. The catalog serves 100.

Tiers: unit 7515 passing, data (sqlite) 592, integration (sqlite) 1425, lint 0 issues across four
modules. Five mutations, all caught. The four-engine data and integration result is the check
suite's on this commit.

One deviation from the plan, recorded in full in the run's log: the operation declares 500, which
the expansion had deliberately left off. `writeJSON` buffers and answers 500 on a failed encode,
and the openapi contract lint reads it by name as a 500 site.

### Stage 4: the audit names and the catalog move to the auth server, with the completeness guard

Commit `f706fae7`.

The 100 `Audit*` names and the `AuditEventTypes` catalog leave `core/constants` for
`authserver/internal/audit/events.go`, beside the `AuditLogger` whose vocabulary they are. That was
104 of the shared package's 139 symbols, and not one of them is referenced by the admin console;
since stage 2 removed the email backfill, no `core` package emits an audit event either, so every
one is auth-server-only by the census. `core/constants/constants.go` goes from 452 lines to 45.

The 85 auth-server files naming one follow. No name and no value changes, so those files change an
import line and a qualifier and nothing else. `audit.AuditAuthFailedPwd` reads redundantly and
stays that way: renaming a symbol in the same change that moves it is what turns a mechanical diff
into one nobody can review. A second package spelled `constants` inside the auth server would have
left the call sites byte-identical, and was rejected because 45 of the 85 also name a constant
staying in `core`, so 45 files would carry two imports of one package name and an alias.

**The #209 gap is closed.** `events_catalog_lint_test.go` parses the declarations and the catalog
out of `events.go` and holds them to each other in both directions, so an event declared and
forgotten fails the auth server's unit tier instead of vanishing from the operator's filter
dropdown. Two hand-maintained lists go with it: `allAuditConstants`, a third copy of the roster
living inside the old test file, and `expectedCount`, a bare number bumped by hand. The first is
why the gap survived -- it made the suite look like it already checked this, while in fact it
compared two hand-written lists to each other and never reached the declarations.

**Folded in.** Nothing pinned an audit event's stored value. Mutating `AuditAuthFailedPwd`'s string
left the whole unit tier green: every test names the constant, including the integration test that
compares the served catalog against the symbol it serves. The value is written to
`audit_logs.audit_event` and matched by the admin API's `auditEvent` filter, so editing one orphans
every row already carrying it. `TestAuditEventTypes_CriticalEventValuesAreWireValues` now pins the
thirteen values the critical-events roster already names, as literals -- thirteen and not 100,
because a literal for every name would be the fourth hand-maintained list.

Tiers: unit 7525 passing, data (sqlite) 592, integration (sqlite) 1425, lint 0 issues across four
modules. Five mutations; four caught first time, and the fifth is the survival described above,
caught after the fix. The four-engine data and integration result is the check suite's on this
commit.

Three deviations from the plan, recorded in full in the run's log. The guard parses with `go/ast`
rather than with the text pattern and `[A-Za-z0-9]` name class the plan specified -- an identifier
read off the syntax tree cannot be truncated at a digit at all, which is the whole property that
class bought, and a test pins it on `AuditUpdatedClientOAuth2Flows` regardless. The seam's
`AssertAuditCatalogComplete` landed unexported, as both nearest guards in this module are. And the
guard checks a third thing the plan did not ask for: the parsed values against the compiled
`AuditEventTypes`, because the two list comparisons are file-against-file and would pass forever
for a file the build no longer sees.

### Stage 5: `core/data` declares its own `DatabaseConfig`, and no core package reads `core/config`

Commit `70bb58f7`.

`core/data` was the whole of core's edge onto `core/config`: `OpenDatabase` and `NewDatabase` both
took a `*config.DatabaseConfig`, and `NewDatabase` read the two data-encryption keys out of the
configuration singleton. Since #351 deletes that package, the edge has to go, which makes this #353's
point 7 arriving early and unavoidably.

`core/data` now declares the eight-field `DatabaseConfig` itself. Each engine package already
declares the same shape for itself, and `OpenDatabase` maps into whichever one `Type` names, so this
is another copy of an established shape rather than a new idea; the engine switch stays where it is,
because moving it is #353. `NewDatabase` takes the current and previous AES keys as `[]byte`,
matching `runStartupDataTasks`, which already took them as parameters and documents why. One
consequence is a branch that could not be tested before: the 32-byte refusal used to see only the
key the process was started with, and a case now hands it four wrong lengths and then a right one.

The six callers are all in the auth server. `cmd/goiabada-authserver` maps in one helper that takes
the source configuration rather than reading the singleton, so the mapping itself is covered — eight
fields copied by hand is eight chances to write one into the wrong place, and a swapped `Host` and
`Name` is a server that cannot open a database the operator configured correctly. The two test
harnesses map their own, and stage 6 collapses the three copies once a package they can share
exists.

`src/core/middleware/test_main_test.go` loses its vestigial `config.Init()` and import: that
package's production code references `config` nowhere, and after this change there is nothing to
call. `grep -rl 'goiabada/core/config"' --include='*.go' src/core/` now returns nothing.

Nothing an operator can see moves: no variable, default, precedence rule, flag or wire value. The
commit is marked breaking for callers of `core/data`, all of which are in this repository and all
updated here.

Unit green (7527), data green on all four engines (2455 per run), integration green on sqlite
(1425), migration rules green. Two mutations, both caught: the key-length guard neutralised, and
`Host` mapped from `Name`.

### Stage 6: `authserver/internal/config`, and the auth server stops importing `core/config`
Commit `ad20bacd`. 184 files changed.

`src/authserver/internal/config` is the auth server's own configuration package, and the 180 files
that named `core/config` name it instead. `core/config` stays for now, because the admin console
still loads from it; stage 7 removes it.

It mirrors what was there — a package-global, a `sync.Once` behind `Init()`, and the
`loadFrom(fs, args)` seam — and every accessor keeps its spelling, so those 180 files changed an
import line and nothing else.

**The auth server now registers 32 flags instead of 45.** Its own 19, the 8 database flags, the 3
initial-setup flags, and the two admin console values it genuinely reads, `-adminconsole-baseurl`
and `-adminconsole-oauth-client-secret`. `goiabada-authserver -adminconsole-log-level=debug` now
exits 2 where it used to parse and be ignored. This was checked on the built binary and not only in
a test: `-h` lists 32 flags, and the refused one answers `flag provided but not defined`.

`AdminConsoleConfig` is two fields here, the peer's base URL and the shared bootstrap OAuth secret,
which is every admin console value the auth server names anywhere. The removed-setting warning names
`GOIABADA_AUTHSERVER_SET_COOKIE_SECURE` alone, so this log stops carrying a line about a setting
whose cookies it never wrote.

`GetDataDatabaseConfig()` collapses the three copies of the `data.DatabaseConfig` mapping stage 5
had to write — in `cmd`, `tests/data` and `tests/integration` — into one, now that a package all
three share exists.

Stage 1's behaviour-locking matrix is carried across and narrowed to the 39 variables and 32 flags
this process loads, drift guard included. The registration test is rewritten to two explicit
literals, the 32 registered and the 13 refused, each checked against the real flag set and against
the table: an expectation derived from the table cannot fail when a flag leaves the table and the
source together, and this narrowing is the one observable change in the whole PR.

Tests: unit 7805 passing, data (sqlite) 598, integration (sqlite) 1425, lint 0 issues across four
modules. Six mutations against the new cases, all caught — including re-adding the peer's removed
setting to the warning.

### Stage 7: `adminconsole/internal/config`, and `core/config` is deleted

`89654180`. The mirror of stage 6 for the admin console, and then the deletion this issue exists
for.

`adminconsole/internal/config` carries what this process reads and nothing else: its own 19
settings, and the two auth server endpoints it talks to. The peer struct is `BaseURL`,
`InternalBaseURL` and `GetEffectiveBaseURL()`, which is every field the module actually read, at 4,
3 and 4 sites. `Init`, the `sync.Once` and the `loadFrom(fs, args)` seam keep their shape and no
accessor is renamed, so the 50 call sites changed an import line and nothing else.

**The admin console now registers 17 flags and refuses 28.** Its own 15, plus
`-authserver-baseurl` and `-authserver-internalbaseurl`; the auth server's listener and logging
flags, the 8 database flags and the 3 initial-setup flags are gone. `goiabada-adminconsole
-db-type=mysql` used to parse and be ignored, and now exits 2. The two lists partition the 45 both
binaries accepted before the split, and a test asserts that sum, so a flag that quietly left both
fails the tier. The deprecated-setting warning names `GOIABADA_ADMINCONSOLE_SET_COOKIE_SECURE`
alone.

`src/core/config/` is deleted, all three files. `go mod tidy` in all four modules moved no `go.mod`
and no `go.sum` line. `ARCHITECTURE.md` loses the `core/config` ownership row, `CLAUDE.md` and
`AGENTS.md` lose the `config/` bullet and point at the two new packages, and
`environment-variables.mdx` gains one sentence about the flags.

Tests: unit 7616 passing, integration (sqlite) 1425, lint 0 issues across four modules. The data
tier is not applicable — no `core/data` file, migration or query is touched. Four mutations against
the new cases, all caught: the flag re-derive dropped, a claimed flag renamed to a refused one, the
warning widened to the peer's setting, and the both-or-neither previous session-key pair weakened
to `&&`.

### Stage 8: the constants partition, and the fourth `ARCHITECTURE.md` table

`4bb4ff6a`. The other half of the issue, and the one the guard is for.

`core/constants` held 33 exported symbols. **Ten leave.** Nine go to
`authserver/internal/constants`, which nothing outside that module named: the auth server's session
name, three session keys, two context keys and the three OIDC authorization error codes. The tenth
is `ContextKeySettings`, now declared once per process at the same string value, because the two
processes store different types under it — `*models.Settings` against `*api.PublicSettingsResponse`
— so either process's assertion panics on the other's value. It was shared in spelling only.

Every value is byte-identical and no accessor is renamed, so the 96 files naming a moved symbol
change an import line. The 21 that also name a surviving one alias core's import as
`coreconstants`, which keeps an unmarked `constants.` meaning the process's own package everywhere.

**The 23 survivors each get a row in a new `ARCHITECTURE.md` table** justified `kernel`,
`both-apps`, `moving` or `contract`, parsed by `AssertArchitecture` and checked against the
production reference graph rather than the import graph. Both directions, plus one more: a row
claiming less than the tree supports fails, so a row states the strongest claim that holds and
`contract` stays reachable only when nothing else does. No row carries it today. The five rows
pinned only by `core/data` and `core/models` carry `moving` and name #359, held to those packages'
own ownership rows, so the rows expire when the packages leave rather than outliving them.

`core/sessionstore`'s cutover matrix names the auth server's session name and authenticated-session
key, and core may not import a process module in a test any more than in production. It carries the
two literals with the reason, and a test in the auth server's package holds the declarations to
them.

Tests: unit 7736 passing, data (sqlite) 598, integration (sqlite) 1425, lint 0 issues across four
modules. Six mutations, all caught: the kernel precedence disabled, the missing-row direction
disabled, the shadowed-identifier check inverted, the moving-issue check short-circuited, the
census counting test files, and the pinned session name renamed. The table itself is taken apart
row by row by a rule test — 23 drops and 69 justification flips, each of which must produce exactly
one finding naming that symbol.

### After the stages: the audit documentation follow-up, folded in rather than filed

`3456c817`. The run drafted this as an issue and did not file it; it is fixed here instead, at the
owner's direction. Two pages, no code, no test tier touched.

Three defects, two of them older than this branch and one it created:

- `concepts/audit-log.mdx` said *"The viewer reads directly from the `audit_logs` table"*. The admin
  console holds no database handle for that page —
  `handler_admin_settings_audit_logs.go:173` calls `GetAuditLogsPaginated`, which is the auth
  server's `GET /api/v1/admin/audit-logs`. The sentence's conclusion was right and only its
  mechanism wrong, which is the kind a reader repeats: "the console reads the table" invites the
  conclusion that it needs database access, which it has not needed since the admin API became its
  only data path.
- **`user_login` was offered as an example event identifier and no such event exists.** Three sites:
  the identifier list on `audit-log.mdx`, and on `integration/rest-api.mdx` both the example
  request's `&auditEvent=user_login` and the example response body. The catalog carries 100 values
  and none of them is that one; it was 0 in `core/constants/constants.go` before those names moved,
  so the example has never matched anything. A consumer copying the request filters on a value
  nothing writes and gets an empty page with no error, indistinguishable from "nothing happened".
  All three now read `auth_success_pwd`, and every other example identifier on both pages was
  checked against the catalog.
- **`GET /api/v1/admin/audit-logs/event-types` reached `openapi.yaml` but not the site.** Stage 3
  added it; `rest-api.mdx` was left describing 56 of the 57 admin paths that file registers. It now
  has a section beside the viewer, the viewer's `auditEvent` gains the field note it never had, and
  `audit-log.mdx` names it in its REST API list.

That last one reverses the agreement: section 2 rejected listing the catalog endpoint on those two
pages, on the reasoning that an API consumer has little reason to call an endpoint that exists to
fill a dropdown. The census is what changed the answer — it is the only undocumented admin route,
and it is the answer to the question the fake example raised. Recorded as a deviation above rather
than absorbed quietly.

Verified by building the docs image: 43 pages, no warnings. Worth saying because the Docs workflow
runs on pushes to `main` and not on pull requests, so no CI job would have caught a broken page.

## For the release notes

- Each binary now accepts only the flags it owns. `goiabada-authserver` keeps `-authserver-*`, the
  database flags, `-admin-email`, `-admin-password`, `-appname` and the two admin console values it
  reads; `goiabada-adminconsole` keeps `-adminconsole-*` and the two auth server endpoints. A command
  line passing the other binary's flag is refused rather than silently ignored. Nothing in the
  project's own compose files, scripts or Dockerfiles passes any of these; they all configure through
  the environment, which is unchanged.
- On the first start after upgrading, the auth server refuses to start — and
  `goiabada-authserver migrate to` refuses to run — if the `users` table holds two addresses
  differing only by letter case, or an address this database engine's own `LOWER()` does not reduce
  the way Goiabada does. Both refusals name every offending row by id and address and leave the
  database untouched: not migrated, not dirty. Resolve the rows and start again. This replaces a
  startup pass that resolved such a collision by disabling accounts and destroying their credentials
  without asking. It can happen once, and never again afterwards.

  The second condition is not SQLite-only, which is worth saying because the design first assumed it
  was: measured against the dev stack, SQL Server's `LOWER()` also leaves U+1E9E and U+212A unchanged
  at the collation migration 000040 installs, where Go lowercases both.
- **Stop traffic for the first start across this upgrade.** Stop every auth server instance that can
  write users, let the upgrade complete, then restore traffic. The case check above is a scan taken
  before the migration runs, so a registration or email change landing between the two can create
  the collision the migration then trips over; that fails the migration and leaves the database
  marked dirty, which has no supported recovery. Later starts are unaffected.


leodip:
## Decision 18: if someone registers an account while the server is upgrading, should the email-lowercasing migration be allowed to fail?

**What this is about.** Goiabada treats email addresses case-insensitively: every sign-in path
lowercases what the user typed before looking the account up. A recent change made the database
compare text case-sensitively on MySQL and SQL Server (for good reasons to do with client IDs), and
the side effect is that an old account stored as `Alice@example.com` stops being findable. The fix
being built here is a one-off database migration that lowercases every stored address, so those
accounts keep working.

**Where you meet it.** On upgrade. Either the auth server migrates the database on its way up, or an
operator runs `goiabada-authserver migrate to 47` by hand.

**What is planned today.** Before migrating, the server scans the `users` table. If two accounts
differ only by case — `Alice@example.com` and `alice@example.com` both exist — it refuses to start,
prints both addresses and their ids, and changes nothing. The operator merges or deletes one and
starts again. That refusal is deliberate: the old code silently picked a survivor, disabled the other
account and destroyed its credentials at startup with nobody asking, and we are removing that.

**The problem the review found.** The scan and the migration are two separate steps. If anything else
is writing to the database in between — another instance still serving traffic during a rolling
upgrade — a new user can register `alice@example.com` in that gap. Registration lowercases before it
writes, so it succeeds. The migration then lowercases the old `Alice@example.com` on top of it, hits
the unique index on the email column, and fails.

**Why a failed migration is worse than it sounds.** The migration runner marks the database "dirty"
before each step and clears the mark after. A dirty database refuses to start, and there is no
supported way to clear it: the `migrate` command has only `version` and `to`, and the library's
`Force` is not wired up to anything. Recovery means editing the `schema_migrations` table by hand in
SQL, on a production auth server that will not start. Avoiding exactly that state is why the scan
runs first at all — so this gap defeats the point of the design rather than nicking it.

**How likely, and can it be aimed?** It needs an old mixed-case address still in the table, plus a
registration for its lowercase twin landing inside the migration window. By itself that is
vanishingly unlikely. But self-registration is an unauthenticated endpoint, so someone who knows an
old mixed-case address exists can keep trying across an upgrade window, and the payoff is a
deployment that will not start. That is what moves it from "rare" to "worth asking about".

**What the answer turns on:** how much the upgrade owes an operator who is still serving traffic
while it runs.

### The three options, on the same axes

|  | **A. Leave it as planned** | **B. Make the migration unable to fail** | **C. Require downtime** |
|---|---|---|---|
| What is built | Nothing more | One extra condition on the `UPDATE` in each of the four engines' migration files, plus a second scan after the migration | A line in the release notes |
| If the race happens | Migration fails, database is dirty, server will not start | Migration skips that one row and succeeds; the second scan then refuses to serve and names the row | Same as A |
| What the operator does | Hand-edits `schema_migrations` in SQL, then fixes the accounts | Fixes the two accounts and restarts — the same thing the scan already asks of them | Takes the outage; if they did not, same as A |
| Can the database end up dirty | Yes | No, in any case | Yes |
| Upgrades with traffic stopped | Fine | Fine | Fine |

**Worked through.** Say you run two auth servers behind a load balancer and upgrade one at a time.
There is one legacy `Alice@example.com` in your table, and while instance 1 is migrating, a visitor
registers `alice@example.com` on instance 2. Under **A**, instance 1 dies mid-migration and your
database is now dirty, so instance 1 will not come back and neither will instance 2 when you get to
it — you are down, and the way out is a manual SQL edit. Under **B**, instance 1 migrates
successfully, then refuses to serve with a message naming both accounts; instance 2 is still up and
serving the whole time, and you fix the two accounts whenever you get to it. Under **C**, you were
told to stop traffic first, so it never happens, and if you did not read that line you get A.

**What does not differ.** All three cost the same to reverse, none changes any configuration
variable, endpoint or response, and none affects stages 1 or 3 to 8 of this work. The scan that
refuses on pre-existing collisions is in all three.

### Recommendation: B

It costs one condition in four SQL files and one repeat of a scan that is already written. It makes
the promise the design was built around — a collision never leaves you with a broken database —
actually true rather than true most of the time, and it keeps the refusal loud, which was the whole
point of replacing the old silent repair.

**Why I am asking rather than just doing it.** A migration that has been applied is never edited
afterwards, so the condition either goes into these four files now or needs a second migration later.
And B is the only option that adds machinery for a case most deployments will never see, so if you
would rather this change stayed small and told operators to stop traffic during the upgrade, now is
when that is free to say.

**What it costs to be wrong.** Wrong about A: one operator, once, hand-edits SQL on a server that
will not start. Wrong about B: four migration files carry a condition that never fires, and the
upgrade reads the `users` table one extra time.

---

Separately, and not a question — while checking this I measured something the design had assumed away
and it is worth knowing: SQL Server's `LOWER()` does not lowercase the same characters Go does. At
the collation this release installs, it leaves `ẞ` (capital sharp s) and `K` (the Kelvin sign)
untouched, where Go turns them into `ß` and `k`. An address containing either would sail through the
migration unchanged and then be unreachable at sign-in. The design had treated this as a SQLite-only
quirk. The startup scan now compares each database's own lowercasing against Go's, row by row, and
refuses on any disagreement — which covers SQL Server, SQLite, and whatever a future engine does,
without anyone maintaining a list of characters. No decision needed; recorded here because it changes
what the scan checks.

leodip:
18:c

leodip:
## Deferred decisions

**None.** Section 3 of the agreement carries 18 decisions and none is open; no stage proceeded past
a question on an assumption it left for you to settle.

One decision was escalated rather than deferred, and you answered it:

- **Decision 18 — migration 000047's time-of-check window.** Raised by the plan-gate review: the
  email-case pre-flight is a scan taken before `Migrate()` runs, so a self-registration landing
  between the scan and the migration's `UPDATE` can insert the lowercase twin of a legacy mixed-case
  address, collide with the unique index, fail the migration and leave the schema dirty. You answered
  `18:c` on this thread at 2026-09-18T00:25:39Z: require downtime, build no machinery. The migration
  is built exactly as design 4.2 states it, nothing verifies after it, and the whole cost of that
  choice is the third bullet under `## For the release notes` in this description.

  Worth recording once: a migration already applied is never edited, so closing that window later is
  a second migration rather than a change to these four files.

leodip:

### Follow-ups — none

The one this section carried was drafted as an issue and has since been **folded into the branch
instead**, at the owner's direction, in `3456c817`. It is recorded below beside the other fold-in.
Nothing is left to file.

### Folded in rather than filed — two

- **The audit log documentation described a viewer and an event that do not exist.** `3456c817`,
  `site/src/content/docs/concepts/audit-log.mdx` and `site/src/content/docs/integration/rest-api.mdx`.
  Raised by the final review beside finding 3 and drafted here as an issue; the owner read the draft
  and chose to fold it in. Three defects, two older than this branch and one it created:

  - `audit-log.mdx` said *"The viewer reads directly from the `audit_logs` table"*. The admin console
    holds no database handle for that page —
    `handler_admin_settings_audit_logs.go:173` calls `GetAuditLogsPaginated`, the auth server's
    `GET /api/v1/admin/audit-logs`. The conclusion was right and only the mechanism wrong.
  - `user_login` was offered as an example event identifier in three places — the list on
    `audit-log.mdx`, and on `rest-api.mdx` both the example request's `&auditEvent=user_login` and
    the example response body. It is not one of the catalog's 100 values and was not one of
    `core/constants`' before the move, so the example has never matched. All three now read
    `auth_success_pwd`, and every other example identifier on both pages was checked against the
    catalog.
  - `GET /api/v1/admin/audit-logs/event-types`, added in `8d85dc50`, reached `openapi.yaml` but not
    the site, leaving `rest-api.mdx` describing 56 of the 57 admin paths that file registers. It now
    has a section beside the viewer, the viewer's `auditEvent` gains the field note it never had, and
    `audit-log.mdx` names it in its REST API list. **This reverses a decision the agreement made**:
    section 2 rejected listing the catalog endpoint on those two pages, reasoning that an API
    consumer has little reason to call it. The census is what changed the answer — it is the only
    undocumented admin route, and it is the answer to the question the fake example raised.

  Verified by building the docs image: 43 pages, no warnings. The Docs workflow runs on pushes to
  `main` and not on pull requests, so no CI job would have caught a broken page here.

- **Nothing pinned an audit event's stored value.** `f706fae7`, stage 4,
  `src/authserver/internal/audit/events_test.go::TestAuditEventTypes_CriticalEventValuesAreWireValues`.
  Found by mutation: changing `AuditAuthFailedPwd` from `"auth_failed_pwd"` to
  `"auth_failed_pwd_moved"` left the whole unit tier green, because every test in this tree names the
  constant rather than its value — including stage 3's integration test, which compares the served
  catalog against the very symbol it serves and is therefore circular. That value is stored data: it
  is written to `audit_logs.audit_event` and is what the admin API's `auditEvent` filter matches, so
  editing one orphans every row already carrying it. Stage 4 is the change that moves all 100 names
  across a module boundary, so it is the change that owed the pin. Thirteen literal values landed —
  the critical-events roster the package already singles out — rather than 100, because a literal
  apiece would be the fourth hand-maintained list decision 10 removed in the same stage.

Two further fold-ins were agreed before the run rather than found by it, and are recorded as
decisions 12 and 15 in the agreement: `core/middleware`'s vestigial `config.Init()` test call, and
the replacement of the Go email backfill with migration 000047 and the pre-flight.

### Ceilings left by this change — one

- `src/core/testutil/constants_ownership.go:79` — a reference to a `core/constants` symbol is read as
  a selector on the identifier the import binds, resolved by name within the file. A package-level
  declaration shadowing that name is caught; a local variable inside a function shadowing it would be
  read as the package. Nothing in this tree does that, and closing it means type-checking every
  package rather than parsing it. **Revisit if a row in the constants-ownership table ever rests on a
  reference that turns out to be a false one** — `go/types` object identity is the next shape, as
  `AssertNoDeadInterfaces` already uses for one package at a time.

leodip:
## The review account

`GBD_REVIEW_MODE` was `end`, so no stage was reviewed on its own: this is the whole of the code
review, three rounds by an independent reviewer, and the verdicts live nowhere else. Pasted in full
rather than linked.

|---|---|---|---|---|---|
| plan | 1 | `openai-codex/gpt-5.6-sol`, xhigh, fresh | 4 (3 blocking, 1 significant) | 3 | 1 escalated to you as decision 18, 3 fixed in the plan before any code landed |
| final | 1 | `openai-codex/gpt-5.6-sol`, xhigh, fresh | 4 (3 significant, 1 minor) | 0 | all fixed in `58fa25fe`; one reviewer claim refuted |
| final | 2 | `openai-codex/gpt-5.6-sol`, xhigh, resumed | 3 (all significant) | 0 | all fixed in `2c628c8b` |

Neither final round moved a production line: `tree-hash.sh --split` reads `production ff49ddb71fea`
before and after both, the same value the reviewer read. Both rounds landed entirely in guard and
test machinery, which is why round 3 was not funded.

---

### Plan gate, round 1 — request `351-20260917T235654Z`

No code had landed; the branch was one empty commit, so this round read `plan.md` and the contract.
All three axes reviewed.

1. **Time-of-check race on migration 000047** — security, blocking. The pre-flight scans for
   case-colliding addresses before `Migrate()` runs, but another instance still serving traffic can
   insert the lowercase twin of a legacy mixed-case address in between; the migration's `UPDATE` then
   trips the unique index on `users.email` and leaves the schema dirty, which refuses to start and has
   no supported recovery. **Escalated to you as decision 18** rather than answered: migration
   territory, the reviewer itself declined to force an answer, and the trigger is steerable
   (unauthenticated self-registration retried against a deployment window) rather than merely rare.
   One refutation was checked and rejected first — the migrator does hold a cross-process advisory
   lock, but that excludes a second migrator, not a serving replica's insert. You answered `C`:
   require downtime, which is the third release-note bullet.

2. **`goiabada-authserver migrate to 47` bypassed the pre-flight entirely** — conformance, blocking.
   **Confirmed and fixed in the plan.** `migrate.go:54` is the only production caller of
   `data.OpenDatabase` besides `NewDatabase`, and goes straight to the migrator, deliberately, so a
   downward step stays possible. Seam 3 already required the pre-flight to be reachable independently
   of `NewDatabase`; the plan had not named the second call site. Stage 2 guards both, gated on the
   step crossing 000047 upward.

3. **The configuration matrix read the package-global `cfg`** — conformance, significant.
   **Confirmed and fixed in the plan.** Section 5 rejects that read by name: it keeps passing when
   every exported accessor is broken, and the accessors are precisely what the package move preserves,
   at 1255 call sites for `GetAuthServer()` alone. Stage 1 reads through all eight accessors, asserts
   decoded bytes for the two AES keys, and adds a mutation that breaks an accessor rather than a
   default.

4. **SQL Server's `LOWER()` disagrees with Go's `strings.ToLower`** — conformance, blocking.
   **Confirmed by independent measurement** (`probe/lower-parity.sh`, re-run rather than taken on
   trust): at the collation migration 000040 installs, SQL Server leaves U+1E9E and U+212A unchanged
   where Go maps them to ß and k, so an address containing either is never selected by the migration
   and stays unreachable by every credential path. The design had treated this as a SQLite-only
   problem. The pre-flight now compares each engine's own `LOWER()` against Go's, per row, which is
   engine-agnostic and absorbs the SQLite case instead of sitting beside it.

Nothing was refuted. `unchecked` carried five entries; two covered, three dismissed with reasons.

---

### Final review, round 1 — request `351-20260918T111645Z`, resolved in `58fa25fe`

**Finding 1, quality / significant: `ScanEmailCase` had no storage-failure coverage.**
`src/core/data/commondb/email_case.go` returns query, scan and iteration failures correctly, but no
test drove any of them through the real method — the pre-flight's own failure case stops at a mock —
and the reviewer neutralized all three exits with the whole core tier (2475 tests) staying green.
**Confirmed, and worse than a lost error message:** `data.CheckEmailCaseBeforeMigrating` compares the
rows it is handed and nothing else, so a storage failure answered as a clean result is the upgrade
crossing 000047 having read a fraction of the table — the exact outcome the pre-flight exists to
prevent, arriving silently. **Fixed:** five cases in a new
`src/core/data/commondb/email_case_test.go` against the scripted driver already in the package — the
mapping of the three selected columns onto `EmailCaseRow` (with the statement asserted to select
`LOWER(email)` rather than compute it), the empty table, `openErr` on the query, an id that will not
scan, and a result set that dies after two good rows. The reviewer's three surviving mutations,
re-run against the fixed tree:

| mutation | result |
|---|---|
| the query failure return replaced by `return nil, nil` | CAUGHT |
| the `rows.Scan` failure return replaced by `continue` | CAUGHT |
| the `rows.Err` failure return replaced by `return result, nil` | CAUGHT |

**Finding 2, conformance / significant: the release notes omitted the downtime for 000047.**
The section described the collision refusal and the flag narrowing but never told an operator to stop
traffic. **Confirmed as a conformance gap against a decided item:** decision 18C says in terms that
what option C adds is "one line in the release-notes entry the pull request body already owes", so the
branch carried the decision's cost without the sentence that makes it survivable. **Fixed** as the
third bullet under `## For the release notes` in this description.

**Finding 3, conformance / minor: `concepts/audit-log.mdx` pointed at the emptied file.** The page
said the full set of event identifiers "is defined in `src/core/constants/constants.go`", which this
branch emptied of audit declarations. **Confirmed by grep** — the only published reference to the old
location; `CLAUDE.md` and `AGENTS.md` had already been repointed. **Fixed in `58fa25fe`:** the
sentence names `src/authserver/internal/audit/events.go`, 35 words before and after, one path token
swapped for another. **The reviewer's preferred alternative was not taken, deliberately:** it
recommended pointing readers at `GET /api/v1/admin/audit-logs/event-types` instead, and decision 14
weighed exactly that and rejected it — that page's endpoint list is what an operator configures and
queries, and this endpoint exists to fill a dropdown, with `openapi.yaml` carrying it for anyone who
does call it.

**Finding 4, quality / significant: the catalog guard could not see declarations outside
`events.go`.** `auditCatalogFile` was one fixed path, so an `Audit*` constant declared anywhere else
in the `audit` package was compared against nothing: the reviewer added
`AuditForgottenOutsideEvents` to `audit/audit.go`, left it out of `AuditEventTypes`, and the auth
server internal tier (3972 tests) stayed green. **Confirmed from the code** independently of that
demonstration: the parse takes a single file path, all three comparisons read that one parse, and the
empty-parse fatal only fires when *every* declaration has left the file. That is the #209 defect
surviving inside the guard written to close it. **Fixed in `58fa25fe`:**
`auditDeclarationsOutsideTheCatalogFile` reads every production Go file in the package except the
catalog file and refuses an `Audit*` name bound to a string literal, telling the author to move it
beside `AuditEventTypes`; it returns the files it read, because an absence is also what a walk that
reached nothing reports. Three cases — must-fail, must-pass, walk-reached-nothing — and two
mutations, both caught (the forgotten outside declaration, which survived before the fix; and the
scan's `.go` suffix test disabled so it reads no files).

**Refuted, with evidence.** The response closed with "PR #375 still says `In progress` and `Stage 8:
Not started` although all eight stages are at HEAD." The live body said neither: `gh pr view 375
--json body` carried a `## Status` opening "All eight stages have landed." and a filled `### Stage 8`
naming `4bb4ff6a`, and a case-insensitive search of that body for "not started" and "in progress"
returned nothing. Whatever the reviewer read was not the body as it stood.

**The one `unchecked` entry — "no interactive browser pass exercised the audit-log dropdown against
both live processes" — is covered** on the one axis such a pass could have reached: an authorization
mismatch, the admin console holding one token while the dropdown now needs a second endpoint.
`routes.go:313` registers `GET /api/v1/admin/audit-logs/event-types` under
`middleware.RequireBearerTokenScopeAnyOf(scopesSettingsRead)`, the same middleware as
`GET /api/v1/admin/audit-logs` two lines above it, which the same handler already calls with the same
token to render the same page. A token that fetches the rows fetches the catalog. The rest of the
path is covered where it can be observed: the endpoint by an integration case at the sqlite tier
including the missing-token and insufficient-scope refusals, the client by the admin console's unit
tier, the handler binding by its own unit test, and the template by not having changed.

Tiers this round: unit across three modules 7744 passing; data on sqlite 598 passing; lint over four
modules 0 issues with mocks and CSS unchanged; anchors all 26 rows resolving.

---

### Final review, round 2 — request `351-20260918T121526Z`, resolved in `2c628c8b`

Round 2 is funded unconditionally, and it is where `58fa25fe` itself got read: round 1's guard change
was new machinery nothing independent had looked at. All three axes reviewed, `unchecked` came back
empty, and every finding arrived with a surviving mutation rather than an opinion. All three are
against the audit catalog guard, so the whole of this round's resolution is test machinery.

**Finding 1 — declaration-kind coverage in `auditDeclarationsOutsideTheCatalogFile`.** The sibling
scan accepts `const` and `var` by design, and every fixture handed it a `const`, so replacing its
condition with a const-only one left every `TestAuditCatalog_` case green. **Fixed** by
`TestAuditCatalog_TheGuardFailsOnAnEventNameDeclaredWithVar/outside_the_catalog_file`, carrying
`var AuditForgottenOutsideAsAVar = "..."`.

**Finding 2 — multi-file traversal in the same scan.** Every fixture and the real package gave the
scan exactly one eligible sibling, and `os.ReadDir` returns it first, so truncating the directory
loop to `entries[:1]` left every case green; the existing walk-reached-nothing test asserts
membership, which cannot tell a complete walk from one that stops after its first entry. **Fixed** by
`TestAuditCatalog_TheSiblingScanReadsEveryProductionFileAndNotJustTheFirst`: two eligible siblings
with the forgotten declaration in the **later** one, and the scanned filename set asserted whole
rather than by membership — which also pins the two exclusions, since the catalog file or a
`_test.go` file leaking into the walk changes the same set.

**Finding 3 — `parseAuditCatalogSource`, and the guard's own advice. The real one.** The other two
are regression-test gaps; this was a hole in the shipped rule. `parseAuditCatalogSource` read no
`var` in `events.go` except `AuditEventTypes`, and the sibling scan deliberately skips `events.go`.
So `var AuditFoo = "foo"` written into the catalog file was in **neither** list — not compared
against the catalog, not reached by the scan — and the sibling scan's message told an author who
declared it elsewhere to *move it into `events.go`*, so following the guard's instruction was what
made the declaration invisible to the guard. **Fixed as the family rather than the instance:** an
audit event name is a `const` declared in `events.go`, checked on both sides of that boundary. The
parse collects every other `Audit*` name bound to a string literal into `src.mutable` and the
reporting half refuses it; the advice now reads "Move it into %s, declared with const", so the
prescribed move lands somewhere the guard can see. The alternative the reviewer also offered — parse
and compare `var` event names in both places — was not taken: all 100 identifiers are already
`const`, so refusing forbids nothing the package does, and a name that can be assigned again after
`AuditEventTypes` was built from it leaves the filtered value and the emitted value as two strings
this guard, which reads source literals, cannot hold to each other.

Four mutations, each through `where.sh mutate`, **all caught**:

| replacement | what it breaks | result |
|---|---|---|
| the const-or-var condition narrowed to const-only | finding 1 | CAUGHT |
| `range entries` narrowed to `range entries[:1]` | finding 2 | CAUGHT |
| `src.mutable = append(...)` replaced by `_ = name` | finding 3 | CAUGHT |
| the `stringLiteralValue` guard in the mutable branch forced false | the new check's accepted kind | CAUGHT |

The fourth is not the reviewer's: it is the pass half `CLAUDE.md`'s guard shape owes. **It survived
on the first attempt**, and why is worth recording as a trap in this shape of fixture: the pass case
declared `var AuditLoggerDefault *AuditLogger`, which has no value at all, so `i >= len(vs.Values)`
short-circuits before the mutated line is ever reached. A var with a non-string *value*
(`var AuditRetryBudget = 3`) was needed to reach it. A fixture can exercise the branch it looks like
it exercises and not reach the line that decides it.

Tiers this round, against `2c628c8b`: unit across three modules 7750 passing; lint over four modules
0 issues with mocks and Tailwind committed; anchors 26 of 26; `git diff --check` clean.

**Why round 3 was not requested.** Nothing left open: three findings, all confirmed, all fixed, no
deferrals, empty `unchecked`. The round budget funds a round past the second only when the previous
one moved production code, and `production` is `ff49ddb71fea` before and after this round as it was
before and after round 1's. Two consecutive rounds found real gaps and neither moved a shipped line
— both landed in the same guard's test machinery, which is the pattern the budget exists to stop
paying for.

---

### What stands behind the tree as it ships

Every stage's locally recorded tier result is stale — each recorded against the tree as it stood, and
the stages after it landed on top — which `tree-hash.sh --verify` reported for all eight before the
final review opened. What replaces them is the check suite at HEAD: **green on
`2c628c8b` across all thirteen jobs**, including `Tests / sqlite`, `Tests / mysql`,
`Tests / postgres`, `Tests / mssql`, `Unit / race`, `Lint`, `Vulnerabilities`, `Build validation` and
`Versions` — [run 35346847761](https://github.com/leodip/goiabada/actions/runs/35346847761). That
covers the four-engine data and integration tiers no local run ever executed.

Issue #354, opened 2026-09-12
[Core refactor 14/16] Move the four persistence adapters and migration tooling to authserver


leodip:
## Goal

Move all four persistence adapters — `core/data/sqlitedb`, `mysqldb`, `postgresdb`, `mssqldb` — together with `core/data/migrator`, `core/data/schemadump` and `core/cmd/schemadump`, into authserver-owned persistence.

## Rationale

The database engines are auth-server storage adapters, not shared services, and migration execution and schema-golden generation exist solely for the auth database. Moving them removes the last heavy third-party dependencies from the shared module.

## What this issue absorbed

This issue replaces the former #355 (MySQL), #356 (PostgreSQL), #357 (SQL Server) and #358 (migration and schema tooling). All four are closed and point here.

## Why this is one issue and not several

This is the strongest merge in the regrouping, on three separate grounds.

**The four adapters have identical import sets.** Every one of them imports exactly `core/constants`, `core/data/commondb`, `core/data/migrator`, `core/errs` and `core/models`. There is no engine-specific coupling to review.

**They have the same importers.** All four are imported by `core/data/database.go`, `core/cmd/schemadump/main.go`, and the same handful of files under `authserver/tests/data` — `migration_testdb_helper.go`, `database_create_test.go`, `blocked_party_test.go`, `browser_session_test.go`, `migration_lock_test.go`. Four separate PRs means editing `migration_testdb_helper.go` four times.

**The migrator cannot sensibly follow them.** `core/data/migrator` is a **production** import of all four `<engine>/db.go` files. Moving the engines without it gives each moved adapter a temporary `authserver/internal/data/<engine>` to `core/data/migrator` edge, which means four exception rows added and then deleted for one conceptual edge that never has to exist if the tooling moves in the same PR. `core/data/sqlitedb/schema_migrations_test.go` already imports `core/data/schemadump` for the same reason.

The per-engine diff is a `git mv` of 28 Go files and roughly 80 SQL files with the SQL byte-for-byte unchanged. The reviewer's question is "did any SQL byte change?", which `git diff -M` answers identically for four engines or one.

## Implementation guidance

1. Move each engine's implementation files, migrations, embedded sources, `schema.golden` and engine-specific tests together, one commit per engine so each rename stays reviewable.
2. Move `core/data/migrator`, the schema dump, parsing and golden support, their tests, and `core/cmd/schemadump` in the same PR, and update every moved engine's `NewMigrator` construction and embedded migration source.
[...]

PR #377, opened 2026-09-18
refactor(data)!: database selection and factory composition move to the auth server

### LINKED ISSUE #353: [Core refactor 13/16] Move database selection and factory composition to authserver

leodip:
Closes #353.

## What this does

Database engine selection and factory composition move out of `core/data` and into the auth server,
as a new `src/authserver/internal/datafactory` package. `src/core/data/database.go` keeps the
`Database` interface and the `ErrUniqueViolation` sentinel; everything below them --
`OpenDatabase`'s four-arm engine switch, `NewDatabase`, the email-case pre-flight, the startup data
tasks, `MigratorProvider` and `DatabaseConfig` -- becomes the auth server's.

The rule this establishes: **`core/data` declares what a database is, and the process that starts
one decides which engine it is.** After this, `core/data` imports none of `sqlitedb`, `mysqldb`,
`postgresdb`, `mssqldb`, and no production file under `core/data` imports `core/data/migrator`
except the four engine packages themselves.

## Why, given #344 already removed the drivers from the admin console

The headline benefit the issue describes has already landed. The four drivers and `sqlbuilder` left
the admin console's production closure at #344, not here: `go list -deps ./...` in
`src/adminconsole` returns 307 packages and none of the five, `src/adminconsole/go.mod` names none
of them, and `ARCHITECTURE.md` already records the five rows as `no` / `—`. So this change deletes
no architecture-guard row and no tier goes red to signal it is done.

What is left is one real edge, and it is a blocker: `core/data` still imports the four engine
packages, at `OpenDatabase`. #354 moves those packages, plus `migrator` and `schemadump`, to the
auth server, and cannot meet its own acceptance criterion while `core/data` imports them -- that
would be core importing authserver, which the module direction refuses and for which the module
graph has no exception mechanism. This change removes that edge, and one more the issue does not
mention: `core/data/email_case_preflight.go` imports `core/data/migrator` for a single sentinel, so
it moves too, leaving the migrator's production importers inside `core/data` to exactly the four
engine files #354 moves alongside it.

Behaviour is intended to be byte-identical: the same log records, the same error strings, the same
`Create` semantics per engine (SQLite has none, by design -- there is no create statement and no
maintenance connection, and `mode=rw` in the operator's DSN is the equivalent), and the same
ordering of open, pre-flight, migrate, key guard, startup tasks.

### Stage 1: the composition moves to `authserver/internal/datafactory`
Landed as `b9e6a927`.

`OpenDatabase`, `NewDatabase`, `MigratorProvider`, `preflightEmailCase` and `runStartupDataTasks`
left `src/core/data/database.go`, which went from 778 lines to 524 and from nine imports to four.
`email_case_preflight.go` and its test moved whole, and `runStartupDataTasks`' fail-closed table
moved with it; neither test was weakened. All nine call sites were repointed.

Two things are deliberate rather than mechanical. Each switch arm now takes its constructor's two
results into a local pair and returns `nil` explicitly on the error path: all four engine
constructors answer a typed nil pointer beside their error, and returning the call directly would
hide that behind a non-nil `data.Database`. And `data.DatabaseConfig` is deleted rather than moved
-- `datafactory` takes `*config.DatabaseConfig` directly, since both sides of that hand-over are the
auth server's now, which also deletes `internal/config/database_config.go` and its test. No coverage
is lost: `load_matrix_test.go` already asserts each of the eight `GOIABADA_DB_*` variables against
its `GetDatabase()` field.

Tiers: unit green on three modules (7747 ok, 0 fail); **data green on mysql, postgres, mssql and
sqlite** (2455 ok, 0 fail), which is the only place a real server is opened through the moved
factory; integration green on sqlite (1425 ok, 0 fail); lint green, with the mock generator
reproducing `core/data/mocks/database_mock.go` unchanged. `go mod tidy` is a no-op in all four
modules, as expected -- the engine packages stay in `core`, so the drivers stay core's direct
requirements and the auth server's indirect ones.

The two moved seams were mutation-tested in their new home, since a move is where a test can quietly
stop reaching its code; both mutations were caught.

Worth knowing: `ARCHITECTURE.md` had two sentences saying `core/data/database.go` imports all four
engine packages, which this stage made false. Both are corrected in place, in past tense. The four
tables are untouched, because the driver rows already read `no`.

### Stage 2: dispatch and field placement become observable

Commit `d9a56da5`. Test-only: no production file changed, and the `production` tree hash is
byte-identical to stage 1's.

`factory_test.go` is new. The four config mappers get a field-by-field table each, over a source
configuration whose eight values are pairwise distinct -- which is what makes them load-bearing,
since a swapped `Host` and `Name` still answers `dial tcp` and `Password` reaches no error at all.
Each table also pins how many fields the engine's own struct declares, so a field added upstream
and left unmapped is a named failure rather than a setting the driver silently never receives.
SQLite asserts the *absence* of `Create` against a source that sets it, because seven fields rather
than eight is a chosen leniency (#293) and not an oversight.

`OpenDatabase` gets one fourteen-case table against `127.0.0.1:1`, observed through Goiabada's own
`errs.Wrap` prefix and never through driver text, which would pin third-party strings that churn on
a dependency bump. Three of the six server prefixes are unique to their engine and name the engine
reached; each engine's pair of prefixes is what shows `Create` changing which connection is opened
first. SQLite opens a temp file for real under both `Create` values. Every error case compares the
returned `data.Database` against a nil interface rather than through `assert.Nil`, which reports a
typed nil pointer as nil -- exactly the regression stage 1's switch arms guard against. The refusal
is pinned byte for byte on a misspelling, a quoted type, a trailing space and an unset one, which is
what makes the printed string length worth printing. The `opening the database` record is pinned as
a rider, never as the dispatch assertion, since it carries the requested type rather than the engine
reached.

Two gaps close beside them: `logSQL`, which the factory owns and nothing observed (every caller
passes `false`), now read through a transaction on the returned SQLite handle; and
`preflightEmailCase`'s leniency, where a handle that cannot produce a migrator is passed rather than
refused.

Tiers: unit green (7801 passing), data green on sqlite (598 passing), integration not applicable --
no handler, route or schema is touched. Nine mutations, all caught, including the typed-nil return
and the SQLite `Create`-absence case.

## Verification

The check suite is **green on `e72dc966`, the final commit**, all thirteen checks:
[run 35387473459](https://github.com/leodip/goiabada/actions/runs/35387473459). That is where the
four-engine claim for the shipped tree is settled.

| Tier | Where it ran |
|---|---|
| Unit, three modules (7801 passing) | locally per stage, and the check suite's `Unit / core`, `Unit / authserver`, `Unit / adminconsole`, `Unit / setup`, `Unit / race` |
| Data, **mysql, postgres, mssql and sqlite** (2455 passing at stage 1) | locally on all four at stage 1; sqlite locally at every later gate; the check suite's four `Tests / *` jobs on the final commit |
| Integration, sqlite (1425 passing) | locally at stage 1; the check suite on the final commit. Not applicable to stages after it — nothing after stage 1 touches a handler, route, template or schema |
| Lint, including the mock and Tailwind regeneration | locally at stage 1 and the check suite's `Lint`. `core/data/mocks/database_mock.go` regenerates unchanged, as expected: `data/iface` is unedited |
| `go mod tidy`, all four modules | locally, a no-op in every one — the engine packages stay in `core`, so the drivers stay core's direct requirements and the auth server's indirect ones |

`check-anchors.sh` passes at every gate, all 16 rows. The `production` tree hash has been
`bcfbf995bfc5` since stage 1 landed: stage 2 and both review rounds are test-only.

## Deviations from the issue as written

Three of the issue's asks were already satisfied before this branch existed, and one thing the issue
does not mention was folded in. None of these is a reduction in scope — they are the issue's text
having been overtaken by #344 and #351.

- **Its "Architecture guard" section is not acted on.** The five rows it asks to delete already read
  `no` / `—`, the guard is already green, and no tier goes red to signal completion. `ARCHITECTURE.md`
  keeps its four tables untouched; two sentences of its prose that claimed `core/data/database.go`
  imports the four engine packages were corrected in place, because this branch made them false.
- **Point 7's `core/config` clause** is already satisfied: #351 deleted `src/core/config` outright.
- **Its last acceptance criterion**, that `go mod tidy` drop five modules from `adminconsole/go.mod`,
  is already met — that file names none of them, and `go mod tidy` changed nothing anywhere.
- **`core/data/email_case_preflight.go` moves too**, which the issue never mentions. It is the last
  production file outside the engine packages importing `core/data/migrator`, and leaving it would
  make #354 either move a file outside its stated scope or fail its own acceptance criterion.

A correcting comment on issue #353 for the first three is drafted in the agreement and is yours to
post; the issue body is deliberately left as written, because the record that the epic's plan was
overtaken is worth keeping.

## Review

Reviewed by `pi openai-codex/gpt-5.6-sol` at effort `xhigh`, 2 rounds over the whole branch. No stage
was reviewed on its own — the run used `GBD_REVIEW_MODE=end`, so these two rounds are the only
reading the code got.

|---|---|---|---|
| 1 | 1 | 0 | confirmed, fixed in `9f9f3a0a` |
| 2 | 1 | 0 | confirmed, fixed in `e72dc966` |

- **significant, quality:** nothing proved `NewDatabase` *calls* `runStartupDataTasks` — neutralizing
  the call left all 598 sqlite data tests green, though the helper owns the legacy key move, the
  env-to-env rotation and the plaintext TOTP backfill. Fixed in `9f9f3a0a`: two cases that seed a
  sqlite file, reopen it through `NewDatabase`, and assert the conversion happened; all three
  arguments crossing the call mutation-tested and caught.
- **significant, quality:** nothing proved `NewDatabase` *propagates a startup-task failure* —
  returning the database instead of the error left the tier green, permitting a server to report
  success and then serve over data it could not convert. Fixed in `e72dc966`, and the family carried
  past the instance: all five of `NewDatabase`'s error arms were mutated to `if false`, which found
  two further uncovered arms (a failed open, and a dirty `schema_migrations` marker) that the verdict
  had not reached. All five now have a caught mutation.

Neither round changed a byte of production code.

Full account, with every mutation table and tier result:
[this comment](https://github.com/leodip/goiabada/pull/377#issuecomment-5735416500).

## Closing comments

- [Deferred decisions](https://github.com/leodip/goiabada/pull/377#issuecomment-5735395036) — none;
  all eight decisions were settled before the run, and nothing escalated. Includes the one thing
  decision 7 asks of you by hand.
- [Follow-ups, fold-ins and ceilings](https://github.com/leodip/goiabada/pull/377#issuecomment-5735398607)
  — no issue to file, one fold-in named, no `ceiling:` marker added.
- [Review account](https://github.com/leodip/goiabada/pull/377#issuecomment-5735416500) — both rounds
  in full.

## Status

**Complete.** Both stages landed, the whole branch was reviewed twice with both findings confirmed
and fixed, and the check suite is green on the final commit. Ready for review.

Nothing here asks an operator to do anything on upgrade: no environment variable, default, log
message, error string or response changes, the ordering of open, pre-flight, migrate, key guard and
startup tasks is identical, and SQLite keeps its seven-field configuration. There are no release
notes to write.


leodip:
## Deferred decisions

**None.** Nothing in this run was left for you to settle after the fact, and nothing was escalated
at any gate.

All eight decisions in the agreement's section 3 were `Decided` during the interview, before the
first stage was built, and the run raised none of its own. The run guard reports `8 total, 0 open,
0 raised by the run` at the close. No stage narrowed itself around an unanswered question, and no
assumption was taken in place of an answer.

For completeness, the four decisions that actually shaped the code, in case any of them reads
differently to you now than it did at sealing — each is reversible, and the cost of reversing is
stated:

| # | Decided | Reversing costs |
|---|---|---|
| 2 | The composition lives in `src/authserver/internal/datafactory`, split `factory.go` / `preflight.go` / `startup_tasks.go` | a package rename and nine call sites; no behaviour |
| 3 | `data.DatabaseConfig` is **deleted**, and `datafactory` takes `*config.DatabaseConfig` directly | restoring the struct, `internal/config/database_config.go`, its test, and one `load_matrix_test.go` case — about 120 lines of hand-copy back |
| 4 | Four pure named mappers (`mysqlConfig`, `sqliteConfig`, …) exist so field placement is testable | inlining them back into the switch arms and losing the field-by-field tables |
| 5 | `core/data/email_case_preflight.go` moves too, though #353 never mentions it | leaving it means #354 either moves a file outside its stated scope or fails its own acceptance criterion |

Decision 7 asked for one thing from you that is **not** a code change and has not been done here: a
correcting comment on issue #353, noting that its "Architecture guard" section, its point 7 clause
about `core/config`, and its last acceptance criterion about `adminconsole/go.mod` were all
overtaken by #344 and #351. The agreement drafted that comment; posting it is yours, since this
skill is read-only against GitHub outside this pull request. The issue body itself is deliberately
left as written, because the record that the epic's plan was overtaken is worth keeping.

leodip:

### Follow-ups to file: none

No issue is proposed by this run. Both review rounds returned `None` under `## Follow-ups` and
`## Bookkeeping`, and nothing surfaced while resolving either finding that belongs to a different
change. There is therefore no `gh issue create` command in this comment, and no duplicate search to
report — there is nothing to check for duplicates of.

Two things that look like follow-ups and are not:

- **#354** (move the four engine packages, `migrator`, `schemadump` and `core/cmd/schemadump` to the
  auth server) already exists and is the reason this change was run at all. It is unblocked by this
  branch, not created by it.
- **`src/core/data/benign_sentinel_lint_test.go`** imports `core/data/migrator` and will need an
  answer when the migrator moves. It is a test file, the module-direction rule catches it the moment
  #354 lands, and the agreement's section 2 already assigns it there. Filing it separately would
  duplicate a known step of an existing issue.

### Folded in rather than filed: one

**`core/data/email_case_preflight.go` and `email_case_preflight_test.go` moved whole**, in `b9e6a927`,
the same stage that moved `preflightEmailCase`. #353 does not mention this file.

Why it fit: after the composition moves, this is the single production file outside the four engine
packages still importing `core/data/migrator`, and it does so for one constant, `migrator.NilVersion`.
#354 moves the migrator, at which point that import becomes core importing authserver — which the
module direction refuses and for which the module graph has no exception mechanism. Folded in,
`core/data`'s production importers of the migrator drop to exactly the four engine files #354 moves
alongside it. It carries `CheckEmailCaseBeforeMigrating`, `LowercaseEmailsVersion` and three
helpers, cost two comment-path corrections in `core/data/commondb/email_case.go` and its test, and
changed nothing an operator can see.

The alternative considered and rejected was declaring a local nil-version constant so the import
could go while the file stayed — which duplicates another package's sentinel where the two can
silently diverge.

One smaller in-scope correction worth naming, since it edits a document rather than code:
`ARCHITECTURE.md` had two sentences stating that `src/core/data/database.go` imports all four engine
packages. Stage 1 made them false, so both were corrected in place, in past tense. The document's
four tables are untouched — the driver rows already read `no` / `—` before this branch existed,
because #344 removed them, which is the whole reason this change deletes no guard row.

**None.**

```
$ echo $?
1
```

No `ceiling:` marker was added on this branch. That is the expected answer here: the change moves
existing code between packages and adds tests, and chooses no bound, limit or threshold that a later
change would need a trigger to revisit.

leodip:
## Review account

This is the code review for the whole branch. The run used `GBD_REVIEW_MODE=end`, so **neither
stage was reviewed on its own**: the two rounds below are the only reading the code got, and this
comment is the only place their result exists outside one machine. Pasted in full rather than
linked, for that reason.

**Reviewer:** `pi openai-codex/gpt-5.6-sol`, effort `xhigh`. Scope both rounds: `git diff
origin/main...HEAD`, the whole branch against its base.

| Round | Request | Findings | Blocking | Significant | Outcome |
|---|---|---|---|---|---|
| 1 | `353-20260918T184949Z` at `d9a56da5` | 1 | 0 | 1 | confirmed and fixed, `9f9f3a0a` |
| 2 | `353-20260918T191443Z` at `9f9f3a0a` | 1 | 0 | 1 | confirmed and fixed, `e72dc966` |

Both rounds were productive. Neither changed a byte of production code: the `production` tree hash
is `bcfbf995bfc5` at stage 1's landing, at the round 1 review, after round 1's fix, and after round
2's fix. Everything below is test coverage that was missing and now ships.

The gate closes at round 2 rather than running a round 3 because round 2 changed no production code
and left nothing open — and because the defect class it found is now closed by enumerating
`NewDatabase`'s control flow rather than by sampling it.

---

### Round 1 — `NewDatabase` could stop invoking the startup data tasks with the whole tier green

**Significant, high confidence.** CONFIRMED and FIXED at `9f9f3a0a`.

The reviewer earned its round by writing its own program rather than re-running the suite: it
mutated four of the five orchestration edges of `NewDatabase` against the real sqlite data tier and
reported three caught and one survived.

**The claim.** Nothing proved that `NewDatabase` calls `runStartupDataTasks` at all. Neutralizing
only that call left all 598 sqlite data tests passing — though the helper owns the legacy data-key
move (#83), the env-to-env key rotation (#83) and the plaintext TOTP secret encryption (#82).

**Verified before acting, not taken on trust.** `startup_tasks_test.go` drives the helper through
`mocks_data.Database`, one case per fail-closed branch, and every one of those passes whether or not
a caller exists. `src/authserver/tests/data/` opened `NewDatabase` in three places — `TestMain`, the
AES key-length refusal, the email-case collision refusal — and none of them puts data in the
database that any of the three tasks would convert, so none could observe the call.

**The fix.** One new file, `src/authserver/tests/data/startup_data_tasks_test.go`, 145 lines,
production untouched. Two cases, each seeding a throwaway sqlite file, closing the seeding handle and
reopening the same file through `NewDatabase`:

- `TestNewDatabase_RunsTheStartupDataTasksOnTheDatabaseItOpened` — seeds a user holding a plaintext
  TOTP secret, asserts it comes back blanked with a ciphertext that decrypts to the original.
- `TestNewDatabase_HandsTheStartupTasksThePreviousKey` — seeds an RSA private key PEM encrypted under
  a previous key, asserts it reads under the current key afterwards and no longer under the old.

**The family, not the instance.** Three arguments cross that call, all members of the same family
the reviewer did not separate. All three were mutated:

| Mutation in `factory.go` | Result |
|---|---|
| `if err := runStartupDataTasks(database, aesKey, previousAESKey); err != nil {` → `if err := error(nil); err != nil {` | **CAUGHT** by both cases |
| `runStartupDataTasks(database, aesKey, previousAESKey)` → `…, aesKey, nil)` | **CAUGHT** by the rotation case |
| `runStartupDataTasks(database, aesKey, previousAESKey)` → `…, make([]byte, 32), previousAESKey)` | **CAUGHT** by both cases |

The second is why the rotation case is not redundant: a dropped previous key is the one member that
fails in silence. Rotation is skipped, startup reports success, and the server then serves against
data it cannot decrypt.

All three ran through `where.sh mutate`, which restores from a trap and compares by hash;
`production` reads `bcfbf995bfc5` afterwards, the value the round was reviewed at, so every mutation
is proven gone.

**Tiers on `9f9f3a0a`:** modules 7801 passing / 0 failing (~90s); data on sqlite 600 passing (598
before, plus the two added) / 0 failing (~70s). Integration not applicable — no handler, route,
template or schema changed. `check-anchors.sh` passes, all 16 rows.

---

### Round 2 — nothing proved a startup-task *failure* stops startup

**Significant, high confidence.** CONFIRMED and FIXED at `e72dc966`.

**The claim.** Round 1's tests prove `NewDatabase` calls `runStartupDataTasks` on a successful
startup, but nothing proved it propagates a failure. Changing that arm from `return nil, err` to
`return database, nil` left the whole sqlite data tier green — so the suite permitted a server to
report a successful startup after the legacy key migration, the env-to-env key rotation or the
plaintext OTP backfill had failed, and then serve requests over data it could not convert.

**Verified before acting.** `factory.go:171-173` is the arm. `runStartupDataTasks` collapses all
three fail-closed branches into one returned error, so that arm is the only thing between any of
them and a usable handle. All nine `TestNewDatabase_*` functions in the tier were grepped: none
supplies data that makes a startup task fail, so no fixture reached the arm.

**The fix.** `TestNewDatabase_RefusesAStartupWhoseDataTasksFailed`. The forcing fixture is the
rotation canary seeded under a *third* key: `RotateEncryptionKeyIfNeeded` treats a canary readable
under the current key as "already rotated" and one readable under the previous key as "rotate now",
so a canary that opens under neither is the one misconfiguration it refuses outright
(`commondb/reencrypt.go:80-83`), and the cheapest real task failure this tier can construct. The
case asserts the error, a nil database, and that both the wrapping message and the variable names
survive to the operator.

**The family, carried past the instance.** The reviewer named the family "caller-to-helper coverage
for the moved `runStartupDataTasks` boundary" and reported it exhausted by that one member. It is
not: `NewDatabase` runs five steps and returns on the first failure, and the reviewer's own defect is
a property *all five* arms have. Each was mutated to `if false`, so the failure is observed and then
ignored, against the sqlite data tier:

| Arm | Mutation | Before | After |
|---|---|---|---|
| 1 `OpenDatabase` error | `if err != nil {` → `if false {` | **SURVIVED** | CAUGHT |
| 2 `preflightEmailCase` error | `if err != nil {` → `if false {` | CAUGHT (email-case collision test) | — |
| 3 `database.Migrate()` error | `if err != nil {` → `if false {` | **SURVIVED** | CAUGHT |
| 4 the 32-byte key guard | `if len(aesKey) != 32 {` → `if false {` | CAUGHT (wrong-length key test) | — |
| 5 `runStartupDataTasks` error | `if err != nil {` → `if false {` | **SURVIVED** (the finding) | CAUGHT |

Two further real gaps, same defect, neither reported by the reviewer:

- **Arm 1.** Nothing anywhere proved `NewDatabase` stops when the open fails.
  `TestNewDatabase_CreateFalse_AbsentDatabaseIsTheConstructorsError` reads as though it covers this
  and does not: it calls `mysqldb.NewMySQLDatabase` and its two siblings *directly*, never through
  `NewDatabase`, so the arm was uncovered on every engine rather than on sqlite alone. Worse than arm
  5 in consequence — the arm below it dereferences the database it was handed, so a lost return ships
  a nil dereference at startup in place of the configuration error the operator must read. Closed by
  `TestNewDatabase_RefusesAStartupWhoseOpenFailed`, using an unsupported engine name: the only
  refusal `OpenDatabase` has that needs no filesystem, so the case runs on all four engines rather
  than sqlite alone.
- **Arm 3.** Nothing proved a migration that will not run stops startup. Closed by
  `TestNewDatabase_RefusesADirtyDatabase`, whose fixture is the dirty `schema_migrations` marker an
  interrupted migration leaves behind. That is the real state the arm exists for, and
  `preflightEmailCase` reads a dirty database's version without complaint — it says so in its own
  comment at `preflight.go:100-102` — so this arm is the only thing between the marker and a startup
  that runs the data tasks over a schema nobody knows the shape of.

All five arms now have a caught mutation. The family is exhausted by enumeration of the function's
control flow rather than by assertion.

**Tiers on `e72dc966`:** modules green, 7801 passing; data on sqlite green, 603 passing, 20.3s;
integration not applicable. `check-anchors.sh` PASS, all 16 rows resolve to exactly one line in the
named function. `tree-hash.sh --split`: `production bcfbf995bfc5` — byte-identical to the tree the
round was reviewed at, so every mutation is proven restored.

---

### The reviewer's `unchecked` entries, all answered

Four across the two rounds. None was left hanging.

1. *"The unchanged PKCE, token validation, session lifecycle, revocation, one-shot resource, endpoint
   authorization, rate limiting, oracle and audit implementations were not re-audited because the
   diff has no request path into them."* (both rounds) — **Dismissed.** Correct as stated, and it is
   the scope the agreement's section 2 drew: the branch edits no handler, route, template or SQL
   statement, and `data/iface` is unedited, so no request path reaches any of those classes. Round
   2's version is self-evident from the hashes — `production` did not move across either round.
2. *"MySQL, PostgreSQL and SQL Server were not rerun locally; CI is green on all four engines at the
   reviewed HEAD, and the relevant recorded source hashes match."* — **Covered**, by the mechanism
   the run is built on rather than by a local re-run. Decision 8 makes the four-engine data tier a
   check-suite claim by design, and the suite is green on all four database jobs at the final commit.
3. *"A direct `go test -count=3` repeat did not reach the tests because the bare container command
   lacked injected AES configuration."* — **Covered.** The configuration comes from the tier's
   `TestMain`, which is why the wrapper is the supported entry point. Repetition came from the
   round's own runs instead: the three new cases executed across seven separate tier invocations —
   two narrowed runs, four full-tier mutation runs, one full green tier — each time passing when the
   code is whole and failing only against their own mutation. Nothing order-dependent or flaky
   appeared.

### What the review did not change

No production code. Two rounds, two confirmed findings, four new test functions, one new file, and a
`production` tree hash that is identical at every point from stage 1's landing to HEAD. Both findings
were about the same thing: a moved composition root whose *orchestration* nobody had pinned, only
its parts.

Issue #289, opened 2026-08-28
BackfillLowercaseEmails scans the whole users table at every startup, forever


leodip:
`commondb.BackfillLowercaseEmails` opens with `SELECT id, email FROM users` carrying no predicate,
filters the rows in Go, and runs at every authserver start through `runStartupDataTasks`. Once the
pass has converged it still reads every row of `users` on every boot to discover there is nothing
to do, and that cost grows with the table and never falls.

The doc comment on the function claims parity with the precedent it names, being "idempotent and
resumable, on `BackfillEncryptedOTPSecrets`' terms (#82)". That is true of the outcome and not of
the cost. `commondb/otp_backfill.go` filters with `sb.Where(sb.NotEqual("otp_secret", ""))`, so
once #82's pass has run its predicate matches nothing and the query is free. This one has no
predicate to become free.

The Go-side filter is correct and is not what should change. The reason it exists is SQLite:
`LOWER()` maps ASCII only through `modernc.org/sqlite`, so a stored `Ädmin@x.com` is not selected
by any `LOWER()`-based predicate there and would survive the pass untouched. That is the exact
defect #283 decision 13 closed by deciding the whole rule in Go.

What makes this a real question rather than a one-line fix: on MySQL, PostgreSQL and SQL Server the
collation is now case-sensitive and `LOWER()` is Unicode-aware, so `WHERE email <> LOWER(email)`
there is a correct **superset** of the rows the Go filter keeps, and it is empty once the pass has
converged. So the cheap remedy is a predicate on three engines and no predicate on the fourth,
which is a per-engine branch inside a pass whose entire design is that it does not have one. #283
decision 13 was answered **A**, "the data layer decides identity uniformly", by the repository
owner. Whether "uniformly" governs which rows are *selected* as well as how they are *decided* is
the question this issue is really asking, and it is the owner's.

Options, if it is worth doing at all:

- **A. Per-engine predicate.** `WHERE email <> LOWER(email)` on MySQL, PostgreSQL and SQL Server;
  the unfiltered scan on SQLite. Cheapest, and it introduces the engine branch decision 13 A reads
  against.
- **B. A converged marker.** A settings column recording that the pass has run to completion, so it
  is skipped entirely on later boots, following the shape of the one-shot data-key move. No engine
  branch, at the price of a migration and a column.
- **C. Leave it.** The scan is two columns and no join, so on any users table this software is
  plausibly deployed against it is milliseconds at boot and nothing else.

[...]

PR #382, opened 2026-09-19
refactor(data): move persistence to authserver, delete the 1.5.x startup conversions

### LINKED ISSUE #98: Drop the plaintext users.otp_secret column (follow-up to #82)

### LINKED ISSUE #262: Remove the two one-shot startup conversions in NewDatabase once the upgrade window allows
`data.NewDatabase` runs two **one-shot legacy conversions** on every boot. Both exist to carry a pre-1.6.0 database forward, both are permanent code paths today, and both can eventually be deleted. Filing this so the removal is remembered and so the condition that gates it is written down somewhere other than a conversation.

## The two

| | Where | What it converts | Guard |
|---|---|---|---|
| **A** | `CommonDatabase.ReencryptDataToNewKey`, `src/core/data/commondb/reencrypt.go` | the data-at-rest key historically stored in `settings.aes_encryption_key` into the `GOIABADA_AES_ENCRYPTION_KEY` environment key, encrypting the RSA private keys on the way and blanking the legacy column (#83) | runs only while `settings.AESEncryptionKeyLegacy` is 32 bytes |
| **B** | `CommonDatabase.BackfillEncryptedOTPSecrets`, `src/core/data/commondb/otp_backfill.go` | plaintext `users.otp_secret` into `users.otp_secret_encrypted`, blanking the plaintext as it goes (#82) | selects only rows whose plaintext is non-empty |

Each carries a `Database` interface entry, a one-line forwarder in each of `sqlitedb/user.go`, `mysqldb/user.go`, `postgresdb/user.go` and `mssqldb/user.go`, and a generated mock.

## Not a candidate, despite sitting between them

`RotateEncryptionKeyIfNeeded` reads as a third legacy conversion because it was added alongside A for the same issue. It is not one. It implements env-to-env key rotation: an operator sets `GOIABADA_AES_ENCRYPTION_KEY_PREVIOUS`, restarts, and the data re-encrypts to the current key. That is the documented way to rotate the key, so deleting it removes a feature. Recorded here because the resemblance is the trap.

## What gates the removal

Both shipped in **v1.6.0**, which is the current latest release. Removing either now breaks a direct **1.5.x to next** upgrade: a database coming from 1.5.2 still holds its data key in `settings.aes_encryption_key` and its TOTP secrets in plaintext, and would start against a binary with no path to convert either. A is worse than B in that state, since the server cannot decrypt anything at all.

So the gate is not elapsed time, it is a policy question the project has not written down:

- **If an upgrade must pass through 1.6.x**, both can go in the release after 1.6.x, and this issue is a small deletion.
- **If direct 1.5.x to N upgrades must keep working**, both stay until that guarantee is dropped, and the release notes need to say which versions you may skip.

Deciding that is most of the work here. Answer it first.

## When it is time

- [ ] Remove A, its `Database` entry, its four forwarders and its mock. The `settings.aes_encryption_key` column drop is a separate schema question, since a blanked column is harmless and dropping it breaks an older binary.
- [ ] Remove B, same five places. **#98 owns the other half of B**: dropping the plaintext `users.otp_secret` column, removing `models.User.OTPSecret` and simplifying the OTP helpers. That issue already records why the column outlived the backfill, that the Go backfill runs after SQL migrations and needs something to read from. Land them together or land #98 second, but do not let them drift.
- [ ] Leave `RotateEncryptionKeyIfNeeded` alone.
- [ ] State the supported upgrade path in the release notes, whichever way the policy lands.

## Why bother

Neither is expensive at runtime: A is skipped on a `len(...) == 32` check and B is one indexed-free scan of a small table, both once per boot. The cost is that they are permanent code with a per-engine forwarder each, they are read as current behaviour by anyone reading `NewDatabase`, and every release that keeps them makes the upgrade guarantee vaguer rather than clearer.

Severity: low, cleanup.

### LINKED ISSUE #359: [Core refactor 15/16] Move persistence contracts, models, common SQL, and startup data services to authserver

leodip:
Closes #359. Closes #262. Closes #98.

## What this does

Two changes, landed in this order.

**The deletion.** Goiabada 1.6.0 converts a 1.5.x database at first boot: it re-encrypts every
stored secret and signing key from the data key held in `settings.aes_encryption_key` to
`GOIABADA_AES_ENCRYPTION_KEY`, and it encrypts the plaintext TOTP seeds in `users.otp_secret`. Both
are permanent code that runs a check on every boot forever. Both are deleted here, along with the
column the second of them fed: migration `000048` drops `users.otp_secret` on all four engines and
`models.User.OTPSecret` goes with it. `runStartupDataTasks` keeps only its
`RotateEncryptionKeyIfNeeded` arm, which is the documented env-to-env rotation and not a legacy
conversion, and after this reads no settings at all.

**The move.** `core` stops holding persistence. The `Database` interface, its transaction contract,
`commondb`, the generated `Database` mock, the persistence models, the seeder and the console half
of audit logging all move under `authserver`:

| from | to |
|---|---|
| `core/auditlog` | folded into `authserver/internal/audit` |
| `core/data` | `authserver/internal/data` |
| `core/data/commondb` | `authserver/internal/data/commondb` |
| `core/data/mocks` | `authserver/internal/data/mocks` |
| `core/models` | `authserver/internal/models` |

The old packages are deleted rather than aliased: no forwarding shim and no new temporary exception
row in `ARCHITECTURE.md`.

## The rule this establishes

**An upgrade must pass through 1.6.x.** That policy did not exist before this change; it is what
makes deleting the two one-shot conversions safe, and it is stated in full under
*For the release notes* below.

## How it works

The move is a rename, and the compiler proves it. Every moved file keeps its contents, `git mv`
keeps its history, and the only edit inside them is the import block. A behaviour change hiding in
the import rewrites would have to compile, pass three module tiers, four data tiers and the
integration tier, and survive the architecture, slog, error, gofmt, mockery-pin and agent-document
guards, which between them read the import graph, the `core/constants` reference graph and the
generated artifacts. So the move adds no test of its own: what it changes is data those guards
already read.

The ordering is forced rather than chosen. `core` may not import `authserver`, so a package cannot
leave core while a package still in core imports it: `core/data` imports `commondb` at one line, so
the two leave together, and `commondb` imports `core/models` in 26 files, so `models` leaves after
them. `core/auditlog` imports none of them and is free.

The deletion goes first so it is reviewed against code sitting where a reviewer expects to find it,
and so that 314 lines about to be deleted never move at all.

The Go half of #98 is safe for a narrower reason: no production code reads or writes
`User.OTPSecret`. Its only two production writes are the blanking lines inside `SetOTPSecret` and
`ClearOTPSecret` themselves, and all four production call sites go through those helpers and
`GetOTPSecret`, whose signatures do not change. The real work is in the tests, where the integration
tier persisted both columns so a fixture could generate codes from the plaintext one; each of those
becomes a local variable carrying the seed.

### Stage 1: delete the two startup conversions and drop `users.otp_secret` (#262, #98)
**Landed** in `ff486cd9`.

Migration `000048` drops `users.otp_secret` on all four engines, and the four `schema.golden` files
are regenerated to exactly that: one column line out of each, plus the recorded version. The two
conversions leave the `Database` interface, the eight engine forwarders and the generated mock;
`commondb/otp_backfill.go` is deleted whole. `runStartupDataTasks` keeps only its rotation arm and
reads no settings at all, which a test asserts explicitly rather than leaving to the absence of a
line. `models.User.OTPSecret` goes with the column, along with 82 references across 21 test files.

`commondb/reencrypt.go` is refactored rather than deleted, which the plan review caught: that file
holds both entry points and every helper, and rotation *calls* the conversion's entry point, so
deleting it would have deleted the one documented env-to-env key rotation. `ReencryptDataToNewKey`
becomes the unexported `reencryptToKey`, and `reencryptAll` drops its
`UPDATE settings SET aes_encryption_key = ''` — that was the conversion's own bookkeeping, which
rotation only ever reached by sharing the helper.

The protected-column coverage moved rather than going with the deleted test. `reencrypt_test.go`
was the only test reaching four of the eight `aesProtectedColumns` entries whose re-keying code
stays, so `rotate_test.go` absorbed its fixture and is now the exhaustive table over all eight plus
the RSA PEM: each must decrypt under the new key and **no longer** under the retired one, which is
what catches a column dropped from an enumeration nothing derives. A new case pins the
plaintext-PEM canary failing closed — the one thing this change can state about a database that
never booted 1.6.x, since it ships no pre-flight.

Tiers: modules green (7799 passing); data green on **mysql, postgres, mssql and sqlite** (2462
passing), run on all four locally because this adds a migration; integration green on sqlite (1425
passing). Six mutations, all caught. The one that survived first is worth naming: it found that the
new migration test captured its "before" column shape by rolling back through the very `.down.sql`
it was checking, so property 3 compared the statement with itself and passed with the down
declaring `BLOB` instead of `TEXT`. Fixed by migrating forward to 000047 instead, and re-verified.

### Stage 2: fold `core/auditlog` into `authserver/internal/audit`
**Landed** in `3379826a`.

`core/auditlog` was 45 lines and one exported function, and it existed for a reason its own doc
comment stated: two callers on opposite sides of the module boundary, the auth server's
`AuditLogger` and an email-collision backfill in `core/data/commondb`, that could not share code
any higher up because core cannot import the auth server. #351 deleted that backfill, so one
production caller is left and this issue puts caller and callee in the same module. The package is
therefore folded in rather than moved one directory across, which is what lets
`slogRequestPathDirs` *lose* its `core/auditlog` entry instead of gaining a renamed one:
`authserver/internal/audit` is already in that list, so `TestSlogConvention` and
`AssertAuditLogContext` keep their coverage by construction.

`auditlog.go` and its five tests move by `git mv`, contents unchanged but for the package clause.
`LogToConsole` keeps its name and its export. The doc comment is rewritten rather than carried:
the "Two sites call this" paragraph has been false since #351, and what survives is why `details`
rides as an attribute value (#320 decision 7) and why it takes a context (#328).

Neither fixture plant was deleted. `slog_lint_test.go`'s follows the real fold to
`authserver/internal/audit/log_to_console.go`; `audit_context_lint_test.go`'s goes to
`core/sessionstore/aliased.go`, a core directory deliberately, since it is the tree's only refused
plant in a listed *core* directory and putting it beside the auth server's would have dropped that
half of the walk's coverage while the test stayed green. Both were mutation-tested by removing the
directory they now sit in from `slogRequestPathDirs`, and both mutations were caught.

`ARCHITECTURE.md`'s `core/auditlog` ownership row goes in the same commit, because
`checkTableHygiene` fails a row holding no production Go file and fails a package holding no row.

Two corrections folded in, both falsified by this fold rather than left for a later commit:
`audit_context_lint.go`'s scope paragraph, which said the scope is the request-path directories
"rather than the two audit packages" and named a backfill #351 deleted as the live example of an
admitted `Background` context (no production site takes that admission today, so it now names the
fixture that pins it); and `CLAUDE.md` / `AGENTS.md`, which list core's `auditlog` among the
request-path directories in pattern 8 — one list item out of each, byte-identical as
`AssertAgentDocs` requires.

Tiers: `go vet` over all four modules, and the three module unit tiers, 7799 tests green. No data
or integration tier — no query, no migration, no handler and no route changed, so neither can
observe this.

### Stage 3: move `core/data`, `core/data/commondb` and `core/data/mocks` under `authserver/internal/data`
**Landed** in `24eed2f5`.

44 tracked files move by `git mv` and 192 are import-rewritten. The only edit inside a moved file
is its import block, which is what makes "this is a rename" a claim the compiler and the guards
check rather than one the commit asserts. The strongest single piece of evidence is the generated
mock: the mockery `Database` entry moves to `src/authserver/.mockery.yaml`, and regenerating in the
dev container produces a 13800-line file byte-identical to the moved-and-rewritten one, so the
config and the tree agree by construction.

What the move forces, all of it data something already reads:

- `ARCHITECTURE.md`'s `core/data` ownership row is deleted in this commit, because
  `checkTableHygiene` fails a row holding no production Go file.
- **All five `core/constants` rows are restated, not the four the plan expected.**
  `strongestJustification` tests `both-apps` before `moving`, so `AdminConsoleClientIdentifier`
  becomes `both-apps` the moment the seeder lands in the auth server beside the admin console's four
  production references — it does not wait for `core/models/client.go` in stage 4. The four
  permission identifiers become `contract`, their only referrers now
  `authserver/internal/server/routes.go` and that same seeder. They are the first `contract` rows in
  the document, and the word is earned rather than conceded to the guard: they are scope strings a
  client asks for and a token carries, and the admin console compiles all four through
  `BuiltInAuthServerPermissionIdentifiers`, which is `both-apps` on its own account and cannot leave
  core. The note predicting these rows would simply stop being backed was wrong, and is rewritten to
  what the tree backs.
- `pageOffsetRoots` collapses to one element as its own comment instructed; `runInTransactionOwners`,
  `pageOffsetOwner` and the id-list walk root are repointed, with every lint fixture path.
- `core/go.mod` loses `github.com/huandu/go-sqlbuilder` and `modernc.org/sqlite` as direct
  requirements, and `go mod tidy` takes eleven indirect lines with them.

Six stale comments corrected, not the three the plan listed. The planned three are
`core/testutil/gofmt.go`, `guard.go` and `slog_lint.go`'s ceiling; the other three were falsified by
this stage's own rename and found by the closing sweep — `datafactory`'s package comment,
`core/oauth`'s #338 counterfactual, and the scripted-driver pointer in the data tier's
`run_in_transaction_test.go`. `guard.go`'s numbers were recounted rather than transcribed, per
decision 9: six reporting halves live outside `core/testutil`, in three directories, against the
four in two the sentence claimed, and it now carries the command that recounts them.

`CLAUDE.md` and `AGENTS.md` are edited together and stay byte-identical: core's directory list loses
`data/`, the auth server's gains `internal/data/`, and the Database Pattern section points at
`src/authserver/internal/data/database.go`.

Tiers: `go vet` over all four modules with test files included; unit 7787 green; **data 2462 green
on mysql, postgres, mssql and sqlite**; integration 1425 green on sqlite. The data tier ran on all
four engines rather than sqlite alone because this stage moved the whole data layer, which is where
an engine difference would hide. One note worth carrying: the first unit run failed
`TestGoSourcesAreGofmted` on 128 files, since the rewrite lengthened import lines and changed their
sort position — `gofmt -w` fixed all 128, and that tier is what catches it, not `go vet`.

### Stage 4: move `core/models` to `authserver/internal/models`
**Landed** in `f35e2b7d`.

31 tracked files move by `git mv` and **423** are import-rewritten, not the 425 the plan counted:
two importers went with the files stage 1 deleted. Every one of the 423 is in `src/authserver` and
every one spells the import unaliased, so the rewrite is a single substitution inside one module --
the admin console and `cmd/goiabada-setup` name the package nowhere, which the grounding's import
census predicted and which held. The generated `Database` mock regenerates to a one-line diff, its
`models` import, and `go mod tidy` moves nothing in any of the four modules: `jwt/v5` is
`core/oauth`'s too, and the two requirements only the moving packages had left with `commondb` in
stage 3.

Flat under `internal/` rather than under `internal/data/`, per decision 1: 423 files import the
package and almost none of them are data-layer files, so filing it beside the `Database` interface
would describe where the types are defined rather than where they are used.

What the move forces:

- `ARCHITECTURE.md`'s `core/models` ownership row is deleted in this commit, the last of the three
  the issue owed. Its two prose mentions are made true at the same word count each: the
  temporary-exceptions preamble, whose worked example was `adminconsole/internal/handlers` granted a
  dependency on `core/models` -- a pair where neither end exists any more -- now uses a live pair
  from the table directly beneath it, and the burn-down note takes the new path.
- The `core/constants` table needs nothing, because stage 3 restated all five rows including
  `AdminConsoleClientIdentifier`. `core/models/client.go` leaving does not disturb it, and
  `AuthServerResourceIdentifier` stays `kernel` on `core/handlerhelpers` and `core/middleware`.
- `CLAUDE.md` and `AGENTS.md` move the `models/` bullet from core's directory list to the auth
  server's, edited together and byte-identical, at an unchanged total word count.
- Three code comments naming `core/models` in prose are corrected by path substitution:
  `apimapping`'s package doc, `AuthCodeReusedError`'s type doc and `openapi_contract_lint_test.go`'s
  section comment. The `core/models` strings in `core/testutil/architecture_rules_test.go` are
  deliberately left: they are synthetic fixtures under a fake `example.test/` module, the
  architecture guard's rule-test input rather than references to this tree, and repointing them
  would be editing a test to match the code it exists to constrain.

Tiers: `go vet` over all four modules with test files included; unit 7787 green; data 602 green on
sqlite; integration 1425 green on sqlite. The last two are bought for a stage that touches no query
and no handler because it is the last stage and the change wanted one end-to-end pass locally before
the four-engine verdict. The unit tier is where this stage is actually verified: `AssertArchitecture`
reads the real import and reference graphs against the edited tables, `AssertAgentDocs` the two
documents, `TestGeneratedMocksArePinned` the regenerated mock.

### Annex: remove `Database.GetLastUserWithOTPState`, which nothing calls (#359)
**Landed** in `e571ac34`, **after the PR was opened and on the user's instruction.**

The run drafted this as a follow-up issue rather than folding it in, on scope alone: §2 puts
"narrowing the `Database` interface" out of scope by name. The user asked for it here instead, which
is what the drafted entry itself said would move it.

`GetLastUserWithOTPState` had no production caller anywhere in the tree. It cost one interface
declaration, one `commondb` implementation, four engine forwarders and 68 lines of the generated
`Database` mock, and a reader had to trace all six to discover it was dead. What kept it looking
alive was a comment: its doc in the data tier said the method was "used by the OTP backfill to walk
users by OTP state", which was already false when written — `BackfillEncryptedOTPSecrets` issued its
own `SELECT` and never called this — and stage 1 then deleted that function outright while closing
#262, so the comment named code that no longer existed.

196 deletions and no addition: the interface entry, the implementation, the four forwarders, the
mock surface and `TestGetLastUserWithOTPState`. The test goes rather than moving, because the only
behaviour it described was this method's. No production behaviour changes, because no production
path reached it, and the regenerated mock is what makes that a claim mockery checks rather than one
the commit asserts — it produced a diff of exactly 68 removed lines and no other change.

**No guard is added, and that is a finding rather than an omission.**
`TestHandlers_NoDeadInterfaces` holds whole interfaces to being referenced, not individual methods,
and a method-level version of it would not be green today: **39 of the interface's 216 methods have
no production caller.** The other 38 are the uniform per-model CRUD surface the data tier exercises
— `GetXById`, `UpdateX`, `DeleteX` — which is a deliberate shape rather than an accident, and
separating the two classes is a judgement no tree walk makes. This one was a bespoke predicate
written for a conversion that has since been deleted, which is a different thing and the reason it
is the only one leaving.

Tiers: `go vet` over all four modules with test files included; modules 7787 green; data **598**
green on sqlite, four fewer than before and exactly the removed test with its three subtests;
integration 1425 green on sqlite; **lint green**, which is the half that runs mockery itself and
fails on a tree it changes. No data tier on the other three engines: this removes a query and adds
none, and CI runs all four.

## For the release notes

**Upgrading to this version must pass through 1.6.x.**

This release removes the two one-shot startup conversions that carried a 1.5.x database forward, and
drops the column the second of them fed. Upgrading a 1.5.x installation directly to this version is
destructive in two ways:

- **The data key is not carried forward.** On 1.5.x the data-at-rest key is stored in
  `settings.aes_encryption_key`. Nothing moves it to `GOIABADA_AES_ENCRYPTION_KEY` any more, so
  client secrets, provider credentials and RSA signing keys will not decrypt.
- **Unencrypted TOTP seeds are destroyed.** On 1.5.x a user's TOTP seed may exist only in
  `users.otp_secret`. Nothing encrypts it into `users.otp_secret_encrypted` any more, and migration
  `000048` then drops the column. Those users keep 2FA enabled with no readable seed and cannot
  complete sign-in.

Upgrade to 1.6.x first, start the server once so both conversions run, then upgrade to this version.
There is no pre-flight check for this: the server will not refuse to start on a 1.5.x database.

Rotating between two `GOIABADA_AES_ENCRYPTION_KEY` values is unaffected and still works exactly as
documented.

## Verification

**The check suite is the authority.** All 13 jobs passed at `4985df66`, the tip of the four-stage
change: `Build
validation`, `Lint`, `Vulnerabilities`, `Versions`, `Unit / core`, `Unit / authserver`,
`Unit / adminconsole`, `Unit / setup`, `Unit / race`, and `Tests / sqlite`, `Tests / mysql`,
`Tests / postgres`, `Tests / mssql`. The four-engine data tier and the integration tier run only
there; nothing local reaches all four engines on every stage. The annex commit `e571ac34` was
pushed afterwards and is running now; it deletes a query and adds none, so the four-engine verdict is
the only thing it is waiting on.

**Run locally, per stage:** `go vet` over all four modules with test files included; the three module
tiers (7787–7799 passing depending on the stage); the data tier; and the integration tier on sqlite
(1425 passing). Stage 1, which carries migration `000048`, ran the data tier on **all four engines**
rather than sqlite alone — 2462 passing — because a migration is the change that tier exists for.
Stage 3 did the same, because it touched `src/core/data/`. Stage 1 also ran **six mutations**, one per
new seam and one per case pinning a section 3 decision; mutation 2 caught a vacuous assertion in the
`000048` down-shape test, which is recorded under *Deviations* below.

The Astro site is not built by CI on a pull request (`docs.yml` triggers on `main`, on release and on
dispatch), so it was built by hand while closing the review: **43 pages, 4.39s, no error**, with both
edited pages rendering.

### Review

Reviewed by pi `openai-codex/gpt-5.6-sol` at effort `xhigh`. Under `GBD_REVIEW_MODE=end` the whole
branch was reviewed at a single final gate — **no stage was reviewed on its own** — after one round at
the plan gate before any code existed.

|---|---|---|---|---|
| plan | 1 | 1 | 1 | confirmed, fixed in the plan before any code landed |
| final | 1 | 1 | 1 | confirmed, fixed in `4985df66` |
| final | 2 | 0 | 0 | clean |

- **blocking, conformance (plan gate):** stage 1 as planned deleted `commondb/reencrypt.go`, which
  holds the surviving `RotateEncryptionKeyIfNeeded` and every helper it uses — the stage could not
  have compiled. Fixed in the plan: the file is refactored rather than deleted, `ReencryptDataToNewKey`
  becoming the unexported `reencryptToKey` that rotation calls. Resolving it also caught that deleting
  `reencrypt_test.go` would have dropped the only coverage of four surviving `aesProtectedColumns`
  entries, so `rotate_test.go` became the exhaustive table over all eight.
- **blocking, conformance (final gate):** `production-deployment/database.mdx` told operators an
  upgrade "needs nothing beyond installing the new version and restarting", which the deletion makes
  false and destructively so. Fixed in `4985df66`: the absolute clause is replaced by a pointer to the
  release notes, mirroring the rollback paragraph six lines below it. The documentation sweep had
  searched for the moving package names and this page names none of them.

Nothing was refuted and nothing was deferred. Full account, with the evidence and the reviewer's own
scope, in [this comment](https://github.com/leodip/goiabada/pull/382#issuecomment-5743216535).

## Deviations

Sixteen in total across the four stages, every one recorded in the run's log rather than left silent.
None changes what section 2 of the agreement promised. The ones a reviewer should know about:

- **Three `## 0. Code anchors` rows were deleted rather than repointed** — `startup/reencrypt`,
  `startup/otp-backfill` and `docs/env-reencrypt` — because each named code decisions 4 and 7
  deliberately delete, and the anchor checker has no marker for a row whose subject is gone on purpose.
- **A fifth documentation surface was corrected**, beyond section 2's "Nothing else": `database.mdx`,
  from the final review's finding above.
- **The `000048` down-shape test was vacuous as first written** and passed with the down declaring
  `BLOB` instead of `TEXT`. It followed `000047`'s fixture pattern, which is safe only because that
  migration's down is a no-op. Fixed by migrating an empty database *forward* to `000047` and never
  down to it; the reason is written into the test's own comment so a later reader does not tidy it back.
- **Stale comments were corrected where a stage's own rename falsified them**, rather than batched:
  three in stage 2, six in stage 3, three in stage 4. Each rides on a file the stage already had open.
- **`core/testutil/architecture_rules_test.go`'s `core/models` strings were deliberately left alone.**
  They are synthetic fixtures under a fake `example.test/` module driving the architecture guard's own
  rule tests; repointing them would be editing a test to match the code it exists to constrain.
- **A line-ending slip in stage 4**, caught by the unit tier: six files landed with CRLF after a
  Python text-mode write. Stripped, re-run green, and each diff re-read to confirm only intended lines
  had moved.

## The run's account

Three comments on this PR carry what the diff cannot:

- [Deferred decisions](https://github.com/leodip/goiabada/pull/382#issuecomment-5743203193) — none
  deferred, and the four judgement calls the run made on its own authority, each with what reversing
  costs.
- [Follow-ups, fold-ins and ceilings](https://github.com/leodip/goiabada/pull/382#issuecomment-5743205840)
  — nothing left to file, the three issues folded in here (the third annexed after the PR was
  opened), one issue for you to close, and no ceilings.
- [Review account](https://github.com/leodip/goiabada/pull/382#issuecomment-5743216535) — all three
  rounds in full.

## Status

**Complete and ready for review.** All four stages landed — `ff486cd9`, `3379826a`, `24eed2f5`,
`f35e2b7d` — with the review repair in `4985df66`, and the check suite was green on all 13 jobs at
that commit. `core` now holds no persistence contract, no common SQL, no audit logging and no
persistence model, which is the whole of the issue's goal item 1, and #262 and #98 are closed with
it.

One commit landed after that, on the user's instruction rather than as part of the sealed agreement:
`e571ac34`, the annex above, which deletes the dead `Database.GetLastUserWithOTPState`. It was
verified locally on every tier this machine reaches and its check suite is running.


leodip:
## Deferred decisions

**None.** Nine decisions were settled at the seal, all nine before any code landed, and the run
raised no new one and escalated nothing. No stage was narrowed to avoid a question, and nothing
here is waiting on you to unblock.

What follows is not a deferral but its neighbour: four judgement calls the run made on its own
authority, each inside work the agreement had already decided. They are listed because each is
reversible and none was yours to approve at the time.

### 1. `reencryptAll` lost its legacy-column blanking

**The call.** Decision 4 deleted `ReencryptDataToNewKey`, but `RotateEncryptionKeyIfNeeded` — which
stays — reached the same helper. The helper ended with `UPDATE settings SET aes_encryption_key = ''`,
whose own comment says it exists "so subsequent startups skip the migration". After stage 1 there is
no migration to skip, so the statement was removed rather than carried into a path that no longer
needs it.

**Why it is safe.** The column is already blank on every deployment that can reach this code, because
1.6.x blanked it. Nothing observable changes.

**Reversing costs** one statement restored — and a red test: stage 1's `rotate_test.go` seeds
`AESEncryptionKeyLegacy` non-empty and asserts rotation leaves it alone, so the assertion would have
to be retired with it.

### 2. `reencryptPrivateKeys`'s plaintext-PEM branch stays, documented as unreachable

**The call.** Rotation picks the first non-empty `PrivateKeyPEM` as its canary and errors when it
decrypts under neither key, so a plaintext PEM fails closed before that branch can run. The branch was
kept and its comment now says it is unreachable and why, citing #262.

**Why.** Deleting it changes a crypto path for no observable gain; leaving it silent reads as live
code. `rotate_test.go` pins the fail-closed claim that makes it unreachable.

**Reversing costs** a few lines deleted, and re-deriving the fail-closed argument to show nothing
depended on them.

### 3. `database.mdx` was corrected, beyond the agreement's "Nothing else"

**The call.** §2 *Documentation owed* ends **"Nothing else."** and names four surfaces. The final
review found a fifth: `production-deployment/database.mdx` promised an upgrade "needs nothing beyond
installing the new version and restarting", which stage 1 made false. The page was edited
(`4985df66`) rather than escalated, per `roles/stage.md` 4.2 for a page the sweep missed.

**Why the sweep missed it.** §2's search was for the moving package names — `core/data`, `core/models`,
`core/auditlog` — and that page names none of them. The conclusion was sound for the move and blind to
the deletion.

**Reversing costs** one sentence, and an operator who reads the page instead of the release notes.

### 4. Three `## 0. Code anchors` rows were deleted, not repointed

**The call.** `startup/reencrypt`, `startup/otp-backfill` and `docs/env-reencrypt` named code that
decisions 4 and 7 deliberately delete. `check-anchors.sh` requires every row to resolve to exactly one
live line and has no marker for a row whose subject is gone on purpose, so the three rows were removed
in stage 3. §0 is 32 rows and the sweep passes.

**Reversing** is not meaningful — the code is gone — but deleting a row from a sealed section is the
kind of edit you are entitled to find an account of, so here it is.

leodip:

### Ready to file — 0

The one entry that stood here, `GetLastUserWithOTPState`, was **folded in instead** on the user's
instruction after this PR was opened, exactly as the entry itself offered. It is now the third item
under *Folded in* below, with what actually landed. Nothing else is left to file.

### Folded in rather than filed — 3

- **#262**, the two one-shot 1.5.x startup conversions. Deleted in `ff486cd9`. Folded in because both
  conversions sat on the interface and in the `commondb` files this move was carrying anyway; moving
  them and then deleting them two stages later is the same work twice. Decision 4 sets the policy that
  makes it safe: an upgrade must pass through 1.6.x. Closed by this PR.
- **#98**, the plaintext `users.otp_secret` column. Migration `000048` drops it on all four engines in
  the same commit, `ff486cd9`. Folded in because deleting `BackfillEncryptedOTPSecrets` takes item 4
  off #98's checklist, and #98 itself says to land them together or not let them drift. Decision 7.
  Closed by this PR.
- **`Database.GetLastUserWithOTPState`**, which had no production caller. Deleted in `e571ac34`:
  the interface entry, the `commondb` implementation, the four engine forwarders, 68 lines of the
  regenerated `Database` mock and the data test that pinned its query — 196 deletions, no addition.
  **Folded in rather than filed because the user asked for it here**, which is the one thing the
  entry above said would move it: three of the four fold-in tests always passed, and the fourth was
  a scope line, not a technical objection.

  Two things found while closing it that the drafted issue did not know. **No guard is added, and
  that is a finding rather than an omission**: a method-level version of `TestHandlers_NoDeadInterfaces`
  would not be green today, because **39 of the `Database` interface's 216 methods have no production
  caller**. The other 38 are the uniform per-model CRUD surface the data tier exercises — `GetXById`,
  `UpdateX`, `DeleteX` — which is a deliberate shape and not an accident, and separating the two
  classes is a judgement no walk makes. `GetLastUserWithOTPState` was a bespoke predicate written
  for a conversion `ff486cd9` deleted, which is why it is the only one leaving. And the regenerated
  mock is the evidence the deletion is exactly what it claims: mockery produced a diff of 68 removed
  lines and no other change.

  Tiers at `e571ac34`: `go vet` over all four modules with test files included; modules 7787 green;
  data **598** green on sqlite, four fewer than before and exactly the removed test with its three
  subtests; integration 1425 green on sqlite; lint green, which is the half that runs mockery itself
  and fails on a tree it changes.

### Bookkeeping — 1, and it is yours to run

- **#289**, `BackfillLowercaseEmails` scanning the users table at every boot. **No work here, and
  nothing to fold in: the code the issue is about does not exist.** #351 (commit `5130cd3a`) deleted
  `commondb.BackfillLowercaseEmails` and its file entirely, with its interface entry, its four engine
  wrappers, its mock and its three test files. What replaced it, `ScanEmailCase`, is reached only
  through `CheckEmailCaseBeforeMigrating`, which returns before scanning once the recorded schema
  version is at or above migration `000047` — so it runs at most once per upgrade and never again,
  which is the "forever" the issue title is about. All three of #289's options are about a startup pass
  that was removed. Decision 5.

```bash
gh issue close 289 --comment "Resolved by #351 (commit 5130cd3a), which deleted BackfillLowercaseEmails outright. Its replacement, ScanEmailCase, is gated by CheckEmailCaseBeforeMigrating on the recorded schema version, so it runs at most once per upgrade across migration 000047 and never on a later boot."
```

**None.** `git diff --unified=0 origin/main...HEAD | grep '^+.*ceiling:'` returns nothing — this branch
added no `ceiling:` marker.

leodip:
## Review account

This change was reviewed under `GBD_REVIEW_MODE=end`: no stage was reviewed on its own, and the whole
branch was read at a single final gate. This comment is that review. The verdicts live in the issue
folder, which is on one machine and pushed nowhere, so this is the only place the reasoning reaches a
reader of the PR.

Three rounds in total: one at the plan gate before any code existed, and two at the final gate over
the complete branch.

|---|---|---|---|---|---|
| plan | 1 | `openai-codex/gpt-5.6-sol`, effort `xhigh`, fresh session | 1 | 1 | confirmed, fixed in the plan before any code landed |
| final | 1 | same, fresh session | 1 | 1 | confirmed, fixed in `4985df66` |
| final | 2 | same, resumed session | 0 | 0 | clean |

Nothing was refuted, nothing was deferred, and no decision was raised by any round.

---

## Plan gate, round 1

**Request** `359-20260919T123315Z`, `gate: plan`, round 1 of a budget of 2. Verdict `findings`:
1 finding, all three axes — conformance, quality, security — `reviewed`. Reviewed `plan.md` against
`origin/main` at `a4e20b2a`, with no code landed and PR #382 holding one empty commit.

### Finding 1 — blocking / conformance — confirmed, fixed in the plan

> Stage 1 step 5 deletes `src/core/data/commondb/reencrypt.go`, which holds the surviving
> `RotateEncryptionKeyIfNeeded` and every helper it uses. The stage cannot compile, and hollowing the
> behaviour instead breaks the env-to-env rotation the agreement keeps.

**Confirmed, and one degree worse than reported.** `RotateEncryptionKeyIfNeeded` is the only
`CommonDatabase` rotation definition, it sat at `commondb/reencrypt.go:55` — inside the file the step
deleted — and at line 87 it **called** `d.ReencryptDataToNewKey`, the method the same stage removes
from the interface. The plan's own hedge, "`RotateEncryptionKeyIfNeeded` lives in its own file", named
a file that does not exist.

**Not escalated,** because it mechanically restores an invariant the agreement had already decided:
decision 4 says the rotation arm stays, so nothing was being chosen.

**Fixed in `plan.md`, not in code** — no code had landed at this gate. Step 5 was rewritten: the file
is refactored rather than deleted, `ReencryptDataToNewKey` becoming the unexported `reencryptToKey`
with its body and transaction wrapper intact and rotation's call following it; `aesProtectedColumns`,
`reencryptStringColumn` and `reencryptPrivateKeys` untouched; `otp_backfill.go` still deleted whole,
verified as declaring exactly one symbol.

**A second defect found while resolving it,** which the finding gestured at: deleting
`reencrypt_test.go` as written would have dropped the only coverage of four `aesProtectedColumns`
entries whose re-keying code survives — `settings.smtp_password_encrypted`,
`users.otp_secret_encrypted`, `users.email_verification_code_encrypted` and
`pre_registrations.verification_code_encrypted`. `rotate_test.go` reached only three of the eight. Step
13 was rewritten so `rotate_test.go` absorbs the fixture and becomes the exhaustive table over all
eight protected columns — two of which neither test reached before.

The two sub-decisions taken inside that repair are in the *Deferred decisions* comment on this PR.

### `unchecked`

Two entries, both dismissed with reasons: no production code or migration exists at a plan gate, so no
tier could run (stage 1 step 19 runs them); and the unaffected endpoint security classes were reviewed
statically because the plan changes no handler, route, token, session or validation path.

### Rounds

One round at this gate. The repair was forced rather than chosen, and the implementation carrying it is
reviewed at the final gate, so the second round was not spent.

---

## Final gate — tree-hash verification before round 1

`tree-hash.sh --verify`, run at `f35e2b7d` with a clean tree, before the round 1 request was written:

```
current   production c73d5f2c6c36   unit-tests c5637bdde5eb   data-tests 5d1956b8dc2b
          integration-tests cb2abcd8b324   docs 531c43f77eb9   all 134b3f4df6b1

stage 1  unit stale   data stale   integration stale   docs stale
stage 2  unit stale   data stale   integration stale   docs stale
stage 3  unit stale   data stale   integration stale   docs stale
stage 4  unit live    data live    integration live    docs live
```

Stages 1 to 3 recorded their tiers against trees that later stages moved, which is what a four-stage
move does by construction: stage 3 rewrote the import block of nearly every file stages 1 and 2 had
tested, and stage 4 did it again.

What that costs the record: the mutations those stages ran are the evidence their new tests can fail,
and those runs are no longer against this tree. **The claims themselves are re-established at HEAD by
the check suite**, green on all thirteen jobs including the four-engine data and integration tiers that
nothing local runs. So the tier results are current; the mutation evidence alone is stale, and no
stage's mutation targeted code a later stage rewrote in substance rather than in import path.

---

## Final gate, round 1

**Request** `359-20260919T150511Z`, the whole branch against `origin/main`. Verdict `findings`:
1 finding, all three axes `reviewed`. Tree at review: `production c73d5f2c6c36 / all 134b3f4df6b1`.
CI green at `f35e2b7d` on all 13 jobs.

### Finding 1 — blocking / conformance — confirmed, fixed in `4985df66`

**Claim.** `site/src/content/docs/production-deployment/database.mdx`, under *Schema migrations and
rollback*, told operators the auth server applies its own migrations "so an upgrade needs nothing
beyond installing the new version and restarting". Stage 1 made that false: with both 1.5.x startup
conversions deleted, a direct 1.5.x upgrade leaves the data key in `settings.aes_encryption_key` with
no reader, and migration `000048` then drops `users.otp_secret` and with it every seed that was never
encrypted. An operator following the page as written walks into both.

**Verified before acting.** The sentence was at `database.mdx:62`, unchanged by the four stages. Both
conditions it promises against are real: `git grep` finds no surviving `ReencryptDataToNewKey` or
`BackfillEncryptedOTPSecrets`, `runStartupDataTasks` carries only the `RotateEncryptionKeyIfNeeded`
arm, and all four `000048_drop_users_otp_secret.up.sql` drop the column.

**Why it was missed.** Section 2's documentation sweep searched for the moving package names —
`core/data`, `core/models`, `core/auditlog` — and this page names none of them, so its "Nothing else"
conclusion was sound for the move and blind to the deletion.

**Fix.** The reviewer's `forced_answer`, applied as given. The absolute clause is gone; no
version-transition section was added, so decision 8's control stays the release notes:

> so the schema needs no step of its own during an upgrade. Anything more a release asks of an
> upgrade, its release notes state.

The second sentence mirrors the rollback paragraph six lines below, which already reads "which its
release notes state", so the page points at the same channel in both directions and a reader has one
concrete action.

**Size.** The *Schema migrations and rollback* section went 260 → 272 words, +12, +5%. The replaced
clause was 12 words and its replacement is 24. `docs.md` section 3 caps this at "the smallest edit that
makes the page true"; no paragraph, section or upgrade heading was added.

**Deviation recorded with it.** Section 2's *Documentation owed* ends "**Nothing else.**" and names four
surfaces. This page is a fifth. Nothing in that list changed; one page was added to it.

### `unchecked`, answered

> "The actual PR body and its `## For the release notes` section were not inspected, being outside the
> checkout."

**Covered.** `gh pr view 382 --json body` shows the section present, headed **"Upgrading to this
version must pass through 1.6.x."**, with both destructive consequences spelled out, the
1.6.x-then-upgrade instruction, and an explicit statement that no pre-flight check exists. That is goal
item 9 and decision 8's deliverable, and it is satisfied.

### Tiers re-run after the fix

The change is prose in one `.mdx` file; no Go source, test, migration or golden moved. `where.sh build`
re-run for the discipline: the four modules compile, test files included. `git diff --check` clean. No
Go tier's result is disturbed; only the `docs` tree-hash category moved.

---

## Final gate, round 2 — clean

**Request** `359-20260919T152850Z`, resumed session, the whole branch against `origin/main`. Verdict
`clean`, **0 findings**. All three axes `reviewed`.

**Tree at ingest:** `production c73d5f2c6c36 / unit-tests c5637bdde5eb / data-tests 5d1956b8dc2b /
integration-tests cb2abcd8b324 / docs 2db50f52be05 / all 6c8ef5913a7f`. `production` is the **same value
round 1 was reviewed at**, which is the checkable form of the reviewer's own note that round 1's repair
touched documentation and nothing else. Only `docs` moved between the two rounds.

### Findings

None. Nothing to confirm, fix, defer or refute.

### What the reviewer executed, so the absence of findings has a stated scope

Recorded because zero findings in a narrow scope reads identically to zero findings in a wide one. Its
`ran` list:

- `review-diff.sh f35e2b7d` and `git diff --check` over the complete round 1 repair commit.
- `tree-hash.sh --verify`, confirming only `docs` moved and that production plus every test surface
  still carry the hashes they were reviewed at.
- `where.sh ci HEAD`, all 13 jobs green.
- A tracked-documentation sweep over the terms the deletion half makes load-bearing: upgrade
  prerequisites, release notes, automatic re-encryption, plaintext OTP, `otp_secret`, `1.5.x`, and
  intermediate-version language — the sweep whose absence produced round 1's finding, now run against
  the deletion's vocabulary rather than the move's package names.
- An exact census of obsolete import paths and of the mockery generator configuration.
- The architecture ownership, exception and `core/constants` documentation diff.
- A census of persisted encoding, reflection, runtime type identity, `embed` directives and the
  generated mock across the relocated packages — the four ways a package move can change behaviour
  while still compiling.

That last item is the one worth noting: it is the reviewer testing the design's central claim, that the
move is a rename, by the mechanisms that could falsify it, rather than re-running suites the request
already reported green.

### `unchecked`, answered — covered by running the build

> "A full Astro site build was not run. `site/node_modules` is absent, and a scratch `npm ci --offline`
> could not obtain the uncached `svgo` package. The changed lines are plain Markdown prose with no MDX
> syntax or structural markup."

Covered by **running the build** rather than by accepting the reasoning. It matters more than it looks:
`docs.yml` triggers on push to `main`, on release and on dispatch, never on a pull request, so nothing
on this branch builds the site and a broken page would first be discovered by the merge to `main` that
publishes it.

```
cd site && npm ci --no-audit --no-fund && npm run build
```

**43 pages built in 4.39s, complete, no error.** Both pages this branch edits rendered:
`/production-deployment/database/index.html` and `/reference/environment-variables/index.html`. The
Pagefind search index rebuilt over all 43 files. The single `[WARN]` is `@astrojs/sitemap` asking for
the `site` option in `astro.config.mjs`; it is pre-existing, unrelated to this branch, and present on
`origin/main` for the same reason.

The mechanical checks behind the reviewer's own reasoning hold too, and were run before the build: no
`<`, `>`, `{` or `}` in any added line, so no JSX or expression syntax was introduced; no heading added
or removed, so no anchor another page links to has moved; no frontmatter touched. `site/node_modules`
and `site/dist` are both gitignored, and `git status --porcelain` is empty after the build, so the tree
the review approved is the tree that ships.

### Why this is the last round

The budget is two unconditional rounds, extended only when the previous round changed production code.
Round 2 reported nothing and changed nothing outside the record, so no round 3 is funded — and that is
the budget working rather than a cap biting: round 1 found a real conformance defect and was answered,
round 2 read the whole branch again across all three axes and found the answer sound.

Deliberately **not** done while closing: no tidying pass over the tree. A clean verdict followed by a
session moving bytes anyway is an unproductive round, and the only cosmetic worth mentioning — one
short wrapped line at `database.mdx:65`, left by round 1's edit — is not worth moving a reviewed tree
for. It is noted here instead so a later editor of that paragraph knows it was seen and left.

Issue #334, opened 2026-09-12
[Core refactor 3/29] Move session-store adapters to their owning applications


leodip:
## Goal

Keep the reusable encrypted session store in core while moving its database and HTTP backend adapters to authserver and adminconsole respectively.

## Rationale

Core/sessionstore combines three responsibilities: shared session encoding/cookie mechanics, an auth-server database adapter, and an admin-console HTTP adapter. The adapters encode application topology and trust boundaries, so co-locating them causes each process to compile dependencies owned by the other. The auth server should own direct browser-session persistence; the admin console should own its authenticated HTTP bridge.

## Implementation guidance

1. Keep Store, Regenerator, Backend, Record, Session, codec/key handling, cookie behavior, expiry rules, and common size/error limits in core/sessionstore.
2. Move backend_db.go and hashSessionId to a focused authserver internal package and update both auth-server construction sites.
3. Move backend_http.go and its HTTP-specific TokenSource boundary to a focused adminconsole internal package.
4. Keep the HTTP request/response types in core/api as stable wire contracts.
5. Preserve clock and HTTP-client injection seams.
6. Move adapter tests to their owner; keep codec and store contract tests in core.
7. Add a compatibility test that writes with the pre-move logical representation and reads after the move. Preserve cookie names, serialized field names, encrypted bytes, TTL behavior, hashing, and error mapping.

## Compatibility and non-goals

- This is a structural refactor only. Do not change business rules, OAuth/OIDC behavior, HTTP routes, status codes, response bodies, persistence behavior, configuration names, logging semantics, or security policy.
- Preserve public serialized representations, including JSON fields, session payloads, cookie names, token claims, database values, and error classification.
- Move or regenerate tests and mocks with the code they cover. Do not weaken assertions to make the move pass.
- Avoid temporary forwarding aliases unless a concrete compatibility consumer requires one; any temporary exception must name its removal issue.
- Keep every commit buildable where practical and finish with gofmt and tidy only in the modules whose imports changed.

## Acceptance criteria

- Core/sessionstore has no imports on core/data, core/models, or the admin-console session API.
- Authserver constructs the DB backend locally; adminconsole constructs the HTTP backend locally.
- Existing encrypted sessions remain readable across deployment of the refactor.

- The unit tier passes for core, authserver, and adminconsole.
- The repository-wide gofmt and error/logging convention guards pass.
- Relevant race, data, and integration tests continue to pass.
[...]

Issue #226, opened 2026-08-16
utils.js is duplicated between authserver and adminconsole, and the authserver copy is unused


leodip:
`src/authserver/web/static/utils.js` and `src/adminconsole/web/static/utils.js` were byte-for-byte identical before #224. That change edited only the admin console copy, because that is where the sink it was closing lives, and the two files now differ.

The auth server copy is dead in the part that matters: `grep -rln sendAjaxRequest src/authserver/web/` matches `static/utils.js` and no template, so nothing in the auth server reaches `sendAjaxRequest` or, through it, `showModalDialog`'s `innerHTML` assignment. There is no vulnerability here today; the problem is that the duplication is silent and the next person to fix one copy has no signal that a second exists.

**Closing this would take** deciding which of two shapes is wanted: delete the unused functions from the auth server copy so each module carries only what it serves, or extract the shared helpers so one file feeds both and a fix cannot land in only one place. The first is smaller and matches the fact that the two modules' pages have diverged; the second is what stops it happening again.

Worth doing because the failure mode is a fix that looks complete and is not. #224's escaping edit is the worked example: it is correct for every page that can reach the sink, and a reader diffing the two files afterwards cannot tell whether the auth server was considered and excluded or simply missed.

## What each copy actually uses

Measured at `9f7677c3`, per function, against `src/authserver/web/template/`:

| Function | authserver templates using it | Notes |
|---|---|---|
| `t` | 32 | live |
| `showModalDialog` | 2 | `forgot_password.html`, `reset_password.html`, both passing literal localized markup only, so no server-supplied string reaches the `innerHTML` assignment |
| `tFormat` | 0 | unused |
| `getTrashCanMarkup` | 0 | unused |
| `getEditMarkup` | 0 | unused |
| `sendAjaxRequest` | 0 | unused, and the reason #224's edit did not need to be made twice |

So the "delete the unused half" option is four functions, and it leaves the auth server copy at `t` plus `showModalDialog`. That confirms the divergence is real rather than accidental: the two modules' pages genuinely need different things, which is an argument for the smaller option and against a shared file, unless the goal is specifically to make a one-sided fix impossible.

Note that the two modules are separate Go modules and each embeds its own `web/static`, so the "extract the shared helpers" option is not just a file move: it needs the shared source to live somewhere both embeds can reach.

Related: #120 covers the admin console's `innerHTML` sinks in templates. This issue is about the two static files diverging, which is next to that inventory rather than part of it.

[...]

PR #383, opened 2026-09-19
refactor(core): move the remaining authserver-only utilities and close the core refactor

### LINKED ISSUE #360: [Core refactor 16/16] Move remaining authserver-only utilities and close the core refactor

leodip:
Closes #360. Absorbs #187 and #226, both closed pointing here.

This is the last of the 16-issue Core Refactor epic. Every other issue in it is closed, so this is
the closer the epic's own framing describes: move what is left, burn the exception table down to
zero, and prove the result rather than assert it.

## What this establishes

**`core` becomes a shared kernel by construction, not by intention.** Six packages leave it for the
auth server, one splits, and `ARCHITECTURE.md`'s temporary-exception table reaches zero rows. Every
surviving ownership row reads `kernel` with no issue attached, and the architecture guard — which
runs from all three module unit tiers and fails in both directions — is what holds it there.

**Concretely: bcrypt stops being compiled into the admin console.** `core/hashutil` held two
unrelated things. The SHA-256 half is called by two kernel packages and reached by both processes,
so it stays. The bcrypt half had no caller outside the auth server, and because the two shipped in
one package, `golang.org/x/crypto/bcrypt` and `golang.org/x/crypto/blowfish` were in the admin
console's production import closure — a password hasher linked into a web UI that never hashes a
password. Nobody chose that; it arrived bundled with a SHA-256 helper. Splitting the package removes
it, and a new `arch/foreign` row naming the *package* (not the module, which the admin console
legitimately compiles for `chacha20poly1305`) keeps it out, so the removal is enforced rather than
measured once.

## How it is kept honest

A move is only safe if it moved nothing else. That is checkable rather than promised: every moved
file must be byte-identical to its old self once the package clause and the import block are
stripped, and a residue probe checks exactly that, per commit, where each rename actually happened.
Over this change's own commits it reports residue for exactly two files — the two halves of the
`hashutil` split — and every other move pure. The same probe is turned on the epic's 127 commits to
answer the issue's "review the complete series diff for accidental logic changes", which is
otherwise an unbounded read of 1324 files. The probe itself is not part of the deliverable and is
not in the tree; stage 7 says why, and the numbers it produced are recorded below.

No behaviour changes, with one recorded exception noted under stage 5.

## Status

**Complete. All seven stages have landed, the check suite is green, and the change has been
reviewed as a whole.**

The exception table emptied at stage 4, so the epic's burn-down is finished; the dead JavaScript
went with stage 5; and stage 6 swept the whole sixteen-issue series for accidental logic changes and
found none; stage 7 then removed the sweep's instrument, which was never part of the deliverable.
`where.sh ci` on `ba278912` — the last commit carrying a code change — reports all 13 checks passing
— [run 35482348074](https://github.com/leodip/goiabada/actions/runs/35482348074) — and the final
review came back clean with zero findings.

### Stage 1: record the constraints — the CI citations and the bcrypt guard row
`7a6e5486`. The repository's own records, made true and enforceable before any code moves.

All 13 dangling references to `docs/ci-v2-design.md` are gone across 8 files — the 10 literal paths
and the 3 bare section numbers a reader can follow even less far. The document was untracked
deliberately in `8511355a` and that reasoning still holds, so the pointers are struck rather than
the file restored. All seven cited sections (4.1, 4.5, 4.6, 4.9, 5.4, 5.8, 5.9) were read first
from the recovered copy to see whether any held something its citing comment does not already
state: none does. Every site already carried its whole constraint and deferred only the supporting
detail behind it.

`ARCHITECTURE.md`'s foreign-modules table gains `golang.org/x/crypto/bcrypt` at `yes | #360`, so
stage 2's removal of bcrypt from the admin console's import closure is enforced rather than
measured once. The row names a package rather than a module because the admin console legitimately
compiles `golang.org/x/crypto` for `chacha20poly1305`; one sentence above the table says so. It is
exercised on the tier that landed it — the guard's row-flipping test runs a subtest per row, which
is why the unit count moved 7787 → 7788, and it now asserts that something does reach bcrypt today.
That something is `core/hashutil`.

Tiers: unit green, 7788 ok 0 fail; build green across the four modules. Data and integration are
not applicable — no Go file, handler, route or schema is touched.

### Stage 2: split `core/hashutil`, and take bcrypt out of the admin console
Commit `384e6af3`.

`core/hashutil` held two unrelated things, and only one of them was shared. The SHA-256 pair,
`HashString` and `VerifyStringHash`, is reached independently by both processes — the auth server
hashes authorization, verification and reset codes to locate rows; the admin console hashes the
nonce it sends and re-checks the `nonce` claim, through `core/handlerhelpers` and `core/oauth` — so
it stays in `core` and its ownership row now reads `kernel` with no issue. The bcrypt half had no
caller outside the auth server and is now `authserver/internal/passwordhash`: `Hash`, `Verify` and
`DummyHash`. Of the 62 files importing `core/hashutil`, 43 repoint, 11 are untouched and 8 carry
both imports with no alias.

**`golang.org/x/crypto/bcrypt` and `blowfish` have left the admin console's binary.** That is the
harm the epic exists to remove, stated in #360 point 5 as "provider-only crypto dependencies ...
pulled into adminconsole through core", and it is measured rather than argued: `go list -deps` on
`./cmd/goiabada-adminconsole` now finds zero packages from either. The `ARCHITECTURE.md` foreign row
flips to `no` and is kept rather than deleted, like the five database-driver rows, because the point
is that it must not come back — without it the split would be a one-time measurement and any later
`core` package could reintroduce bcrypt with nothing going red. Both `core/hashutil`
temporary-exception rows go in the same commit, leaving one row in that table.

The three declarations moved verbatim, the pre-computed dummy hash copied character for character.
Verified rather than asserted, since a split is the one move the residue probe cannot pair:
the pre-split file was reconstructed with the three renames applied and diffed against both halves,
production and test, and all four reconstructions came back identical. No caller changed an
argument, a result or its control flow.

Tiers: unit green, 7786 ok 0 fail; build green across the four modules with test files included.
The count is two below stage 1's, and both are accounted for — the architecture guard runs one
subtest per exception row, dropping each in turn to check the guard still names the edge it covered,
and two rows went. Two mutations caught: `Verify`'s comparison flipped fails `TestVerify`, and the
bcrypt row flipped back to `yes` fails `TestArchitectureBoundaryHolds` with "nothing reaches it any
more".

### Stage 3: move `encryption`, `oidc` and `rsautil` to the auth server
Commits `d53bf3c1` (the move) and `e2480c13` (the package docs).

Three whole-package moves to `src/authserver/internal/`, 13 files with their tests, and 89 importer
files repointed. None of the three has a consumer outside the auth server, production or test: the
admin console and the setup module import none of them, and no core package does either. `oidc`
moves whole, `WellKnownConfig` with it — the discovery document is the auth server's published
contract with arbitrary OIDC clients rather than a contract between these two processes, so the
process that publishes it should own it.

The move is checkable rather than argued, and the check is the point of the stage.
the residue probe over both commits reports **13 renames, 13 pure, 0 with residue**, and
**0 of the 13 changed its import set**, so none of them is pure only because the probe strips
import blocks. A second detection pass at `-M10%` finds no rename the default threshold missed.
The three packages import only the standard library, `core/errs` and testify, so no `go.mod` moves
and nothing third-party leaves `core`.

The three `ARCHITECTURE.md` ownership rows go in the same commit as the code, because the guard
fails in both directions: a row for a package with no production Go file is a finding, and so is a
core package with no row. 21 rows remain, two of them still `authserver | #360` for stage 4.

Two commits rather than one, deliberately: a package doc sits above the package clause, and the
residue probe strips only the clause and the import block, so landing the prose with the move would
have made all three files read as residue and cost the move the one property that makes it safe to
wave through.

Tiers: build green (four modules, test files included), unit green at **7786 ok / 0 fail**,
unchanged from the count stage 2 left. Integration not applicable — no handler, route, template or
schema is touched, and a package's address is not something an endpoint can observe.

### Stage 4: move `urlutil`, `uuidutil` and `testutil/fake`, and empty the exception table
Commit `1f28f874`.

The last three authserver-only packages leave `core`, in one commit because they cannot be
separated: `core/testutil/fake` imports `core/uuidutil`, so the moment either sits under
`authserver/internal/` while the other stays behind, the core module stops compiling — the import
crosses a module edge into an internal package `core/go.mod` does not even require. `fake` is the
auth server's fixture package by every measure, with 165 importers, every one an authserver test
file and no adminconsole file among them, so moving it is what keeps `fake.UUID` drawing from the
generator production uses. Eight files moved, 187 had their imports repointed.

**`ARCHITECTURE.md`'s temporary-exception table now has no rows**, which is what this whole change
was for. The ownership table is down to 19 rows, every one `kernel` with no issue, and five prose
passages that asserted a burn-down still in flight are corrected. The file is the same 355 lines it
was.

the residue probe over the commit reports **8 moves, 8 pure, 0 with residue**, and 0 renames
missed at the `-M10%` second pass. Two of the eight changed their import set, both inside `fake`,
both repointing at companion packages that moved in the same commit; the probe lists those
separately rather than letting the stripped import block hide them.

Two things the plan did not foresee, both in the commit message in full. The guard's own
`TestArchitecture_TheRealTreeIsHeldByItsExceptions` required the real exception table to be
non-empty and then proved the guard was wired to this repository by dropping each row; at zero rows
it would have passed by iterating nothing, so it now holds the other direction — adding back the
exact row this commit deleted must be reported as a row to delete, against the real import graph.
And six code comments pointed at the old paths by name, repointed rather than left as traps.

Tiers: unit 7786 ok 0 fail, unit `--race` the same, lint clean with CSS and mocks unchanged, and
the **four-engine data tier** green at 2446 ok 0 fail on mysql, postgres, mssql and sqlite. The new
case is mutation-tested: inverting `reconcileExceptions`'s stale-row branch is caught.

### Stage 5: delete the dead JavaScript and the JSBootstrap chain

`b939243a`. The auth server's `utils.js` shipped seven functions and its templates called one. `t`
and `tFormat` were reached only from inside `sendAjaxRequest`, which nothing called; `debounce`,
`getTrashCanMarkup` and `getEditMarkup` had no caller at all. The dead set is therefore six, not the
four #226 predicted, and `showModalDialog` survives alone — `forgot_password.html` and
`reset_password.html` call it with strings the server has already localized through the Go `T`.

Both auth server layouts also lose `{{ JSBootstrap $.ctx }}`. That chain was dead end to end before
this change: the block populates `window.i18n`, only `t` read it, and eleven of its nineteen keys
are for an `image-upload.js` this server does not serve. `JSBootstrap` itself stays in
`core/handlerhelpers` — the admin console calls it from both its layouts and every one of the
nineteen keys keeps a consumer there, so neither catalog changes.

Two new guards in `authserver/web`, reading the byte streams the router serves rather than the files
on disk, because no JavaScript runs in any tier here and the auth server has no render test package:
every function a served script declares must be reachable from a call a served template makes,
transitively; and a layout emits `JSBootstrap` if and only if a served script reads `window.i18n`.
The first is reachability rather than a list of forbidden names because every name deleted here was
referenced somewhere — a denylist reads `t` as live, which is how it survived #226. The second is
stated as an implication, so a future page needing client-side strings adds both halves and passes,
while either half alone fails.

**The recorded exception to "no behaviour changes" is here, and it is two things rather than one.**
Every auth server page stops emitting a `<script>window.i18n={...}</script>` block. And the
regenerated `main.css` drops `h-5`, `w-5` and `align-middle`, which appeared only in the SVG markup
inside the two deleted helpers — Tailwind scans the served JavaScript, so this is a consequence of
the deletion rather than a choice, and the committed file must match what the generator writes.
Nothing reads either, and no test pinned either.

Tiers: build green; unit 7788 ok 0 fail; integration green on sqlite at 1425 ok 0 fail, which
decision 10 puts on this stage because it is the only tier that reaches a rendered auth server page;
lint green including the Tailwind and mock regeneration checks. Four mutations against the two new
guards, all four caught — including restoring `t` together with the dead `sendAjaxRequest` that
calls it, which a denylist would have admitted — plus a hand-run negative control confirming that
both halves of the `JSBootstrap` rule pass together.

### Stage 6: the series sweep, and the closing record

Commit `ba278912`. Point 7's acceptance criterion — review the complete series diff for accidental
logic changes — answered with an instrument rather than a reading. The probe, `move-residue.sh`,
walks the series commit by commit, finds each Go rename where it happened, and compares `parent:old`
against `commit:new` with the package clause and the import block stripped. What survives is content
the moving commit changed as well, which is the only thing "accidental logic change" can mean. The
base-to-head diff was rejected as the instrument: at 83 files and 5698 lines it is not merely large
but wrong, because it blames each move for every later commit that touched the file.

**Over the sixteen-issue series: 325 moves, 287 pure, 38 with residue — and no defect.** All 38 were
read. The 11 production non-mock files (199 lines) and the 17 test files (~1269 lines) in full; the
10 generated mocks (514 lines) skipped, because mockery writes them and `--type lint` regenerates
and fails on any difference, so a hand edit there is already caught by a check that runs. Every
residue turned out to be a package qualification, a symbol rename, a doc comment that had stopped
being true, a relocation inside the same commit, or a design change explained in the code with its
issue number. The two that are real design changes are the session backend's owner-fixed
constructors (#334) and the rate limiter answering through `apiresponse.WriteError`, which encodes
the same `api.ErrorResponse` at the same status with the same `Content-Type`. Two test files gained
cases rather than losing them, and the one test with no successor,
`TestSessionKeyValidatorsAreIndependent`, became unexpressible when the config split gave each
module its own struct — its property is structural now.

**Over this change's own commits: 21 moves, 21 pure, 0 with residue.** That is §4's claim for these
moves, measured rather than asserted. The `hashutil` split is invisible to a rename walk by
construction — a split leaves the source `M` and the destination `A`, so there is no rename to read
— and was verified directly in stage 2 by reconstructing all four resulting files and diffing them
against disk.

The script reports its two exclusion boundaries rather than hiding them: every import delta is
printed, and a second pass at `-M10%` reports any rename the default similarity threshold missed.
Both lists were read too — the 7 moves that are pure only because imports are stripped are all one
companion package moving in the same commit, and of the 2 unpaired renames one is the genuine
`core/config` split (checked separately: 26 functions in, 26 out across the two destinations) and
the other is not a move at all. The script was wired into no tier, and said so in its own header:
it reviews history, which is fixed. Stage 7 draws the conclusion that follows from that.

Closing checks: `go mod tidy -diff` clean in all four modules, and the admin console's production
import closure re-measured a final time — `bcrypt` and `blowfish` absent, all seven foreign-module
rows still absent, `chacha20poly1305` still present for `core/sessionstore/codec.go`. Build green;
unit green, 7788 ok 0 fail.

### Stage 7: remove the sweep's instrument

Commit `05893b52`. Stage 6 landed `scripts/move-residue.sh` alongside its result. On reflection the
script is the means and not the deliverable, so it goes and the result stays.

The reasoning is the script's own, followed one step further. It reviews git history, which is
fixed, so it cannot produce a different answer than the one stage 6 already wrote down — which is
why it was wired into no tier in the first place. A sweep whose output is a constant does not need
its instrument kept on disk to stay true, and every number it produced is recorded in `ba278912`'s
message and in this description, including both exclusion boundaries and what reading them found.

Nothing referenced it — no tier, no workflow, no document, no sibling script — so this is a deletion
and not a burn-down, and no pointer is left dangling. That last part is the standard stage 1 of this
same change set for itself when it struck the 13 references to `docs/ci-v2-design.md` rather than
restore the file. `scripts/` keeps `i18n/lint-error-codes.sh`, which is a different kind of thing:
it reads live paths under `src/core/i18n`, so it can go out of date and still has something to say.

No tier is affected and none was run. The file is not Go, sits under no module, is reached by no
test and is executed by no workflow; the tree-walking guards all walk `src/`. 115 lines deleted, no
other file touched.

## How it was verified

**Locally, in the dev container.** Every stage ran `--type modules` and `where.sh build`. Stage 5,
which lands the layout and JavaScript deletion, also ran `--type integration --db sqlite`, because
the auth server has no render test package and integration is the only tier that reaches its
rendered pages. Stage 4, which finishes the moves, also ran `--type lint`, `--type modules --race`
and `--type data --db all`, the four-engine data tier. The last of those is a deliberate departure
from `testing.md`, which scopes the four-engine tier to a stage touching `src/core/data/` or a
migration — this change touches neither. It was run anyway, by the user's decision, because this is
the epic's closing issue with 325 file moves behind it. Cost, not risk: about seven minutes.

**In the check suite**, on `ba278912` — the last commit carrying a code change, stage 7 being a
lone deletion of an unreferenced shell script — green on all 13 checks: Build validation;
`Tests / sqlite`, `mysql`, `postgres`, `mssql`; `Unit / race`; `Unit / core`, `authserver`,
`adminconsole`, `setup`; Vulnerabilities; Lint; Versions. That is where the four-engine integration
matrix, the schema golden comparison, the binary builds and the container and image checks report;
no local run in this branch covered them.

**Deviations from the agreement: none.** Every stage built what its steps specified, and no decision
was reopened, deferred or escalated. Two things were deliberately *not* read and say so where they
are claimed: the ten generated mock residues in the series sweep, because mockery writes them and
the lint tier regenerates and fails on any difference; and no interactive browser pass, answered in
the review comment from the served bytes instead.

### Review

Reviewed by `openai-codex/gpt-5.6-sol` at effort `xhigh`, 1 round over the whole branch. No stage
was reviewed on its own.

|---|---|---|---|
| 1 | 0 | 0 | clean |

No finding, blocking or otherwise, was raised. The round is worth more than that sentence suggests,
because its evidence is program-writing rather than suite-re-running: an independent normalizer over
all 259 changed Go files that reverses the expected path and symbol rewrites and diffs the result
against `origin/main` — a second implementation of this change's byte-identity argument, reaching
the same answer — plus **four mutations, all caught and restored**: a dead served JavaScript
function, a lone restored `JSBootstrap` emitter, a flipped bcrypt reachability row, and inverted
password verification. Those are the four seams this change adds, each shown to fail when its
property is broken. Two `unchecked` entries were answered rather than left open.

Full account, with the evidence and the tree hashes, in
[this comment](https://github.com/leodip/goiabada/pull/383#issuecomment-5747015727).

### Also in this thread

- [Deferred decisions](https://github.com/leodip/goiabada/pull/383#issuecomment-5747010803) — none;
  all twelve settled before any code was written.
- [Follow-ups, fold-ins and ceilings](https://github.com/leodip/goiabada/pull/383#issuecomment-5747012987)
  — no issue is owed; four things were annexed rather than filed; no `ceiling:` marker added.


leodip:
## Deferred decisions

**None.** This run deferred nothing and took no assumption that a reader has to accept on trust.

All twelve decisions in the agreement were settled during the interview, before any code was
written, and none was reopened by a stage. The guard reports `12 total, 0 open`. No stage escalated,
so the pull request thread carries no unanswered question and nothing here needs reversing.

The one assumption worth naming, because it is recorded in the agreement's section 1 rather than in
a decision, is about the epic checklist and not about the code: the issue body still shows #339,
#344, #346, #350, #351, #353, #354 and #359 unticked, while `gh issue list` shows all of them
closed. Treated as a stale checklist rather than as work undone. Reversing costs nothing in this
branch — it is a statement about the issue body, which this change does not edit.

Two decisions are worth repeating here because they are the ones a reviewer would most reasonably
have expected to go the other way, and both are already argued in full in the agreement:

- **Decision 1 — `core/hashutil` splits rather than moving whole.** The issue and the old ownership
  table both said the package moves out of `core` entirely. The census refutes that: the SHA-256
  half has two production kernel callers, both reached by the admin console. Only the bcrypt half
  moved. Reversing means either shipping bcrypt in the admin console's binary again or spelling
  SHA-256 twice.
- **Decision 10 — the four-engine data tier was run locally.** The user's call against the
  recommendation in `testing.md`, which scopes that tier to a stage touching `src/core/data/` or a
  migration; this change touches neither. Taken anyway because this is the epic's closing issue.
  Cost, not a risk: about seven minutes.

leodip:

### Follow-ups to file

**None.** No issue is owed by this change.

The interview settled all twelve decisions without deferring work, and the two adjacent problems it
found were folded in rather than filed (below). The stage 6 series sweep — the acceptance
criterion asking that 127 commits of moves be reviewed for accidental logic change — read 38 residue
files at the commit that moved each and found no defect, so it drafted nothing either. Nothing was
refuted or deferred at the final review, which came back clean. The duplicate search that would
normally run here has no query to run.

Four things were annexed into this change instead of becoming issues. Each is argued in the
agreement; the one-line reason is why it fit here.

- **#187, the dangling `docs/ci-v2-design.md` citations** — `7a6e5486`. The issue was closed
  pointing at #360, and the citation sites are point 6's own file set, so they were already open.
  13 sites across 8 files, not the 11 the issue predicted: 10 literal paths plus 3 bare section
  numbers a reader can follow even less far. Each of the seven cited sections was read from the
  recovered document before its pointer was struck; none held a constraint the surrounding comment
  did not already state in full. Items 2 and 3 of #187 were already fixed in the tree and are
  recorded as such rather than re-done.
- **#226, the dead client-side JavaScript** — `b939243a`. Also closed pointing here, and it is
  point 7's "remove obsolete aliases, mocks and directories" in a different language. The census
  corrected the issue in both directions: the dead set is **six** functions, not four, and `t`
  survives on zero auth server templates rather than 32 — what those templates call is the Go
  template function `T`, server-side. `showModalDialog` alone survives.
- **Decision 7, the `JSBootstrap` chain** — `b939243a`, same commit, because the two are one
  property. Both auth server layouts lost their `{{ JSBootstrap $.ctx }}` line. The chain was dead
  end to end *before* this change: the emitted `window.i18n` table is read only by `t`, which is
  called only from inside the already-dead `sendAjaxRequest`, and 11 of its 19 keys are served by an
  `image-upload.js` the auth server does not have. Deleting the functions and keeping the layout
  lines would have left every page emitting a table no served script mentions — harder to read than
  today. **This is the one edit in the change that alters a response body:** every auth server page
  stops emitting a `<script>` tag. Nothing reads it and no test pinned it, but it is a real
  difference and is stated in the PR body too. `JSBootstrap` itself stays in `core/handlerhelpers`
  for the admin console, and neither catalog changed.
- **Decision 12, the bcrypt foreign-module row** — added `yes | #360` in `7a6e5486`, flipped to
  `no` by `384e6af3`, the commit that landed the split. Without it, taking bcrypt out of the admin
  console's binary would have been a one-time measurement: any later `core` package the console
  imports could put it straight back with nothing going red, which is exactly how it arrived. The
  row names a *package* rather than a module, which is new for that table and needed, because the
  admin console legitimately compiles `golang.org/x/crypto` for `chacha20poly1305`.

Nothing else was found next to the work. The sweeps that looked are recorded in the agreement's
section 1: no diagram of any kind exists in the tree, so point 6's "diagrams" has no referent, and
no non-Go file outside `ARCHITECTURE.md` names any of the six moved packages.

**None.** `git diff --unified=0 origin/main...HEAD | grep '^+.*ceiling:'` returns nothing: this
branch added no `ceiling:` marker.

leodip:
## The review account

This change was reviewed once, over the whole branch, at the final gate. Under end-of-run review
mode that is the only reading the code got, so the account is pasted here in full rather than
pointed at: the verdicts live in the issue folder, which is on one machine and pushed nowhere.

### Round 1 — final gate, whole change

- **Reviewer:** `openai-codex/gpt-5.6-sol`, effort `xhigh`, fresh session. Request
  `360-20260920T015721Z`, gate `final`, round 1 of a 5-round budget.
- **Verdict: clean.** 0 findings. All three axes read `reviewed` (conformance, quality, security).
  `unchecked` held two entries, both answered below.
- **Tree at ingest**, unmoved since the request was written, so the verdict was produced against
  exactly the tree that ships: production `ba742440a098`, unit-tests `fe6b90164118`, data-tests
  `2f7a0eb58d63`, integration-tests `cc6e81ecae95`, docs `ca8965dfaece`, all `c3830de4a6d8`.

**What the reviewer ran.** Recorded because zero findings only means something with a stated scope:
read the agreement, plan, closing log and the stage 5 and 6 accounts; `review-diff.sh origin/main`,
diff statistics and hunk classification, `git diff --check`, `check-anchors.sh`, `tree-hash.sh`;
`where.sh ci HEAD`, confirming all 13 checks green at `ba278912`; repository greps for old and new
package paths, bcrypt symbols, package fanout, CI citations, JavaScript callers, `JSBootstrap`
emitters and readers, CSS class residue, docs, catalogs and security-sensitive changed lines; a
scratch normalizer over the 259 changed Go files that reverses the expected path and symbol
rewrites and diffs the result against `origin/main`; `move-residue.sh` over the whole history and
over #360's own 21 moves, reconstructing and reading the historical production and non-generated
test residues; `go list -deps` for the admin console in the dev container, confirming bcrypt and
blowfish absent; `go mod tidy -diff` clean in all four modules; targeted current-tree tests for the
embedded JavaScript and layout guards, the architecture and foreign-module guards, and
`passwordhash`; and **four `where.sh` mutations, all caught and restored** — adding a dead served
JavaScript function, restoring a lone `JSBootstrap` emitter, flipping the bcrypt reachability row,
and inverting password verification.

That last line is the part worth keeping. The four seams this change added are the only new claims
it makes, and each was shown to fail when its property is broken. The normalizer is the independent
version of the design's byte-identity property — a second implementation of the run's own
`move-residue.sh` argument, reaching the same answer.

**`unchecked`, both entries answered:**

1. *"No interactive browser pass was performed. The rendered-layout change was checked through
   embedded-file source, targeted guards, mutations, and green current-HEAD integration CI."*
   **Covered, in the ingest session.** The only browser-observable surface this change touches is
   the two pages that call the one surviving function. Read back from the served bytes:
   `static/utils.js` now defines `showModalDialog` alone, and its body references DOM APIs only — no
   `t`, no `tFormat`, nothing from the deleted set, so the survivor cannot have been broken by the
   deletions. Its two callers are `forgot_password.html:6` and `reset_password.html:7`, exactly the
   census's two, and both pass strings already localized server-side by the Go `{{ T $.ctx ... }}`,
   so no client-side catalog is owed. No file under `src/authserver/web/` calls any of the six
   deleted functions, and no served file emits or reads `window.i18n` — the only remaining mentions
   anywhere in the auth server are inside `static_lint_test.go`, which is the seam stating the rule.
   A browser would add the rendering of an unchanged modal; it would not reach a property the
   embedded-FS guards do not already hold.
2. *"The ten historical generated-mock residue hunks were not read line by line."*
   **Dismissed, and it was decision 8's call rather than the reviewer's omission.** The stage
   skipped those 514 lines deliberately: mockery writes them from `versions.yaml`'s pin, and
   `./run-tests.sh --type lint` regenerates and fails on any difference, so a hand edit there is
   already caught by a check that runs rather than by a reading. Lint passed at `ba278912`.

**Nothing was refuted**, there being nothing to refute, and nothing was deferred. No follow-up came
out of this round.

**Why round 2 was not spent.** The verdict meets the definition of clean on every limb — `findings`
empty, three axes `reviewed`, and `unchecked` holding nothing the gate depended on once the two
entries above are answered — and the round's evidence is program-writing rather than
suite-re-running: four mutations and an independent normalizer over the whole diff. A second round
would re-read a tree nothing has changed since.

### What the request itself recorded, and what it costs

The request went out against `ba278912` with CI green on all 13 checks. `tree-hash.sh --verify` at
that moment:

```
current   production ba742440a098   unit-tests fe6b90164118   data-tests 2f7a0eb58d63
          integration-tests cc6e81ecae95   docs ca8965dfaece   all c3830de4a6d8

stage 1  unit ?              data ?              integration ?       docs stale
stage 2  unit stale          data ?              integration ?       docs -
stage 3  only `all` recorded, 026dd9e7c54a, now c3830de4a6d8: moved, and which tier cannot be told apart
stage 4  unit stale          data stale          integration ?       docs live
stage 5  unit stale          data ?              integration stale   docs -
stage 6  unit live           data ?              integration ?       docs -

5 of 6 entries record a result against a tree that has since moved.
```

Read plainly: five of the six stage entries claim a tier result against a tree that later stages
moved, which is the ordinary consequence of six stages landing in sequence and not a defect. What it
costs is that the mutations those stages ran are evidence about a tree that no longer exists. Only
stage 6's unit hash is live, and stage 4's docs hash. Stage 3 recorded `all` alone, so its tiers
cannot be told apart at all — a recording gap worth not repeating.

The independent check that does still hold at `ba278912` is the check suite, which ran every tier on
the current tree, including the four-engine data and integration matrices no local run covered. It
concluded green: https://github.com/leodip/goiabada/actions/runs/35482348074

PR #388, opened 2026-09-20
refactor(core): enforce symbol-level ownership in core

### LINKED ISSUE #385: [Architecture follow-up 1/3] Enforce symbol-level ownership in core
## Goal

Finish the shared-kernel refactor at **symbol and responsibility level**, not only at Go-package level.

Production implementation used by only one executable should normally be owned by that executable. `core` should contain only:

1. code with real production consumers in both applications;
2. an intentionally shared wire contract between processes; or
3. repository-wide test/build infrastructure.

A stable HTTP contract is not, by itself, a reason for its Go implementation to live in `core`. If a type is kept there as a public Go SDK surface, that must be an explicit, documented decision.

This is a structural follow-up to #332–#360. Those issues successfully fixed the module graph; this issue removes the application policy that remains inside otherwise-shared packages.

## Why this is one issue

The remaining problems are small fragments inside packages that the architecture table correctly calls `kernel`. Filing one issue per package would obscure the single boundary decision and cause the architecture guard, imports, tests, and naming to be revised repeatedly.

The unit of work is one rule: **a shared package must not hide one-application implementation merely because another symbol in the package is shared**.

## Current census and intended ownership

Re-run the production consumer census before implementation and treat it as the source of truth. The following is the reviewed starting point.

| Current surface | Finding | Intended direction |
|---|---|---|
| `core/handlerhelpers/auth_helper.go` | Implements the admin console's OAuth redirect, PKCE and scope flow; constructed by `adminconsole/internal/server/routes.go` | Move to an admin-console package such as `internal/oauthclient` or `internal/authflow` |
| `core/handlerhelpers/template_funcs.go` | `isAdmin*Page` helpers and `JSBootstrap` are admin-console presentation policy | Move to adminconsole; let each process contribute its own template functions |
| `core/handlerhelpers/http_helper.go` | Adds admin-specific `loggedInUser` / `isAdmin` page data | Move that enrichment to adminconsole; keep only a neutral renderer if it remains genuinely shared |
| `core/middleware/middleware_jwt.go` | Bearer-token middleware is authserver-owned; console JWT/session middleware is adminconsole-owned | Move each half to its owning application |
| `core/middleware/middleware_csrf.go` | Shared code hard-codes authserver and adminconsole route names | Keep a neutral CSRF mechanism if useful, but make exemption/policy configuration application-owned |
| `core/i18n/middleware.go` | Machine-request classification hard-codes authserver routes; locale middleware/refinement has different single-process consumers | Keep catalog/locale primitives in core and move request policy to each process |
| `core/oauth/token_exchanger.go` and `token_parser_jwks.go` | Concrete OAuth-client behavior used by adminconsole | Move to adminconsole's OAuth client boundary |
| `core/oauth/response_type.go` | Provider-side response-type parsing used by authserver | Move to authserver protocol validation/provider code |
| Neutral OAuth types, JWK representations and PKCE primitives | Used by both processes | Keep in a narrowly defined shared protocol package |
| `core/enums` | `TokenType`, `AcrLevel`, `AuthMethod`, `PasswordPolicy`, `SMTPEncryption` and `ThreeStateSetting` are authserver-owned; `Gender` and `KeyState` are shared | Move types by domain; do not replace the current enum drawer with another generic package |
| `core/customerrors` | Shared API error detail is mixed with authserver-only sentinels (`ErrNoAuthContext`, disabled-user/client errors, redirect URI error) | Move sentinels to their owning authserver packages; rename the remaining shared contract only if its responsibility becomes clear |
| `core/stringutil` | Random letter/number strings are authserver-only; the security-random primitive is shared | Move the one-process helpers and retain/rename only the cohesive shared primitive |
| DCR types in `core/api` | RFC 7591 endpoint DTOs are authserver protocol types, not the admin API contract | Move to authserver protocol/DCR code while preserving the HTTP/OpenAPI representation |

Admin API DTOs can stay together even when one endpoint has no current console caller: they form the deliberate authserver/adminconsole wire contract. `core/testutil` and other repository-wide guard support are also explicitly outside the production-consumer rule.

## Implementation guidance

1. Produce a production-only symbol census for every package above. Tests do not make production code shared.
2. Choose destination packages by responsibility, not by mechanically recreating the core package name under `internal/`.
3. Move application route lists, claim interpretation and page-data policy with the application that owns them.
4. Prefer dependency injection for the small neutral mechanisms that remain shared: template `FuncMap` additions, page-data enrichment, CSRF policy and locale selection should be supplied at composition roots.
5. Preserve one definition for algorithms genuinely shared by both processes, notably PKCE challenge generation and neutral token/JWK wire values.
6. Move tests and mocks with their implementation. Add focused import/census tests where they make the ownership decision durable.
7. Update `ARCHITECTURE.md` in the same change. If a package remains intentionally mixed, record the exceptional symbols with the same explicitness used for `core/constants`; otherwise split it so the package-level guard is sufficient.
8. Do not add forwarding aliases solely to make the move look non-breaking. These Go modules are composed in this repository; retain an alias only for a demonstrated external compatibility requirement and give it a removal issue.
9. Apply idiomatic initialism/package naming only to code already being moved. Avoid an unrelated repository-wide rename.

- This is a structural refactor. Do not change OAuth/OIDC semantics, routes, status codes, response bodies, JSON keys, token claims, cookie/session behavior, CSRF coverage, localization behavior, or template output.
- Moving DCR Go types must not change the published OpenAPI schema or RFC 7591 behavior.
- Do not turn `core` into a speculative SDK. A reusable package needs a real contract and consumer, not merely possible future reuse.
- Do not duplicate security-sensitive logic between the applications to satisfy the ownership rule. Share the neutral primitive and keep application policy local.
- This issue does not narrow the large `Database` or `ApiClient` interfaces and does not extract authserver use cases from handlers; those are separate follow-ups in this three-part architecture series.

## Acceptance criteria

- Every production symbol remaining in a top-level core package is either used by both applications or is explicitly classified as shared wire contract/kernel infrastructure.
- `core` contains no concrete application route names other than values deliberately defined as inter-process/public contract.
- The admin OAuth login/client implementation is owned by adminconsole.
- Provider-side OAuth parsing and authserver-only domain/settings enums are owned by authserver.
- Application-specific JWT, CSRF and locale policy is composed inside the owning process.
- DCR Go DTO ownership no longer relies on describing them as part of the admin API contract; their serialized contract remains unchanged.
- `ARCHITECTURE.md` and `AssertArchitecture` are updated so the resulting boundary cannot silently regress at the same granularity.
- No temporary architecture exception is left without a numbered removal issue.

## Verification

- Run the unit tier for core, authserver and adminconsole.
- Run the architecture, generated-code, gofmt, error and logging guards.
- Run auth flow, callback, CSRF, localization, DCR and admin-console rendering tests affected by the moves.
- Run the integration tier because ownership changes cross both executable composition roots.
- The PR description should include a before/after package map and the final production consumer census.

## Series

Part 1 of 3. Recommended first because it establishes ownership before interfaces and use cases are reshaped. The other two issues cover narrow contextual infrastructure ports and extraction of application services from transport/persistence.

leodip:
Closes #385.

## What this is for

`core` is shared by two binaries, and package granularity cannot tell whether a package hides
implementation only one of them uses. `core/constants` is the proof: it reached 139 symbols, 108 of
them named by a single process, with every import rule in `ARCHITECTURE.md` green at every step.
#351 answered that for one package with a per-symbol table. This change generalises the answer to
all of `core`, and then moves out everything the answer cannot justify.

## The rule it establishes

Every exported symbol `core` declares carries a row in `src/core/OWNERSHIP.md` saying why core
declares it. Seven justifications, four computed from the reference graph and three asserted by a
human who has to write a note beside the word:

| justification | who says it | means |
|---|---|---|
| `kernel` | computed | another `core` package references it in production |
| `both-apps` | computed | both applications reference it in production |
| `own-package` | computed | the declaring package's own production code references it |
| `reachable` | computed | a justified symbol's declaration names it, or it is a const or var of a justified type |
| `test-support` | asserted, checked | no production reference anywhere, and at least one test names it |
| `contract` | asserted | none of the above; an intentionally stable cross-process value |
| `moving` | asserted | the symbol leaves core in the named issue |

`src/core/cmd/ownershipdump` writes the computed rows, preserves the asserted ones, and refuses to
invent a justification for a symbol that has none, so a single-application symbol in a shared
package stops the tier until somebody moves it or argues for it. All three module unit tiers check
the file against the tree in both directions, the way they already check the four tables in
`ARCHITECTURE.md`.

## Behaviour changes

This is a refactor, and nothing a caller can reach changes except these three. They are deliberate,
each is a defect the boundary exposed, and each closes inside a stage the plan already had.

1. **The audit row of an admin API action names the administrator who performed it.** 39 production
   sites read the caller's subject through an accessor that reads a context key nothing in the auth
   server writes, so every audited admin API action recorded `loggedInUser` as the empty string. The
   38 sites using `middleware.GetValidatedToken` recorded it correctly, which is why the same key
   appeared filled on some events and empty on others. The four `/auth/logout` rows keep the empty
   value they carry today: that path is not behind the bearer middleware, so there is no subject to
   read there, and the actor of a browser-flow logout is the row's own `userId`.
2. **A cross-origin unsafe-method request to a path an application does not mount is refused 403 by
   the origin check** instead of reaching chi for a 404 or 405. Both servers currently install one
   CSRF exemption table listing both applications' routes, so each exempts the other's. Safe methods
   and every mounted route are unchanged.
3. **A profile form carrying a gender index outside 0 to 2 re-renders with its validation message
   instead of answering 500.** `String()` on all seven enums indexed a slice literal and panicked
   out of range; it now returns the empty string, which is already the established "no gender"
   value.

No route, status, response body, JSON key, token claim, cookie, session, localized string or
rendered template moves.

## Status

**Complete. All 8 stages landed, the whole change was read five times end to end, and the check
suite is green on `c4a5b66b`.**

`src/core/OWNERSHIP.md` carries 333 rows, every one either computed from the reference graph or
argued in a note, and **none of them `moving`** — nothing is parked on its way out. `core/enums` and
`core/handlerhelpers` no longer exist; `core/gender`, `core/sessionstore/sessiontest` and
`core/cmd/ownershipdump` do. No forwarding alias was added anywhere, and no temporary exception row
was added to `ARCHITECTURE.md`.

Three closing comments carry what the diff cannot:

- [**Deferred decisions**](https://github.com/leodip/goiabada/pull/388#issuecomment-5754754116) —
  none. All 17 decisions were settled before the run started; nothing here is waiting on you.
- [**Follow-ups, fold-ins and ceilings**](https://github.com/leodip/goiabada/pull/388#issuecomment-5754762385)
  — one issue drafted ready to file, six defects folded in rather than filed, two ceilings shipped
  with their revisit triggers.
- [**The review account**](https://github.com/leodip/goiabada/pull/388#issuecomment-5754767511) —
  all five code rounds and both plan rounds, every finding, and the evidence behind each fix.

### Stage 1: The ownership guard, `OWNERSHIP.md` and `ownershipdump`

`6fadf86b`. `src/core/OWNERSHIP.md` now carries one row per exported symbol every `core` package
declares — **420 rows**, 245 computed from the reference graph (`kernel`, `both-apps`,
`own-package`, `reachable`) and 175 asserted with a note the guard refuses to find empty
(`contract`, `moving`, `test-support`). `core/testutil.AssertSymbolOwnership` checks it from all
three module unit tiers, in both directions; `src/core/cmd/ownershipdump` writes the computed rows
from the same census and refuses to invent a justification for a symbol that has none; the lint
tier runs the tool and fails on a tree it changed. `ARCHITECTURE.md` gains rule 8 and a `core/cmd`
row, and `CLAUDE.md`/`AGENTS.md` gain the regeneration step.

The guard is built first, per the agreement's decision 12, so every later stage regenerates the
table and its own diff carries the rows for the symbols it moved. The 175 asserted rows are the
burn-down list the rest of this pull request works through.

Two deviations from the plan, both forced by the tree and recorded in full in the run log. An
asserted row does not seed the justification fixpoint — seeding from one makes two mutually
referring symbols computed on one regeneration and unjustified on the next, so the table would not
be stable and the lint tier could not compare the tool's output against the committed file. And
`test-support` asks whether a binary links the referrer rather than whether a production file names
it, because three packages here are test support written in files with no `_test.go` suffix, so
`adminconsole/internal/handlertest` naming `testutil.Reporter` is not a reason to refuse `Reporter`
the word.

Tiers: unit green (7819 ok, 0 fail), lint green including the new regeneration check. Data and
integration not applicable — no migration, no schema change, no query, no handler, no route. Eight
mutations against the new tests, all caught, two of them against the committed table itself.

### Stage 2: Dead localization scaffolding, `i18n.WithLocale`, and `sessiontest`

`a6612afe`. `core/i18n` now exports one locale primitive,
`WithLocale(ctx, explicit, tags...)`, in place of three policy helpers.
`EmailContext`, `RefineLocalizerWithUILocales` and `RefineLocalizerWithUserLocale` are gone and
their eight production call sites repointed; `defaultBundle`, `localizerFor`, `attachLocale` and
`hasExplicitIntent` all stay unexported, which is the point — the three stragglers were in a shared
package only because setting a locale needed four private things. `MiddlewareLocaleFromJWT` and
`localeClaimFromJwt` move to `adminconsole/internal/middleware`, where the middleware's body
collapses to one `WithLocale` call because that primitive is all three of its early returns, and
`core/i18n` stops importing `core/constants` and `core/oauth`.

Three surfaces are deleted rather than moved, having no caller at all: `system_entities.go`, whose
registry is an empty map so both exported functions always return the fallback; `IsMachineRequest`;
and the `SysName`, `SysDesc` and `DirAttr` template functions. `MemoryBackend` moves to a new
`core/sessionstore/sessiontest`: its own doc said no binary constructs it, and until now both
binaries linked it anyway.

**One correction to the plan, measured rather than assumed.** The primitive hands *every* usable tag
to the matcher, not just the first. `ui_locales=fr pt-BR` resolves pt-BR today — `fr` has no
catalog, and the matcher reads the list as one preference order — so matching `usable[0]` alone
would have answered English and silently discarded the RP's second choice. A trailing `"en"` never
changes the pick, which is what makes the email sites' recipient-then-English form exact. Both
readings come from a probe over the two loaded catalogs, and the first is pinned by a test and a
mutation.

Nothing a caller can observe changes. Unit green (7805 ok), integration green on sqlite (1425 ok,
run because eight handler call sites changed), data green on sqlite (598 ok). Six mutations against
the new cases, all caught; one survived first and found two cases asserting a value that held
either way, both fixed.

### Stage 3: The admin console's OAuth client, its JWT session middleware, and the audit-subject fix

`70570fae`. `core/middleware/middleware_jwt.go` held both applications' JWT middleware in one file
and is now deleted; `core/handlerhelpers/auth_helper.go` held the admin console's whole authorize
redirect and is gone with it. The admin console's client of the protocol — the redirect with its
PKCE, state and nonce minting, the code-for-token exchanger, the JWKS parser and the two bounds
they share — is `adminconsole/internal/oauthclient`; the session middleware that calls the redirect
is `adminconsole/internal/middleware`, beside the other `http.Handler` wrappers; the six
`SessionKey*` are `adminconsole/internal/constants`. The bearer half and `ContextKeyBearerToken`
are `authserver/internal/middleware`, moved in this stage rather than stage 6 so that core is never
left holding a parser port and a mockery entry for one application. The credential-read **policy**
moved with it, not just the file: `"access_token"` joins the auth server's forbidden list carrying
the RFC 6750 2.3 and RFC 9700 4.3.2 paragraph, and core's caller stays, rewritten to refuse that
read coming back.

**The second of the three observable changes this pull request makes lands here.** An audit row for
an admin API action now names the administrator who performed it instead of naming nobody.
`AuthHelper.GetLoggedInSubject` read `ContextKeyJwtInfo`, a key only the admin console's session
middleware writes, so all 39 of its `apihandlers` call sites recorded `loggedInUser` as `""`. They
read the caller's own bearer token now, through `middleware.GetValidatedToken`, the way the other
38 sites already did; the accessor is deleted from the implementation, the port and the generated
mock. The four `/auth/logout` rows deliberately keep the literal `""` — that route carries no
bearer token, and since `AuditLogResponse.Details` is returned verbatim by
`GET /api/v1/admin/audit-logs`, dropping the key would be a fourth observable change.

Two repairs this move required and the plan did not name. The #266 session-fixation guard held
every writer of the authenticated session key in the admin console to reaching
`sessionstore.Regenerator` — the same sentence as "the sign-in writer rotates" only while that
module had one writer. The moved middleware's token refresh is a second writer and must *not*
rotate, because the session it rewrites was already the administrator's. Rather than exempt a path,
which is where the next real sign-in site would hide, the guard now classifies writers and fails
both on one it does not know and on a named writer that has stopped writing. And a comment in
`handler_api_users_sessions_test.go` promised a no-token case that did not exist, leaving
`callerSubject`'s fallback arm entered by nothing; it is written and mutation-checked.

Unit green (7775 ok), data green on sqlite (598 ok), integration green on sqlite (1425 ok, run
because the audit payload of every admin API action changes), lint green including the mock and
`OWNERSHIP.md` regeneration checks. Five mutations against the new and moved seams, all caught —
including the one the plan named, `r.PostFormValue("access_token")` → `r.FormValue`, which reddened
core's credential lint before this stage and now reddens the auth server's.

### Stage 4: CSRF exemption policy supplied per application

`448ecda5`. `core/middleware` declared three exemption tables naming eleven concrete routes, and
both servers registered the same middleware, so each binary exempted the other's endpoints from the
origin check: the admin console exempted `/auth/authorize`, `/auth/token`, `/userinfo` and
`/connect/register`, none of which it mounts, and the auth server exempted `/auth/callback`, which
it does not. `MiddlewareSkipCsrf` takes a `CsrfPolicy` now. core keeps the mechanism — the origin
check, `explainCsrfFailure`, the localized 403 and the matching — and keeps the three-way
distinction between an exact path, a prefix and a conditional, because that distinction is
load-bearing rather than stylistic and collapsing it to one predicate per application would make
each module re-establish the reasoning. Each server declares its own policy at its composition
root, with every per-endpoint rationale comment moved intact.

The `/auth/logout` `id_token_hint` predicate is `authserver/internal/middleware` as
`LogoutIdTokenHintPresent`, still reading through
`core/handlerhelpers.LookupFromUrlQueryOrFormPost`, so the property #109 exists for — one reading
shared by the middleware and the logout handler, since disagreeing would exempt a cross-site POST
and then route it down the branch that tears the session down with no consent — stays structural
rather than promised. Both ends are in the auth server now, which also removes `core/middleware`'s
only import of `core/handlerhelpers`.

**The CSRF narrowing, the second of the three observable changes listed above, lands here.** For
every route a binary mounts nothing changes; for a path it does not mount, a cross-origin
unsafe-method request is refused 403 by the origin check instead of reaching chi for a 404 or 405.
Each chain test states it as a difference rather than as a status, pairing every cross-site row
with a same-origin control that still answers 404 on the unmounted paths.

Two additions the plan did not name, both closing a hole a policy *value* can express where the
package-level tables could not: `newCsrfSkipper` refuses a path listed both exactly and
conditionally, which is silently unconditional because the exact entry wins and the predicate is
never called — precisely the hole the conditional shape exists to close — and refuses an empty
prefix, which exempts the whole server. Both are literals at a composition root evaluated once at
startup, so both fail the boot rather than a request.

The auth server gains the chain-level CSRF test it has never had, and the admin console's gains the
same three claims. core's own table is over a fixture policy now, because core declares no route to
assert about. Unit green (7824 ok), integration green on sqlite (1425 ok, run because routes change
hands). Seven mutations, one per behaviour claimed, all caught — including the presence-versus-value
drift #109 warns about, which the two empty-hint rows catch.

### Stage 5: One renderer per application; `core/handlerhelpers` deleted

`92a2a48f`. `core/handlerhelpers` was one package hiding two applications' renderers. Its bind map
carried `loggedInUser` and `isAdmin`, read off the admin console's browser session, and no auth
server template binds either; its template `FuncMap` held twenty-two entries of which the auth
server's templates call four — `T`, `Lang`, `args`, `versionComment` — so every auth server page
was parsed with eighteen functions it cannot reach, five of them predicates over the admin
console's own URL paths. Each application now owns its renderer: the auth server's copy joined
`authserver/internal/handlerhelpers` and kept the two form-post readers, the admin console got
`adminconsole/internal/handlerhelpers` and kept the enrichment and all twenty-two entries. About
238 lines are the same in both and can drift, which is what owning a renderer costs and is cheaper
than one package behaving two ways for its two callers.

`ContextKeyJwtInfo` moved to `adminconsole/internal/constants` — this commit deleted its last
reader outside that module — so `core/constants/context_key.go` is gone with its `ctxKey` type and
core declares no context key at all. `core/stringutil.ConvertToString` moved in beside
`addUrlParam`, its only caller, as an unexported function; core had published it for that one
caller. Each module now declares its own `HttpHelper` port and generates its own mock, so no
`.mockery.yaml` writes outside its own module any more: the arrangement that ended worked only
because the auth server's port is the wider of the two, which left the console asserting against a
mock of an interface it does not have.

Two table consequences the guards forced rather than the plan chose. Five `core/constants` rows
dropped from `kernel` to `both-apps` — `Version`, `BuildDate`, `GitCommit`,
`AuthServerResourceIdentifier`, `ManagePermissionIdentifier` — each having been `kernel` on
`core/handlerhelpers` alone, so no row in that table reads `kernel` now. And seven `core/i18n`
formatters gained argued `contract` rows in `OWNERSHIP.md`: with the shared renderer gone only the
admin console's `FuncMap` names them, and they cannot follow it, because each reads the localizer
and the bundle `core/i18n` keeps private — moving them would export that state and put one message
catalog behind two renderers.

Three new tests pin the split rather than restating what the moved code does: the auth server's
renderer binds no `loggedInUser` and no `isAdmin`; the console's `isAdmin` follows the access
token's scope, both arms, which nothing pinned before; and each `FuncMap`'s exact key set. Unit
7852 passing, integration on sqlite 1425 passing, lint 0 issues with the mocks and `OWNERSHIP.md`
regenerated clean. Four mutations against the new tests, all caught. No behaviour changes.

### Stage 6: The auth server's provider-side symbols leave core

`96759db5`. Sixteen exported symbols that only the auth server names left `core`, each to the
package that owns the thing it is about: `MiddlewareNoStore` to `authserver/internal/middleware`;
`ParseResponseType` and `ResponseTypeInfo` to `authserver/internal/protocolvalidation`, beside
their only callers; the four `customerrors` sentinels split between
`authserver/internal/handlerhelpers` and `protocolvalidation`, beside the helper that returns one
and the validator that constructs the other three; `ValidateNoAngleBrackets` to
`authserver/internal/accountvalidation`; and the five RFC 7591 registration types to
`authserver/internal/oidc`, beside the `well_known_config.go` whose own package doc already argued
this case word for word and whose discovery document advertises the registration endpoint.

Two of the moves needed something built rather than carried. The
`WWW-Authenticate` constructor went to `authserver/internal/apiresponse` — issuing an RFC 6749
section 5.2 challenge is provider-side by definition — but `ErrorDetail` keeps its data in an
unexported map and core published only a getter, so nothing outside `core/customerrors` could
build the value. `WithWWWAuthenticate` is `WithDescription`'s exact sibling and, like the deleted
four-argument constructor, adds no entry for an empty value; that guard is load-bearing, because
`Is` compares the number of details before comparing any of them and three sentinels are matched
through it. And the two non-security random-string helpers could not move unchanged, since both
wrap a sampler `GenerateSecurityRandomString` also uses and which therefore cannot leave core. So
`core/stringutil` exports `RandomStringFromAlphabet` rather than the rejection sampling being
duplicated, and the wrappers went to a new `authserver/internal/randomstring` as `Letters` and
`Digits`. Exporting the sampler publishes its precondition, so the precondition is now enforced:
the rejection limit is `256 - (256 % n)`, zero once `n` exceeds 256, so an alphabet above 256 bytes
made every draw rejected and the call never return. It answers `""` instead, which is already what
an empty alphabet and a non-positive length answer.

`core/oauth`'s package doc argued that ownership follows what each binary ends up containing rather
than which process names a symbol, and `response_type` parsing stayed in core on exactly that
argument. That is the rule this issue replaces, so the doc is rewritten rather than left to give
the next reader the old answer. `ARCHITECTURE.md`'s `core/validators` "one rule, packaged twice"
note and its `core/oauth` split note both retire: every symbol left in those two packages is named
by both applications. `CLAUDE.md` and `AGENTS.md` lose two now-false statements each. Three entries
came out of `openapi_contract_lint_test.go`'s `apiStructsWithNoSchema` in the same commit, because
that map is exact in both directions and they name structs `src/core/api` no longer declares.

New coverage where the move could otherwise have been silent: a wire test asserting each DCR
struct's serialized form against a literal in both directions, which is the only thing in the tree
that would notice a tag change — the integration suite reads the fields back through these same
structs, so a tag renamed on both sides at once passes there; the exported sampler at 1, 256 and
257 bytes; and the token endpoint's challenge header, which nothing covered before, placed at
`jsonErrorConformed`, the one line where a header carried outside the `*ErrorDetail` would be
dropped in silence. Unit 7881 passing, integration on sqlite 1425 passing, lint 0 issues with the
mocks and `OWNERSHIP.md` regenerated clean. Four mutations, one per new seam, all caught. No
behaviour changes.

### Stage 7: Enums by domain, `core/gender`, total `String()`; `core/enums` deleted
`26054669`. Six of the eight types `core/enums` held were named by the auth server alone, so each
went to the package that owns its domain: `AcrLevel`, `PasswordPolicy`, `KeyState` and
`ThreeStateSetting` to `authserver/internal/models`, forced there because `internal/data` writes all
four and imports nothing above `models`; `AuthMethod` to `ceremony`, beside the `AuthContext` field
that accumulates it; `TokenType` to `issuance`, which produces every value it has; `SMTPEncryption`
to `emaildelivery`, the only thing that acts on it. `Gender`, its constants and `IsGenderValid`
became `core/gender`, a fourth shared reference vocabulary beside `countries`, `locales` and
`timezones`. 101 files carry a changed import line and a changed qualifier.

What freed `KeyState` was deleting the admin console's key-ordering loop: the keys page re-sorted a
list `GET /api/v1/admin/settings/keys` already returns in that order, using a structurally identical
copy of the producer's own loop. Both ends are now tested rather than one — the producer against
deliberately unsorted fixtures, the page against orders the API would not produce, which is the half
that goes red if the loop comes back.

**The third behaviour change is here** (see *Three deliberate behaviour changes* above): `String()`
on all seven integer enums returns `""` outside its declared range instead of panicking on the slice
index. **`AcrLevel.String()` is deliberately excluded** and says so in a comment and a test case: it
is string-backed, has no range to be outside of, and OIDC Core defines `acr` as a case-sensitive
string whose meaning can be deployment-specific, so collapsing a value this server does not rank
would discard what a token carries. `Priority()` already answers 0 for one, which is what keeps
comparison safe without throwing the value away.

`ARCHITECTURE.md`'s package table swaps `core/enums` for `core/gender` rather than only losing a row,
because rule 6 wants exactly one row per top-level `core` package. `OWNERSHIP.md` regenerated to 333
rows. Unit green (7981), integration green on sqlite (1425), data green on sqlite (598, because the
repoint edited 9 files the four CI data jobs execute), lint green including the `ownershipdump`
regeneration check. Four mutations, one per new seam, all caught.

**CI was red on `26054669`, and `68c2bb1b` fixed it.** The `Lint` job failed on
`unparam -exported`: `ThreeStateSettingFromString - result 0 is never used`. In `core` the function
had no caller inside its own module and unparam could conclude nothing about it; in `authserver`
its two callers are both in `HandleAPIClientTokensPut` and both discarded the parsed value,
validating with it and then writing the column from a second independent `TrimSpace` of the same
request field. The move made an existing shape visible rather than creating one. The handler now
writes the columns from the parsed values' `String()`, which is byte-identical to what it stored
before — a string that parsed is one of the three the type declares. The second half is why only CI
saw it: `run-tests.sh`'s lint tier claims in its own comment to run "the exact command CI's Lint job
runs per module", and that job runs gofmt, `go vet`, `unparam` and golangci-lint while the tier ran
only the last. `unparam` now runs there too, over the same four modules; gofmt and vet were already
held by `TestGoSourcesAreGofmted` and `where.sh build`. Proven by reverting the handler by hand and
watching the tier exit 1 with the same finding, then restoring and confirming the tree by hash.

### Stage 8: The closing sweep and the recount

`1f2781e2`. The recount, taken with `ownershipdump` against the finished tree rather than with the
probe — the probe is not recursive and would have missed `core/sessionstore/sessiontest` outright.
**333 rows**, against decision 15's estimate of about 343. The dump left no diff, so the committed
table already described the finished tree.

The 10-row delta has two halves, both recorded rather than absorbed. **87 exported symbols left
`core`** where the estimate said about 74, measured by diffing the table against its own first
commit: `core/api` −5 (the DCR types), `core/i18n` −9, `core/enums` −39, `core/constants` −8,
`core/oauth` −8, `core/middleware` −7 net, `core/customerrors` −5, `core/handlerhelpers` −8,
`core/sessionstore` −2, `core/stringutil` −2, `core/validators` −1, against `core/gender` +5 and
`core/sessionstore/sessiontest` +2. And **`core/testutil` gained three**, because the guard is
written in it: 420 rows at stage 1 against the probe's count of 417.

The three properties the goal asserts and nothing else checks were each verified rather than
assumed. **Zero `moving` rows**, so no row is a temporary exception left behind an issue that is
closing. **All 112 asserted rows carry a note**, the shortest 92 characters; the guard only checks
non-empty, so these were read. **No forwarding alias anywhere** — none added on this branch and none
in the tree at all, and no `core` production file imports either application module.
`ARCHITECTURE.md`'s exception table is empty and its package table is nineteen `kernel` rows.

The sweep of `ARCHITECTURE.md`, `CLAUDE.md`, `AGENTS.md`, `README.md` and the docs site for every
name this change deleted or moved found **one stale sentence**, which is this stage's only edit: the
temporary-exceptions preamble illustrated its "both ends name a package" rule in the present tense
with `core/handlerhelpers`, a package the change deleted. Swapped for `core/middleware`. Everything
else the sweep turned up is deliberate past tense. The five new packages each carry a doc in #360's
voice; `CLAUDE.md` and `AGENTS.md` are byte-identical and were already current, pattern 8's
directory list diffed against the code.

Unit green (7981 passing) and lint green, including the `OWNERSHIP.md` regeneration check that is
the recount's own guard. Data and integration not run and not applicable: no Go file changed, which
the split tree hashes confirm against stage 7's.

## Review

Reviewed by `openai-codex/gpt-5.6-sol` at effort `xhigh`, **five rounds over the whole branch**, plus
two earlier rounds over the plan before any code existed. `GBD_REVIEW_MODE=end` was in force for the
whole run, so **no stage was reviewed on its own**.

|---|---|---|---|---|
| plan | 1 | 13 | 5 | all resolved in `plan.md` |
| plan | 2 | 6 | 4 | 4 applied, 2 withdrawn on code; gate closed |
| code | 1 | 3 | 0 | all fixed, `c53d2f4b` |
| code | 2 | 2 | 0 | all fixed, plus 3 more found by enumerating the family, `c72e94c7` |
| code | 3 | 5 | 0 | all fixed, `1195bc73` |
| code | 4 | 2 | 0 | all fixed, plus 1 settled by a surviving mutation, `8b107725` |
| code | 5 | 4 | 0 | all fixed, `9b2cde93`; gate closed at its 5-round budget |

Nothing was refuted at the code gate, nothing deferred, nothing escalated: every code finding came
back `judgement: false` with a populated forced answer, at `high` confidence, on all three axes
reviewed.

**Every one of the sixteen code findings is in the guard this change introduces**, not in the code it
moves — `src/core/testutil/symbol_ownership.go` and the helper it shares with `constants_ownership.go`.
The reviewer's round-1 note: *"Production behavior, package boundaries, wire compatibility, and all
reviewed security controls are sound."* And `src/core/OWNERSHIP.md` regenerates byte for byte
identical after all five repairs, checked each time, so **no row in this repository was ever wrong**.
What was defective is what the guard would have accepted next — which, for a guard whose value is
firing on the addition nobody looked at, is the failure mode rather than a lesser one.

The significant findings, one line each:

- **`c53d2f4b`** — evidence was read by spelling, not object identity, so a local or a struct field
  named like an exported symbol let dead code park in core behind `test-support`; one `ValueSpec`
  sibling vouched for another; and reachability came off the syntax rather than the Go type, so
  `ColourRed = Colour("red")` and its explicitly-typed twin got different answers. Fixed by
  type-checking the declaring package, pairing names to values positionally, and asking `go/types`
  for the type — with the value's own declared type then deleted from its references, because that
  is the loop that let `core/enums` justify itself.
- **`c72e94c7`** — missing declaration edges: a value spec's type slot, a type parameter's
  constraint, the expression list an omitted `ConstSpec` repeats, and **a body-less function
  declaration that crashed the walk outright** on a typed nil the `node == nil` check cannot see.
- **`1195bc73`** — an unaliased import binds the imported package's **declared** name, not its
  path's last element, so a core package named for anything but its directory had every selector on
  it read as no reference at all; an omitted `ConstSpec` inherits the preceding type slot as well as
  the expression list; and the declared-type exclusion was erasing independent uses of the type in
  the same initializer.
- **`8b107725`** — a generic type slot (`var Default Box[int]`) and a type-assertion head
  (`var ToneWarm = source.(Tone)`) both escaped that exclusion, letting a value justify its own type
  — the exact circularity the exclusion exists to refuse, reached by spelling the type differently.
- **`9b2cde93`** — `receiverTypeName` did not strip parentheses, so `func (l *(Level)) Build()`
  compiles, passes `gofmt`, and had its method silently discarded from the graph. Three more
  fixtures for arms whose mutations had been surviving.

Full account, with every mutation and its result, in
[this comment](https://github.com/leodip/goiabada/pull/388#issuecomment-5754767511).

## Tiers

**Run locally**, in the dev container through `where.sh`, on every stage and every review round:

- `build` — all four modules compile, test files included.
- `test --type modules` — the three module unit tiers. **8010 passing, 0 failing** at the end;
  `AssertSymbolOwnership` is called from all three, which is the tier that matters for the new guard.
- `test --type lint` — golangci-lint and `unparam -exported` over the four modules, plus the
  Tailwind, mocks and **`ownershipdump`** regeneration checks. The last of those is what proves the
  committed table is the generated one rather than a hand-edited copy.
- `test --type data --db sqlite` — run on the stages that edited files under `tests/data/`, and on
  round 2 because it costs seventeen seconds.
- `check-anchors.sh` — 30/30 rows resolve to exactly one line, in the named function.
- **Mutation testing** on every guard fixture: each new test was proven to fail against the code it
  pins before the fix landed, and each arm proven to be caught after. Two mutations survived and
  were resolved by deleting unreachable code rather than by widening a test.

**The check suite's**, because no local tier covers them: the **four-engine data and integration
runs** (`Tests / sqlite`, `mysql`, `postgres`, `mssql`), `Unit / race`, `Lint`, `Vulnerabilities`,
`Versions` and `Build validation`. All thirteen jobs are green on `c4a5b66b`.

**Honest limit on the record.** Six of the eight stages recorded a tier result against a tree a later
stage then moved — what landing eight stages in sequence does. The green four-engine suite settles
the *tiers* on the current tree; it does not re-establish the per-stage *mutation* evidence, so what
is unproven is narrow and specific: that a test written in stage 3 would still fail if stage 3's
production code were broken today. Named rather than glossed, and the final gate's five rounds all
mutated against the current tree.

Section 2's goal stayed true as written throughout; what moved is section 4's fanout. Each is
recorded in full in the stage logs.

- **The bearer-token middleware moved in stage 3 rather than stage 6**, and the random-string helpers
  went to a new `authserver/internal/randomstring` rather than beside their one caller, with
  `core/stringutil` exporting `RandomStringFromAlphabet`. Both from the plan review.
- **Two symbols were renamed on the way out**: `GenerateRandomLetterString` and
  `GenerateRandomNumberString` are `randomstring.Letters` and `randomstring.Digits`, matching the
  #360 leaf packages beside them (`uuidutil.New`, `urlutil.CanonicalOrigin`). §2 allows a rename on
  code the change already moves.
- **The `GetLoggedInSubject` rewrite covers 39 sites, not the 44 §1 counted**: four logout sites take
  the literal `""` because there is no caller to name, and one was already correct.
- **`AcrLevel` is excluded from decision 16's totality**, because it is string-backed rather than
  index-backed; a case pins that an unranked value survives.
- **`ARCHITECTURE.md`'s package table gains a row rather than only losing two**: `core/gender` is a
  new top-level core package and rule 6 wants exactly one row per such package. Net effect is one row
  swapped, and `core/enums` is gone as planned.
- **`core/customerrors` gained a `WithWWWAuthenticate` method**, and three DCR entries came out of
  `openapi_contract_lint_test.go`'s `apiStructsWithNoSchema`, which is exact in both directions and
  would have red-lit the authserver tier if left.
- **The console's share of the template `FuncMap` is 22, not 21**, after stage 2 deletes three dead
  entries; the auth server's four are also the console's.
- **`unparam -exported` was added to the local lint tier.** This change moved an exported parse into
  the module whose two call sites discard its result, and only CI could see the finding — which is
  the same reason the Tailwind and mock regeneration checks are there.


leodip:
## Deferred decisions: none

All 17 decisions in the agreement were settled at specification time, before the run started. The
run raised none of its own, deferred none, and escalated none — nothing on this pull request is
waiting on an answer from you, and no part of it was built on an assumption that could still be
reversed.

For completeness, the three decisions whose consequence is user-visible are **3** (an audit row
names the administrator instead of naming nobody), **5** (a cross-origin unsafe-method request to
a path an application does not mount is refused 403 by the origin check instead of reaching chi for
a 404 or 405) and **16** (a profile form carrying a gender index outside 0–2 re-renders with its
validation message instead of answering 500). All three are argued in the agreement and set out
under **Behaviour changes** in the body above. They are decided, not open.

leodip:

Everything this run found next to the work, in three blocks: what is drafted for you to file, what
was fixed here instead of filed, and what was shipped with a stated limit.

---

### 1. Follow-ups drafted — none

The run never files issues, and the one it drafted was folded in instead. It is the last row of
§2. Nothing else is outstanding.

---

### 2. Folded in rather than filed — seven

Five were folded in at specification time and are named in the agreement's §2 goal, so the diff
explains itself; the sixth and seventh were found mid-run and are recorded here because nothing
else would show them. Each is a real defect this change's boundary exposed, and each closed inside a stage the plan
already had.

| What | Commit | Why it fit |
|---|---|---|
| **The audit row of every admin API action named nobody.** 39 call sites read `GetLoggedInSubject`, which reads a context key only the admin console's middleware ever writes, so it returned `""` in the auth server. Replaced by `middleware.GetValidatedToken`, which the other 38 sites already used | `70570fae` | Decision 3. The stage that moves the console's JWT middleware out of core is the stage that proves the auth server never writes that key, and an issue would otherwise have been written against a file this change deletes |
| **`Gender.String()` panicked out of range**, so `gender=3` on either profile form answered 500 instead of re-rendering with its validation message. Made total, and the same one-line guard put on all seven enums, since all seven were being moved anyway | `26054669` | Decision 16. A one-line fix whose test is the probe already written, against a package this change deletes |
| **Inert localization scaffolding deleted:** `core/i18n/system_entities.go` whole, `IsMachineRequest`, and the `SysName`, `SysDesc` and `DirAttr` template entries — about 160 lines with no caller in either binary | `a6612afe` | Decision 4. The new guard forces most of it: none of the seven justifications is available for a symbol nothing references |
| **`MemoryBackend` stopped being linked into both production binaries**, moving to `core/sessionstore/sessiontest`. Its own doc already said "No binary constructs it" | `a6612afe` | Decision 13, and the same guard: `test-support` is the only word available, and it belongs in a test-support package |
| **The admin console stopped re-sorting a list the API already sorts** — 18 lines structurally identical to the auth server's own ordering pass, and idempotent on its output | `26054669` | Decision 7, and what let `KeyState` leave core at all |
| **A function declaration with no body crashed the ownership walk.** `collect` was passed a typed nil `*ast.BlockStmt`, which its own `node == nil` check cannot see, so `ast.Inspect` panicked. There are none in core today; the day one landed it would have taken all three module unit tiers down with a nil dereference instead of reporting anything | `c72e94c7` | Found mid-run at the final gate's round 2, by enumerating the family the reviewer's two findings belonged to. A verified crash on a surface this change already touches, fixed narrowly per the final-review rule |
| **Three guards read their imports by path base**, so a package named for something other than its directory bound a name none of them expects. `defaultImportName` returned the last path element and its doc claimed that was what an unaliased import binds; Go binds the package clause. Five versioned modules contradict it (`chi/v5` binds `chi`, 83 unaliased imports in non-test files), as do the ten generated mocks directories and `tests/data`. Both failure directions are a guard going quiet: a call site written through a name the map lacks, and — worse — a bogus entry occupying a watched name and silencing the real calls beside it | `fb3655b1` | Drafted as a follow-up, then folded in once the fix turned out not to need the tree-wide package-name index the draft assumed. Nothing enumerates these maps, so recording the watched paths alone is exact and closes the collision by construction — and it reaches the versioned modules, which an index over `src/` could not. `slog_lint.go` was already in this diff, and the same defect was fixed twice inside this change's own census (`1195bc73`) |

---

### 3. Ceilings left by this change — two, both on the ownership census

Both are limits that were weighed and shipped, each with the event that would make it worth
revisiting.

- `src/core/testutil/symbol_ownership.go:544` — a reference from outside the declaring package is
  read as a selector on the identifier the import binds, resolved by name within the file. A
  package-level declaration shadowing that name is caught, because the parser resolves it; a local
  variable inside a function shadowing it would be read as the package. Closing it means
  type-checking every package in four modules rather than parsing them. **Revisit when** a row ever
  rests on a reference that turns out to be a false one. (`constants_ownership.go` records the same
  ceiling for the same technique. References from *inside* the declaring package are not read this
  way at all — those go through `go/types`, because there an identifier matched by spelling would
  let a local, a parameter or a struct field justify its namesake.)
- `src/core/testutil/symbol_ownership.go:559` — that `go/types` pass runs with a stub importer, so
  nothing an import contributes is resolved and every selector on an imported package is an error
  the checker is told to continue past. In-package name resolution does not depend on imports, so
  `Uses` and `Defs` are sound for exactly the identifiers that arm reads. A dot import would break
  that, and this tree has none. **Revisit when** a dot import is written under `src/`, which would
  need a real importer here rather than a stub.

Both are mirrored in `src/core/OWNERSHIP.md`'s own **ceiling** paragraph, so a reader of the table
meets them without reading the guard. The second one's trigger was missing and was added at the
finish, in `c4a5b66b`: a ceiling nobody can revisit is a TODO nobody owns.

leodip:
## The review account

`GBD_REVIEW_MODE=end` was in force for the whole run, so **no stage was reviewed on its own**: the
five rounds below are the only reading the code got, each over the whole branch against
`origin/main`. Two earlier rounds reviewed the plan before any code existed. The verdicts live in
the issue folder, which is on one machine and pushed nowhere, so this comment is the only place they
reach anyone.

Reviewer throughout: `openai-codex/gpt-5.6-sol` at effort `xhigh`, all three axes (conformance,
quality, security) reported `reviewed` on every round.

### Summary

|---|---|---|---|---|---|
| plan | 1 | 13 | 5 | 6 (+1 minor) | all 13 resolved in `plan.md`; 1 flagged as needing a human |
| plan | 2 | 6 | 4 | 2 | 4 applied, 2 withdrawn on code; gate closed at its 2-round budget |
| code | 1 | 3 | 0 | 3 | all fixed, `c53d2f4b` |
| code | 2 | 2 | 0 | 2 | all fixed, plus 3 more found by enumerating the family, `c72e94c7` |
| code | 3 | 5 | 0 | 5 | all fixed, `1195bc73` |
| code | 4 | 2 | 0 | 2 | all fixed, plus 1 settled by a surviving mutation, `8b107725` |
| code | 5 | 4 | 0 | 4 | all fixed, `9b2cde93`; gate closed at its 5-round budget |

**Nothing was refuted at the code gate. Nothing was deferred. Nothing escalated** — every code
finding carried `judgement: false` with a populated `forced_answer`, meaning a question the code
settles rather than a choice for a human. Confidence was `high` on all 16.

### The one thing worth knowing about all sixteen code findings

**Not one of them is in the code this change moves.** Every one is in
`src/core/testutil/symbol_ownership.go`, the reference resolver the whole issue rests on, and its
helper in `constants_ownership.go`. The reviewer's own round-1 note: *"Production behavior, package
boundaries, wire compatibility, and all reviewed security controls are sound. Three independent
resolver defects weaken the central ownership guard and need targeted fixes and fixtures."*

And `src/core/OWNERSHIP.md` **regenerates byte for byte identical after every single one of the five
repairs**, checked with `go run ./cmd/ownershipdump` and `git status` each time. So no row in this
repository was ever wrong. What was defective is what the guard would have accepted *next* — which,
for a guard whose whole value is firing on the addition nobody looked at, is the failure mode rather
than a lesser one.

---

### Round 1 — `c53d2f4b`, three findings

1. **`recordOwnPackageTest` read a matching spelling as a reference.** A local variable, parameter,
   field or test function spelled like an exported symbol let genuinely unreferenced production code
   pass as `test-support`. The code carried a comment *defending* the spelling match, and that
   argument has the direction backwards: matching by spelling is more permissive than object
   identity, so the arm could only ever grant the word wrongly. Fixed by splitting the two kinds of
   test file — an internal test (`package p`) is type-checked with the production files and counts
   only when `info.Uses` resolves the identifier to the package scope; an external test
   (`package p_test`) is read as a selector like any other importer. Mutation
   (`obj.Parent() == scope` → `scope != nil`, the old behaviour): **CAUGHT**.
2. **`readDecl` let one `ValueSpec` sibling vouch for another.** The `collect` over `s.Values` sat
   inside the loop over `s.Names`, so every name got every expression. Reproduced before fixing:
   `var Ready, discarded = true, Orphan{}` computed `Orphan` as `own-package` off an edge the source
   does not have. Fixed by pairing positionally when the counts match, keeping shared attribution
   for the tuple form `var a, b = pair()`. Mutation: **CAUGHT**.
3. **Reachability was read off the syntax, not the Go type.** `namedTypeIn` read the spec's type
   slot, which is nil for `ColourRed = Colour("red")`, so the identical constant written two ways
   got two different answers and the second invited an asserted `contract` row for something the
   tree justifies. Fixed by asking `go/types` for the object's type. The forced answer attached a
   constraint — the edge runs from a justified type to its values and never the other way — which
   made a previously inert loop live, so the declared type is now deleted from the value's own
   references. *That is the loop that let `core/enums` justify itself, so it is not a detail.* Two
   mutations: both **CAUGHT**.

### Round 2 — `c72e94c7`, two reported and three found

Both reported findings confirmed: an omitted `ConstSpec` (`const ( discarded = int(Tick(1));
Rearmed )`) carried no initializer edges at all, and a production package legally named `odd_test`
had its internal tests misread as external ones and its genuine `test-support` evidence refused.

The reviewer's `why_exhaustive` argued only about `ConstSpec` omission, so **the family was
enumerated rather than the two instances fixed**. Three more, each demonstrated with a failing probe
first: a value spec's type slot was never read (`var Registry map[string]Slotted` left `Slotted`
unreferenced); a type parameter list was never read (`type Box[T Bounded]`); and **a body-less
function declaration crashed the walk outright** — `collect` was handed a typed nil `*ast.BlockStmt`
that its own `node == nil` check cannot see, and `ast.Inspect` panics on it. There are none in core
today; the day one landed it would have taken all three module unit tiers down with a nil
dereference instead of reporting anything.

**Eight mutations, all CAUGHT.** Mutation residue was then grepped out of the production file, and
`git status` showed the two intended files and nothing else.

### Round 3 — `1195bc73`, five findings

All five confirmed by reading and then demonstrated by a fixture that fails against the unrepaired
code. Four of the five new fixtures were run against the old resolver first and failed exactly where
the reviewer said they would:

```
--- FAIL: TestSymbolOwnership_AnOmittedConstSpecInheritsItsTypeSlotToo   expected "reachable", actual ""
--- FAIL: TestSymbolOwnership_AnIndependentUseOfTheDeclaredTypeIsStillAReference  expected "own-package", actual ""
--- FAIL: TestSymbolOwnership_AnExternalTestOfAPackageNamedLikeATestIsEvidence
        records core/odd.Unused as test-support, but nothing names it
--- FAIL: TestSymbolOwnership_APackageNamedForSomethingOtherThanItsDirectory  expected "both-apps", actual ""
```

The substantive one: **an unaliased import binds the imported package's *declared* name, not its
path's last element.** A core package named for anything other than its directory had every
production selector on it read as no reference at all. Also: an omitted `ConstSpec` inherits the
preceding *type slot* as well as the expression list; the declared-type exclusion was deleting the
type's name from the value's whole declaration, erasing independent references in the initializer;
and nothing distinguished a const repeating the list above it from a var taking its zero value
(mutating `case d.Tok == token.CONST` to `case true` left the whole core tier green).

Two corrections to the findings as written, neither changing what was fixed: finding 2 expected
`own-package` where a *declaration* edge gives `reachable`, and findings 4 and 5 named one helper in
two places. **Seven mutations after the repair, all CAUGHT.**

Where the family stops was decided here rather than assumed: `core/testutil` has a third site with
the same assumption, `defaultImportName` in `errors_lint.go`, and it is left alone deliberately and
drafted as the follow-up in the comment above.

### Round 4 — `8b107725`, two findings and one settled by mutation

Both in `declaredTypeOccurrences`. **The generic type slot:** `mark` accepted only a bare
`*ast.Ident`, and Go spells an instantiated named type as an `IndexExpr` or an `IndexListExpr`, so
`var Default Box[int]` never reached the exclusion and a `both-apps` `Default` made `Box`
`reachable` off nothing but itself — the exact circularity the exclusion exists to refuse, reached
by writing the type generically. **The type-assertion head:** `var ToneWarm = source.(Tone)`
promoted `Tone` to `own-package` off its own member, while the byte-for-byte equivalent
`Tone("warm")` was correctly excluded. Those three heads — conversion, composite literal, type
assertion — are the whole of it, and that argument is now written above the helper rather than left
implicit.

**A third defect, found by mutation rather than by the review.** The first draft of the comma-ok
fallback carried `&& i == 0`, on the reviewer's forced answer that only the first result takes the
asserted type. The mutation removing it **SURVIVED**: the restriction can never decide anything,
because the helper returns early on an empty `named` and `named` is the value's own `go/types` type.
For it to matter, a package-level object would have to be both a type and a function, which Go
forbids. The condition was deleted rather than kept as defensiveness no test can reach. This was the
second time in this gate that a surviving mutation, not a verdict, settled a rule.

Four mutations **CAUGHT**, one **SURVIVED** and resolved by deleting what it proved unreachable.

### Round 5 — `9b2cde93`, four findings, and the gate closed

Three were about the previous round's *coverage* rather than its behaviour, each verified by
mutating the arm and watching the core tier stay green (1143 passing) before the fixture was
written: the `IndexListExpr` arm of `declaredTypeOccurrences` (Go writes `Box[Knob]` as an
`IndexExpr` and `Pair[Knob, Toggle]` as an `IndexListExpr`, and round 4's fixture covered only the
first), the `CompositeLit` arm, and both generic `receiverTypeName` arms — blind because no core
method has a generic receiver.

**The fourth was a defect in shipped code:** `receiverTypeName` did not strip parentheses.
`func (l *(Level)) Build()` compiles and `gofmt -l` lists it, so nothing this repository runs would
flag the spelling; unstripped, the walk fell to its default arm, returned `""`, and `readDecl`
discarded the method. That is the quiet direction — edges lost silently, and the table then asks a
human to assert a symbol the tree justifies. Verified with the toolchain rather than from memory.

**Evidence: five mutations after the fix, every one CAUGHT; four mutations before it, every one
SURVIVED** — which is what makes these findings about the tests rather than about the reviewer.

**What closing at five costs, stated rather than implied.** `9b2cde93` is 304 lines that no reviewer
has read, in the same position round 4's repair was in when round 5 opened — and round 5 found two
surviving mutations in that repair's neighbourhood. The argument for stopping anyway: all four of
round 5's findings are in the guard's own coverage rather than in what it computes; the table it
produces is byte for byte unchanged across every one of the five rounds' repairs; and the five
mutations run there are the evidence a sixth round would mostly re-derive.

---

### The `unchecked` lists, answered

Round 1 raised two; rounds 2 to 5 returned `unchecked` empty.

- *"No manual interactive browser pass was run… visual layout was not inspected."* — **Dismissed.**
  The change's property is that no rendered template changes, and the renderer split moved the two
  `FuncMap`s and the enrichment without editing a template or a class. What would catch a regression
  is `adminconsole/internal/rendertest`, which renders every console page through
  `RenderTemplateToBuffer` including the moved `loggedInUser` bind, and it is green. A visual pass
  would be judging the design rather than the change.
- *"The full four-engine suites were not rerun locally…"* — **Covered**, and it is this run's own
  discipline rather than a gap: the four-engine data and integration tiers are the check suite's,
  read from GitHub at a named commit.

### What the record could not settle, said plainly

At the final gate, `tree-hash.sh --verify` showed that six of the eight stages recorded a tier
result against a tree a later stage then moved. That is what landing eight stages in sequence does,
not a defect in any of them, but it costs the **mutation** evidence: every `MUTATION CAUGHT` in
stages 1 to 7 was established against a `production` hash that has since changed. The green
four-engine suite settles the tiers on the current tree; nothing short of re-running each mutation
settles the mutations. So what is unproven is narrow — that a test written in stage 3 would still
fail if stage 3's production code were broken today — and it was named for the reviewer rather than
glossed. Stage 8 moved no production, test or docs bytes, so its claims stand as written.

Across the five code rounds, `data-tests 95b4af9e8d3a`, `integration-tests 9a560ed96300` and
`docs f5a84e6c8fbe` never moved: every repair was confined to `src/core/testutil`, a test-support
package no binary links.

Issue #419, opened 2026-09-21
The security-critical path sits in 444 to 994-line functions, which is where #137's ordering defect hid


leodip:
## Summary

The four largest functions on the OAuth path are 444 to 994 lines. Each is readable line by
line and none is navigable: you cannot see the shape of the token-request decision or the
authorize decision without reading the whole thing. The test ratios are the measurable cost —
every branch of a monolithic handler needs a full harness to reach, which is why `handlers` has
three times as much test code as production code.

Found in a full-tree review on `0c322a83`, finding E3. Measured on that commit.

## Measured

| Lines | Function | Location |
|---|---|---|
| 994 | `ValidateTokenRequest` | `authserver/internal/protocolvalidation/token_validator.go:98` |
| 592 | `main` | `cmd/goiabada-setup/main.go:140` |
| 532 | `HandleTokenPost` | `authserver/internal/handlers/handler_token.go:78` |
| 531 | `HandleAuthorizeGet` | `authserver/internal/handlers/handler_authorize.go:78` |
| 504 | `HandleIssueGet` | `authserver/internal/handlers/handler_auth_issue.go:28` |
| 444 | `HandleAuthCompletedGet` | `authserver/internal/handlers/handler_auth_completed.go:21` |
| 415 | `Seed` | `authserver/internal/data/database_seeder.go:98` |
| 384 | `initRoutes` | `authserver/internal/server/routes.go:30` |

`ValidateTokenRequest` is a single `switch` on grant type with four cases of roughly 100 to 460
lines and a nested switch inside one of them.

## Why this is worth an issue rather than a shrug

**It hides ordering defects, and has.** #137 exists because the code-exchange arm runs the
user-enabled and generation checks before client authentication and PKCE. The comments in the
same file at `:681`, `:701` and `:774` describe an order of checks the code no longer follows —
they were accurate when written. The refresh arm has the same defect (`:614-624` and `:641-655`
run before the ownership check at `:658`), and the only reason it was found in this review is
that someone read 500 consecutive lines looking for it. In a 60-line function whose name is
`validateRefreshGrant`, the ordering is the first thing a reader sees.

**The test ratio is the cost, and it is measurable:**

```
handlers               9.5k production   29.6k test
protocolvalidation     1.9k production    9.1k test
issuance               1.5k production    7.2k test
```

Those are not thorough-test ratios; they are reach-the-branch ratios. To exercise one arm of
[...]

Issue #417, opened 2026-09-21
Five verified drifts in the unguarded parts of CLAUDE.md, and the register problem behind them


leodip:
## Summary

`AssertAgentDocs` holds the parts of `CLAUDE.md` / `AGENTS.md` that are machine-checkable — the
auth-state roster, the module tables, the ownership table. The parts it does not check have
drifted, and five instances are verifiable on `0c322a83`. Each is a small edit; filing them
together because the pattern is the point.

Found in a full-tree review on `0c322a83`, finding E2. All five verified against the source.

## The five

**1. The route table says 500 where the code writes 400.**

`CLAUDE.md`'s route gate table:

> | `/auth/level1completed` | GET | `level1_password_completed`, `level1_existing_session` | **500**, the one gate c499ec51 missed when it converted the other ten to 400. #248 part 1 |

`handler_auth_level1.go:40` calls `rejectAuthStateMismatch`, and
`auth_ceremony.go:121` sets `"_httpStatus": http.StatusBadRequest`. The gate answers **400**
like the other ten. `auth_ceremony.go:103` cites #248 part 1 as landed. The table was not
updated with the fix, so it now describes a defect that no longer exists — and it describes it
as the one exception to a rule the reader is being taught.

**2. `AuthContext.AuthStateGeneration` is absent from the AuthContext field rule.**

The field appears in 22 production files. The "AuthContext field rule" section enumerates which
fields accumulate across hops, which are recomputed, and which are written once at
`/auth/authorize` — `AuthMethods`, `AuthenticatedAt`, `OtpConfigGeneration`, `AcrLevel`,
`ConsentedScope`, `OTPKeyURL`. `AuthStateGeneration` is named nowhere, so the one section that
exists to say what happens to each field on a ceremony restart is silent about one of them.
Given that the section's whole purpose is the #140 class, that omission is the expensive kind.

**3. `handler_authorize.go` still describes a cookie the session no longer travels in.**

Around line 119:

```go
// ... Sanitize first — BCP 47 shape filter, capped at
// 10 tags / 256 bytes — so we don't bloat the session cookie or
// accept attacker-controlled junk.
```

#266 replaced `ChunkedCookieStore` with the server-side SQL store. The `AuthContext` lives in a
database row keyed by a cookie; there is no session cookie to bloat. This is the exact drift
`AssertAgentDocs` was written to catch, one file away from where it looks.

**4. The core directory list is incomplete.**

[...]

Issue #386, opened 2026-09-20
[Architecture follow-up 2/3] Narrow database and API client interfaces and propagate context


leodip:
## Goal

Replace the authserver's monolithic persistence capability and the admin console's monolithic HTTP-client capability with **consumer-owned, narrow interfaces**, and use that boundary change to propagate cancellation and deadlines through SQL and outbound HTTP.

Current starting point:

- `src/authserver/internal/data/database.go` declares a `Database` interface of roughly 215 methods and passes it through handlers, services and middleware.
- Its generated mock is almost 12,000 lines, so an unrelated method change creates broad generated-code and review churn.
- `src/adminconsole/internal/apiclient/auth_server_client.go` declares an `ApiClient` interface of roughly 106 methods and passes it through nearly every handler area.
- Database methods accept `*sql.Tx` but not `context.Context`; common SQL code therefore uses `Query` / `Exec` rather than context-aware variants.
- Admin API client methods do not accept context; implementations create requests without the incoming request context, the general client has no explicit overall timeout, and many response paths read bodies without a common maximum.

The objective is not merely smaller files. It is to make each consumer declare the capability it needs and to make every external operation cancellable and bounded.

## Why interface narrowing and context propagation are one issue

Doing these separately would cause two repository-wide API migrations: first split every signature, then revisit the same interfaces, implementations, mocks and call sites to add `context.Context`. The context belongs on the consumer-owned port from its first version.

The database and HTTP halves may land in separate PRs or commits, but they share one architectural rule and acceptance model.

### Consumer-owned ports

Interfaces belong at the package that consumes them. A handler or use case that reads one user and revokes sessions should depend on those capabilities, not on every operation supported by every database engine.

Representative shape only:

```go
type UserReader interface {
    GetUserByID(ctx context.Context, tx *sql.Tx, id int64) (*models.User, error)
}

type SessionRevoker interface {
    RevokeUserSessions(ctx context.Context, tx *sql.Tx, userID int64) error
}
```

Names and grouping should follow real use cases. Do not create a mechanical one-interface-per-method mirror. A cohesive consumer may own a small multi-method port; unrelated consumers should not share an interface only because the same concrete adapter implements it.

The concrete database adapters and auth-server HTTP client may still expose many methods. What must disappear is passing one omnipotent interface through the application.

### Composition

- Concrete implementations are assembled in `internal/server` / application composition roots.
- Go's implicit implementation should satisfy the narrow ports without adapter boilerplate.
[...]

Issue #420, opened 2026-09-21
The guard suite is 26% of the production tree, and one guard reimplements a go/types resolver


leodip:
## Summary

The guard and lint machinery is now about 25k lines against 96.6k lines of hand-written
production Go — roughly 26% of the tree, maintained to the same standard as the product. Most of
it earns that. One guard does not, and the marginal cost of the next one is now high enough that
it is worth setting a policy rather than deciding case by case.

This is a proposal, not a defect report. Filed from a full-tree review on `0c322a83`, finding
E1. All figures measured on that commit.

## Measured

```
production Go (excluding _test.go, _mock.go, mocks/)   96,637
core/testutil                                          14,620
module-level *lint*_test.go and *rules_test.go         10,372
                                                       ------
guard and lint total                                   24,992   (25.9% of production)
```

The largest module-level ones:

```
authserver/internal/data/migration_rules_test.go                      2,674
authserver/internal/server/openapi_contract_lint_test.go              1,972
authserver/internal/audit/events_catalog_lint_test.go                   828
authserver/internal/handlers/apihandlers/api_error_code_lint_test.go    747
authserver/internal/data/page_offset_lint_test.go                       479
authserver/internal/data/benign_sentinel_lint_test.go                   416
adminconsole/internal/handlers/api_error_forwarding_lint_test.go        393
...and nine more
```

Since 1 June: **64 of 691 commits** touch `core/testutil` — about one commit in eleven.

## What is working, and should not change

The guard shape is genuinely good and the reasoning in `guard.go`'s header comment is correct:
splitting finder from reporter, taking a `testutil.Reporter` instead of `testing.TB`, and
requiring a must-fail tree, a must-pass tree and an empty-walk case per guard. The
`AssertNoDeadInterfaces` incident on `8883642d` — a blinded guard that reported nothing with
nothing going red — is exactly the failure that discipline prevents, and it happened, so the
discipline is paid for.

The cheap guards each encode a real incident in a few hundred lines and are clearly worth it:
architecture edges, credential-from-query (255 lines), no-stdlib-errors, the slog convention,
gofmt, the mock pin. `credential_read_lint.go` is the model — narrow rule, small
implementation, obvious value.

## The one that has passed the point of return

`core/testutil/symbol_ownership.go` is **1,412 lines**, plus **2,129 lines** of rule tests:
3,553 lines in total, or about 3.7% of the production tree, for one guard. It reads a
[...]

PR #422, opened 2026-09-22
refactor(authserver): separate application services from transport and persistence

### LINKED ISSUE #387: [Architecture follow-up 3/3] Separate application services from transport and persistence
## Goal

Separate authserver application workflows from HTTP transport packages and persistence records, so package boundaries describe responsibilities rather than access to convenient functions.

The concrete targets found in the architecture review are:

- `src/authserver/internal/handlers/revocation.go`: substantial transaction, revocation and audit orchestration inside the transport package.
- `src/authserver/internal/handlers/otp_enrolment.go`: shared OTP lifecycle behavior kept in the parent handler package because browser and API handler packages need it.
- `src/authserver/internal/models/user.go`: a persistence record that also encrypts/decrypts OTP secrets and maps itself into OIDC address claims.
- Child handler packages importing `internal/handlers` for application services, interfaces and helpers, making the parent transport package an accidental service layer.
- Very large root and API handler packages whose size is a symptom of mixed responsibilities, not a reason to split files arbitrarily.

This issue extracts the use cases first, then lets the handler structure become smaller around real boundaries.

## Design rule

HTTP handlers should normally:

1. parse and validate transport input;
2. authenticate/authorize the caller;
3. invoke one application operation through a narrow port;
4. translate the result into the correct page, redirect or JSON response.

Transaction ownership, multi-repository orchestration, credential lifecycle, audit-event decisions and claim construction should live outside the HTTP package unless they are intrinsically transport behavior.

Persistence records should represent stored data and persistence-specific conversions. Possessing fields does not make a record the right owner of encryption, credential changes or protocol claim construction.

## Why this is one issue

The revocation and OTP functions live in `handlers` primarily so sibling/child handler areas can reach them. Moving either in isolation creates another temporary shared-helper location. The model methods participate in those same workflows.

This issue establishes one application boundary and removes the package-visibility workaround. It is intentionally not a wholesale domain-driven-design conversion.

## Intended package direction

Use capability/domain names, not a generic `services` package. Exact names should follow the final dependency graph, but likely destinations are:

- `internal/revocation` — user/client/session/token revocation orchestration and transaction ownership;
- `internal/otpcredential` or an equally precise credential package — OTP enrollment, replacement and removal lifecycle;
- an issuance/userinfo claims mapper — conversion from stored user/profile data to OIDC claims;
- a small `internal/websupport` package only for genuinely transport-specific seams shared by multiple handler areas.

Do not move every helper into these packages. A package earns existence from one cohesive responsibility and an acyclic consumer graph.

### 1. Characterize behavior before moving it

Pin the current workflows with tests before extraction:

- transaction statement order and rollback behavior;
- which sessions, authorization codes, access tokens and refresh-token families each revocation scope affects;
- audit event name, fields and the rule that success is logged only after a successful commit;
- error identity/classification reaching page and JSON handlers;
- OTP generation changes, enrollment/replacement/removal atomicity and secret handling;
- OIDC address/profile claim omission and serialization behavior.

The extraction must make those tests easier to point at the use case directly; do not leave all proof only at the HTTP layer.

### 2. Extract revocation as an application capability

- Move transaction and repository orchestration out of `handlers/revocation.go`.
- Give each operation narrow persistence/audit dependencies from #386 rather than the full database capability.
- Keep transaction boundaries and ordering exactly as today, including the deliberate session-delete/issuance ordering described in `AGENTS.md`.
- Return typed/classifiable outcomes; do not return HTTP status codes or write responses from the application package.
- Keep authorization decisions that depend on the authenticated HTTP caller at the boundary, while enforcing invariant/domain authorization in the use case where all callers need it.

### 3. Extract OTP credential lifecycle

- Move enrollment, replacement and removal behavior out of the handler hierarchy.
- The application capability owns secret generation/protection, `otp_config_generation` changes and their transaction atomicity.
- Browser and admin/API handlers invoke the same operation rather than sharing an unexported handler helper.
- Keep form parsing, continuation/ceremony identifiers, rate-limit response selection and HTML/JSON rendering in the appropriate handler.
- Never expose a plaintext OTP secret beyond the minimum operation result required to render enrollment or construct a QR code.

### 4. Remove protocol and cryptographic behavior from persistence records

For `models.User` and similar records:

- Move OTP secret encryption/decryption to the credential capability or a narrow secret protector used by it.
- Move OIDC address/profile claim mapping to issuance/userinfo code.
- Keep database tags, nullable database representation and trivial record-local predicates where they genuinely describe stored state.
- Check other methods while the package is open, but add them to this issue only when they exhibit the same transport/protocol/credential mixing. Do not turn this into a speculative rewrite of every model.

Renaming `models` to `model` without changing responsibilities is explicitly not a deliverable.

### 5. Simplify handler dependencies after extraction

- Remove imports from `handlers/apihandlers` and `handlers/accounthandlers` back to the root `handlers` package when they exist only for the extracted services/interfaces/helpers.
- Keep shared response writers and form/continuation parsing in a transport-focused package if multiple handler areas truly use them.
- Reassess the large root handler package after dependency edges are removed. Group protocol ceremony, account recovery/self-service and admin API handlers by bounded feature only where that produces a durable package boundary.
- Avoid a mass file shuffle whose only result is shorter directories.

### 6. Naming while touched

Apply Go initialism spelling (`API`, `HTTP`, `ID`, `JSON`) to moved/new identifiers where compatibility permits. Remove names that stutter with their package, for example a `client` package need not expose `HandleAdminClient...` if a clearer transport-level name is available.

Do not make repository-wide naming churn part of this issue.

- Preserve all routes, methods, status codes, redirects, templates, JSON, error codes, translated messages and OpenAPI shapes.
- Preserve authentication ceremony transitions, ACR/AMR behavior, session promotion, OTP generation semantics, revocation reach and audit records.
- Preserve rate-limit accounting: successful credential checks spend nothing and the same failures are counted.
- Preserve transaction boundaries, statement order, deadlock retry behavior and post-commit audit rules.
- Do not introduce a universal service layer, repository abstraction framework, event bus or dependency-injection framework.
- Do not require persistence records to become pure domain aggregates. The aim is to remove clear application/protocol/cryptographic orchestration, not to erase useful record methods.
- Do not combine page and JSON error writers; existing response-format guards remain authoritative.
- Handler package reorganization is secondary to service extraction and must not obscure behavioral review.

## Acceptance criteria

- Revocation transaction/orchestration code is no longer owned by an HTTP handler package.
- OTP credential enrollment/replacement/removal is one application capability used by browser and API handlers.
- Persistence records do not encrypt/decrypt OTP secrets or construct OIDC claims.
- Application packages do not import HTTP handler packages or accept `http.ResponseWriter`, `*http.Request`, template data or HTTP status codes.
- Feature handler packages do not import the root `internal/handlers` package merely to reach application services or shared interfaces.
- Extracted use cases depend on narrow consumer-owned ports and accept `context.Context` as established by #386.
- Unit tests exercise the application operations directly, including failure/rollback/audit cases.
- HTTP tests demonstrate unchanged page-versus-JSON classification and unchanged externally visible behavior.
- No new generic `helpers`, `common`, `utils` or catch-all `services` package is introduced.

## Verification

- Unit tests for core, authserver and adminconsole.
- Authserver data tests for every affected transaction and database engine.
- Race tier for revocation/session/OTP paths with concurrent writes.
- Integration tests for user-, client- and session-scoped revocation; OTP enable/replace/disable; login step-up after OTP generation changes; userinfo/ID-token claims.
- Audit catalog, API error-code, handler-interface, transaction, architecture, gofmt, logging and generated-mock guards.
- The PR description should show the old and new dependency graph for each extracted capability and state which responsibilities deliberately remain in HTTP and persistence packages.

## Dependencies and series

Part 3 of 3. Depends on #386 so the extracted operations are born with narrow contextual ports rather than being migrated twice. It also assumes the ownership decisions in #385 are complete, preventing a new application package from being placed in core.

leodip:
Closes #387.

## What this is

Five pieces of the auth server behave as application services and sit in an HTTP handler package
anyway. `internal/handlers/revocation.go` is 688 lines with zero `net/http` references; it declares
an exported 12-method database port, five operations and three audit helpers, and never touches a
`ResponseWriter`, a template or a status code. `internal/handlers/otp_enrolment.go` says the reason
out loud in its own doc comment: "It lives here rather than beside `disableUserOTP` because the
browser handler cannot reach an unexported function in `apihandlers`." `models.User` encrypts a TOTP
seed and builds an OIDC `address` claim; `models.KeyPair` decrypts and parses a private key. And the
userinfo endpoint and the token issuer each carry their own copy of the OIDC claim block.

The thesis is one sentence: **the application layer sits in the transport package because nothing
else was reachable from all its callers.** 42 production files in `apihandlers` and `accounthandlers`
import the parent handler package, and every symbol they reach is one of seven groups, each of which
belongs somewhere else.

## The rule this establishes

An application operation takes a `context.Context` and a port the *consumer* declares, naming only
the methods that consumer calls. It never takes a `ResponseWriter`, a `*http.Request`, template data
or an HTTP status code, and it never imports a handler package. Where two packages need the same
collaborator, each declares its own port rather than sharing one — which is what `internal/middleware`
already does with its one-method `AuditLogger`, and what #386 established for databases one commit
before this. No `helpers`, `common`, `utils` or `services` package is created; each new package is a
lowercase compound noun naming the one thing it owns.

Four new packages: `internal/revocation`, `internal/otpcredential`, `internal/userclaims`,
`internal/emaillinks`. `ParsePrivateKey` joins `internal/signingkeys`, which already owns key
generation and public-key lookup.

## What does not change

No route, status code, redirect, template, JSON body, error code, translated message or OpenAPI
shape. No migration: this moves code that reads and writes the existing columns. No handler file
moves between directories and no function is split — `HandleIssueGet` and `HandleAuthCompletedGet`
stay for #419, which describes that work for the same package for the same reason. `routes.go` is not
edited, because the concrete types it builds satisfy whatever port each consumer declares.

Two places could change behaviour without a compile error: merging the duplicated claim block, and
extracting the OTP verify arms. Both are pinned by characterization tests written against the code as
it stands, before either moves. The claim block's three measured divergences — where `updated_at` is
gated, which include flag filters groups and attributes, and where the base URL comes from — are
preserved as explicit inputs rather than merged away.

## Status

In progress. No stage has landed yet.

### Stage 1: pin the claim divergences and the OTP enrolment replay arm
_Not started._

### Stage 2: `internal/revocation`
_Not started._

### Stage 3: `internal/emaillinks`
_Not started._

### Stage 4: `internal/otpcredential`
_Not started._

### Stage 5: `internal/userclaims`
_Not started._

### Stage 6: `ParsePrivateKey` to `signingkeys`, and the `models` import assertion
_Not started._

### Stage 7: consumer-owned ports, and the child packages stop importing `handlers`
_Not started._

### Stage 8: the package layout in the agent docs, #417's three live drifts, and the issue comments
_Not started._


leodip:
## Deferred decisions

**None.** Nothing in this branch was built on an assumption waiting for an answer, and nothing here
is a decision the reviewer has to ratify before merging.

The agreement carries ten decisions, all of them settled before the first commit and all of them
raised by the planning pass rather than by the run: how far the handler-package reorganization goes
(extraction only), whether the two large ceremony handlers are split here (no, they are #419's),
where the shared collaborator interfaces live (each consumer declares its own), the OTP capability's
four-operation surface, whether the claims mapper merges the two copies (yes, with the three
divergences as inputs), whether those divergences are pinned before the move (yes, stage 1), whether
a transport-shared package exists and what is in it, whether `KeyPair.ParsePrivateKey` moves with the
`User` methods (yes), which of #417's drifts are folded in, and what the four new packages are
called. Zero were opened during the run, so zero were carried on an assumption.

Two things the run decided for itself, inside a stage rather than as a decision, are worth naming
here because each disagrees with what the agreement predicted and each is visible in the diff:

- **`models`' allowed imports are three paths, not two.** The plan wrote acceptance criterion 3 as
  "the standard library and `core/constants`". Four enumerations that #385 moved into `models` raise
  their refusals through `core/errs`, as pattern 7 requires of every error this tree constructs, so
  `models/import_lint_test.go` allows three. The property the criterion is about — no cipher, no
  parser, no database, no service behind a persistence record — is unchanged, and is what the guard's
  failure message states. Reversing this means moving four enumerations out of `models`, which is a
  different change.
- **`revocation.RevokeRefreshTokens` is exported, one rename more than planned.** The token
  endpoint's auth-code-reuse arm (#77) calls what used to be a package-private helper and is now one
  package away. The alternative is duplicating the sweep at the call site. Reversing it costs that
  duplication.

leodip:
## Follow-ups

Two problems were found while doing this work. **Both have since been folded into this branch, so
there is nothing left to file.** Each is recorded under **Folded in** below, with what the fix
turned out to be and where the original draft was wrong.

This comment was first posted with both as follow-ups, and then with one. The duplicate searches
behind them were re-run against the tracker after the last commit at that time:
`gh issue list --state all --search "RunInTransaction stub test helper"`,
`--search "expectRunInTransaction"`, `--search "updated_at claim access token"` and
`--search "authserver:userinfo scope"` returned nothing that reaches either, and a read of the 30
open issues found no overlap — #420 is about the bulk of the guard suite and the `go/types`
resolver inside one guard, not about duplicated test helpers.

The three `CLAUDE.md` drifts taken from #417 are not a fold-in: they are documentation this change's
own goal already owed, settled as decision 9 before the first commit, and described in the body's
stage 8 section.

---

## Folded in

### The `RunInTransaction` stub, merged from seven copies into one that refuses nil

Commit `9697f035`. Drafted here as a follow-up, then done on this branch instead: it is test
scaffolding for the mock `Database`, two of the copies were added by this change, and the fix is
confined to the tier this branch already owns.

**It was seven, not six.** The count in the draft came from
`git grep -ln "func expectRunInTransaction"`, which misses
`signingkeys/signing_key_rotator_test.go`: the same stub under a different name,
`expectRotatorTransaction`, with its own `rotatorTx` sentinel.

**The defect is narrower than the draft states, and real where it lands.** The four packages without
a `txSentinel` were not all weak — their call sites each declare their own non-nil sentinel
(`revokeTx`, `issuanceTx`, `otpTx`, `clientUpdateTx` and the rest), so the identity assertion is
there, just declared per file. **Two call sites actually passed nil**, and both were asserting the
in-transaction property they could not see, through `(*sql.Tx)(nil)` expectations that a call made
outside any transaction matches equally well:
`TestHandleTokenPost_AuthCodeReuse_RevocationFailureReturns500` in `handler_token_test.go` and
`TestHandleTokenPost_WrappedAuthCodeReuseStillRevokes` in `handler_token_wrapped_errors_test.go`.
Both now name `revokeTx`.

**The placement question the draft said needed settling, settled.** The one copy is
`mocks_data.ExpectRunInTransaction`, hand-written beside the generated mock in
`src/authserver/internal/data/mocks/run_in_transaction_stub.go`, with
`ExpectRunInTransactionThenFail` and `ExpectRunInTransactionRefused` beside it. That placement costs
no import edge, which is what the draft's two candidates each cost: all twenty call-site files
already import `mocks_data`, so the change is a rename at 105 call sites and not one new import.
Nothing there is generated, so `AssertGeneratedMocksArePinned` leaves it alone — it keys on
mockery's own `DO NOT EDIT` marker rather than on a filename, which is what already leaves
`core/mocks/test_fs_mock.go` alone — and the lint tier, which runs the generator and fails on a
dirty tree, confirmed the generator does not touch it. The build tag is the generated mock's, for
the generated mock's reason.

**A nil transaction is now refused at registration**, panicking with a message that names the fix, so
the test that wrote one is the test that fails. What was two packages' convention is the stub's rule.

**The stub's own behaviour is now under test**, which nothing stated while it was seven copies:
the transaction the body is handed, the error the body returns, the commit that fails after a body
that did not, the edges a `note` sees on all three arms, the transaction that never opens, the nil
refusal, and that exactly one call is registered.

Two stubs stay local on purpose: `usercreation`'s rerun stub and `apihandlers`' deadlock-loop stub
run the body more than once, which is not what the shared one does. `usercreation`'s now hands over
`txSentinel` rather than nil.

`CLAUDE.md` and `AGENTS.md` gained a **Transaction stub** paragraph under Testing, so the next test
that needs the mock to answer `RunInTransaction` finds the one stub rather than writing an eighth.

Results: unit tier green, 8693 passing 0 failing, against 8630 at this branch's stage 1 baseline.
Lint tier green. Three mutations against the shared stub, all `MUTATION CAUGHT` — dropping the nil
refusal, handing the body nil instead of the caller's transaction, and noting a rollback as a commit.
Net −78 lines across 31 files.

---

### `updated_at`, emitted under the `profile` scope at every site

Commit `d94520c7`. Drafted here as an issue naming three possible answers; the answer chosen is the
second of them, made deliberate for both token types by gating on the scope the claim actually
belongs to.

**The draft's mechanism is correct, and I verified it before changing anything.**
`generateAccessTokenCore` appends `authserver:userinfo` to its `scopes` slice for the audience, at
`token_issuer.go:734`, before the claim block reads that same slice at line 767;
`generateIdTokenCore` passes the granted scope verbatim. Measured rather than assumed: pointing the
access token's claim call at the granted scope flipped **exactly one leaf case** in the whole
authserver tier, `TestClaimCharacterization_UpdatedAtGate/openid_alone`, because `AddGroupClaims`
and `AddAttributeClaims` gate on membership of a named scope and only the one gate that counts
elements can see the appended entry.

**Two things the draft got wrong, and they are why the fix is not the one it proposed.**

*It understates the problem, and misses the half that happens by default.* Per OIDC Core 5.4,
`updated_at` is a claim requested by the **`profile`** scope: *"name, family_name, given_name,
middle_name, nickname, preferred_username, profile, picture, website, gender, birthdate, zoneinfo,
locale, and updated_at"*. Issuance's gate is not that rule, so it is wrong in the other direction
too: **`scope=openid email` put `updated_at` in both tokens with no `profile` granted**, and that is
the default path, since `IncludeOpenIDConnectClaimsInIdToken` is seeded on while
`IncludeOpenIDConnectClaimsInAccessToken` is seeded off. The divergence the draft names appears only
where an operator turned access-token claims on; the conformance gap it does not name appears
everywhere. This repository's own documentation had already assigned the claim to `profile`, in
three tables across `site/src/content/docs/concepts/openid-connect.mdx` and
`integration/endpoints.mdx`, so the old behaviour contradicted Goiabada's own published contract.

*Its "three answers" are not equally weighted, and its "what it needs" points at the wrong file for
two of them.* The tree already contained the conformant rule — `/userinfo` used
`UpdatedAtWithProfileScope`. Adopting it in issuance makes all three sites agree and **removes**
divergence 2 from #387's structure rather than adding to it: `UpdatedAtGate`,
`UpdatedAtWithProfileScope`, `UpdatedAtBeyondOpenidScope` and `Mapper.UpdatedAt` are all deleted,
and the claim moves inside the profile arm. That fix lands in `userclaims`, not in
`generateAccessTokenCore`'s scope handling as the draft says; only the narrow answer lands there.

**Severity, stated plainly.** No normative MUST or MUST NOT in OIDC Core forbids returning
unrequested claims, so the old behaviour was off-model rather than a spec violation. RFC 9068 §6 is
the relevant pressure for the access-token half: *"it's important to evaluate whether the resource
server gained the proper entitlement to have access to any content received in the form of claims,
for example, through user consent."* The claim is a timestamp, so the exposure is small. This is
conformance and data minimization, not a vulnerability.

**What changed on the wire:** a token for `openid` alone or `openid email` no longer carries
`updated_at`; `openid profile` is unchanged. Breaking for a client reading the claim without asking
for `profile`. The PR body's **What does not change** section now carries this as its one exception,
and stage 10 has the detail.

**What was measured.** The profile gate flips 8 leaf cases, all of them pinning the old behaviour,
which is how the blast radius was chosen before the decision rather than discovered after it. Three
mutations, all `MUTATION CAUGHT`: the old issuance gate restored against the new integration case,
failing exactly the `openid alone` and `openid email` rows; the same against the unit tier; and a
gate widened to any multi-scope grant. Unit 8695 passing 0 failing, sqlite integration 1429, sqlite
data 634, lint 0 issues over the four modules; the other three engines from CI.

---

### Comments already posted to issues this work touches

Not follow-ups — these are done, posted with this pull request, one per issue whose territory this
change entered:

| Issue | What it was told |
|---|---|
| [#419](https://github.com/leodip/goiabada/issues/419#issuecomment-5779713341) | `HandleIssueGet` and `HandleAuthCompletedGet` are back in its court; extraction took almost nothing out of either |
| [#417](https://github.com/leodip/goiabada/issues/417#issuecomment-5779713741) | which three of its five drifts landed here, which two were wrong or already fixed, and that `CLAUDE.md` grew 787 words so the halving proposal's measurements need re-taking |
| [#144](https://github.com/leodip/goiabada/issues/144#issuecomment-5779714089) | the OTP claim work it describes now sits in `internal/otpcredential` |
| [#259](https://github.com/leodip/goiabada/issues/259#issuecomment-5779714417) | `RevokeClientGrants` now lives in `internal/revocation`, and its cited `src/core/data` path has moved twice since it was filed |
| [#131](https://github.com/leodip/goiabada/issues/131#issuecomment-5779714735) | the refresh/generation coordination it describes now sits in `internal/revocation`, not `handlers/revocation.go` |
| [#386](https://github.com/leodip/goiabada/issues/386#issuecomment-5779715034) | closed, but its decision 8 exported two ports for a reason this change removed |

**None.** `git diff --unified=0 origin/main...HEAD | grep '^+.*ceiling:'` returns nothing: this branch
added no `ceiling:` marker, which is the expected answer for a change that relocates code without
choosing a new bound.

leodip:
## The review account

The whole branch was reviewed once, at the final gate, against `origin/main...HEAD`. No stage was
reviewed on its own. This is the entire reading the code got, so it is posted here in full rather
than left in a local run directory nobody else can open.

| Round | Reviewer | Effort | Findings | Blocking | Outcome |
|---|---|---|---|---|---|
| 1 | `openai-codex/gpt-5.6-sol`, fresh session | `xhigh` | 0 | 0 | clean |

Request `387-20260922T160909Z`, gate `final`, round 1 of a maximum of 5, started
2026-09-22T16:10:11Z, against `9395b4bc6925990f25a91ca3d827e9703b6a8e30`.

**Verdict: clean, zero findings.** Nothing to fix, nothing to refute, nothing deferred. There are no
per-finding entries below because there are no findings; everything that follows is the evidence
that the verdict is worth its word.

`axis_coverage` reports conformance, quality and security all `reviewed`, with `unchecked` empty —
the reviewer asserting it covered every axis rather than reporting a clean read of part of the
change. The verdict was recorded by the driver rather than by a session, because it met the clean
definition exactly and no session was needed to say so; nothing in the tree was touched between the
verdict and this comment.

### What the clean verdict rests on

A clean verdict is only worth its stated execution, so here is the execution the reviewer reported:

- the agreement itself — entry, contract, digest, closing tail and the full closing log — plus the
  testing and security review references
- the diff twice, `origin/main...HEAD` and the working tree, and `git diff --check`
- `tree-hash.sh --verify` and `check-anchors.sh`
- changed-file, import-graph, reuse and caller-fanout checks
- targeted test runs on the current tree: the OTP credential, `userclaims`, `signingkeys`, stage 1's
  characterization cases and the two new import guards
- `AGENTS.md` and `CLAUDE.md` checked byte-identical
- **five behaviour mutations run against the current tree**, all caught: the `updated_at` gate, the
  access-token inclusion wiring, the userinfo base URL, OTP enabled-state enforcement, and the OTP
  replay outcome. A sixth, a base-URL mutation, did not compile; the reviewer discarded it rather
  than counting it and replaced it with a compile-safe one, which was also caught

Reviewer's own note: *"Clean on all three axes. The exact HEAD tree hash matches the final recorded
CI tree, targeted new cases pass, and five current-tree behavior mutations were caught."*

### The gap the request named before the round started, and what closed it

The request did not ask for a clean read of a tree it had to discover for itself. A staleness check
ran first, and it found that **7 of the 8 stage entries record a tier result against a tree that has
since moved** — which is ordinary for an eight-stage branch where later stages repoint what earlier
ones moved, but it decides what a final review can still rely on. Two halves, handled differently:

- **Every tier claim in stages 1 through 7 is stale, and every one is re-established at HEAD by CI**,
  which is green on `9395b4bc` across all thirteen jobs: `Tests / {sqlite,mysql,postgres,mssql}`,
  `Unit / {core,authserver,adminconsole,setup,race}`, `Lint`, `Vulnerabilities`, `Versions` and
  `Build validation`. The same tiers ran again, on the final tree, on four engines, after the last
  stage landed. A stale tier result is not a gap here.
- **The ten stage mutations are the part CI does not re-establish.** Stages 1 through 8 recorded ten
  mutations, all `MUTATION CAUGHT`, each against its own stage's production hash, and seven of those
  eight hashes have since moved. A mutation proves a claim about a *pair* — a test and the production
  line it is aimed at — and later stages moved both sides of several: stage 2's revocation mutations
  ran before stage 7 repointed the ports that reach them, stage 4's OTP mutations before stage 7 gave
  `apihandlers` its own collaborator ports. Nothing suggests any of the ten has gone soft, since
  every guard and seam they target is green at HEAD, but "still green" is the weaker claim.

That gap was named in the request rather than left for the reviewer to discover, so it could decide
whether to spend a round on it. It did: the five current-tree mutations listed above are aimed at
exactly the pairs the staleness check flagged — the `updated_at` gate and the inclusion wiring from
stage 5 sitting on stage 1's characterization cases, the OTP enabled-state and replay outcomes from
stage 4 sitting behind stage 7's repointed ports. All five caught. The stronger claim is re-made on
the final tree for the pairs where it had most to lose.

### What no reviewer looked at, stated plainly

One round, so there is no second opinion on any of it, and an unproductive round does not become a
second reviewer by being counted. The parts of this branch with the most room for a silent behaviour
change were the claim-block merge and the OTP verify extraction, and those are the two the round's
mutations went at; the rest of the diff is moves that a compiler and thirteen CI jobs constrain
rather than a reviewer does.

